PR #18, opened 2019-02-07
quick fix to the docs for conda install

andrewkern:
just a quick fix up to the install directions for conda. we needed to specify to the user that they should add the `conda-forge` channel before installing the requirements. I'm also tempted to ask the users to start a new virtual environment in conda for stdpopsim, but perhaps that is overkill.


jeromekelleher:
Good call, thanks @andrewkern!

> I'm also tempted to ask the users to start a new virtual environment in conda for stdpopsim, but perhaps that is overkill.

It's probably a good idea to suggest it anyway, but I'm not sure how much detail we should be supplying here. Let's merge this much and we can add to it easily enough.

PR #23, opened 2019-03-04
At model

arundurvasula:
Hi! Here's our first stab a model for Arabidopsis thaliana. Chris Kyriazis converted the genetic map from Salome et al 2012 (now hosted on Kirk's website) to HapMap format. We implemented the demographic model inferred using MSMC from my 2017 paper.  We've also added tests based on Andy's fly example and tried to adhere to PEP8 (though please let me know if anything doesn't conform to the standard!).

Please let me know if we're missing anything or if anything seems off!

Thanks!
Arun

### DIFF
--- a/athaliana_example.py
+++ b/athaliana_example.py
@@ -1,0 +1,22 @@
+"""
+A. thaliana msprime example
+"""
+import msprime
+
+from stdpopsim import arabidopsis_thaliana
+
+chrom = arabidopsis_thaliana.genome.chromosomes["chr5"]
+recomb_map = chrom.recombination_map()
+
+model = arabidopsis_thaliana.Durvasula2017MSMC()
+model.debug()
+
+samples = [
+    msprime.Sample(population=0, time=0), msprime.Sample(population=0, time=0)]
+
+ts = msprime.simulate(
+    samples=samples,
+    recombination_map=chrom.recombination_map(),
+    mutation_rate=chrom.mean_mutation_rate,
+    **model.asdict())
+print("simulated:", ts.num_trees, ts.num_sites)
--- a/stdpopsim/arabidopsis_thaliana.py
+++ b/stdpopsim/arabidopsis_thaliana.py
@@ -1,0 +1,115 @@
+"""
+Genome, genetic map and demographic model definitions for Arabidopsis thaliana.
+"""
+import msprime
+import numpy as np
+
+import stdpopsim.models as models
+import stdpopsim.genomes as genomes
+import stdpopsim.genetic_maps as genetic_maps
+
+
+###########################################################
+#
+# Genetic maps
+#
+###########################################################
+
+
+class Salome2012(genetic_maps.GeneticMap):
+    """
+    Genetic map from Salome 2012 averaged across population crosses.
+    Please see this repo for details on how this was done: https://github.com/
+    LohmuellerLab/arabidopsis_recomb_maps
+
+    """
+
+    url = ("http://www.eeb.ucla.edu/Faculty/Lohmueller/data/"
+           "uploads/salome2012_maps.tar.gz")
+    file_pattern = "arab_{name}_map_loess.txt"
+
+
+genetic_maps.register_genetic_map(Salome2012())
+
+
+###########################################################
+#
+# Genome definition
+#
+###########################################################
+
+# Data for length information based on:
+# https://www.arabidopsis.org/portals/genAnnotation/
+#   gene_structural_annotation/agicomplete.jsp
+# Lengths from TAIR 10 although its unclear what reference the genetic map used
+#   -- follow up on this with Salome 2012 authors
+
+_chromosome_data = """\
+chr1 30427671
+chr2 19698289
+chr3 23459830
+chr4 18585056
+chr5 26975502
+"""
+# mutation rate from Ossowski 2010 Science
+# recombination value from Huber et al 2014 MBE
+# rho=200/Mb, assume Ne=124,000, rho=2*Ne*r
+_chromosomes = []
+for line in _chromosome_data.splitlines():
+    name, length = line.split()[:2]
+    _chromosomes.append(genomes.Chromosome(
+        name=name, length=int(length),
+        mean_mutation_rate=7e-9,
+        mean_recombination_rate=8.1e-9))
+
+genome = genomes.Genome(
+    species="arabidopsis_thaliana",
+    chromosomes=_chromosomes,
+    default_genetic_map=Salome2012.name)
+
+
+###########################################################
+#
+# Demographic models
+#
+###########################################################
+
+
+class Durvasula2017MSMC(models.Model):
+    """
+    Model estimated from two homozygous individuals from the South Middle Atlas
+    using MSMC
+    """
+
+    def __init__(self):
+
+        # the size during the interval times[k] to times[k+1] = sizes[k]
+        self.times = np.array([
+            699, 2796, 6068, 9894, 14370, 19606, 25730, 32894, 41275,
+            51077, 62544, 75958, 91648, 110001, 131471, 156584, 185960, 220324,
+            260520, 307540, 362541, 426879, 502139, 590173, 693151, 813610,
+            954517, 1119341, 1312147, 1537686, 1801500, 2110100])
+        self.sizes = np.array([
+            42252426, 42252426, 60323, 72174, 40591, 21158, 21442,
+            39942, 78908, 111132, 110745, 96283, 87661, 83932, 83829, 91813,
+            111644, 143456, 181571, 217331, 241400, 246984, 238593, 228222,
+            217752, 198019, 165210, 121796, 121796, 73989, 73989, 73989])
+
+        # MSMC is accurate from 40Kya-1.6Mya for A.thaliana(Durvasula et al 2017)
+        # set the first 7 sizes
+        # equal to the size at 8 (~40Kya)
+        self.sizes[:8] = self.sizes[8]
+        # set the last 2 entries equal
+        # to the size at 30 (~1.6Mya)
+        self.sizes[30:32] = self.sizes[30]
+        self.demographic_events = []
+        for idx, t in enumerate(self.times):
+            self.demographic_events.append(
+                msprime.PopulationParametersChange(
+                    time=t, initial_size=self.sizes[idx], population_id=0))
+
+        self.migration_matrix = [[0]]
+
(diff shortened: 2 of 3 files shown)


stdpopsim/arabidopsis_thaliana.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class Durvasula2017MSMC(models.Model):
+    """
+    add info about model here
+
+    """
+
+    def __init__(self):
+       self.times = [699.1107142857144, 2796.4428571428575, 6067.585714285714,
```
andrewkern: hey @jeromekelleher are you cool with this style for hard coded arrays? looks like this passes flake8 linting

andrewkern:
@arundurvasula do you have a ballpark estimate for how long `athaliana_example.py` takes to run? the drosophila example takes a very long time to run as well so I've added a couple quick checks as well.

athaliana_example.py
```diff
@@ -0,0 +1,23 @@
...
+import stdpopsim
+from stdpopsim import arabidopsis_thaliana
+
+chrom = arabidopsis_thaliana.genome.chromosomes["chr5"]
+recomb_map = chrom.recombination_map()
+
+model = arabidopsis_thaliana.Durvasula2017MSMC()
+model.debug()
+
+samples = [msprime.Sample(population=0, time=0),msprime.Sample(population=0,
```
jeromekelleher: Usually the best way to break lines like this is to use hanging indents:
```python
samples = [
    msprime.Sample(population=0, time=0), msprime.Sample(population=0, time=0)]
```
But, style is subjective...

stdpopsim/arabidopsis_thaliana.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class Durvasula2017MSMC(models.Model):
+    """
+    Model estimated from two homozygous individuals from the South Middle Atlas
+    using MSMC
+    """
+
+    def __init__(self):
+        self.times = [
```
jeromekelleher: This might be more convenient as a numpy array, .e.g
```python
    self.times = np.array([...])
```

stdpopsim/arabidopsis_thaliana.py
```diff
@@ -0,0 +1,112 @@
...
+        self.sizes = [
+            42252426, 42252426, 60323, 72174, 40591, 21158, 21442,
+            39942, 78908, 111132, 110745, 96283, 87661, 83932, 83829, 91813,
+            111644, 143456, 181571, 217331, 241400, 246984, 238593, 228222,
+            217752, 198019, 165210, 121796, 121796, 73989, 73989, 73989]
+
+        # MSMC is accurate from 40Kya-1.6Mya for A.thaliana(Durvasula et al 2017)
+        # set the first 7 sizes
+        # equal to the size at 8 (~40Kya)
+        self.sizes[0:8] = [self.sizes[8]]*7
```
jeromekelleher: We can make this more readable usign the numpy array now:
```
self.sizes[:8] = self.sizes[8]
```
similarly for the other ones.

stdpopsim/arabidopsis_thaliana.py
jeromekelleher: If it passes flake8 and it's not likely to *not* pass flake8 in the future I'm fine with it.

jeromekelleher:
I understand the frustration with pep8, but it really is the only way to make sure that code is reasonably uniform and readable. Otherwise, everyone has different ideas on what constitutes good style and not, and we end up spending 1/2 of our valuable development time arguing over whether style nuance (a) is better than style nuance (b).

You can run flake8 locally to avoid having to find out during CI. I add this to my git pre-commit hook, which helps a *lot*:
```
exec flake8 [all pyhon files in repo] --max-line-length=89
```

Another alternative is for us to switch to automatically formatting the code using [black](https://black.readthedocs.io/en/stable/) which might be save time and energy...

jeromekelleher:
Also, would be nice (but not essential; don't worry if you're not familiar with rebasing) to squash the commits before merging (see quick description [here](https://github.com/popgensims/stdpopsim/pull/24#pullrequestreview-210600846))

arundurvasula:
Thanks for the quick replies @jeromekelleher and @andrewkern! Some responses below:

1. regarding the PEP8 stuff: I do think it's worth having since as Jerome says, the codebase can quickly devolve into a mess if we don't conform to something (and I am definitely aware that I can write some _bad_ python). In general I think this sort of thing leads to front-loading the annoying bits but saves time and headaches overall. Installing flake locally and rebasing seems like it'll help with this.

2. The model takes about a few hours to run on chromosome 5. @andrewkern what checks are you referring to? I'd be happy to add these to the `test_athaliana.py` file.

3. I can add in the changes @jeromekelleher suggested and do a rebase -- should be able to get to that later today.

Thanks again for the quick turnaround!

stdpopsim/arabidopsis_thaliana.py
```diff
@@ -0,0 +1,112 @@
...
+#
+###########################################################
+
+
+class Salome2012(genetic_maps.GeneticMap):
+    """
+    Genetic map from Salome 2012 averaged across population crosses.
+    Please see README for details on how this was done.
+    http://www.eeb.ucla.edu/Faculty/Lohmueller/data/uploads/
+        README_athaliana_map_salome2012.txt
```
jradrion: I'm getting a URL not found, but I'm assuming that this might just be an issue with permission to access UCLA servers.

jeromekelleher: I just tried it from an Oxford university address and it worked fine, but then tried it from my home IP and got a 
```
Failed to establish a new connection: [Errno -2] Name or service not known',
```
Looks like the UCLA server isn't known outside academic circles?

jeromekelleher: I just tried it from an Oxford university address and it worked fine, but then tried it from my home IP and got a 
```
Failed to establish a new connection: [Errno -2] Name or service not known',
```
Looks like the UCLA server isn't known outside academic circles?

jeromekelleher: Err nope, ignore that. Turned it was just my network wasn't on properly. I can download the map fine from my home IP.

andrewkern: @jradrion is this working for you now?

jeromekelleher:
FYI @arundurvasula, the proposed rebase docs over in #27 might come in useful here.

arundurvasula:
> FYI @arundurvasula, the proposed rebase docs over in #27 might come in useful here.

Thanks! Those docs were great and really well written. I've squashed the commits into a single one.

stdpopsim/arabidopsis_thaliana.py
arundurvasula: The most recent version is a little cleaner: https://github.com/popgensims/stdpopsim/pull/23/files#diff-579791a7ad461a14fdba75fc383c1ae7R85

arundurvasula:
I added a link to the scripts Chris used to make the recombination map. Should be good to go now.

stdpopsim/arabidopsis_thaliana.py
```diff
@@ -0,0 +1,113 @@
...
+#
+# Demographic models
+#
+###########################################################
+
+
+class Durvasula2017MSMC(models.Model):
+    """
+    Model estimated from two homozygous individuals from the South Middle Atlas
+    using MSMC
```
andrewkern: Maybe some more information here: how are the `.times` and `.sizes` related? E.g., is the effective population size in time interval `.times[k]` to `times[k+1]` equal to `.sizes[k]`?  With the last size covering `.times[n]` to infinity?

andrewkern: It looks from the code like the pop size is `.sizes[0]` on `[0, .times[0])` but also on `[times[0], times[1])`; is this a bug?

tests/test_athaliana.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_basic_attributes(self):
+        genome = arabidopsis_thaliana.genome
+        self.assertEqual(genome.species, "arabidopsis_thaliana")
+        self.assertEqual(genome.default_genetic_map, "Salome2012")
+        self.assertEqual(len(genome.chromosomes), 5)
+
+    def test_str(self):
+        s = str(arabidopsis_thaliana.genome)
+        self.assertGreater(len(s), 0)
+        self.assertIsInstance(s, str)
```
andrewkern: Is this doing anything (seeing as how `s` is the output of `str( )`)?

andrewkern: looks like this unit test was coming straight from the earlier examples. not sure we want the `.assertIsInstance()`

petrelharp:
This looks good to me pending resolution of the `.times` and `.sizes` issue @andrewkern pointed out.

andrewkern:
okay @arundurvasula if you can double check that that `.times` and `.sizes` thing isn't a bug we can merge.

stdpopsim/arabidopsis_thaliana.py
jradrion: Sorry, just getting back to looking at this now. No, it still doesn't work for me; must be something on my end. I have no issues accessing "http://www.eeb.ucla.edu/Faculty/Lohmueller/data/uploads/salome2012_maps.tar.gz", however.

tests/test_athaliana.py
jeromekelleher: This is testing the ``__str__`` implementation on the Genome class. I'll tidy these all up in another pass after we've got this round of models in.

stdpopsim/arabidopsis_thaliana.py
arundurvasula: Thanks Andy! Just added a comment about that. This behavior is because MSMC's output can produce inaccurate estimates in the very recent and very ancient past (e.g. http://www.g3journal.org/content/7/11/3605). Instead of having users simulate 40 million arabidopsis individuals in the most recent generations, we thought it best to use the population size from the first time bin we are confident in and extend that to the present (and likewise for the ancient time bins). So the first 8 time bins have the same population size and the last 3 do as well.

I'm happy to discuss this point further though -- it could be that what we did isn't what we want to have the default model doing. We discuss this point in the supplement of the 2017 paper (line 336) https://www.pnas.org/content/pnas/suppl/2017/05/03/1616736114.DCSupplemental/pnas.1616736114.sapp.pdf

PR #25, opened 2019-03-05
added ecoli model from Lapierre et al

fbaumdicker:
I added a simple constant model for the evolution of Escherichia coli as described in Lapierre et al, MBE, 2016. Lapierre also considers models with gene conversion and selection which is so far not included.
The evolutionary rates have been taken from Table 2 in Lapierre et al.
In the example file, a sample of size 100 is generated.
At the moment this takes less than a minute. This will increase when gene conversion is included.

### DIFF
--- a/ecoli_example.py
+++ b/ecoli_example.py
@@ -1,0 +1,23 @@
+"""
+Example of using the stdpopsim library with msprime.
+"""
+import msprime
+
+
+import stdpopsim
+from stdpopsim import e_coli
+
+chrom = e_coli.genome.chromosomes["chr"]
+#recomb_map = chrom.recombination_map()
+
+model = e_coli.LapierreConstant()
+model.debug()
+
+samples=[msprime.Sample(population=0, time=0)]*100
+
+ts = msprime.simulate(
+    samples=samples,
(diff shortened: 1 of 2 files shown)


fbaumdicker:
I started today and hope to have the python version running within the next 2-3 weeks.

fbaumdicker:
Here is the clickable link to the original publication, which I used to infer the parameters:

[(https://doi.org/10.1093/molbev/msw048)]

I also added the link to the e_coli.py file.

jeromekelleher:
Looks good, thanks @fbaumdicker. Can you [rebase and squash](https://stdpopsim.readthedocs.io/en/latest/development.html) please?

@andrewkern, my take would be to merge this and see how it fits in with the bigger picture later. What do you think?

andrewkern:
sounds good to me. @fbaumdicker after the rebase let's merge.

fbaumdicker:
Ok, I rebased and squashed. @jeromekelleher Hope this is fine? Next time I will also use a separate branch instead of master.

jeromekelleher:
Perfect, thanks @fbaumdicker!

PR #30, opened 2019-03-09
added qc/homosap

jgallowa07:
Not sure if this is the best place for quality checks - and I'm open to suggestions. I have started a Jupyter notebook to do some quality check on the Gutenkunst Three Pops out of Africa model. 

It's just a start, and I plan on doing more next week, but I figured it would be good to start the conversation and find a home for QC.


andrewkern:
seems totally reasonable to me. what do you think @jeromekelleher?

jeromekelleher:
Thanks @jgallowa07, this is really excellent! We probably want to do a bit of replication I'd imagine, so perhaps more, smaller simulations would be better than one large one? Looks like this is on the right track to me though.

Here's some nitpicky feedback:
- I don't think we should store the tree sequence file in this repo. Keeping derived binaries in git causes headaches, and it'll make the repo blow up quickly, and it can be resimulate quickly enough.
- Likewise, we definitely don't nee the checkpoints directory here.
- You're recomputing the genotype matrix several times in the [7] code chunk. Much more efficient to have ``genotype_matrix = ts.genotype_matrix()`` and reuse it.
- Let's keep all imports in the first cell; that way, at least you know straight away if you're missing some libraries.

jgallowa07:
Awesome, some quick follow up questions:
1. by more, smaller, sims do you think I should simply run more replicates of the same simulation w/ fewer samples? 
2. Should I be doing this over multiple chromosomes?

ndukler:
Chromosomes can basically be treated as replicates right?

jeromekelleher:
> by more, smaller, sims do you think I should simply run more replicates of the same simulation w/ fewer samples?

I was thinking more of shorter chromosomes; we probably don't need to simulate all of chr22 for this I'd imagine. Do we need an need an empirical recombination map here or should a (say) 1Mb chunk with the mean recombination rate do?

Simulation time grows quadratically with chromosome length, so unless we need whole chromosomes there's not much point in doing it. My 10c would be that lots small of replicates would be better then one big sim (but, definitely depends on application!)

jgallowa07:
@andrewkern and I talked about using the genome-wide (rather than just chrom22) average recombination rate and doing something like 500 replicates? does this sounds reasonable?

In this case might a genome-wide recombination rate be a nice attribute to the `genome` class?

jeromekelleher:
> @andrewkern and I talked about using the genome-wide (rather than just chrom22) average 
 recombination rate and doing something like 500 replicates? does this sounds reasonable?

Yep, sounds perfect.

> In this case might a genome-wide recombination rate be a nice attribute to the genome class?

Excellent idea! Could you open an issue for this please? Presumably we can calculate this directly from the chromosomes and won't need to store it.

andrewkern:
simple question before we go further-- are folks happy with the placement of this dir in the project? if so I am tempted to merge it and we can work on details moving forward.

jeromekelleher:
> simple question before we go further-- are folks happy with the placement of this dir in the project? if so I am tempted to merge it and we can work on details moving forward.

Fine by me.

jeromekelleher:
Eek, I think we've just added 20MB to the repo here as we didn't rebase after removing the big trees file, which is really slowing it down. 

Anyone object if I do some quick git surgery before we go any further to squash these commits?

jgallowa07:
Please do. I didn't see this was happening until just now.

jeromekelleher:
OK, done. Sorry, I should have been clear that rebasing would be needed to completely get rid of big .trees file from the repo history.

Can you take a quick look at the notebook to see if it went OK please?

PR #33, opened 2019-03-12
flat (0) maps for non-existing chrom maps

jgallowa07:
Simple fix for defining a recombination map across all chromosomes on a species genome.

### DIFF
--- a/stdpopsim/genetic_maps.py
+++ b/stdpopsim/genetic_maps.py
@@ -169,6 +169,14 @@ class GeneticMap(object):
             logger.info("Storing map in {}".format(self.map_cache_dir))
             shutil.move(extract_dir, self.map_cache_dir)
 
+    def contains_chromosome_map(self, name):
+        """
+        Just a quick check to see whether of not
+        this genetic map contains a genetic map for `name`
+        """
+        map_file = os.path.join(self.map_cache_dir, self.file_pattern.format(name=name))
+        return os.path.exists(map_file)
+
     def get_chromosome_map(self, name):
         """
         Returns the genetic map for the chromosome with the specified name.
@@ -176,5 +184,7 @@ class GeneticMap(object):
         # TODO look up a list of known names to give a good error message.
         if not self.is_cached():
             self.download()
+        if not self.contains_chromosome_map(name=name):
+            raise ValueError("Chromosome map for '{}' not found".format(name))
         map_file = os.path.join(self.map_cache_dir, self.file_pattern.format(name=name))
         return msprime.RecombinationMap.read_hapmap(map_file)
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -4,6 +4,8 @@ species.
 """
 
 import stdpopsim.genetic_maps as genetic_maps
+import msprime
+import warnings
 
 
 class Genome(object):
@@ -65,4 +67,12 @@ class Chromosome(object):
         if map_name is None:
             map_name = self.default_genetic_map
         genetic_map = genetic_maps.get_genetic_map(self.species, map_name)
-        return genetic_map.get_chromosome_map(self.name)
+        if genetic_map.contains_chromosome_map(self.name):
+            ret = genetic_map.get_chromosome_map(self.name)
+        else:
+            warnings.warn(
+                "Warning: recombination map not found for chromosome: '{}'"
+                " on map: '{}', substituting a zero"
+                "-recombination map.".format(self.name, map_name))
+            ret = msprime.RecombinationMap.uniform_map(self.length, 0)
+        return ret
--- a/tests/test_homo_sapiens.py
+++ b/tests/test_homo_sapiens.py
@@ -7,6 +7,7 @@ import io
 import msprime
 
 from stdpopsim import homo_sapiens
+from stdpopsim import genetic_maps
 
 
 class TestGenome(unittest.TestCase):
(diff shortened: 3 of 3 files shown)


jgallowa07:
oops, looks like I'm not passing some tests. And I think I need to rebase ...

stdpopsim/genetic_maps.py
```diff
@@ -177,4 +177,9 @@ def get_chromosome_map(self, name):
         if not self.is_cached():
             self.download()
         map_file = os.path.join(self.map_cache_dir, self.file_pattern.format(name=name))
-        return msprime.RecombinationMap.read_hapmap(map_file)
+        if(os.path.exists(map_file)):
```
jeromekelleher: Does msprime not already raise a FileNotFound error? Or is a better message you're looking for? This is probably a ValueError I'd say:
```python
if not os.path.exists(map_file):
    raise ValueError("Chromosome map for '{}' not found".format(name)
 return msprime.RecombinationMap.read_hapmap(map_file)
```

stdpopsim/genomes.py
```diff
@@ -65,4 +66,8 @@ def recombination_map(self, map_name=None):
         if map_name is None:
             map_name = self.default_genetic_map
         genetic_map = genetic_maps.get_genetic_map(self.species, map_name)
-        return genetic_map.get_chromosome_map(self.name)
+        try:
```
jeromekelleher: OK, I see what you're doing here. We need to be careful about masking other errors here, so I think we should add a method to query whether the chromosome map is defined and do it based on that. This would be more readable as well, and explain what's happening more clearly to the reader.

jeromekelleher: Also, we're not checking to see if it's a chromosome that we know about in the genome. We need to check the chromosome names to see if the are good, then see if there is a map associated with it. If there is no map, then return a flat 0 map. This will just return a flat map for any old key at all, which is definitely a bad idea.

tests/test_homo_sapiens.py
```diff
@@ -53,6 +53,12 @@ def test_chromosome_lengths(self):
         self.assertEqual(genome.chromosomes["chrX"].length, 156040895)
         self.assertEqual(genome.chromosomes["chrY"].length, 57227415)
 
+    def test_chromosome_recombination_maps(self):
+        for chrom in homo_sapiens.genome.chromosomes:
+            chrom = homo_sapiens.genome.chromosomes[chrom]
+            cm = chrom.recombination_map()
+            assert(type(cm) == msprime.simulations.RecombinationMap)
```
jeromekelleher: We're using unittest here, so we need to use the built int assert methods, e.g.
```
self.assertIsInstance(cm, msprime.RecombinationMap)
```

Also, we need to test the error cases here. What happens when you don't know the key?

stdpopsim/genomes.py
```diff
@@ -65,4 +66,8 @@ def recombination_map(self, map_name=None):
...
+            ret = genetic_map.get_chromosome_map(self.name)
+        except FileNotFoundError:
+            ret = msprime.RecombinationMap.uniform_map(self.length, 0)
```
petrelharp: I think this should produce a warning, because there's any number of things that could go wrong, so we don't want to silently substitute a zero-recomb map. Like: "Warning: recombination map not found for chromosome {}, substituting a zero-recombination map."

jeromekelleher: There's two things here:
1. If it's a string that doesn't correspond to any know chromosome (name="blahblah") then it should raise a ValueError
2. If it is a known chromosome and we don't have a map, then we insert a zero map.

Should we still raise a warning in case 2? I guess it's safer, in case someone did something like deleting all the recombination map files but didn't remove the map directory.

jgallowa07: OK, to sum up, it seems I should:

1. add a method, say, `contains_chromosome_map(self, name)` to `GeneticMap` that returns a boolean flag indicative of whether or not it contains a recombination_map for a certain chromosome.

2. In `Chromosome.recombination_map(self, mape_name=None)` we first check that a map exists,
if so, return the recombination_map
if not, raise a warning & return the flat map

3. in `GeneticMap.get_chromsome_map(self, name)` 
We check if the name exists, and throw ValueError if not.

tests/test_homo_sapiens.py
jgallowa07: got it, will finish after I make a few changes discussed below

jgallowa07: so here are you talking about what we should do when we don't know the key of a specific chromosome? like if a user were to do
`chrom = homo_sapiens.genome.chromosomes["blahblah"]`

andrewkern: the code should quit with a KeyError

stdpopsim/genomes.py
petrelharp: I think we should raise a warning if we default to zero recombination.

It seems like for step (1) we need a way of recording/standardizing what the legal chromosome names are (grrr, "chr1" versus "chr01" versus "1").

jeromekelleher: We've touched on the idea of synonyms in another issue @petrelharp (https://github.com/popgensims/stdpopsim/issues/16), so this can be handled as part of that.

stdpopsim/genetic_maps.py
```diff
@@ -169,12 +169,24 @@ def download(self):
             logger.info("Storing map in {}".format(self.map_cache_dir))
             shutil.move(extract_dir, self.map_cache_dir)
 
+    def contains_chromosome_map(self, name):
+        """
+        Just a quick check to see whether of not
+        this genetic map contains a genetic map for `name`
+        """
+        map_file = os.path.join(self.map_cache_dir, self.file_pattern.format(name=name))
+        if not os.path.exists(map_file):
```
jeromekelleher: ``return os.path.exists(map_file)`` is more concise and clearer

Also, the doc string comment should describe what the method does: Returns True if this genetic map contains a chromosome with the specified name.

stdpopsim/genomes.py
```diff
@@ -65,4 +67,12 @@ def recombination_map(self, map_name=None):
         if map_name is None:
             map_name = self.default_genetic_map
         genetic_map = genetic_maps.get_genetic_map(self.species, map_name)
-        return genetic_map.get_chromosome_map(self.name)
+        try:
```
jeromekelleher: I think it would be clearer if this explicitly tested to see if the map exists:
```
if genetic_map.contains_chromosome_map(self.name):
    # same as before
else:
    warnings.warn()
```

Otherwise, I wonder *why* is this raising a ValueError, and why am I defaulting to a 0 map if it's raised?

tests/test_homo_sapiens.py
```diff
@@ -7,9 +7,11 @@
 import msprime
 
 from stdpopsim import homo_sapiens
+from stdpopsim import genetic_maps
 
 
 class TestGenome(unittest.TestCase):
+    # TODO Should this have it's own file?
```
jeromekelleher: No, there's no need to keep different unit test classes in different files.

tests/test_homo_sapiens.py
```diff
@@ -53,6 +55,40 @@ def test_chromosome_lengths(self):
...
+            cm = chrom.recombination_map()
+
+    def test_chromosome_errors(self):
+        # Assert that unknown chromosomes throw a KeyError
+        with self.assertRaises(KeyError):
+            homo_sapiens.genome.chromosomes["jibberish"]
+
+
+class TestGeneticMap(unittest.TestCase):
+    # TODO Should this have it's own file?
```
jeromekelleher: Same comment as above

tests/test_homo_sapiens.py
```diff
@@ -53,6 +55,40 @@ def test_chromosome_lengths(self):
         self.assertEqual(genome.chromosomes["chrX"].length, 156040895)
         self.assertEqual(genome.chromosomes["chrY"].length, 57227415)
 
+    def test_chromosome_recombination_maps(self):
+
+        with self.assertWarns(Warning):
```
jeromekelleher: This assertWarns is pretty scattergun, and it's not at all obvious why it should raise a warning. You need to make it locally obvious what code is doing, a reader shouldn't need to keep the whole codebase in their head to understand a local function.

Really you want  to split this into two unit tests, one where we're testing chromosomes where we know we have maps and another where we know we don't.

jgallowa07: Agreed, I think here I wanted to make sure that every `chrom in homo_sapiens.genome.chromosomes:` could call `chrom.recombination_map()`  and get an instance of `msprime recombination_map` with nothing *more* than a warning. That is the desired behavior?

I'll break the tests up to make it a little more concise.

jeromekelleher: OK, cool. Ping me when it's ready to look at again.

jgallowa07: @jeromekelleher ping

jeromekelleher: Great! OK, looks good to go for me. Thanks for the updates.

jgallowa07:
@petrelharp @andrewkern does this look good to you?

andrewkern:
looks good @jgallowa07!

PR #34, opened 2019-03-13
Added mean recombination rates to chromosome & genome

jgallowa07:
In addition, I changed out the defined chromosome lengths with those found here : 
`ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/working/20110106_recombination_hotspots/README_hapmapII_GRCh37_map`

### DIFF
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -18,10 +18,12 @@ class Genome(object):
         self.species = species
         self.default_genetic_map = default_genetic_map
         self.chromosomes = {}
+        self.length = 0
         for chromosome in chromosomes:
             self.chromosomes[chromosome.name] = chromosome
             chromosome.default_genetic_map = default_genetic_map
             chromosome.species = species
+            self.length += chromosome.length
 
     def __str__(self):
         s = "Genome for {}:\n".format(self.species)
@@ -31,6 +33,20 @@ class Genome(object):
             s += "\t{}\n".format(chrom)
         return s
 
+    @property
+    def mean_recombination_rate(self):
+        """
+        This method return the weighted mean recombination rate
+        across all chomosomes in the genome.
+        :rtype: float
+        """
+        mean_recombination_rate = 0
+        for chrom in self.chromosomes.values():
+            normalized_weight = chrom.length / self.length
+            cont = chrom.default_recombination_rate*normalized_weight
+            mean_recombination_rate += cont
+        return mean_recombination_rate
+
 
 class Chromosome(object):
     """
@@ -38,21 +54,21 @@ class Chromosome(object):
 
     .. todo:: Define the facilities that this object provides.
     """
-    def __init__(self, name, length, mean_recombination_rate, mean_mutation_rate):
+    def __init__(self, name, length, default_recombination_rate, default_mutation_rate):
         self.name = name
         self.length = length
-        self.mean_recombination_rate = mean_recombination_rate
-        self.mean_mutation_rate = mean_mutation_rate
+        self.default_recombination_rate = default_recombination_rate
+        self.default_mutation_rate = default_mutation_rate
         self.species = None
         self.default_genetic_map = None
 
     def __repr__(self):
         return (
             "{{'name': {}, 'length': {}, "
-            "'mean_recombination_rate': {}, "
-            "'mean_mutation_rate': {}}}".format(
-                self.name, self.length, self.mean_recombination_rate,
-                self.mean_mutation_rate))
+            "'default_recombination_rate': {}, "
+            "'default_mutation_rate': {}}}".format(
+                self.name, self.length, self.default_recombination_rate,
+                self.default_mutation_rate))
 
     def __str__(self):
         return repr(self)
--- a/stdpopsim/homo_sapiens.py
+++ b/stdpopsim/homo_sapiens.py
@@ -40,48 +40,53 @@ genetic_maps.register_genetic_map(HapmapII_GRCh37())
 #
 ###########################################################
 
-# List of chromosomes. Data for length information based on GRCh38,
-# https://www.ncbi.nlm.nih.gov/grc/human/data
+# List of chromosomes.
 
-# FIXME: add mean mutation and recombination rate data to this table.
+# FIXME: add mean mutation rate data to this table.
+# Name  Length  mean_recombination_rate mean_mutation_rate
+
+# length information can be found here
+# <http://hgdownload.soe.ucsc.edu/goldenPath/hg19/database/chromInfo.txt.gz>
+
+# mean_recombination_rate was computed across all windows of the GRCh37 genetic map
+# <ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/working/20110106_recombination_hotspots>
 _chromosome_data = """\
-chr1   248956422
-chr2   242193529
-chr3   198295559
-chr4   190214555
-chr5   181538259
-chr6   170805979
-chr7   159345973
-chr8   145138636
-chr9   138394717
-chr10  133797422
-chr11  135086622
-chr12  133275309
-chr13  114364328
-chr14  107043718
-chr15  101991189
-chr16  90338345
-chr17  83257441
-chr18  80373285
-chr19  58617616
-chr20  64444167
-chr21  46709983
-chr22  50818468
-chrX   156040895
-chrY   57227415
+chr1 	 249250621 	 1.1485597641285933e-08
+chr2 	 243199373 	 1.1054289277533446e-08
+chr3 	 198022430 	 1.1279585624662551e-08
+chr4 	 191154276 	 1.1231162636001008e-08
+chr5 	 180915260 	 1.1280936570022824e-08
+chr6 	 171115067 	 1.1222852661225285e-08
+chr7 	 159138663 	 1.1764614397655721e-08
+chr8 	 146364022 	 1.1478465778920576e-08
+chr9 	 141213431 	 1.1780701596308656e-08
+chr10 	 135534747 	 1.3365134257075317e-08
+chr11 	 135006516 	 1.1719334320833283e-08
+chr12 	 133851895 	 1.305017186986983e-08
+chr13 	 115169878 	 1.0914860554958317e-08
+chr14 	 107349540 	 1.119730771394731e-08
+chr15 	 102531392 	 1.3835785893339787e-08
+chr16 	 90354753 	 1.4834607113882717e-08
+chr17 	 81195210 	 1.582489036239487e-08
+chr18 	 78077248 	 1.5075956950023575e-08
+chr19 	 59128983 	 1.8220141872466202e-08
+chr20 	 63025520 	 1.7178269031631664e-08
+chr21 	 48129895 	 1.3045214034879191e-08
+chr22 	 51304566 	 1.4445022767788226e-08
+chrX 	 155270560 	 1.164662223273842e-08
+chrY 	 59373566 	 0.0
 """
 
 _chromosomes = []
 for line in _chromosome_data.splitlines():
-    name, length = line.split()[:2]
+    name, length, mean_rr = line.split()[:3]
     _chromosomes.append(genomes.Chromosome(
         name=name, length=int(length),
-        mean_mutation_rate=1e-8,  # WRONG!,
-        mean_recombination_rate=1e-8))  # WRONG!
+        default_mutation_rate=1e-8,  # WRONG!,
+        default_recombination_rate=float(mean_rr)))
 
 
-#: :class:`stdpopsim.Genome` definition for humans. Chromosome length data is
-#: based on `GRCh38 <https://www.ncbi.nlm.nih.gov/grc/human/data>`_.
+#: :class:`stdpopsim.Genome` definition for humans.
 genome = genomes.Genome(
     species="homo_sapiens",
     chromosomes=_chromosomes,
--- a/tests/test_homo_sapiens.py
+++ b/tests/test_homo_sapiens.py
@@ -5,6 +5,7 @@ import unittest
 import io
 
 import msprime
+import numpy as np
 
 from stdpopsim import homo_sapiens
 from stdpopsim import genetic_maps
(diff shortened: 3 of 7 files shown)


jgallowa07:
@jeromekelleher I can't seem to get the codecov from my test, any thoughts?

stdpopsim/homo_sapiens.py
```diff
@@ -43,41 +43,41 @@ class HapmapII_GRCh37(genetic_maps.GeneticMap):
 # List of chromosomes. Data for length information based on GRCh38,
 # https://www.ncbi.nlm.nih.gov/grc/human/data
 
-# FIXME: add mean mutation and recombination rate data to this table.
+# FIXME: add mean mutation rate data to this table.
 _chromosome_data = """\
-chr1   248956422
```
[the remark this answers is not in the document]
jeromekelleher: Also, where does the recomb rate info come from? This needs to be clearly documented.

tests/test_homo_sapiens.py
```diff
@@ -26,33 +26,42 @@ def test_str(self):
...
+        self.assertEqual(genome.chromosomes["chr15"].length, 102506108)
+        self.assertEqual(genome.chromosomes["chr16"].length, 90163588)
+        self.assertEqual(genome.chromosomes["chr17"].length, 81045269)
+        self.assertEqual(genome.chromosomes["chr18"].length, 78015180)
+        self.assertEqual(genome.chromosomes["chr19"].length, 59097160)
+        self.assertEqual(genome.chromosomes["chr20"].length, 62949445)
+        self.assertEqual(genome.chromosomes["chr21"].length, 48100155)
+        self.assertEqual(genome.chromosomes["chr22"].length, 51229805)
+        self.assertEqual(genome.chromosomes["chrX"].length, 154929412)
         self.assertEqual(genome.chromosomes["chrY"].length, 57227415)
```
jeromekelleher: Similar tests for recombination rate?

tests/test_homo_sapiens.py
```diff
@@ -26,33 +26,42 @@ def test_str(self):
...
 
+    def test_mean_recombination_rate(self):
+        genome = homo_sapiens.genome
+        highest_rr = 0
+        lowest_rr = 10000
+        for chrom, chrom0 in genome.chromosomes.items():
```
jeromekelleher: What is this testing?

tests/test_homo_sapiens.py
```diff
@@ -26,33 +26,42 @@ def test_str(self):
...
+            rr = chrom0.mean_recombination_rate
+            lowest_rr = min(lowest_rr, rr)
+            highest_rr = max(highest_rr, rr)
+        mean_genome_rr = genome.mean_recombination_rate
+        assert((mean_genome_rr < highest_rr) & (mean_genome_rr > lowest_rr))
```
jeromekelleher: need to use assert methods here rather than assert statements as we're using nose.

stdpopsim/homo_sapiens.py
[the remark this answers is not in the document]
jgallowa07: The lengths were changed to match  `GRCh37` as requested by @andrewkern. 

The recombination rate was computed explicitly from the data; I'll document now if you're okay with it sticking around ... should we be getting average recombination rate from somewhere else?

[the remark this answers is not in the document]
jeromekelleher: > The recombination rate was computed explicitly from the data; I'll document now if you're okay with it sticking around ... should we be getting average recombination rate from somewhere else?

From what data though? It's great that we're adding this stuff, but we need to be really explicit about where the numbers are coming from.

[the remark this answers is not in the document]
andrewkern: @jgallowa07 did you end up taking these? http://hgdownload.soe.ucsc.edu/goldenPath/hg19/database/chromInfo.txt.gz

@jeromekelleher these are the chromosome lengths for GRCh37 aka hg19.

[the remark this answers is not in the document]
jeromekelleher: Sure, sounds good to me @andrewkern, I just want to make certain there's a clear provenance trail (we'll want to incorporate links in the documentation also).

tests/test_homo_sapiens.py
jgallowa07: It just tests that the mean lies with the lowest and greatest value. It was the first sanity check I could think of.

open to suggestions for a better test.

tests/test_homo_sapiens.py
jgallowa07: gotcha, I'll do that now

tests/test_homo_sapiens.py
jeromekelleher: Recompute the mean using numpy and see if it's close?

stdpopsim/homo_sapiens.py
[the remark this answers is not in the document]
jgallowa07: Sorry, mine was the max position found in the genetic map,

ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/working/20110106_recombination_hotspots/

 But I'll use the physical one,

http://hgdownload.soe.ucsc.edu/goldenPath/hg19/database/chromInfo.txt.gz

And be sure to note where it is from in the documentation.
I'll fix this now.

tests/test_homo_sapiens.py
jgallowa07: will do.

tests/test_homo_sapiens.py
jgallowa07: done

vsbuffalo: Sorry for my ignorance here, but why are such tests needed? Aren't you're just as likely to accidentally change a hard coded value in a test like this as you are the underlying file?

jgallowa07: I was wondering that as well tbh. maybe because we want to make sure no added code accidentally modifies these values. In other words, we always want these values to be fixed and the tests will make sure they are.

jeromekelleher: Yeah, it's arguable. You have to get it wrong the same way twice with these tests though. Personally I'm a bit dyslexic, so I don't trust myself to get this sort of thing right.

tests/test_homo_sapiens.py
jgallowa07: I've got the code for this but I'm going to hold off on pushing it until #33 is pulled so I can iterate through chromosomes and get an `msprime` recombination_map for each one.

stdpopsim/genomes.py
```diff
@@ -29,6 +31,20 @@ def __str__(self):
             s += "\t{}\n".format(chrom)
         return s
 
+    @property
+    def mean_recombination_rate(self):
+        """
+        This method return the weighted mean recombination rate
+        across all chomosomes in the genome.
+        :return: float
```
jeromekelleher: This should probably be ``:rtype:``.

stdpopsim/genomes.py
```diff
@@ -29,6 +31,20 @@ def __str__(self):
...
+        """
+        mean_recombination_rate = 0
+        for chrom, chrom0 in self.chromosomes.items():
```
jeromekelleher: This should be ``for chrom in self.chromosomes.values()``

andrewkern:
was just talking to @petrelharp about this PR. We have some additional changes that we would like to see, in particular making mean_recombination_rate a property of chromosomes rather than the genome as currently written

petrelharp:
Here's what I think should happen here. Basically, `mean_recombination_rate` should be derived from a `GeneticMap`, rather than being an explicit property set at creation of a `Chromosome`. Otherwise, there's conflicting information - someone might use a genetic map for simulation, and then think that the `.mean_recombination_rate` goes along with that. (Similarly, maybe `.length` should always derive from the recombination map as well?)

To do that, I'd propose that we:

-  remove `mean_recombination_rate` from [`Chromosome.__init__()`](https://github.com/popgensims/stdpopsim/blob/003a5cbc47c6287ca1aaecfe8056f0c75569ad16/stdpopsim/genomes.py#L39)

- make a `.mean_recombination_rate` method/`@property` method for [`GeneticMap`](https://github.com/popgensims/stdpopsim/blob/003a5cbc47c6287ca1aaecfe8056f0c75569ad16/stdpopsim/genetic_maps.py#L73) that divides the total genetic length by the total physical length for that chromosome

- make a `.mean_recombination_rate` method/`@property` for `Chromosome` also, that takes a genetic map name as an argument, to make it easy to get this for a chromsome

- instead of passing in mean recombination rates at initialization (as [here](https://github.com/popgensims/stdpopsim/pull/34/commits/4db96abf8f3bc65312f76c394ab9d684d8b6be63#diff-1b018f8c54343a41e051a8f902702c11R53)) to be a property of the Chromsome, use these to define and register a uniform genetic map with this rate.

Thoughts?

jgallowa07:
@petrelharp This makes sense to me ... 

The only thing that might be confusing about conceptually binding chromosomal (and hence, genomic) recombination rates to genetic maps is that every other major defining aspects of the chromosome (names, lengths & mutation rates) are presumably going to be hard coded. - meaning the user might have a tendency towards wanting a (baseline) definition as well. 

Also, the maps require that you have internet access to download them

So, the alternative is that it's nice to have a fixed simple definition (including mean recombination rate) of each chromosome that works offline - and then if they want the mean recombination rate from a map we could provide a function in `genetic_maps` that takes an msprime `recombination_map` and computes the mean (or maybe this is even a good method for `msprime` to have).

I'm indifferent, and happy to impliment whatever. Just my two cents.

petrelharp:
That's a good point, @jgallowa07 - in that case, I propose renaming `mean_recombination_rate` to `default_recombination_rate`.

andrewkern:
i agree with @petrelharp's suggestion of renaming, but just want to point out that recombination rates are *explicitly* a function of a genetic map so that it makes sense they should be tied there.

jgallowa07:
how's this sound @jeromekelleher? 

If we go with my comment above, the renaming should be a trivial change to the code at the latest commit of this PR: 

However, I just re-did the calculations for mean recombination rate (using chromosome.length to calculate weights)  and got different numbers than those coded in now.

For QC, I'd like to show you all how I've calculated the weighted mean recombination rate (in two separate ways), and maybe @jeromekelleher would like to put some of these in `msprime` as a method of `recombination_map`. Or, if any of you have suggestions about where these functions should(/already do) live. 

First, we have a brute force approach to computing the "weighted mean of per-base recombination rate given a `genetic_maps.Chromosome` using `numpy`
```
def get_mean_rr_numpy(chromosome):
    chrom_length = chromosome.length
    recombination_map = chromosome.recombination_map()    

    positions = np.array(recombination_map.get_positions())
    positions_diff = recombination_map.get_positions()[1:]
    positions_diff.append(chrom_length)
    positions_diff = np.array(positions_diff)
    
    window_sizes = positions_diff - positions
    
    weights = window_sizes / chrom_length
    rates = recombination_map.get_rates()

    return np.average(rates,weights=weights)
```

Next, by suggestion of @petrelharp, A nice little trick where you can get mean recombination rate by dividing total genetic length by total physical length.

step 1: calculate total genetic distance in cM
```
def get_total_genetic_distance(chromosome):
    rec_map = chromosome.recombination_map()
    positions = rec_map.get_positions()
    rates = rec_map.get_rates()
    cm = 0.0
    for i in range(1,len(positions)):
        num_bp = positions[i] - positions[i-1]
       #the rates are alread in Morgans/bp. so mutiply by the number of bases in the window, 
       #mutiply by 100 
        cm += (rates[i-1] * num_bp) * 100
    return cm
```
Then simply
```
def get_mean_rr_genetic_distance(chromosome):
    total_genetic_distance_cm = get_total_genetic_distance(chromosome)
    return (total_genetic_distance_cm/chromosome.length) * 0.01   
```
Both have similar timing. -- The `get_total_genetic_distance` function is easily generalized to map all physical positions to a position in cM. I looked for a way to do this in msprime but the `RecombinationMap.physical_to_genetic(self, physical_x)` function was confusing me a bit.  ¯\_(ツ)_/¯

It would be great if I could get some of you to check these for correctness before I use the numbers produced by them in this PR. Thanks! (and sorry for the legthy comment on a friday afternoon, no rush)

jeromekelleher:
Calculation method looks good to me @jgallowa07, and this is something we should add to msprime (would you mind opening an issue to flag this please?).

jgallowa07:
Will do @jeromekelleher. 

I'll finish the changes to this pull request now.

tests/test_homo_sapiens.py
```diff
@@ -27,36 +28,77 @@ def test_str(self):
...
+        """
+        # We catch the warning that will be thrown when we iterate over
+        # Y chromosome.
+        with self.assertWarns(Warning):
+            for chrom in homo_sapiens.genome.chromosomes.values():
+                default_rr = chrom.default_recombination_rate
+                numpy_rr = self.get_mean_rr_numpy(chrom)
+                # Assert that the difference in mean recombination rate
+                # is small when computed with numpy.
+                self.assertGreater(1e-11, abs(numpy_rr-default_rr))
```
jeromekelleher: Use ``np.allclose`` here, comparing floating point numbers for approximate equality is subtle and tricky.

tests/test_homo_sapiens.py
```diff
@@ -27,36 +28,77 @@ def test_str(self):
...
+        positions_diff = recombination_map.get_positions()[1:]
+        positions_diff.append(chrom_length)
+        positions_diff = np.array(positions_diff)
+        window_sizes = positions_diff - positions
+        weights = window_sizes / chrom_length
+        rates = recombination_map.get_rates()
+        return np.average(rates, weights=weights)
+
+    def test_default_recombination_rates(self):
+        """
```
jeromekelleher: Test functions are the one case where it's better to **not** use docstrings. This is because nose will write out the docstring rather than the test name in the log, making it more difficult to find tests. You can use single-line-# comments instead.

tests/test_homo_sapiens.py
jgallowa07: Oh, interesting, totally makes sense. will do.

tests/test_homo_sapiens.py
jgallowa07: Noted.

jgallowa07: @jeromekelleher I made the changes on this branch then did a 
```
git commit -a --amend -m "updated tests"
```

Now it doesnt want me to push 
```
MacDre:stdpopsim jaredgalloway$ git push origin mean_genome_recombination_rate
To https://github.com/jgallowa07/stdpopsim.git
 ! [rejected]        mean_genome_recombination_rate -> mean_genome_recombination_rate (non-fast-forward)
error: failed to push some refs to 'https://github.com/jgallowa07/stdpopsim.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```
should I just try to force push this? I'm finding mixed answers to this question on stack overflow.

jgallowa07:
@jeromekelleher amended and pushed

jeromekelleher:
Thanks @jgallowa07, looking good. I'll let someone else do the merging (perhaps @petrelharp or @andrewkern).

By-the-by, I'm not crazy about ``default_recombination_rate`` etc, this seems a bit weird to me. I think we need to refactor and clarify the Genome/Chromosome stuff anyway though, so I'll come up with a concrete proposal before going any further with it.

petrelharp:
LGTM. Thanks, @jgallowa07!

PR #40, opened 2019-03-18
Fix race condition in map cache.

### LINKED ISSUE #38: Renaming does not work properly if several threads are downloading maps concurrently.
The code for figuring out when to download a genetic map depends on checking for the corresponding files. So does substituting default zero maps, as introduced in #33. This means we are always returning zero maps at the moment. 

(HT to @petrelharp here; I wouldn't have spotted this except for your warning!)

jeromekelleher:
Closes #38.

WIP. This should resolve the bug, but I need to figure out how to test it (plus test the map downloading code better also).

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -8,4 +8,3 @@ sphinx-argparse
 sphinx_rtd_theme
 msprime
 appdirs
-requests
--- a/requirements/rtd.txt
+++ b/requirements/rtd.txt
@@ -1,4 +1,3 @@
 # Requirements needed when building on readthedocs.
 setuptools_scm
 appdirs
-requests
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     keywords='simulations, recombination map, models',
     packages=['stdpopsim'],
     include_package_data=True,
-    install_requires=["msprime", "requests", "appdirs"],
+    install_requires=["msprime", "appdirs"],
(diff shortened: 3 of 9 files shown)


andrewkern:
i notice that CI is bitching here. is this okay @jeromekelleher?

jeromekelleher:
Yeah, codecov is complaining because I haven't written any tests. I need to figure out how to test this, so the PR is still a WIP. It won't make any difference upstream anyway, the extra rule in Snakemake will prevent the race condition happening.

jeromekelleher:
This is still a WIP --- trying to get a decent way of testing map downloads with out mucking up the user cache dir, and having to re-download the maps every time.

jeromekelleher:
OK, I think this is ready for review. A couple of things have ended up in here:

1. Decoupled the cache handling from genetic maps. This allows us to specify a fixed cache directory, not just ~/.cache/stdpopsim which will be useful in the analysis repo (i.e., we can have a proper target now to represent the downloaded genetic maps).
2. Switched from requests to the builtin urlretrieve. This is mainly because urlretrieve allows us to download from a file URL, which makes testing a whole pile easier.
3. A fix for the original race condition.

jeromekelleher:
OK, thanks @jgallowa07. I'm going to merge this --- the changes are all improvements I think and there's other stuff I want to do that depends on it.

PR #41, opened 2019-03-25
add Fu 2013 2-pop model

carjed:
Added the model from [Fu et al., 2013](https://dx.doi.org/10.1038%2Fnature11690). AFAIK this is identical to the Tennessen 2012 model, but I've included the African growth starting @5.1kya and AFR<-->EUR migration parameters.

### DIFF
--- a/stdpopsim/homo_sapiens.py
+++ b/stdpopsim/homo_sapiens.py
@@ -168,44 +168,77 @@ class GutenkunstThreePopOutOfAfrica(models.Model):
         ]
 
 
-class TennessenEuropean(models.Model):
+class TennessenTwoPopOutOfAfrica(models.Model):
     """
     The model is derived from the Tennesen et al.
     `analysis <https://doi.org/10.1126/science.1219240>`_  of the jSFS from
     European Americans and African Americans.
+
+    Model parameters are taken from Fig. S5 in
+    `Fu et al. (2013) <https://doi.org/10.1038/nature11690>`_.
 
     .. todo:: document this model, including the original publications
         and clear information about what the different population indexes
         mean.
 
     """
+
     def __init__(self):
         super().__init__()
-        # Population sizes
+
+        generation_time = 25
+        T_AF = 148e3 / generation_time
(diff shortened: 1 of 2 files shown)


andrewkern:
Nice! I’ll get some reviewers assigned to this PR.

carjed:
I initially thought the Fu et al. model was an extension of the model published in Tennessen et al., but looking at the supplements for the two papers (fig. S5 in both), they are in fact identical. Since Tennessen et al. was published first, would it make more sense to just delete the `TennessenEuropean()` model and rename `FuTwoPopOutOfAfrica()` as `TennessenTwoPopOutOfAfrica()`?

andrewkern:
So the `TennessenEuropean()` model as written should be generalized to handle samples from each of the two populations-- the only thing keeping from that at this point is the `PopulationConfiguration` which defines samples from the Euro popn only. Your implementation @carjed provides a good sanity check on our original implementation-- I haven't looked closely, but are the parameters equivalent between your implementation and what is already in the repo? 

As for renaming, my vote would be for you to just create a new PR where you just change the name of the existing Tennessen model as well as the default `PopulationConfiguration`. It would make for a cleaner commit history. Thoughts @jeromekelleher @petrelharp ? We can talk about this on the call too.

carjed:
`TennessenEuropean()` is missing parameters for the migration rates and recent growth in Africans, so those need to be updated too, not just the `PopulationConfiguration`. 

Otherwise, all of the parameters specified in `TennessenEuropean()` should be identical to what is in my implementation (after a couple of bugfixes coming in the next commit).

jeromekelleher:
> As for renaming, my vote would be for you to just create a new PR where you just change the name of the existing Tennessen model as well as the default PopulationConfiguration. It would make for a cleaner commit history. Thoughts @jeromekelleher @petrelharp ? We can talk about this on the call too.

I'm agnostic here on PRs. Once there's a single commit doing the renaming I think the history should be pretty clear.

carjed:
Does `TennessenTwoPopOutOfAfrica()` sound ok for the renamed class?

PR #42, opened 2019-03-26
Interactive notebooks on Binder 

carjed:
Thought it would be cool to make this available on [Binder](https://mybinder.org/) so we can use cloud-based Jupyter notebooks for interactive documentation/examples (e.g., [qc/homo_sapiens_qc/homo_sapiens.ipynb](https://github.com/popgensims/stdpopsim/blob/master/qc/homo_sapiens_qc/homo_sapiens.ipynb)). The basic setup just requires an `environment.yml` file in the root of the repository, specifying Conda dependencies.

A working demo from my fork is available at:
https://mybinder.org/v2/gh/carjed/stdpopsim/feature/binder_setup (it takes a few moments to load).

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,20 @@
+# This file contains the Conda environment configuration for running stdpopsim 
+# in a Dockerized image on Binder, accessible through https://mybinder.org/.
+# Binder enables any Jupyter notebooks in the repository to be run interactively and
+# reproducibly in a stable cloud-based environment.
+#
+# Developer Note: New Jupyter notebooks that are added to this repository should confirm
+# that any package dependencies are specified in the list below.
+
+name: stdpopsim-environment
+channels:
+        - conda-forge
+dependencies:
+        - numpy
+        - pandas
+        - msprime
+        - scikit-allel
+        - matplotlib
+        # load the stdpopsim Github repository as a library using pip
+        - pip:
+            - "--editable=git+https://github.com/popgensims/stdpopsim.git#egg=stdpopsim-master"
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -0,0 +1,11 @@
+name: stdpopsim-environment
```
jeromekelleher: Maybe a comment at the top explaining what this file is for?

jeromekelleher:
Great idea @carjed, we should definitely do this! Your example doesn't work for me though.

carjed:
hmm, it's working for me...it does take a minute or two to launch, and annoyingly gives a 404 page while it's loading, but should go to a Jupyter directory after a few moments.

jeromekelleher:
Ah, I see, I mistook the 404 for not working.

andrewkern:
I'm just getting 404 and spinning icons... doesn't seem to be working. love the idea though.

carjed:
It does sometimes take 2-3 minutes for the Binder instance to launch, and I confirmed that it's still working on my end—the 404 error is occurring because it tries to embed a non-interactive version of a default notebook using nbviewer (typically `index.ipynb` in the root of the repository). I'll push an example later today that should clear up the 404 errors.

carjed:
a bit more detail on the 404 errors—apparently [nbviewer](https://nbviewer.jupyter.org/) doesn't like having forward slashes in the branch name, as is the convention for git-flow that I have been using for creating branches for PRs. I'll open a new PR with a different branch name for testing.

PR #47, opened 2019-03-28
Equality comparison for models.

### LINKED ISSUE #39: Equality test for models
We want a way to compare two different stdpopsim demographic models for equality, as discussed in #12.

jeromekelleher:
Closes #39.

Still some work to be done to test all the corner cases, but it basically works I think.

cc @ndukler, @andrewkern.

### DIFF
--- a/stdpopsim/__init__.py
+++ b/stdpopsim/__init__.py
@@ -15,3 +15,6 @@ from . genomes import *  # NOQA
 # Species definitions.
 from . import homo_sapiens  # NOQA
 from . import pongo  # NOQA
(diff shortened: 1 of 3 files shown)


ndukler:
Thanks! I'll look into reimplementing one of models, but I'll have to familiarize myself with msprime first, so definitely not the out of Africa model.

jeromekelleher:
This is ready for review and merge @andrewkern. Should be uncontroversial I think, just adding some needed infrastructure.

andrewkern:
lgtm!

Issue #46, opened 2019-03-27
Tracking list of published demographic models to include


carjed:
Working on a big PR with several additional demographic models from the literature (mostly in human). Opening this issue to track progress. Here's the list I'm working on so far:

Human:
- [ ] [Gazave et al., 2014](https://www.pnas.org/content/111/2/757)
- [ ] [Chen et al., 2015](https://academic.oup.com/mbe/article/32/11/2996/981518) (basically a revised Gazave model extended to include African population)
- [x] [Browning et al., 2018](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1007385) (admixture in American populations)
- [x] [Ragsdale & Gravel, 2018](https://www.biorxiv.org/content/10.1101/489401v1) (modified Gutenkunst 3-pop model + Neanderthal and ghost archaic branches)
- [ ] [Gao & Keinan, 2016 (maybe)](http://www.genetics.org/content/genetics/202/1/235.full.pdf) (this models recent growth in Europeans as faster-than-exponential, but I'm not sure `msprime` can handle this)

Chimp:
- [ ] [de Manuel et al., 2016](http://science.sciencemag.org/content/354/6311/477.long) (4 chimp populations + bonobo outgroup)

andrewkern:
Sounds good but I wonder if there is a subset of these we can prioritize? The issue being we need to QC each of these and that can become a bottleneck.

carjed:
Sure, we can probably drop the Gazave and Gao models. Gazave is superseded by the Chen model, and AFAIK `msprime` isn't currently able to handle FTE growth needed for the Gao model.

I'd probably prioritize the remaining 4 models by ease of implementation—I already have code for the Chen model, and the Browning paper has a supplement with `msprime` code used in their simulations, so it should be relatively easy to port these over. The Ragsdale human+archaic model will mainly be just tweaks to the parameters of the Gutenkunst model, plus the archaic branches and corresponding migration rates. The de Manuel chimp model will likely take the most work.

jeromekelleher:
Can we close this issue off @carjed? Perhaps we could open specific issues for these remaining models, and mark them as Good First Issue or whatever for new contributors?

ndukler:
I agree we can close this

andrewkern:
i closed this but feel free to reopen

PR #50, opened 2019-04-06
Adding more human models

carjed:
Adding the model of American admixture from Browning et al.—see #46. This paper included msprime code in the supplement, which I ported over and modified only for formatting.


jeromekelleher:
This looks excellent to me, thanks @carjed. I guess we need to decide on how to proceed with QCing these models, but unless someone volunteers to do a blind implementation of the model now, I vote we merge this now and QC it later.

andrewkern:
Yes I agree. Will merge these now. Thanks @carjed!

PR #56, opened 2019-04-10
Method to raise an exception if models are different.

### LINKED ISSUE #54: Change model.equals to 'verify' and raise an exception
Copied from #52.

> @jeromekelleher As I kinda expected, my implementation of the model did not match the production version. Would it be a good idea to have a diff() function that would report model differences per epoch in a format similar to the DemographyDebugger() program? Otherwise I'm just comparing two DemographyDebugger() outputs which is doable, but feels clumsy.

What we could do is change the definition of ``Model.equals`` so that it raises an exception if models differ rather than returning true/false; this would at least say *where* the difference occurs). So, we'd change the name to ``check_equal``, which would raise a ValueError with meaningful messages telling us which bits of the models differ. The existing tests can be rejigged to use this easily enough.

Would this be useful for the QC process @carjed, @ndukler?

jeromekelleher:
Closes #54

Adds another method ``verify_equal`` which tells the user which parts of their models are different.

### DIFF
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -12,15 +12,41 @@ DEFAULT_ATOL = 1e-05
 DEFAULT_RTOL = 1e-08
 
 
+class UnequalModelsError(Exception):
+    """
+    Exception raised models by verify_equal to indicate that models are
+    not sufficiently close.
+    """
+
+
 def population_configurations_equal(
         pop_configs1, pop_configs2, rtol=DEFAULT_RTOL, atol=DEFAULT_ATOL):
     """
     Returns True if the specified lists of msprime PopulationConfiguration
     objects are equal to the specified tolerances.
 
+    See the :func:`.verify_population_configurations_equal` function for
+    details on the assumptions made about the objects.
+    """
+    try:
+        verify_population_configurations_equal(
+            pop_configs1, pop_configs2, rtol=rtol, atol=atol)
+        return True
+    except UnequalModelsError:
+        return False
+
+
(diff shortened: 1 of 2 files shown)


jeromekelleher:
> The functionality looks good. I tried test the function by cloning the repo and checking out the 'check_equal' branch, then ran python setup.py install --force. I must be doing something wrong because I get a "Model has no attribute verify_equal" error.

OK, thanks. Installing stuff into your global Python installation with ``python3 setup.py install`` can be problematic, it's usually easiest to just run stuff from the repo root, which will always use the local version of the package then. FYI, it's handy to use a [remote](https://help.github.com/en/categories/managing-remotes) for doing this sort of review. If you ``git add remote`` for my fork then you can checkout this branch in the same directory as your own fork. Saves a bit of hassle.

ndukler:
> > The functionality looks good. I tried test the function by cloning the repo and checking out the 'check_equal' branch, then ran python setup.py install --force. I must be doing something wrong because I get a "Model has no attribute verify_equal" error.
> 
> OK, thanks. Installing stuff into your global Python installation with `python3 setup.py install` can be problematic, it's usually easiest to just run stuff from the repo root, which will always use the local version of the package then. FYI, it's handy to use a [remote](https://help.github.com/en/categories/managing-remotes) for doing this sort of review. If you `git add remote` for my fork then you can checkout this branch in the same directory as your own fork. Saves a bit of hassle.

Ah, thanks for the advice! It's my first time working on a collaborative repo of this scale so I'm learning a lot.

PR #58, opened 2019-04-12
QC module and TennessenTwoPopOOA QC implementation

ndukler:
Implemented the TennessenTwoPopOOA blind for QC and created the QC module.


ndukler:
I did not add the model comparison test so it shows as passing. Should I do that?

PR #60, opened 2019-04-12
QC module and Tenessenn OOA model test

ndukler:


jeromekelleher:
I think it did run, looking at the timestamps anyway. Do the tests pass if you pull from your fork and run locally?

ndukler:
The test for model equality failed but I've looked at both models in the demography debugger and, unless I've overlooked something, they have identical parameterizations. The only difference is in the print order of events in epochs 0 and 1. For the production model it looks like this 
> Events @ generation 204.6
>    - Migration rate change for (0, 1) to 2.5e-05
>    - Migration rate change for (1, 0) to 2.5e-05
>    - Population parameter change for 1: initial_size -> 9279.212349452768 growth_rate -> 0.00307
>    - Population parameter change for 0: initial_size -> 14474 growth_rate -> 0

Whereas for the qc model it looks like this:
> Events @ generation 204.6
>    - Population parameter change for 0: initial_size -> 14474 growth_rate -> 0
>    - Population parameter change for 1: initial_size -> 9279.212349452768 growth_rate -> 0.00307
>    - Migration rate change for (0, 1) to 2.5e-05
>    - Migration rate change for (1, 0) to 2.5e-05

Something similar happens for epoch 1.

ndukler:
> I think it did run, looking at the timestamps anyway. Do the tests pass if you pull from your fork and run locally?

You're right, I realized it right after I posted so I deleted it hoping no-one noticed :)

jeromekelleher:
The ordering of the events within a specific time is significant (the events are applied in this order), so it's probably best to change one or the other models so they are the same. In this case it's clear that the ordering is arbitrary, but I'm not sure this is guaranteed to be true, so I'd rather not change the definition of 'equality'. I agree it's annoying to have to make them pair up though.

ndukler:
Sorry if this is a dumb question since I don't have much experience with this, but aren't the events simultaneous from a mathematical perspective? So, nothing has a chance to happen within or between populations during the time they are applied? If not, that's kinda interesting from a model implementation perspective b/c it's weird that something meaningful could happen based on order in 0 time.

carjed:
also curious about this—would an example scenario be when modeling bottlenecks with instantaneous recovery? you could, for example, model a mass migration from pop 1 into pop 2 that takes place in the middle of an instantaneous bottleneck/recovery in pop 2. All 3 demographic events would be specified as occurring at the same time, but would need to be ordered properly.

ndukler:
Could we add an option to `equals()` that allows for the ordering to not matter. It wouldn't be the default but it could save a lot of headaches.

jeromekelleher:
Suppose we had a two mass migrations at time t: one moves 0.5 from population A to C and the other moves 0.5 of population A to population D. The order these occur in *will* matter, and result in subtly different population models down the line. I agree with you @ndukler that this is weird mathematically, but this is how things work --- the only other alternative is to rule out events that occur at the same time entirely (which is surely worse). So, I would rather not change the definition of equals, as figuring out what are significant and not significant differences is is non-trivial.

ndukler:
I suppose I still don't really understand your example if it results in all of A going to C or D since it still seems like a coin flip for each member of the population. Unless it actually results in 0.5 of the initial population of A going to C and then 0.5 of the remaining population of A going to D thus leaving a quarter of the original population in A? I get @carjed's example even though I think instantaneous bottlenecks and recoveries are weird. That being said I'll reorder events in my QC implementation. I know this may be a bit of work but would it be possible to have `verify_equal()` still fail when the models are only different in ordering, but return a message indicating that? Do you think it would be worthwhile?

ndukler:
Ok, so I have no idea why it's still failing, the models are equivalent as far as I can tell and when I run `model.equals(homo_sapiens_qc.TennessenTwoPopOutOfAfrica())` manually it returns `True`. The failure message is:

> ======================================================================
> FAIL: test_known_models (tests.test_models.TestModelsEqual)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/circleci/stdpopsim/tests/test_models.py", line 250, in test_known_models
>     self.assertEqual(j == k, known_models[j].equals(known_models[k]))
> AssertionError: False != True

jeromekelleher:
> I know this may be a bit of work but would it be possible to have verify_equal() still fail when the models are only different in ordering, but return a message indicating that? Do you think it would be worthwhile?

I think it is a bit tricky, and it would take a long time to think through all the possible corner cases, so I'd be happier with a definition of ``equals`` that's overly strict but definitely correct rather than one which is a bit more convenient for developers but could contain subtle problems. It just doesn't seem like it's on the critical path to me, and there's a bunch of other things that need to be developed.

jeromekelleher:
> Ok, so I have no idea why it's still failing, the models are equivalent as far as I can tell and when I run model.equals(homo_sapiens_qc.TennessenTwoPopOutOfAfrica()) manually it returns True. The failure message is:

This was a problem with the way I had defined the ``all_models`` function @ndukler, I hadn't considered that we'd have identical QC models as well. There's a fix in #61. Once this is merged you should be able to rebase this PR off upstream/master and all will be good.

jeromekelleher:
OK, that's done; if you rebase and force push now @ndukler I think it should all work.

ndukler:
Ok, looks good to merge!

qc/homo_sapiens_qc/tennesen_two_pop_qc.py
```diff
@@ -0,0 +1,11 @@
+#%% [markdown]
```
jeromekelleher: I don't think this file does anything; can we remove it?

ndukler: Ah yeah, oops. Delete away.

ndukler: Or actually I'll take care of it

ndukler:
So are we going to merge this in, then in future separate out the pull requests?

jeromekelleher:
I would vote to merge, yes. @andrewkern, what do think? We can add in the documentation stuff later, as it's clear the model is QC'd now with the addition of the model in the qc module.

andrewkern:
yes definitely lets merge this. sorry to be behind on Github right now.

PR #61, opened 2019-04-13
Refine all_models to only return from within stdpopsim.

jeromekelleher:
Changes definition of all_models so that we only return those defined within stdpopsim. Fixes problem encountered in QC process in #60.


jeromekelleher:
I'm going to merge this as it's a minor bugfix change.

PR #63, opened 2019-04-25
Ground truth

jgallowa07:
This is not completed, It needs tests and some re-naming;  I thought I'd put it up so others could take a look. 

@petrelharp maybe you can pull this in and have a closer look? I'm happy to write formal tests once we're getting sane estimates.

### DIFF
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -168,7 +168,8 @@ class Model(object):
             N_t[j] = N
         return steps, N_t
 
-    def coalescence_rate_trajectory(self, end, num_samples, num_steps=10, min_pop_size=1):
+    def coalescence_rate_trajectory(self, end, num_samples,
+                                    num_steps=10, min_pop_size=1):
         """
         This function will calculate the ground truth
         coalescent rate trajectory, r (Ne = 1 / 2r), for a population with
@@ -183,10 +184,10 @@ class Model(object):
         `num_steps` parameter will determine how many points along the
         time axis are returned.
 
-        `min_pop_size` is the smallest a population size can be represented as during 
-        the computation of coalescent rates. In models where populations 
+        `min_pop_size` is the smallest a population size can be represented as during
+        the computation of coalescent rates. In models where populations
         grow exponentially, very small effective population sizes skew the
-        calculation when we want to compute coalescent rates within a population (1/2Ne)
+        calculation when we want to compute coalescent rates within a population (1/2Ne).
 
         Returns a tuple where the first value
         is the steps (in generations) and the second
@@ -194,10 +195,10 @@ class Model(object):
         steps[j], i.e., the probability per unit time that a randomly
         chosen pair of lineages from the original samples coalesces
         around that time, conditioned on the pair not having yet coalesced.
-        The third is the probablility that the pair of lineages 
+        The third is the probablility that the pair of lineages
         has not yet coalesced.
         """
-        
+
         num_pops = self.num_pops()
         assert(len(num_samples) == num_pops)
         P = np.zeros([num_pops**2, num_pops**2])
@@ -225,7 +226,7 @@ class Model(object):
             M = N_M["migration_matrix"]
             C = np.zeros([num_pops**2, num_pops**2])
             for idx in range(num_pops):
-                C[IA[idx, idx], IA[idx, idx]] = 1 / (2 * max(min_pop_size,N[idx]))
+                C[IA[idx, idx], IA[idx, idx]] = 1 / (2 * max(min_pop_size, N[idx]))
             for idx, row in enumerate(M):
                 M[idx][idx] = -1 * sum(row)
             G = (np.kron(M, Identity) + np.kron(Identity, M)) - C
@@ -247,7 +248,7 @@ class Model(object):
                 P = np.matmul(P, scipy.linalg.expm(ds * G))
                 P = np.matmul(P, S)
                 P = np.matmul(P, scipy.linalg.expm((dt - ds) * G))
-            else: 
+            else:
                 P = np.matmul(P, scipy.linalg.expm(dt * G))
             p_t[j] = np.sum(P)
             r[j] = np.sum(np.matmul(P, C)) / np.sum(P)
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -6,7 +6,7 @@ import inspect
 
 import msprime
 import numpy as np
-
+import scipy.linalg
 
 # Defaults taken from np.allclose
 DEFAULT_ATOL = 1e-05
@@ -136,6 +136,155 @@ class Model(object):
             "population_configurations": self.population_configurations,
             "migration_matrix": self.migration_matrix,
             "demographic_events": self.demographic_events}
+
+# ###############################################################
+# UNDER CONSTRUCTION - GROUND TRUTH N_t
+# ###############################################################
+
+    def population_size_trajectory(self, end, num_steps=10):
+        """
+        This function computes the defined effective population size
+        for each sub-population, for any demographic model.
+
+        `end` specifies at which generation to stop
+        collecting effective population sizes among subpopulations.
+        The sampling will always begin at generation 0
+
+        `num_steps` parameter will determine how many points from
+        0 to `end` to sample.
+
+        Returns a tuple where the first value
+        is the steps (in generations) and the second is a
+        num_steps-by-num_pops ndarray that will contain the effective
+        population sizes for each subpopulation individually, for each step,
+        as defined by the model which inherits from this class.
+        """
+
+        num_pops = self.num_pops()
+        N_t = np.zeros([num_steps, num_pops])
+        steps = np.linspace(0, end, num_steps)
+        for j, t in enumerate(steps):
+            N = self.pop_size_and_migration_at_t(t)["population_sizes"]
+            N_t[j] = N
+        return steps, N_t
+
+    def coalescence_rate_trajectory(self, end, num_samples, num_steps=10, min_pop_size=1):
+        """
+        This function will calculate the ground truth
+        coalescent rate trajectory, r (Ne = 1 / 2r), for a population with
+        multiple sub-populations. This takes a
+        demographic model and returns a
+        function of time that gives haploid coalescence rate.
+
+        `num_samples` should be a list of the same length as the number
+        of populations, so that num_samples[j] is the number of
+        samples in subpopulation j.
+
+        `num_steps` parameter will determine how many points along the
+        time axis are returned.
+
+        `min_pop_size` is the smallest a population size can be represented as during 
+        the computation of coalescent rates. In models where populations 
+        grow exponentially, very small effective population sizes skew the
+        calculation when we want to compute coalescent rates within a population (1/2Ne)
+
+        Returns a tuple where the first value
+        is the steps (in generations) and the second
+        is an array whose jth element is the coalescence rate at time
+        steps[j], i.e., the probability per unit time that a randomly
+        chosen pair of lineages from the original samples coalesces
+        around that time, conditioned on the pair not having yet coalesced.
+        The third is the probablility that the pair of lineages 
+        has not yet coalesced.
+        """
+        
+        num_pops = self.num_pops()
+        assert(len(num_samples) == num_pops)
+        P = np.zeros([num_pops**2, num_pops**2])
+        IA = np.array(range(num_pops**2)).reshape([num_pops, num_pops])
+        Identity = np.eye(num_pops)
+        for x in range(num_pops):
+            for y in range(num_pops):
+                K_x = num_samples[x]
+                K_y = num_samples[y]
+                P[IA[x, y], IA[x, y]] = K_x * (K_y - (x == y))
+        P = P / np.sum(P)
+        r = np.zeros(num_steps)
+        p_t = np.zeros(num_steps)
+        steps = np.linspace(0, end, num_steps)
+        dt = steps[1] - steps[0]
+        mass_migration_objects = []
+        mass_migration_steps = []
+        for demo in self.demographic_events:
+            if type(demo) == msprime.simulations.MassMigration:
+                mass_migration_objects.append(demo)
+                mass_migration_steps.append(int(demo.time/dt))
+        for j in range(num_steps):
+            N_M = self.pop_size_and_migration_at_t(steps[j])
+            N = N_M["population_sizes"]
+            M = N_M["migration_matrix"]
+            C = np.zeros([num_pops**2, num_pops**2])
+            for idx in range(num_pops):
+                C[IA[idx, idx], IA[idx, idx]] = 1 / (2 * max(min_pop_size,N[idx]))
+            for idx, row in enumerate(M):
+                M[idx][idx] = -1 * sum(row)
+            G = (np.kron(M, Identity) + np.kron(Identity, M)) - C
+            if j in mass_migration_steps:
+                idx = mass_migration_steps.index(j)
+                ds = mass_migration_objects[idx].time - (j * dt)
+                a = mass_migration_objects[idx].source
+                b = mass_migration_objects[idx].dest
+                p = mass_migration_objects[idx].proportion
+                S = np.eye(num_pops**2, num_pops**2)
+                for x in range(num_pops):
+                    if x == a:
+                        S[IA[a, a], IA[a, b]] = S[IA[a, a], IA[b, a]] = p * (1 - p)
+                        S[IA[a, a], IA[b, b]] = p ** 2
+                        S[IA[a, a], IA[a, a]] = (1 - p) ** 2
+                    else:
+                        S[IA[x, a], IA[x, b]] = S[IA[a, x], IA[b, x]] = p
+                        S[IA[x, a], IA[x, a]] = S[IA[a, x], IA[a, x]] = 1 - p
+                P = np.matmul(P, scipy.linalg.expm(ds * G))
+                P = np.matmul(P, S)
+                P = np.matmul(P, scipy.linalg.expm((dt - ds) * G))
+            else: 
+                P = np.matmul(P, scipy.linalg.expm(dt * G))
+            p_t[j] = np.sum(P)
+            r[j] = np.sum(np.matmul(P, C)) / np.sum(P)
+        return steps, r, p_t
+
+    def num_pops(self):
+        """
+        This function returns the number populations
+        defined by the demographic model
+        """
+        ddb = msprime.DemographyDebugger(**self.asdict())
+        return len(ddb.epochs[0].populations)
+
+    def pop_size_and_migration_at_t(self, t):
+        """
+        Given a time, t (in generations), find and return
+        a dictionary containing:
+        1: "population_sizes" the vector N which should represent effective
+        population size for each populations at time t and,
+        2: "migration_matrix" The migration matrix for for the populations at time t
+        """
+
+        ddb = msprime.DemographyDebugger(**self.asdict())
+        epochs = ddb.epochs
+        j = 0
+        while epochs[j].end_time <= t:
+            j += 1
+        N = ddb.population_size_history[:, j]
+        for i, pop in enumerate(epochs[j].populations):
+            s = t - epochs[j].start_time
+            g = pop.growth_rate
+            N[i] *= np.exp(-1 * g * s)
+        return {"population_sizes": N, "migration_matrix": epochs[j].migration_matrix}
+
+# ###############################################################
+# END CONSTRUCTION - GROUND TRUTH N_t
+# ###############################################################
 
     def equals(self, other, rtol=DEFAULT_RTOL, atol=DEFAULT_ATOL):
         """


stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        return G
+
+    # TODO Rename
+    def _get_Mp_from_M(self, M):
+        """
+        Simply compute Mp. which represents the **,
+        From the migration matrix.
+        """
+
+        Mp = copy.deepcopy(M)
```
petrelharp: What's wrong with
```
Mp = M.copy()
```
?

stdpopsim/models.py
```diff
@@ -6,7 +6,7 @@
 
 import msprime
 import numpy as np
-
+import copy
```
petrelharp: I think we don't need this (see below).

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
             "migration_matrix": self.migration_matrix,
             "demographic_events": self.demographic_events}
 
+# ###############################################################
+# UNDER CONSTRUCTION - GROUND TRUTH N_t
+# ###############################################################
+
+    # TODO Rename
+    def get_Ne_through_time_single_pop(self, end, num_steps=10):
```
petrelharp: vote to rename this `population_size_trajectory()`

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        num_pops = self.num_pops()
+        N_t = np.zeros([num_steps, num_pops])
+        steps = np.linspace(0, end, num_steps)
+        for j, t in enumerate(steps):
+            N = self._get_N_M_at_t(t)["population_sizes"]
+            N_t[j] = N
+        return steps, N_t
+
+    # TODO Rename
+    def get_Ne_through_time_all_pops(self, end, num_samples_per_location, num_steps=10):
```
petrelharp: vote to rename this `coalescence_rate_trajectory( )` and have it return coalescence rates (so when plotting you have to do 1/(this)

stdpopsim/models.py
petrelharp: And, vote to call this "population size" not "Ne" in the docs.

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        """
+        This function will calculate the ground truth
+        effective population size (Ne) for a population with
+        multiple sub-populations. This takes a
+        demographic model and returns a
+        function of time that gives inverse coalescence rate.
+
+        `num_samples_per_location` should be a list containing the
+        number of samples in each subpopulation organized by population index.
```
petrelharp: Vote to rename this `num_samples` with documentation: "`num_samples` should be a list of the same length as the number of populations, so that `num_samples[j]` is the number of samples in subpopulation `j`."

stdpopsim/models.py
jgallowa07: well, because python doesn't pass by reference we don't need to copy at all, I'll fix this.

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+
+        `num_steps` parameter will determine how many points along the
+        time axis are returnes.
```
petrelharp: need to say the time steps are uniformly spaced from 0 to `end`

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+
+        Returns a tuple where the first value
+        is the steps (in generations) and the second is a numpy
+        ndarray that will contain the effective population sizes for
+        each subpopulation individually, for each step, as defined
+        by the model which inherits from this class.
```
petrelharp: need to say the dimensions of the returned array (eg are rows populations or are columns?)

stdpopsim/models.py
jgallowa07: agreed

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+
+        num_pops = self.num_pops()
+        assert(len(num_samples_per_location) == num_pops)
+        P = np.zeros([num_pops**2, num_pops**2])
+        index_array = np.array(range(num_pops**2)).reshape([num_pops, num_pops])
+        for x in range(num_pops):
+            for y in range(num_pops):
+                K_x = num_samples_per_location[x]
+                K_y = num_samples_per_location[y]
+                P[index_array[x, y], index_array[x, y]] = K_x * (K_y - self._delta(x, y))
```
petrelharp: It'd be more readable (don't need to look up `_delta`) to say it this way:
```
                 P[index_array[x, y], index_array[x, y]] = K_x * (K_y - (y == x))
```

stdpopsim/models.py
jgallowa07: agreed.

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        steps = np.linspace(0, end, num_steps)
+        dt = steps[1] - steps[0]
+        for j in range(num_steps):
+            N_M = self._get_N_M_at_t(steps[j])
+            pop_sizes = N_M["population_sizes"]
+            mig_matrix = N_M["migration_matrix"]
+            C = self._get_C_from_N(pop_sizes)
+            Mp = self._get_Mp_from_M(mig_matrix)
+            G = self._get_G_from_Mp_C(Mp, C)
+            P = P * np.exp(dt * G)
```
petrelharp: This needs to be (a) matrix multiplication, and (b) matrix exponentiation, instead of the "usual" ones. So:
```
P = np.matmul(P, scipy.linalg.expm(dt * G))
```

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        Given a time, t (in generations), find and return
+        a dictionary containing:
+        1: "population_sizes" the vector N which should represent effective
+        population size for each populations at time t and,
+        2: "migration_matrix" The migration matrix for for the populations at time t
+
+        <class 'int'> -> <class 'dict'>
+        """
+
+        epochIndex, epoch = self._get_epoch_at_t(t)
```
petrelharp: if you moved the `get_epoch` code into here (the only place that function is used) then you wouldn't have to create an extra DemographyDebugger...

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        """
+        Compute the matrix C, which reprents coalescent probabilities
+        in each population, from Individual effective population sizes.
+        """
+
+        n = len(N)
+        C = np.zeros([n**2, n**2])
+        index_array = np.array(range(n**2)).reshape([n, n])
+        for idx in range(n):
+            C[index_array[idx, idx], index_array[idx, idx]] = 1 / N[idx]
```
petrelharp: this should be `1/(2*N[idx])` - but, you already found this, I think?

stdpopsim/models.py
jgallowa07:
```>>> x = [[1,2],[3,4]]
>>> def yoo_copy(alist):
...  a = alist.copy()
...  a[0][0] = 500
...  return a
... 
>>> yoo_copy(x)
[[500, 2], [3, 4]]
>>> x
[[500, 2], [3, 4]]
>>> x = [[1,2],[3,4]]
>>> def yoo_deepcopy(alist):
...  a = copy.deepcopy(alist)
...  a[0][0] = 500
...  return a
... 
>>> yoo_deepcopy(x)
[[500, 2], [3, 4]]
>>> x
[[1, 2], [3, 4]]```

File: stdpopsim/models.py [hunk 12]
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        return C
+
+    # TODO Rename
+    def _get_G_from_Mp_C(self, Mp, C):
```
petrelharp: maybe these two don't need to be separate functions.

stdpopsim/models.py
jgallowa07: deep copy is reccursive and python is pass by reference with mutable object. Because the migration matrix is a list of lists, it needs to be reccursivly copied

stdpopsim/models.py
```diff
@@ -278,7 +278,7 @@ def _get_C_from_N(self, N):
         C = np.zeros([n**2, n**2])
         index_array = np.array(range(n**2)).reshape([n, n])
         for idx in range(n):
-            C[index_array[idx, idx], index_array[idx, idx]] = 1 / (2 * N[idx])
```
petrelharp: Huh, this is troubling. There *should* be a factor of 2 there. I think this means we're missing a factor of 2 somewhere else.

stdpopsim/models.py
jgallowa07: No, that's what I changed it from. It is correct as `1 / N[idx]` by my checks.

stdpopsim/models.py
jgallowa07: Agreed.

stdpopsim/models.py
jgallowa07: Ahhh, That might mess up the multi-loc for sure.

petrelharp:
For reference, here's what we are implementing: 
image

stdpopsim/models.py
```diff
@@ -137,6 +137,176 @@ def asdict(self):
...
+        P = P / np.sum(P)
+        r = np.zeros(num_steps)
+        steps = np.linspace(0, end, num_steps)
+        dt = steps[1] - steps[0]
+        for j in range(num_steps):
+            N_M = self._get_N_M_at_t(steps[j])
+            pop_sizes = N_M["population_sizes"]
+            mig_matrix = N_M["migration_matrix"]
+            C = self._get_C_from_N(pop_sizes)
+            Mp = self._get_Mp_from_M(mig_matrix)
```
petrelharp: gee, since we don't use `mig_matrix` again, there's no need for that `deepcopy()` below of it is, there?

petrelharp:
Oh! I know where the factor of 2 is. The Ne(t) that you are getting from the other methods are *diploid*, so if X is our coalescence rates, you should compare Ne(t) to 1/(2*X).

stdpopsim/models.py
petrelharp: So, confirmed; this line *should* read `1 / (2 * N[idx])`

stdpopsim/models.py
jgallowa07: not really, sg.

stdpopsim/models.py
jgallowa07: agreed, I'll probably get rid of the functions to clean it up.

petrelharp:
For testing, next
 - do a two-pop model with one big and one small population. Ne for samples from the big pop should start at big Ne then decline to something in the middle; Ne for samples from the small pop should start at small Ne then increase to the same thing

And, we need to do something different for mass migrations. Suppose that a time increment that has a mass migration in it is from `t` to `t + dt`, and the mass migration occurs at time `t + ds` (with `ds < dt`). Then we should
1. Do `P = np.matmult(P, expm(ds * G))`.
2. Compute the matrix `S` that has the same dimensions as `P`, defined as follows. Suppose the mass migration has `source` population `a` and `dest` population `b` and `proportion` moving `p`, and writing `J[x,y]` for `index_array[x,y]`:
```
S = np.eye(n**2, n**2)
for x in range(n):
  if x == a:
    S[J[a,a], J[a,b]] = S[J[a,a], J[b,a]] = p * (1 - p)
    S[J[a,a], J[b,b]] = p ** 2
  else:
    S[J[x,a], J[x,b]] = S[J[a,x], J[b,x]] = p
    S[J[x,a], J[x,a]] -= p
    S[J[a,x], J[a,x]] -= p     
```
3. Do `P = np.matmult(P, S)`
4. Do `P = np.matmult(P, expm((dt - ds) * G))`.

stdpopsim/models.py
petrelharp: ah-ha. thanks! scary that I didn't realize that.

jeromekelleher:
This is great functionality, and we should definitely support it. I wonder if the functionality should be in msprime though? All the basic information should already be available in the DemographyDebugger and we can add a method to return the population sizes for a specific array of times (saves the problem of deciding how to space the times: maybe someone would want to logspace them).

jgallowa07:
> This is great functionality, and we should definitely support it. I wonder if the functionality should be in msprime though? All the basic information should already be available in the DemographyDebugger and we can add a method to return the population sizes for a specific array of times (saves the problem of deciding how to space the times: maybe someone would want to logspace them).

I think It would be great in `msprime`. Once @petrelharp and I get some formal tests (and cleanup) I'm happy to add the method to DemographyDebugger.

andrewkern:
if this goes into `msprime` we should probably delete that bit of code that I added to `DeomgraphyDebugger` as this is the ultimate solution for what I was going for

jeromekelleher:
> f this goes into msprime we should probably delete that bit of code that I added to DeomgraphyDebugger as this is the ultimate solution for what I was going for

There's no need I think --- the extra attributes are simple and could be useful for other stuff.

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+        for demo in self.demographic_events:
+            if type(demo) == msprime.simulations.MassMigration:
+                mass_migration_objects.append(demo)
+                mass_migration_steps.append(int(demo.time/dt))
+        for j in range(num_steps):
+            N_M = self.pop_size_and_migration_at_t(steps[j])
+            N = N_M["population_sizes"]
+            M = N_M["migration_matrix"]
+            C = np.zeros([num_pops**2, num_pops**2])
+            index_array = np.array(range(num_pops**2)).reshape([num_pops, num_pops])
```
petrelharp: this should be defined outside the for loop

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+            for idx in range(num_pops):
+                C[index_array[idx, idx], index_array[idx, idx]] = 1 / (2 * N[idx])
+            for idx, row in enumerate(M):
+                M[idx][idx] = -1 * sum(row)
+            Identity = np.eye(len(M[0]))
```
petrelharp: `Identity` could also be defined outside the for loop.

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+            G = (np.kron(M, Identity) + np.kron(Identity, M)) - C
+            if j in mass_migration_steps:
+                idx = mass_migration_steps.index(j)
+                ds = mass_migration_objects[idx].time - (j * dt)
+                a = mass_migration_objects[idx].source
+                b = mass_migration_objects[idx].dest
+                p = mass_migration_objects[idx].proportion
+                J = np.array(range(num_pops**2)).reshape([num_pops, num_pops])
```
petrelharp: `J` is the same thing as `index_array`.

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+                S = np.eye(num_pops**2, num_pops**2)
+                for x in range(num_pops):
+                    if x == a:
+                        S[J[a, a], J[a, b]] = S[J[a, a], J[b, a]] = p * (1 - p)
+                        S[J[a, a], J[b, b]] = p ** 2
```
petrelharp: Whoops, I forgot a term here: there should be another line:
```
S[J[a, a], J[a, a]] = (1 - p) ** 2
```

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+                    else:
+                        S[J[x, a], J[x, b]] = S[J[a, x], J[b, x]] = p
+                        S[J[x, a], J[x, a]] -= p
+                        S[J[a, x], J[a, x]] -= p
```
petrelharp: Instead of 
```
                        S[J[x, a], J[x, a]] -= p
                        S[J[a, x], J[a, x]] -= p
```
it might be more clear to have
```
                        S[J[x, a], J[x, a]] = S[J[a, x], J[a, x]] = 1 - p
```

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+                P = np.matmul(P, S)
+                P = np.matmul(P, scipy.linalg.expm((dt - ds) * G))
```
petrelharp: Also, whoops, we forgot the first `ds` here: this should be
```
                P = np.matmul(P, scipy.linalg.expm(ds * G))
                P = np.matmul(P, S)
                P = np.matmul(P, scipy.linalg.expm((dt - ds) * G))
```

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+            else:
+                P = np.matmul(P, scipy.linalg.expm(dt * G))
+            r[j] = np.sum(np.matmul(P, C)) / np.sum(P)
+        return steps, 1/(2*r)
```
petrelharp: This function is called `coalescence_rate_trajctory` but it returns "effective population sizes" instead of coalescence rates. I vote to have this function return `steps, r` and let the user plot `1/(2*r)` if that's what they want.

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+        of populations, so that num_samples[j] is the number of
+        samples in subpopulation j.
+
+        `num_steps` parameter will determine how many points along the
+        time axis are returned.
+
+        Returns a tuple where the first value
+        is the steps (in generations) and the second is a numpy
+        ndarray that will contain the coalescence rate for
+        the entire population.
```
petrelharp: "the second is an array whose `j`th element is the coalescence rate at time `steps[j]`, i.e., the probability per unit time that a randomly chosen pair of lineages from the original samples coalesces around that time, conditioned on the pair not having yet coalesced."

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+                        S[J[x, a], J[x, a]] -= p
+                        S[J[a, x], J[a, x]] -= p
+                P = np.matmul(P, S)
+                P = np.matmul(P, scipy.linalg.expm((dt - ds) * G))
+            else:
+                P = np.matmul(P, scipy.linalg.expm(dt * G))
+            r[j] = np.sum(np.matmul(P, C)) / np.sum(P)
+        return steps, 1/(2*r)
+
+    # Eventually just grab the mass migration events.
```
petrelharp: this function isn't used.

petrelharp:
Looks good! I found some errors (deriving from what I wrote down); let's get some sanity checks? I think these can be formal tests:

1) Have a history with two populations, with no migration; check that coalescence rate of lineages from a single pop reflects the size history of that pop:
 - population sizes today are [100, 1000]
 - population 1 had a size change to 200 at t=100 generations ago
 - there is a mass migration event that entirely merges population 2 (source) into population 1 (dest) at t=200 generations ago
 - population 1 has a size change to 100 at t=300 generations ago
Then, we can test the following:
 a) The coalesce rates through time of two samples from population 1 is (1/(2*100) until time 100), (1/(2*200) between times 100 and 300), (1/(2*100) past t=300)
 b) The coalesce rates through time of two samples from population 2 is (1/(2*1000) until time 200), (1/(2*200) between times 200 and 300), (1/(2*100) past t=300)
 c) The coalesce rates through time of one sample from population 1 and one sample from population 2 is (0 until time 200), (1/(2*200) between times 200 and 300), (1/(2*100) past t=300)
As another basic check, we should check that
 d) The coalescence rates through time of one sample from each population is the same as the rates for 10 samples from each population.

2) Have two populations with very high migration; check they act as a single population:
 - population sizes today are [100, 1000]
 - migration rate from population 1 to 2 of 20, and from 2 to 1 of 10
Then, we test
 a) The coalescence rate for two samples from population 1 is (1/3) * (1/(2*100)) + (2/3) * (1/(2*1000))
 b) Same for two samples from population 2 and for one sample from each population.

These things all should be equal up to some degree of numerical precision.

stdpopsim/models.py
```diff
@@ -137,6 +137,153 @@ def asdict(self):
...
+                        S[J[a, a], J[b, b]] = p ** 2
+                    else:
+                        S[J[x, a], J[x, b]] = S[J[a, x], J[b, x]] = p
+                        S[J[x, a], J[x, a]] -= p
+                        S[J[a, x], J[a, x]] -= p
+                P = np.matmul(P, S)
+                P = np.matmul(P, scipy.linalg.expm((dt - ds) * G))
+            else:
+                P = np.matmul(P, scipy.linalg.expm(dt * G))
+            r[j] = np.sum(np.matmul(P, C)) / np.sum(P)
```
jgallowa07: @petrelharp I keep getting `RuntimeWarning: invalid value encountered in double_scalars
  r[j] = np.sum(np.matmul(P, C)) / np.sum(P)` for this line when running on a model with mass migration. Shall we check that np.sum(P) != 0 before dividing?

jgallowa07:
@petrelharp sounds good, I'll work on those tests today.

stdpopsim/models.py
petrelharp: It should never be the case that `np.sum(P)` is zero. Can you check where the warning is coming from? Is it really from the `1/np.sum(P)`? Is it because of a `nan` somewhere? Or what?

jgallowa07: No, It is because of the np.sum(P) == 0, (I checked with an assertion). I'll investigate more now.

petrelharp: A good sanity check while debugging would be to set C to zero (or, just have very, very large population sizes) and verify that the rows of `P` all sum to 1.

jgallowa07: This should be true for every time step? There must be a bug somewhere because all the rows are summing close to 0.1 when C is kept as zeros for each iteration.

petrelharp: Sorry, I meant that the sum of *all* of `P` should be 1 at all times if coalescence rates are zero.

petrelharp: At all times, sum(P) should be the probability that the lineages of a randomly chosen pair of samples have not yet coalesced, so at time `t`, if all population sizes are greater than `N`, we should always have that `1 - t / (2*N) <= sum(P) <= 1`.

jgallowa07: Looks like my head cold is preventing me from thinking about this too clearly, but I can say that P converges to 0 with any model if the `end` time is far enough.

jgallowa07: So this should never be a problem so long as `end` isn't too far being when the species demographic defines to start. Maybe we should just limit how large `end` can be?

jgallowa07: because if you go back to far, the probability that the lineages of a randomly chosen pair of samples have not yet coalesced converges to zero, no?

petrelharp: My first suspicion is that this is a bug, because `end` would have to be really, really big for this to be a legitimate problem.

jgallowa07: Hmm, doesn't seem to be the case in the models I have tried - for Gutenkunst, P converges around 9500 years ago

jgallowa07:
@petrelharp @jeromekelleher now that we have some good sanity checks and some real tests should I focus on integrating the tests into `stdpopsim`, or should I start integrating all this all into `msprime`'s `DemographyDebugger`?

petrelharp:
I think what we have tests it good enough; so I'd say move it all over to `msprime` now. Should be almost copy-and-paste at this point.

jgallowa07:
Should probably close this PR, due to https://github.com/tskit-dev/msprime/pull/742

andrewkern:
okay closing it down now

PR #72, opened 2019-05-10
added decode 2010 genetic map

andrewkern:


jeromekelleher:
> Eventually, I believe we should choose one location to host all these maps along with simulated data. But I think this is separate discussion

Agreed on both counts!

PR #76, opened 2019-05-15
Dev docs

ndukler:
Updated dev docs with a few more specific details

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -381,12 +381,23 @@ follow these steps for it to be officially supported by stdpopsim:
 
     3. M creates an issue for the CLI implementation of the model.
 
-    4. Developer B creates a blind implementation of the model as PR within the
-       unit testing module, and adds the automatic checking of this model for
-       equality with the production model to the suite of unit tests. All being
-       good, this PR is merged and the QC issue is closed.
+    4. Developer B creates a blind implementation of the model in the
+       ``qc/species_name_qc.py`` file. Note that if you are adding a new species
+       you will have to add a new import to ``qc/__init__.py``.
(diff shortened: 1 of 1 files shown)


docs/development.rst
```diff
@@ -381,12 +381,23 @@ follow these steps for it to be officially supported by stdpopsim:
...
 
-    4. Developer B creates a blind implementation of the model as PR within the
-       unit testing module, and adds the automatic checking of this model for
-       equality with the production model to the suite of unit tests. All being
-       good, this PR is merged and the QC issue is closed.
-
-    5. Someone then makes a PR updating the CLI, checking that the
+    4. Developer B creates a blind implementation of the model in the
+       ``qc\species_name_qc.py`` file. Note that if you are adding a new species
+       you will have to add a new import to ``__init__.py``.
```
jeromekelleher: Two things here:
1. It's not clear that it's ``qc/__init__.py``
2. I'm not sure it's necessary to do this anyway though. Since we're only importing in the test modules, I don't think we need to worry about flattening the namespace (which is what this does). Not a strong opinion though, just a thought.

jeromekelleher:
LGTM, thanks @ndukler. Minor point about ``__init__.py`` above.

ndukler:
I'll fix the ``__init__.py`` text. I added it to match the formats of the other sections (without really understanding deeply what was going on) but I kinda like it and I don't think it does any harm so if it's ok I'll leave it for now?

ndukler:
@jeromekelleher It's done

jeromekelleher:
Merged - thanks @ndukler!

Issue #67, opened 2019-05-07
QC for Li-Stephans model (Drosophila)


ndukler:
When you have a chance can you just throw in the paper of reference and the population id indices for the production msprime implementation @andrewkern. Thanks!

ndukler:
Li and Stephan:
https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.0020166

The Li and Stephan model is a two population case. In that case index 0 is the African population and index 1 is the european popn

ndukler:
Are the epoch time off by a factor of 10 in the production model @andrewkern ? In the paper they say " per generation (assuming ten generations per year)", and give all the event times in years so I multiplied the event times by 10. Would this be correct?

ndukler:
Actually it looks like we're off by a factor of 100. I think you may have divided by 10 instead of 0.1.

ndukler:
Submitted PR #77

andrewkern:
Ah yes! It looks like I divided by 10 rather than by 0.1!! Great catch @ndukler. So what is our SOP at this point? Do I make changes to the `master` branch? Do we merge your PR?

ndukler:
Submit a PR to the master branch, then once it's accepted, I'll rebase my PR against it

andrewkern:
okay will do right now

andrewkern:
@ndukler just pushed my PR to master for your review

ndukler:
@andrewkern Reviewed, you merge it into master now

ndukler:
@andrewkern Thanks! It looks like everything passed so PR #77 can be merged in and I'll close this issue.

PR #77, opened 2019-05-19
Li stephans qc

ndukler:
Li-Stephans QC implementation. Currently failing equality test due to discrepancy in generation time calculation. @andrewkern @jeromekelleher


jeromekelleher:
LGTM, thanks @ndukler. Looks like tests are failing due to a genuine difference in model, so this should be resolved over in #67.

PR #80, opened 2019-05-21
added mean_recombination_rate to genome and respective tests

jgallowa07:
This is trivial at the moment because we don't have any varying default mutation rates for chromosomes - Still, I think we'll need it at some point and it seems to make sense (It makes my analysis code cleaner because we need genome-wide average for inference). Open to suggestions if anyone thinks otherwise. 

@jeromekelleher, @andrewkern thoughts/review?


andrewkern:
lgtm. i'm going to go ahead and merge this small change.

PR #81, opened 2019-05-28
Sheehan song qc

ndukler:
Added QC model for Sheehan-Song. There is a difference with the production model based on t2 which will need to be resolved, probably by making `` (t_1_coal + t_2_coal) * 4 * N_ref``. Note also that there is a discrepancy in the number of generations for the epoch times estimated from ``PSMC`` (which uses 2N) and ``msms`` (which uses 4N) in the paper itself. We went with the generation counts based on the 4N coalescence unit since that is what the deep learning model was actually fit on.

### DIFF
--- a/qc/__init__.py
+++ b/qc/__init__.py
@@ -2,4 +2,4 @@
 
 # Species definitions.
 from . import homo_sapiens_qc  # NOQA
-
+from . import drosophlia_melanogaster_qc # NOQA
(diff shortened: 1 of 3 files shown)


andrewkern:
I'll get on the time discrepancy right now and request a PR

ndukler:
@andrewkern Looks good to merge

PR #84, opened 2019-06-04
Documentation of human demographic models

ndukler:
First try at documenting human demographic models. Any thoughts?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -114,7 +114,12 @@ Demographic models
 
 .. autoclass:: stdpopsim.homo_sapiens.GutenkunstThreePopOutOfAfrica
     :members:
-
+.. autoclass:: stdpopsim.homo_sapiens.TennessenTwoPopOutOfAfrica
+    :members:
+.. autoclass:: stdpopsim.homo_sapiens.TennessenOnePopAfrica
+    :members:
+.. autoclass:: stdpopsim.homo_sapiens.BrowningAmerica
+    :members:
 
 ***********************
 Drosophila Melanogaster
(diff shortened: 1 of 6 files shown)


jeromekelleher:
This is excellent, thanks @ndukler. I think you've hit just the right level of detail.

I'm happy to merge whenever.

ndukler:
@jeromekelleher @andrewkern Can you take a look at the new documentation I put in and tell me what you think? I settled on inserting a download link for a parameter table, I think it's cleaner and allows for nice LaTeX tables. 

P.S. I know that it's failing b/c of a flake thing (which I would really appreciate if you help me with!)

jeromekelleher:
I'm not so crazy about the linked PDF and latex --- it's going to be a maintenance headache going forward, and it'll be hard to get model authors to follow the correct template.

Were RST tables too fiddly?

ndukler:
I couldn't figure out how to get math symbols in but I guess I can get rid of that whole column since it doesn't really do anything without an accompanying figure. I'll try to get an RST version up soon for comparison!

ndukler:
I redid it with an embedded csv table (which gets auto-converted to RST format apparently). I can manually convert it to rst format and copy in the text, but this seems cleaner. What do you think @jeromekelleher ?

jeromekelleher:
The CSV solution is perfect - that's really handy!

Re the flake8 violation, you can either put a ``#  noqa`` on the end of the function line, or just shorten the paths a little bit (you can get rid of the ``../docs`` prefix, as this is always built in the docs directory. I see it's a bit easier for people to follow by leaving it in though.)

ndukler:
Ok, I think the documentation for the human models is good! Thoughts? @andrewkern @jeromekelleher

P.S. Gutenkunst hasn't been QC'ed, I should probably get on that since it's documented ...

andrewkern:
This looks amazing @ndukler! Can you rebase and squash your commits and then I will merge?

ndukler:
Done! @andrewkern

PR #87, opened 2019-06-05
Added TennessenOnePopAfrica QC model

ndukler:
QC for TennessenOnePopAfrica model


ndukler:
QC for one pop model looks good! @andrewkern

andrewkern:
that was easy!!

PR #88, opened 2019-06-07
Cleanup of qc module

ndukler:
Just cleaning out some of the old QC stuff that we didn't end up using.


jeromekelleher:
Thanks @ndukler, merging.

PR #92, opened 2019-06-18
added generation times as class variable to demog models

andrewkern:
this PR attempts to clean up the inconsistency noted [here](https://github.com/popgensims/analysis/issues/32) by adding `generation_time` as a class variable to each of the demographic models. 

the rationale for having `generation_time` under a specific demographic model rather than a property of the species is that each published model might assume a different generation time.

once we merge this, we can update the `Snakefile` on the `analysis` repo to pull this variable directly from the `stdpopsim` module, rather than have it set by the config file.

### DIFF
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -124,6 +124,7 @@ class Model(citations.CitableMixin):
         self.demographic_events = []
         # Defaults to a single population
         self.migration_matrix = [[0]]
+        self.generation_time = -1
 
     @property
     def name(self):
--- a/dmel_example.py
+++ b/dmel_example.py
@@ -7,7 +7,8 @@ import msprime
 import stdpopsim
 from stdpopsim import drosophila_melanogaster
 
-chrom = drosophila_melanogaster.genome.chromosomes["chr2R"]
+chr_str = "chr2L"
+chrom = drosophila_melanogaster.genome.chromosomes[chr_str]
 recomb_map = chrom.recombination_map()
 
 print("Testing Li and Stephan (2006) model")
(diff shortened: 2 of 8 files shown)


andrewkern:
sure I can do that and add to this PR.

andrewkern:
okay @ndukler check out these changes that i made

andrewkern:
ugh having issues with codecov and that new property. any idea how i can hit that code?

ndukler:
Is this link helpful? https://stackoverflow.com/questions/45680483/python-unit-testing-class-properties . Essentially just test that the generation times field value is numeric.

andrewkern:
will work on this. apologies to all that will be getting codecov emails now..

andrewkern:
okay @ndukler I think this is now set. can you give it a once over? a bunch of little changes in this PR

andrewkern:
will wait for @jeromekelleher to take a look before I merge...

stdpopsim/models.py
```diff
@@ -124,6 +124,7 @@ def __init__(self):
         self.demographic_events = []
         # Defaults to a single population
         self.migration_matrix = [[0]]
+        self._generation_time = -1
```
jeromekelleher: I don't see the point of making this a 'private' property - everything else is represented as a standard attribute so I think we should do the same for ``generation_time``. Why make a special case of generation_time?

stdpopsim/models.py
```diff
@@ -192,3 +200,15 @@ def all_models():
         if mod.startswith("stdpopsim"):
             ret.append(cls())
     return ret
+
+
+def all_model_generation_times():
```
jeromekelleher: What's the idea of this function? It's simple to reproduce (``[model.generation_time for model in stdpopsim.all_models()]``), so I don't see the need to have it.

andrewkern:
> Probably what we should do is have a species generation time which a model inherits _by default_
> unless it's set explicitly.

okay I can get this together this morning.

stdpopsim/models.py
andrewkern: yeah no reason I guess. I added it as a class `property` of each model emulating `_name`. can switch it.

stdpopsim/models.py
andrewkern: this was just a helper function I added in my struggles with codecov. it's definitely bloat. 

related-- is there a simple way to run codecov locally?

andrewkern:
okay @jeromekelleher take a look. i've now defined a species level `default_generation_time` that is scoped within each organism module.

andrewkern:
ok ready for real @jeromekelleher

stdpopsim/models.py
jeromekelleher: You can run coverage locally using some arguments to nose like [here](https://github.com/popgensims/stdpopsim/blob/master/.circleci/config.yml#L32). This will tell you how many lines you're missing and stuff, but won't give the detailed line-by-line analysis.

PR #94, opened 2019-06-21
Added D. mel model documentation

ndukler:
Documentation for the two d. mel. models. Also added the metadata for the populations.

### DIFF
--- a/stdpopsim/drosophila_melanogaster.py
+++ b/stdpopsim/drosophila_melanogaster.py
@@ -77,16 +77,41 @@ default_generation_time = 0.1
 
 class SheehanSongThreeEpoch(models.Model):
     """
-    The three epoch model estimated for African samples from Sheehan and Song
+    Model Name:
+        SheehanSongThreeEpoch
 
-    .. todo:: document this model, including the original publications
-        and clear information about what the different population indexes
-        mean.
+    Model Description:
+        The three epoch (modern, bottleneck, ancestral) model estimated for a
+        single African Drosophila Melanogaster population from `Sheehan and Song <https:/
+        /doi.org/10.1371/journal.pcbi.1004845>`_ . Population sizes are estimated by a
+        deep learning model trained on simulation data. NOTE: Due to differences in
+        coalescence units between PSMC (2N) and msms (4N) the number of generations were
+        doubled from PSMC estimates when simulating data from msms in the original
+        publication. We have faithfully represented the published model here.
 
-    """
+    Model population indexes:
+        - African D. melanogaster: 0
+
+    Parameter Table:
+        .. csv-table::
+            :widths: 15 8 20
+            :header: "Parameter Type (units)", "Value", "Description"
+            :file: ../docs/parameter_tables/drosophila_melanogaster/SheehanSongThreeEpoch_params.csv
+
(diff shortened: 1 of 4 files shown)


stdpopsim/drosophila_melanogaster.py
```diff
@@ -75,13 +75,44 @@ class Comeron2012_dm6(genetic_maps.GeneticMap):
...
-    """
+    Model Name:
+        SheehanSongThreeEpoch
+    
+    Model Description:
+        The three epoch (modern, bottleneck, ancestral) model estimated for a
+        single African Drosophila Melanogaster population from `Sheehan and Song <https:/
+        /doi.org/10.1371/journal.pcbi.1004845>`_ . Population sizes are estimated by a 
+        deep learning model trained on simulation data. NOTE: Due to differences in 
+        coalescence units between PSMC (2N) and msms (4N) generation times were doubled
```
jeromekelleher: Does this last note help? Everything is in units of generations here, so the scaling used in the original publication doesn't really enter into it, does it?

ndukler:
Think this is good to go @jeromekelleher

ndukler:
Actually hold off for a bit. I want to run the wording of the note on the Sheehan et al. model by @andrewkern .

jeromekelleher:
> Actually hold off for a bit. I want to run the wording of the note on the Sheehan et al. model by @andrewkern .

Cool - merge whenever you're happy with it @andrewkern.

stdpopsim/drosophila_melanogaster.py
```diff
@@ -77,16 +77,40 @@ class Comeron2012_dm6(genetic_maps.GeneticMap):
...
-    .. todo:: document this model, including the original publications
-        and clear information about what the different population indexes
-        mean.
+    Model Description:
+        The three epoch (modern, bottleneck, ancestral) model estimated for a
+        single African Drosophila Melanogaster population from `Sheehan and Song <https:/
+        /doi.org/10.1371/journal.pcbi.1004845>`_ . Population sizes are estimated by a
+        deep learning model trained on simulation data. NOTE: Due to differences in
+        coalescence units between PSMC (2N) and msms (4N) the number of generations were
+        doubled from PSMC estimates when simulating data from msms.
```
andrewkern: I think we should say in the note "Due to differences in
        coalescence units between PSMC (2N) and msms (4N) the number of generations were
        doubled from PSMC estimates when simulating data from msms in the original publication. We have faithfully represented the published model here"

andrewkern: I just want to make sure this doesn't sound like a bug that we have injected for no reason...

stdpopsim/drosophila_melanogaster.py
andrewkern: so unfortunately the authors seem to have had their own factor of two error and we should probably document it....

stdpopsim/drosophila_melanogaster.py
ndukler: I agree. I figured more experienced word-smithing was called for for this :)

ndukler:
Ok, wording changed. @andrewkern @jeromekelleher

PR #98, opened 2019-06-21
Add e. coli model documentation

ndukler:


andrewkern:
oops looks like there is a conflict here @ndukler. maybe you need to rebase against latest master?

ndukler:
Yeah, the conflict popped up when the d. mel documentation was merged. Should be fixed now.

PR #100, opened 2019-06-27
Generalise CLI run and decompose docs

jeromekelleher:
Some nice changes here I think:

1. The CLI uses the model object to create an argument parser and run the specific model. So, we don't need any model specific code in the CLI any more (just for you @reedacartwright!)
2. The documentation is split up into smaller chunks so that we can reuse it between the online HTML documentation and CLI. This requires a little dancing around to build the model docsstrings, but otherwise is reasonably clean. 
3. Add a "Population" object which allows us to handle metadata and documentation for populations much better.

The upshot now is we now get this in the CLI:
```
$ python3 -m stdpopsim homo-sapiens BrowningAmerica --help
usage: __main__.py homo-sapiens BrowningAmerica [-h] [--num-AFR NUM_AFR]
                                                [--num-EUR NUM_EUR]
                                                [--num-ASIA NUM_ASIA]
                                                [--num-ADMIX NUM_ADMIX]
                                                output

Demographic model for American admixture, taken from `Browning et al.
<http://dx.doi.org/10.1371/journal.pgen.1007385>`_. This model extends the
`Gravel et al. (2011) <https://doi.org/10.1073/ pnas.1019276108>`_ model of
African/European/Asian demographic history to simulate an admixed population
with admixture occurring 12 generations ago. The admixed population had an
initial size of 30,000 and grew at a rate of 5% per generation, with 1/6 of
the population of African ancestry, 1/3 European, and 1 2 Asian. This code was
ported over from `Supplementary File 1
<https://doi.org/10.1371/journal.pgen.1007385.s005>`_

positional arguments:
  output                The file to write simulated tree sequence to

optional arguments:
  -h, --help            show this help message and exit
  --num-AFR NUM_AFR     The number of samples to take from the AFR population
  --num-EUR NUM_EUR     The number of samples to take from the EUR population
  --num-ASIA NUM_ASIA   The number of samples to take from the ASIA population
  --num-ADMIX NUM_ADMIX
                        The number of samples to take from the ADMIX
                        population
```

I'm sure we can improve the population names and various bits of documentation.

If we agree this is a good way to go I'll apply it to the other organisms as well and tidy up some loose ends. 

One important thing to note here is that we now allow for sampling from the ancient admixed population. Since this will be at time 0, it's almost certainly wrong. There's a few things we can do here:

1. Add a ``sampling_time`` attribute to the populations, which is equal to the start time of these populations (0 by default, but we can set it to something else for these ancient pops).
2. Add a ``allow_sampling`` attribute to populations, and don't include options in the CLI if this is False (easy, but possibility stopping people doing things they want to do).
3. Add a ``sampling-time [population] [time]`` option to the CLI, which allows the user to manually specify when they want the sample to go in.

None of these cover the case where someone wants to put samples in at arbitrary times in the simulation. I think we can just assume someone who wants to do this can use the Python interface; it's too much complexity otherwise.


reedacartwright:
This looks great.

On the  'ancestral population' issue: would it be possible for  `add_model_runner` to detect if a population is not present at t=0, and remove that from the interface. I think allowing users to take historical samples requires a more comprehensive interface to specify both time and population, and maybe something that should not be supported in the CLI.

jeromekelleher:
Thanks @reedacartwright 

> On the 'ancestral population' issue: would it be possible for add_model_runner to detect if a population is not present at t=0, and remove that from the interface. I think allowing users to take historical samples requires a more comprehensive interface to specify both time and population, and maybe something that should not be supported in the CLI.

Yeah, you're right. We can add a flag to the Population object which allows the CLI to tell whether it should be sampled from or not, and just filter out those that aren't flagged as sampleable. We might do something more sophisticated later, but this will do fine for now and prevent people from obvious mistakes.

jeromekelleher:
Thanks @ndukler, some responses below.

> Remove the Ragsdale model documentation since I havn't done QC yet

Will do.

> Maybe have some automated way to iterate over the add_model_runner calls based on iterating though subclasses of the Model class?

We could definitely do this - I'll add this once the other species models follow the new framework.

> As part of the automated addition of models to CLI and documentation add finalized type property to model to ensure something isn't added to the documentation unless it has been QC'd

Yeah, good idea. Maybe we should be explicit about it and add a ``qc_done`` flag, which is fipped by the QCer.

> Is there a similar way to automate things at the level of species and per species genetic maps?

That's where I'd like to get to all right. I think we should wrap the model side of things up first though, and revisit the species/genomes/genetic maps then.

> I suppose my main thought is this is a potentially odd mix of automated documentation generation and the need for custom additions of code to specific functions. I'm wondering if we can get around some of this by more fully automating and consolidating all the properties that need to be set for this to happen in the species_name.py file. Does this seem reasonable or like over automation?
> At the same time I don't want to lock us into a structure which may prove inflexible for popSim phase II so I want to make sure that we think about how this framework will handle things like selection and mutation models. I apologize if I'm missing something here due to my relative inexperience with python.

Nothing to apologise about, these are all good ideas. The short answer is "I don't know yet, but that's the basic direction I think we should head in". Once we get a solid and clean user-facing API, I'm not so worried about getting all the back end structures exactly right. The main thing always is to make sure that users have a consistent experience and that we never break **their** code. Breaking our own code, on the other hand, is totally fine and a necessary part of evolving a code base.

jeromekelleher:
> Looks great to me @jeromekelleher. One question looking forward-- do we want to build different simulation model possibilities into the CLI at this stage? For instance, perhaps we want to be able to run the various demographic models under the DTWF instead of the Hudson model?

Good question - one to discuss later I think.

jeromekelleher:
> Good question - one to discuss later I think.

Ah yes; by later @andrewkern, I thought we had a call today so we could discuss it then. This is a tricky one, we should probably get this much fleshed out and merged first before thinking about it.

We could add a simple switch I guess, but how well that'll generalise to other stuff I don't know.

jeromekelleher:
I've updated this and made a few tweaks. I haven't updated the non-human models as we need to fix up the genome/chromosome infrastructure in general; this still needs refactoring. I'd suggest we merge this much now though @andrewkern, so we can't start working with it and figure out how the generic models are going to fit into this framework.

jeromekelleher:
ps. I'm not sure how to test that one line of code in the diff that's not being covered. I'd suggest we ignore codecov and merge.

reedacartwright:
I see that you also added logic to prevent the sampling of archaic samples. Great.

jeromekelleher:
I'm going to go ahead and merge this. There'll be a follow up PR which should tidy up the loose ends and start finalising the external API.

PR #103, opened 2019-07-10
small note on downloading recomb maps

petrelharp:
There's a lot more to document, but I noticed this step was missing.


jeromekelleher:
We shouldn't have to do this (they should be downloaded automatically on demand), but until this is fixed it's good to have the docs - thanks @petrelharp.

PR #105, opened 2019-07-23
Add SchiffelsZigzag demographic model

Chris1221:
I've included a basic model that is often used to test demographic inference software (used in the MSMC paper and others), essentially converting an `ms` command to an `msprime` model. Here is the model run through the analysis pipeline with msmc (and with the `steps` modified for plotting to show the full behaviour):

image

I've used a different `N0` from figure 10 of Schiffel's supplementary material but otherwise the model is unchanged. 

The `ms2msp_nt` is a convenience function for calculating the new `n_t` after a given number of generations given a growth rate and `N0`. 

I'm not sure if this will be stupendously useful to others, but I thought I would include it just in case anyone else wants to benchmark against it.

### DIFF
--- a/stdpopsim/homo_sapiens.py
+++ b/stdpopsim/homo_sapiens.py
@@ -704,15 +704,9 @@ class SchiffelsZigzag(models.Model):
         SchiffelsZigzag
 
     Model Description:
-        A validation model used by Schiffels and Durbin (2014) and Terhorst and Terhorst, Kamm, and Song (2017) with periods
+        A validation model used by Schiffels and Durbin (2014) and
+        Terhorst and Terhorst, Kamm, and Song (2017) with periods
         of exponential growth and decline in a single population.
-        The original SCRM command for this model was:
-
-        .. code_block::
-
-            scrm 8 1 -N0 14312 -t 1431200 -r 400736 2000000000
-            -eN 0 1 -eG 0.000582262 1318.18 -eG 0.00232905 -329.546 -eG 0.00931619 82.3865
-            -eG 0.0372648 -20.5966 -eG 0.149059 5.14916 -eN 0.596236 0.1 -seed 1 -T -L -p 10 -l 300000.
 
     Model population indexes:
         - Single population: 0
@@ -721,49 +715,42 @@ class SchiffelsZigzag(models.Model):
         python -m stdpopsim homo-sapiens SchiffelsZigzag -h
 
     Citation:
-       Schiffels, S., & Durbin, R. (2014). Inferring human population size and separation history from multiple genome sequences. Nature Genetics. https://doi.org/10.1038/ng.3015
-    """  # noqa: E501
+       Schiffels, S., & Durbin, R. (2014). Inferring human population size and
+       separation history from multiple genome sequences. Nature Genetics.
+       https://doi.org/10.1038/ng.3015
+    """
     author = "Schiffels and Durbin"
     year = 2014
     doi = "https://doi.org/10.1038/ng.3015"
-
-    def ms2msp_nt(self, n0, s, a, N0=14312):
-        """
-        Convert the ms growth rates to those more appropriate for
-        msprime and calculate n(t) according to them.
-        The 4N0 correction is relative to the absolute N0, rather
-        than the epoch specific one. """
-        return n0 * math.exp(-s*a)
 
     def __init__(self):
         super().__init__()
 
         self.generation_time = 29
         N0 = 14312
-        scale = 4 * N0
 
-        g_1 = 1318.18 / scale
-        t_1 = 0.000582262 * scale  # (generations)
+        g_1 = 0.023025
+        t_1 = 33.333
         n_1 = N0
 
-        g_2 = -329.546 / scale
-        t_2 = 0.00232905 * scale
-        n_2 = self.ms2msp_nt(n_1, t_2 - t_1, g_1)
+        g_2 = -0.005756
+        t_2 = 133.33
+        n_2 = N0/10
 
-        g_3 = 82.3865 / scale
-        t_3 = 0.00931619 * scale
-        n_3 = self.ms2msp_nt(n_2, t_3 - t_2, g_2)
+        g_3 = 0.0014391
+        t_3 = 533.33
+        n_3 = N0
 
-        g_4 = -20.5966 / scale
-        t_4 = 0.0372648 * scale
-        n_4 = self.ms2msp_nt(n_3, t_4 - t_3, g_3)
+        g_4 = -0.00035977
+        t_4 = 2133.33
+        n_4 = N0/10
 
-        g_5 = 5.14916 / scale
-        t_5 = 0.149059 * scale
-        n_5 = self.ms2msp_nt(n_4, t_5 - t_4, g_4)
+        g_5 = 8.99448e-5
+        t_5 = 8533.33
+        n_5 = N0
 
-        n_ancient = 0.1 * N0
-        t_ancient = 0.596236 * scale
+        n_ancient = N0/10
+        t_ancient = 34133.31
 
         self.population_configurations = [
             msprime.PopulationConfiguration(
--- a/stdpopsim/homo_sapiens.py
+++ b/stdpopsim/homo_sapiens.py
@@ -696,3 +696,95 @@ class RagsdaleArchaic(models.Model):
             msprime.MassMigration(
                 time=T_nean_split, source=3, destination=0, proportion=1.0)
         ]
+
+
+class SchiffelsZigzag(models.Model):
+    """
+    Model Name:
+        SchiffelsZigzag
+
+    Model Description:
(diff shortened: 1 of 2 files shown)


Chris1221:
Sorry, I forgot to run `flake` and the tests locally but it should be fixed up now.

andrewkern:
this is cool and IMO a good addition, but I wonder what others think-- do we want test cases rather than models estimated from data as part of the project? my take is "yes"

Chris1221:
Sorry @andrewkern, maybe I should have brought it up in a call first and fielded everyone's opinions. If test cases are not within the scope of the project, please don't worry about it!

jeromekelleher:
I think this is a good idea, and I agree with @andrewkern on keeping it out of ``homo_sapiens``. ``generic_models`` sounds like a good place to put this to me.

andrewkern:
@Chris1221 if you can make this simple change and then squash your commits I think this is ready to merge in.

Chris1221:
Great! I like this direction. Should I make a single large chromosome with a flat recombination map? That seems the least controversial and a decent place to start generic simulations.

andrewkern:
i think we are covered by using the chromosome definitions that exist in other species along with a constant rate? would we need anything more?

Chris1221:
I could be wrong, but to be consistent with the analysis pipeline, I think we may have to redefine them in the new file. I don't think we explicitly need anything else, but at the moment (in `n_t/Snakefile`):

```py
# The analysis species
species = getattr(stdpopsim,config["species"])

# The specific model you would like to run
model_class = getattr(species,config["model"])
model = model_class()

# The genetic map you would like to use.
genetic_map_class = getattr(species,config["genetic_map"])
genetic_map = genetic_map_class()

# The names of all chromosomes to simulate, separated by commas
# Use "all" to simulate all chromsomes for the genome
chrm_list = [chrom for chrom in species.genome.chromosomes]
if(config["chrm_list"] != "all"):
    chrm_list = [chr for chr in config["chrm_list"].split(",")]
```

the recombination map and model are both pulled from the specific species submodule of interest. I'm not sure what the best way to address this is, but to my eye we can either make `generic_models` a special case and address it in the analysis repo in a different way (a seperate config parameter for "species of interest" which is decoupled from the generic model from which to pull the chromosome definitions for example) or not make it a special case, and define these within the `generic_models` file (and then allow to tweak these to address specific concerns). 

Sorry if I've made this more complicated, if you would prefer I just move over the code to a new file, that's fine by me as well!

jeromekelleher:
I don't think we should define chromosomes in the ``generic _models`` @Chris1221, this seems artificial to me. The ``analysis`` repo is separate, and we can figure out how to include this model in there later if we need it.

stdpopsim/homo_sapiens.py
```diff
@@ -696,3 +696,95 @@ def __init__(self):
...
+            -eG 0.0372648 -20.5966 -eG 0.149059 5.14916 -eN 0.596236 0.1 -seed 1 -T -L -p 10 -l 300000.
+
+    Model population indexes:
+        - Single population: 0
+
+    CLI help:
+        python -m stdpopsim homo-sapiens SchiffelsZigzag -h
+
+    Citation:
+       Schiffels, S., & Durbin, R. (2014). Inferring human population size and separation history from multiple genome sequences. Nature Genetics. https://doi.org/10.1038/ng.3015
```
jeromekelleher: This line can be split. We should only be breaking the 90 char limit here where we can't avoid it, like when we have a long URL.

stdpopsim/homo_sapiens.py
```diff
@@ -696,3 +696,95 @@ def __init__(self):
...
         ]
+
+
+class SchiffelsZigzag(models.Model):
+    """
+    Model Name:
+        SchiffelsZigzag
+
+    Model Description:
+        A validation model used by Schiffels and Durbin (2014) and Terhorst and Terhorst, Kamm, and Song (2017) with periods
```
jeromekelleher: This line is also too long.

stdpopsim/homo_sapiens.py
```diff
@@ -696,3 +696,95 @@ def __init__(self):
...
+
+    Model population indexes:
+        - Single population: 0
+
+    CLI help:
+        python -m stdpopsim homo-sapiens SchiffelsZigzag -h
+
+    Citation:
+       Schiffels, S., & Durbin, R. (2014). Inferring human population size and separation history from multiple genome sequences. Nature Genetics. https://doi.org/10.1038/ng.3015
+    """  # noqa: E501
```
jeromekelleher: I think we should remove the ``noqa`` here, there's no need for long lines in this case.

stdpopsim/homo_sapiens.py
```diff
@@ -696,3 +696,95 @@ def __init__(self):
...
+
+class SchiffelsZigzag(models.Model):
+    """
+    Model Name:
+        SchiffelsZigzag
+
+    Model Description:
+        A validation model used by Schiffels and Durbin (2014) and Terhorst and Terhorst, Kamm, and Song (2017) with periods
+        of exponential growth and decline in a single population.
+        The original SCRM command for this model was:
```
jeromekelleher: What's the reasoning for including this in the documentation? I don't see how it's going to help users TBH; presumably this provenance is somewhere in the original papers if we need to track it down?

stdpopsim/homo_sapiens.py
```diff
@@ -696,3 +696,95 @@ def __init__(self):
...
+        return n0 * math.exp(-s*a)
+
+    def __init__(self):
+        super().__init__()
+
+        self.generation_time = 29
+        N0 = 14312
+        scale = 4 * N0
+
+        g_1 = 1318.18 / scale
```
jeromekelleher: What are the units of these hard-coded values? I think it would be more readable if put in the values directly in the native scaling --- it's going to be very confusing for readers of the code when they are used to working in generations for the rest of the models. 

I can see how this is simpler when you start with the ``ms`` description and want to translate from that, but I think it'd be a lot clearer to do the translation fully and remove the ``ms2msp_nt`` function.

stdpopsim/homo_sapiens.py
Chris1221: You're right, given that it's probably best to just natively code the model this is superfluous. I've removed it.

stdpopsim/homo_sapiens.py
Chris1221: The unit of the growth parameter is 4N_0 generations, so I've divided out the 4N0 to fit with `msprime`. I agree with your point, it's definitely going to be more readable and maintainable if it's simply natively coded. I've done this in the new commit.

Chris1221:
I definitely see your point, and I think that natively coding it is the more maintainable approach. I was doing it quickly and as you say, just converting over the `ms` command. I've addressed your comments (thank you for the quick and thorough review) in the latest commit.

FWIW I agree that the model as it is implemented is intentionally human-centric, though I do think there is value in starting a generic models module.

andrewkern:
> I guess this model is also somewhat human centric, since we're assuming a few human parameters within it (generation time, Ne is human-ish), so perhaps it would be better to keep it in homo_sapiens after all. @andrewkern, what do you think?

I've been thinking about the generic cases a bit. I think it would be ideal to be able to use generic models as mix-in functions, such that they could be brought in in an organism agnostic way. E.g., coming up with a way to extend the set of demographic models available to each organism through importing a mixin. what do you think of that idea?

ndukler:
> > I guess this model is also somewhat human centric, since we're assuming a few human parameters within it (generation time, Ne is human-ish), so perhaps it would be better to keep it in homo_sapiens after all. @andrewkern, what do you think?
> 
> I've been thinking about the generic cases a bit. I think it would be ideal to be able to use generic models as mix-in functions, such that they could be brought in in an organism agnostic way. E.g., coming up with a way to extend the set of demographic models available to each organism through importing a mixin. what do you think of that idea?

Apologies if I'm misunderstanding something but as much as I like the modularity of this idea I think there is too much complexity in the potential range of generic models to really do this well. For example in this model alone in addition to generation time, we need two Ne values and potentially a bunch of generation time to make this sensible for different organisms. It might make sense to just implement organism specific version of these models. The only way around this might be to define a bunch of standard multipliers (e.g. Ne, # of generations, etc. ) and use one organism (human for example) as the reference level. But this might be too involved. We can also just try implementing this separately for each species and then look at it afterwards and see if we can abstract it in any meaningful way.

jeromekelleher:
I'm with @ndukler here - this pariticular model is quite human centric so I think it's best to acknowledge this up front. We can do generic stuff later on, but it does get hard to to generally and well.

I say we merge.merg

jeromekelleher:
On the call today we decided to merge this as in into homo_sapiens, and then figure out how to generalise it to the genetic mixin strategy over in #106.

andrewkern:
okay as per our call I'm merging this PR.

PR #108, opened 2019-08-02
first commit for generic_models bundle

andrewkern:
did a bit of work pulling together a `generic_models` bundle that will provide a set of simple models for `stdpopsim` that will be accessible for each of the organisms in a way that should not break the current usage paradigms. that is to say that this set of generic models will be accessible for each organism such that a user can say `homo_sapiens.generic_constant_size()` and get the generic model back. see `generic_example.py` for a usage example. 

the way I have done this involves calling `import` on `generic_models.py` in each of the organism specific files. thus these generic models will also be available at the level of `stdpopsim.generic_model`. Hopefully this is a point of convenience rather than confusion?

additionally I think the generic models are ready to go for the CLI. thoughts on this?

### DIFF
--- a/stdpopsim/arabidopsis_thaliana.py
+++ b/stdpopsim/arabidopsis_thaliana.py
@@ -7,7 +7,7 @@ import numpy as np
 import stdpopsim.models as models
 import stdpopsim.genomes as genomes
 import stdpopsim.genetic_maps as genetic_maps
-
+import stdpopsim.generic_models as generics # NOQA
 
 ###########################################################
 #
--- a/stdpopsim/drosophila_melanogaster.py
+++ b/stdpopsim/drosophila_melanogaster.py
@@ -7,7 +7,7 @@ import msprime
 import stdpopsim.models as models
 import stdpopsim.genomes as genomes
 import stdpopsim.genetic_maps as genetic_maps
-
+import stdpopsim.generic_models as generics # NOQA
 
 ###########################################################
 #
@@ -129,8 +129,8 @@ class SheehanSongThreeEpoch(models.Model):
 
         # Population metadata
         metadata_afr = {
-            "name": "AFR_dmel",
-            "description": "African D. melanogaster population"
+           "name": "AFR_dmel",
+           "description": "African D. melanogaster population"
         }
 
         # Single population in this model
--- a/stdpopsim/generic_models.py
+++ b/stdpopsim/generic_models.py
@@ -1,0 +1,62 @@
+# generic_models
+import msprime
+
+
+#########
+# Generic Model Mixin class
+# provides generic models that can be
+# initialized with methods
+#
+#
+########
+
+
+class ConstantSizeMixin(object):
+    def __init__(self, N):
+        """
+        Model Name:
+            ConstantSizeMixin
+
+        Model Description:
+            Generic model of constant size. Uses default_population_size
+            depending on organism
+
+        Model population indexes:
+            - Population 0: 0
+
+        CLI help:
+            python -m stdpopsim homo-sapiens GenericConstantSize -h
+        """  # noqa: E501
+        self.population_configurations = [
+            msprime.PopulationConfiguration(initial_size=N)
+        ]
+        self.migration_matrix = [[0]]
+        self.demographic_events = []
+
+
+class TwoEpochMixin(object):
+    def __init__(self, N1, N2, t):
+        """
+        Model Name:
+            TwoEpochMixin
+
+        Model Description:
+            Generic model of a single population with
+            piecewise constant population size with a single change.
+            Defaults depend on organism, but population sizes and time
+            should be set by user
+
+        Model population indexes:
+            - Population 0: 0
+
+        CLI help:
+            python -m stdpopsim homo-sapiens GenericTwoEpoch -h
+        """  # noqa: E501
+        self.population_configurations = [
+            msprime.PopulationConfiguration(initial_size=N1)
+        ]
+        self.migration_matrix = [[0]]
+        self.demographic_events = [
+            msprime.PopulationParametersChange(
+                time=t, initial_size=N2, growth_rate=0, population_id=0),
+        ]
--- a/stdpopsim/homo_sapiens.py
+++ b/stdpopsim/homo_sapiens.py
@@ -9,7 +9,7 @@ import msprime
 import stdpopsim.models as models
 import stdpopsim.genomes as genomes
 import stdpopsim.genetic_maps as genetic_maps
-
+import stdpopsim.generic_models as generic_models
 logger = logging.getLogger(__name__)
 
 
@@ -119,6 +119,7 @@ genome = genomes.Genome(
     chromosomes=_chromosomes,
     default_genetic_map=HapmapII_GRCh37.name)
 
+
 #
 # Experimental interface used to develop the CLI.
 #
@@ -166,7 +167,34 @@ def chromosome_factory(name, genetic_map=None, length_multiplier=1):
 default_generation_time = 25
 
 
-class GutenkunstThreePopOutOfAfrica(models.Model):
+class HomoSapiensModel(models.Model):
+    """
+    TODO: documentation
+    """
+    def __init__(self):
+        super().__init__()
+        self.generation_time = default_generation_time
+        self.default_population_size = 10000
+
+
+class GenericConstantSize(HomoSapiensModel, generic_models.ConstantSizeMixin):
+    def __init__(self):
+        HomoSapiensModel.__init__(self)
+        generic_models.ConstantSizeMixin.__init__(self, self.default_population_size)
+
+
+class GenericTwoEpoch(HomoSapiensModel, generic_models.TwoEpochMixin):
+    def __init__(self, n2=None, t=None):
+        HomoSapiensModel.__init__(self)
+        n1 = self.default_population_size
+        if n2 is None:
+            n2 = n1 / 2.0
+        if t is None:
+            t = n1 / 100
+        generic_models.TwoEpochMixin.__init__(self, n1, n2, t)
+
+
+class GutenkunstThreePopOutOfAfrica(HomoSapiensModel):
     """
     Model Name:
         GutenkunstThreePopOutOfAfrica
@@ -212,7 +240,7 @@ class GutenkunstThreePopOutOfAfrica(models.Model):
         N_EU0 = 1000
         N_AS0 = 510
         # Times are provided in years, so we convert into generations.
-        self.generation_time = default_generation_time
+        # self.generation_time = default_generation_time
         T_AF = 220e3 / self.generation_time
         T_B = 140e3 / self.generation_time
         T_EU_AS = 21.2e3 / self.generation_time
@@ -260,7 +288,7 @@ class GutenkunstThreePopOutOfAfrica(models.Model):
         ]
 
 
-class TennessenTwoPopOutOfAfrica(models.Model):
+class TennessenTwoPopOutOfAfrica(HomoSapiensModel):
     """
     Model Name:
         TennessenTwoPopOutOfAfrica
@@ -301,7 +329,6 @@ class TennessenTwoPopOutOfAfrica(models.Model):
     def __init__(self):
         super().__init__()
 
-        self.generation_time = default_generation_time
         T_AF = 148e3 / self.generation_time
         T_OOA = 51e3 / self.generation_time
         T_EU0 = 23e3 / self.generation_time
@@ -359,7 +386,7 @@ class TennessenTwoPopOutOfAfrica(models.Model):
         ]
 
 
-class TennessenOnePopAfrica(models.Model):
+class TennessenOnePopAfrica(HomoSapiensModel):
     """
     Model Name:
         TennessenOnePopAfrica
@@ -396,7 +423,6 @@ class TennessenOnePopAfrica(models.Model):
     def __init__(self):
         super().__init__()
 
-        self.generation_time = default_generation_time
         T_AF = 148e3 / self.generation_time
         T_EG = 5115 / self.generation_time
 
@@ -426,7 +452,7 @@ class TennessenOnePopAfrica(models.Model):
         ]
 
 
-class BrowningAmerica(models.Model):
+class BrowningAmerica(HomoSapiensModel):
     """
     Model Name:
         BrowningAmerica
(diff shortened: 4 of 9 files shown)


stdpopsim/drosophila_melanogaster.py
```diff
@@ -7,6 +7,7 @@
 import stdpopsim.models as models
 import stdpopsim.genomes as genomes
 import stdpopsim.genetic_maps as genetic_maps
+from stdpopsim.generic_models import * # NOQA
```
jeromekelleher: As a style thing, I tend to avoid ``import *``. The idea is that it's nice to know the difference between a function that's defined in the current module and one that's come from somewhere else. Then, when you refer to it, *where* the function is coming from it explicit, so you know exactly where to find it when you go looking.

stdpopsim/drosophila_melanogaster.py
```diff
@@ -195,18 +197,20 @@ def __init__(self):
...
-            "description": "European D. melanogaster population"
-        }
+        # metadata_afr = {
+        #     "name": "AFR_dmel",
+        #     "description": "African D. melanogaster population"
+        # }
+        # metadata_eu = {
+        # "name": "EU_dmel",
+        # "description": "European D. melanogaster population"
+        # }
```
jeromekelleher: I'm not seeing any changes in here other than removing the metadata - is the idea to inherit from one of the constant models?

stdpopsim/generic_models.py
```diff
@@ -0,0 +1,87 @@
...
+import stdpopsim.models as models
+
+
+#########
+# Generic Demographic Models
+#
+#
+########
+
+class generic_constant_size(models.Model):
```
jeromekelleher: These are classes, so we should follow the SnakeCase convention like the other classes. -> GenericConstantSize

stdpopsim/generic_models.py
```diff
@@ -0,0 +1,87 @@
...
+    Model Name:
+        generic_constant_size
+
+    Model Description:
+        single-size, single-population model
+
+    Model population indexes:
+        - Population1 : 0
+
+    Parameter Table:
```
jeromekelleher: We can probably just ditch the Parameter Table section here, probably won't be relevant.

stdpopsim/generic_models.py
```diff
@@ -0,0 +1,87 @@
...
+            :file: ../docs/parameter_tables/homo_sapiens/GutenkunstThreePopOutOfAfrica_params.csv
+
+    CLI help:
+        python -m stdpopsim homo-sapiens generic_constant_size -h
+
+    Citation:
+
+    """  # noqa: E501
+
+    def __init__(self, N=100):
```
jeromekelleher: Having a default seems pretty arbitrary here. I'd say we force people to choose an N.

stdpopsim/generic_models.py
```diff
@@ -0,0 +1,87 @@
...
+            :file: ../docs/parameter_tables/homo_sapiens/GutenkunstThreePopOutOfAfrica_params.csv
+
+    CLI help:
+        python -m stdpopsim homo-sapiens generic_two_epoch -h
+
+    Citation:
+
+    """  # noqa: E501
+
+    def __init__(self, N1=1000, N2=100, t=10):
```
jeromekelleher: Ditto with the defaults here.

tests/test_homo_sapiens.py
```diff
@@ -298,6 +298,46 @@ def test_debug_runs(self):
         self.assertGreater(len(s), 0)
 
 
+class TestGenericConstantSize(unittest.TestCase):
+    """
+    Basic tests for the TestGenericConstantSize model.
```
jeromekelleher: Typo here - testing the GenericConstantSize model

tests/test_homo_sapiens.py
```diff
@@ -298,6 +298,46 @@ def test_debug_runs(self):
...
+        model = homo_sapiens.generic_constant_size(100)
+        output = io.StringIO()
+        model.debug(output)
+        s = output.getvalue()
+        self.assertGreater(len(s), 0)
+
+
+class TestGenericTwoEpoch(unittest.TestCase):
+    """
+    Basic tests for the TestGenericTwoEpoch model.
```
jeromekelleher: Ditto

stdpopsim/drosophila_melanogaster.py
andrewkern: needed to remove the meta-data. Not sure why but it was throwing errors.

stdpopsim/generic_models.py
andrewkern: ugh. my bad here

stdpopsim/generic_models.py
andrewkern: so I needed to set defaults to deal with the tests that use `models.all_models()`. `all_models()` works fine for all of the organism specific models because they carry around their own parameters. For the generics, I went with defaults to get around that. suggestions?

stdpopsim/drosophila_melanogaster.py
andrewkern: I agree this is ugly. I can get around this if generic models inherit from the model class directly and then the organism specific models inherit from the generic model class as you suggest above. My goal with this inherit was to allow for calls like `homo_sapiens.generic_constant_size()` without having to write a new `import` line for each new model.

andrewkern:
> I like it @andrewkern, it's a clean use of OO design. It looks to me like organism specific models should be able to inherit directly from the generic models (although, so far the models are so simple that there's not much to be gained from this inheritance).

Any preference between defining a `GenericModel` class that inherits from `Model` and that in turn each of the organism specific models inherit from, or defining the generic models as I have right now? I think the former might be cleaner.

stdpopsim/drosophila_melanogaster.py
jeromekelleher: Yeah - we could also just put the generic models code into the ``models.py`` file --- probably overkill to have a separate module since, as you say, a lot of concrete models will be inheriting directly from these (so, we'd have models.ConstantSize, etc).

stdpopsim/generic_models.py
jeromekelleher: Hmm. Maybe we should make the default 1 then so, so that we're working in coalescent units by default like msprime?

stdpopsim/generic_models.py
jeromekelleher: But yeah, how do we deal with situations like this? Maybe the tests using ``all_models`` are just over generalising, and we should ditch it? Some things will need parameters, and putting in fake ones by default just to please a test case doesn't seem like the right approach to me.

jeromekelleher:
> Any preference between defining a GenericModel class that inherits from Model and that in turn each of the organism specific models inherit from, or defining the generic models as I have right now? I think the former might be cleaner

I don't see why we'd need this --- why not something like (e.g.) homo_sapiens.SomebodySomething -> models.ConstantSize -> models.Model (or, perhaps models.BaseModel)? I think the OO hierarchy already captures the idea that it's generic, we don't need to highlight this.

stdpopsim/drosophila_melanogaster.py
andrewkern: what about this proposal: we keep the `generic_models.py` module and then have each organism `import generic_models as generics` such that one could call `homo_sapiens.generics.GenericConstantSize()`? 

That's sort of a middle of the road solution.

petrelharp:
I'd vote for either `homo_sapiens.generics.ConstantSize` or `homo_sapiens.GenericConstantSize`, because if it was under `models` I'd always be wanting to look for *all* the models there, e.g. typing `homo_sapiens.models.GutenkunstEtAlThreePop`.

And, I think either of those is fine.

jeromekelleher:
I'm getting a bit lost here to be honest and will need to try stuff out to see what works. The main I want to avoid is having too much substructure as part of the public API within the ``stdpopsim`` package. Once you add hierarchy into the public namespace you're stuck with it forever.

stdpopsim/generic_models.py
andrewkern: easy enough

stdpopsim/generic_models.py
andrewkern: having this limited to the generic cases doesn't bother me much and having the tests that use `all_models` are important. can you see a clean solution here?

andrewkern:
I just pushed a version where I have moved the `import` scheme around. take a look at `generic_example.py` or the tests. also maybe a quick call to get this straight?

jeromekelleher:
> also maybe a quick call to get this straight?

Call would be good - have you got some time tomorrow?

andrewkern:
yep. will email you.

stdpopsim/arabidopsis_thaliana.py
```diff
@@ -7,7 +7,7 @@
 import stdpopsim.models as models
 import stdpopsim.genomes as genomes
 import stdpopsim.genetic_maps as genetic_maps
-
+import stdpopsim.generic_models as generics # NOQA
```
jeromekelleher: Doesn't  this mean that we're defining exactly the same models (with no modifications) multiple times just with different paths? This seems like a bad thing to me. What's wrong with defining a generic population model and letting users set the generation time, etc, themselves using the values defined in the various species modules? I feel like we're over-designing this.

andrewkern:
Okay @jeromekelleher take a look at what I have done to refactor this as a mixin. Right now I have two models going in `homo_sapiens.py`. if you like this design I will expand it to the other organisms.

stdpopsim/arabidopsis_thaliana.py
andrewkern: shouldn't need the `NOQA`s now that i've moved it to a Mixin

stdpopsim/generic_models.py
```diff
@@ -1,70 +1,34 @@
...
-
-    CLI help:
-        python -m stdpopsim homo-sapiens generic_constant_size -h
-
-    Citation:
-
-    """  # noqa: E501
-
-    def __init__(self, N=1):
+class GenericMixin(object):
```
jeromekelleher: Might be a bit more OO to have 
```
class ConstantSizeMixin(object):
    # Standard docs so that the CLI etc will pick these up and we don't have to duplicate them
    def __init__(self, N):
          self.population_configurations = [
            msprime.PopulationConfiguration(initial_size=N)
        ]
        self.migration_matrix = [[0]]
        self.demographic_events = []

class TwoEpochMixin(object):
     # as before
```

stdpopsim/homo_sapiens.py
```diff
@@ -169,7 +169,38 @@ def chromosome_factory(name, genetic_map=None, length_multiplier=1):
...
+    TODO: documentation
+    """
+    def __init__(self):
+        super().__init__()
+        self.generation_time = default_generation_time
+        self.default_population_size = 10000
+
+
+class GenericConstantSize(HomoSapiensModel, generics.GenericMixin):
+    """
```
jeromekelleher: No docstring - should be inherited from the Mixin (I'm not sure whether it needs to come first or last for this to happen), and so the online docs and CLI will get the standard documentation.

stdpopsim/homo_sapiens.py
```diff
@@ -169,7 +169,38 @@ def chromosome_factory(name, genetic_map=None, length_multiplier=1):
...
+    TODO: documentation
+    """
+    def __init__(self):
+        super().__init__()
```
jeromekelleher: Probably clearest to do:
```
def __init__(self):
    HomoSapiensModel.__init__()
    generics.ConstantSizeMixin(self.default_population_size)
```

stdpopsim/homo_sapiens.py
```diff
@@ -169,7 +169,38 @@ def chromosome_factory(name, genetic_map=None, length_multiplier=1):
...
+class GenericTwoEpoch(HomoSapiensModel, generics.GenericMixin):
+    """
+    TODO: documentation
+    """
+    def __init__(self):
+        super().__init__()
+        n1 = self.default_population_size
+        n2 = n1 / 2.0
+        t = n1 / 100
+        self.generic_two_epoch(n1, n2, t)
```
jeromekelleher: How about
```
 def __init__(self, n2=None, t=None):
        HomoSapiensModel.__init__()
        n1 = self.default_population_size
        if n2 is None:
             n2 = n1 / 2.0
        if t is None:
            t = n1 / 100
        generics.TwoEpoch.__init__(n1, n2, t)
```

stdpopsim/generic_models.py
andrewkern: can do

stdpopsim/homo_sapiens.py
jeromekelleher: By "it" I mean the order of the parent classes. We want ot inherit the docstring from the Mixin.

petrelharp:
I had a look at the suggested "mixin" way of doing this. My two cents here is that I can see how the UI will be somewhat cleaner, but the structure of the code seems significantly more confusing (and more voluminous).

jeromekelleher:
> I had a look at the suggested "mixin" way of doing this. My two cents here is that I can see how the UI will be somewhat cleaner, but the structure of the code seems significantly more confusing (and more voluminous).

:man_shrugging: I hear ya, but I'm not sure there's a better way to do it if we want to abstract these generic models out.

andrewkern:
okay guys here are the two generic human models cleaned up to have separate mixin classes per model which sit in `generic_models.py`. take a look and if you like it I will keep going.

IMO this requires a bit more coding but is clean

andrewkern:
mind you there is a bunch of clean up I now have to do before we can merge. just want to get high level opinions as before on this use of the mixins

jeromekelleher:
I realised looking through this that I haven't finished #100 yet, which will feed into the design of this. I'll get stuck into that this evening and ping you @andrewkern.

andrewkern:
cool. should I built out the generics for the other organisms and do cleanup in the meanwhile? are you sufficiently happy with the `ConstantSizeMixin` class and the homo_sapiens specific versions?

jeromekelleher:
OK, let's get this cleaned up and merged @andrewkern and then get #100 in. Let's get it all working for the human models first; there's a fair bit of reactoring to be done and mights as well do it all at once.

jeromekelleher:
Squasharoo?

andrewkern:
> Squasharoo?

squashed

jeromekelleher:
Thanks @andrewkern, merging. I'll update #100 ASAP and aim to get that merged in the next day or twol.

Issue #21, opened 2019-02-10
Species.genome not rendering correctly on RTD


jeromekelleher:
The documentation for the ``species.genome`` variable is supposed to give details abut the variable, but is giving details about the class instead. This is possibly due to RTD using an older version of sphinx. 

One easy workaround is to put the genome documentation directly into the RST file rather than as source comments (probably less brittle as well, given the non-standard #: comment format required).

jeromekelleher:
Not relevant to new API in #110.

PR #110, opened 2019-08-10
Simplify external API with factory functions

jeromekelleher:
This is not ready for review yet, but the idea is to build on #100 to generalise the change made there to all the species, and to simplify the external API. In the end, it should look something like:

```python
import stdpopsim as sps

contig = sps.contig_factory("homo_sapiens", "chr22", genetic_map="GRCh37")
model = sps.homo_sapiens.BrowningAmerica()
ts = model.run(contig, num_afr=10)
```
The idea is the Contig object contains the details about what's to be simulated (which might be a subsection of the chromosome, or many chromosomes). The contig_factory will then be the only documented external part of the API, hiding away all the messy details of how we're setting this stuff up.

It's not there yet, but that's general direction of travel.


andrewkern:
just looked through this. i think my continuing to build out the generics and modify that `all_models()` function should not impact this PR at all.

jeromekelleher:
Made some good progress here @andrewkern --- going to try and finish things up interface-wise tomorrow.

jeromekelleher:
Am working my way through a hefty refactor here. The idea is to isolate the user from the internal details of how we organise the models, and to make the "species" object the central hook. So, we can do stuff like this now:
```python
import stdpopsim

species = stdpopsim.get_species("homo_sapiens")
contig = species.get_contig("chr22", length_multiplier=0.1)

model = species.get_model("constant")
ts = model.run(contig, samples)

model = species.get_model("2_epoch")
ts = model.run(contig, samples)

model = species.get_model("ooa", 3)
# 2 sampes from pop 0, 1 from pop 2
samples = model.get_samples(2, 0, 1)
ts = model.run(contig, samples)
```

The first argument for `species.get_model` is the "kind" of model. This is a short identifier, which will lead to less typing on the CLI and hopefully easier to remember then GutenkunsThreePopOutOfAfrica or whatever. Notice the generics are just normal classes now which we instantiate and register with the species definition.

We should be able to build a nice lookup table of all the models for each species with this as well, which will be good for documentation.

andrewkern:
I really like the idea of moving to the species as the hook. this is leading to much cleaner code. I take it this isn't yet ready for a review?

jeromekelleher:
Great, thanks @andrewkern. No, it's not ready for review. I've realised we're doing some pretty convoluted stuff just for the sake of getting sphinx documentation. It might be simpler to get into how sphinx works. I'll ping you when it's ready to view.

The goal anyway is to decouple things from the static Python class structure; it's all too bound up in details of this.

jeromekelleher:
This is coming along nicely, resolving quite a few outstanding issues. Notes to self

- Change CLI to simulate a constant size model by default, and make other models options. The sample sizes are positional arguments, and the number of args depends on the model

- Can then also add commands for cache management, which are badly wanted. (i.e., add a ``download-maps`` command, which will download all genetic maps to the cache)

- Need to make the population size for the species a properly citable property to that we can write out citations properly in the default case.

jeromekelleher:
I think this is ready for review now @andrewkern, @petrelharp, @ndukler. I've changed a fair bit, so I'm not sure there's much point in going through the diff; probably best to check out a local copy of the code and have a quick play with it. In particular, you'll need to build the docs to see the new catalog, which is central.

Main changes:

- Add the concept of a "catalog", which contains the species that we have defined in stdpopsim, which then provide methods to get genetic maps and simulation models. The catalog gives full details of these species and models as part of the online documentation. Part of making this possible is breaking away from defining models/genetic maps etc *statically* as subclasses, but rather making them *instances* of a particular Model or GeneticMap. (This process hasn't been completed, but it's done enough to hide these messy details from end users.)

- Add the idea of an "id" which is used as a human-readable handle for the various objects. For species, we use the 3+3 abbreviation scheme, giving "homsap" and "dromel". (Details of this are up for discussion, of course).

- Refactor the CLI so that ``--model`` is an option and we simulate a constant size population by default.

- Lots of other refactorings.

Note that this **will** totally break any code using the current API. I'm happy to go through the ``analysis`` repo and fix it up though, if we agree on making these changes. 

Ideally I'd like to merge this, and perhaps release an alpha version of stdpopsim to PyPI this week. I think it'd be good to give a demo on where the software is at on Sunday also, so that we can get rapid feedback from people.

andrewkern:
nice @jeromekelleher! I'll look at this first thing in the morning

stdpopsim/cli.py
```diff
@@ -7,6 +7,7 @@
 import logging
 import platform
 import sys
+import textwrap
 import resource
```
andrewkern: TODO: need some documentation on the CLI

stdpopsim/cli.py
```diff
@@ -138,100 +138,137 @@ def summarise_usage():
...
-                parser.add_argument(f"--num-{name}", help=help_text, default=0, type=int)
-                sample_arg_names.append("num_{}".format(name))
+                models_text += indent * 2
+                models_text += f"{population.name}: {population.description}\n"
+        models_text += "\n"
+
+    description_text = textwrap.fill(header) + "\n" + models_text
+
+    species_parser = parser.add_parser(
+        f"sim-{species.id}",
```
andrewkern: adding the `sim-` species id means a bit of extra typing. can we ditch the `sim-` prefix here and instead have the user just start with the species name?

stdpopsim/cli.py
```diff
@@ -96,7 +95,7 @@ def write_output(ts, args):
     ts.dump(args.output)
 
 
-def write_citations(chromosome, model):
+def write_citations(contig, model):
```
andrewkern: this is an awesome, thoughtful touch.

stdpopsim/models.py
```diff
@@ -242,40 +204,53 @@ def debug(self, out_file=sys.stdout):
             demographic_events=self.demographic_events)
         dd.print_history(out_file)
 
-    # TODO deprecate/remove this, as the recommended interface is now 'run'.
-    def asdict(self):
-        return {
-            "population_configurations": self.population_configurations,
-            "migration_matrix": self.migration_matrix,
-            "demographic_events": self.demographic_events}
+    def get_samples(self, *args):
```
andrewkern: using the `*args` syntax here is a bit cryptic for the user. can the doc string be a bit more descriptive?

stdpopsim/models.py
```diff
@@ -242,40 +204,53 @@ def debug(self, out_file=sys.stdout):
...
-    ret = []
-
-    for scls in Model.__subclasses__():
-        for cls in scls.__subclasses__():
-            mod = inspect.getmodule(cls).__name__
-            if mod.startswith("stdpopsim"):
-                ret.append(cls())
-    return ret
+# Resuable generic population
+_pop0 = Population(name="pop0", description="Generic population")
```
andrewkern: so this is a bit strange. will we have a `_pop1` when we add a generic IM model?

stdpopsim/species.py
```diff
@@ -0,0 +1,262 @@
...
+        is to be simulated based on empirical information for a given species
+        and chromosome.
+
+        :param str species: The name of the species to simulate.
+        :param str chromosome: The name of the chromosome to simulate.
+        :param str genetic_map: If specified, obtain recombination rate information
+            from the genetic map with the specified name. If None, simulate
+            a flat recombination rate on a region with the length of the specified
+            chromosome. (Default: None)
+        :param float length_multiplier: If specified simulate a contig of length
```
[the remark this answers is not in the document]
jeromekelleher: Done, and opened new issue to track: #21.

stdpopsim/species.py
```diff
@@ -0,0 +1,262 @@
...
+        self.genetic_maps.append(genetic_map)
+
+    def get_genetic_map(self, name):
+        for gm in self.genetic_maps:
+            if gm.name == name:
+                return gm
+        raise ValueError("Genetic map not found")
+
+
+class Genome(object):
```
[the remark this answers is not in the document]
jeromekelleher: I've moved Genome, Chromosome and Contig back into the genomes.py file, and used attrs for Chromosome and Genome.

stdpopsim/catalog/pongo.py
```diff
@@ -6,40 +6,13 @@
 import msprime
 
 import stdpopsim.models as models
-import stdpopsim.generic_models as generic_models
 
 
-default_generation_time = 27
+# TODO we need to add a Genome here. Should we just take a copy of the human
+# genome?
```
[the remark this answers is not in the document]
jeromekelleher: Well, it's not doing any harm in there and if we remove it we'll forget about his model that someone went to the trouble of implementing. I'm happy for someone else to remove it though, if they want to open a PR.

stdpopsim/cli.py
jeromekelleher: I struggled with this a bit, and decided on the prefix after thinking a bit about the cache management commands we'll want as well (say, ``download-genetic-maps``, ``cache-ls``, ``cache-rm``). It seems nice to distinguish the simulate commands from non-simulate.

We can also add in synonyms for each of the simulate commands like ``sim-homsap`` as ``homsap``, which people can use?

I dunno, I'm happy with either tbh. If you said "I definitely think it's a better idea to drop the prefix", I'd very happily make the change.

stdpopsim/cli.py
jeromekelleher: Citations be important --- we want model writers to be motivated to contribute stuff. If their paper won't get cited ever again after it gets implemented in stdpopsim then there isn't much motivation to contribute...

jeromekelleher:
> one point-- i think we might rename the catalog to be species_catalog simply for clarity.

Not convinced on this one: does it add much clarity? Users aren't going to be exposed to this internal package structure, and it means a fair bit of extra typing for developers who will figure out pretty quickly what's in there (especially when we rename the modules to use the 3+3 species IDs.)

stdpopsim/models.py
jeromekelleher: Yeah, I guess so. This would have a different description to make it clear which population was simulated (this class really just exists to create metadata for the downstream TreeSequence.populations table).

stdpopsim/models.py
jeromekelleher: Agreed, this is cryptic. I've added a TODO for now.

stdpopsim/cli.py
jeromekelleher: Opened issue to track:  #118

stdpopsim/cli.py
jeromekelleher: You're right, let's ditch the prefix. Thanks!

jeromekelleher: Done.

docs/_ext/speciescatalog.py
```diff
@@ -0,0 +1,340 @@
...
+
+    thead.append(row)
+
+    rows = []
+    for genetic_map in species.genetic_maps:
+        row = nodes.row()
+        rows.append(row)
+
+        # map_id = f"sec_catalog_{species.name}_genetic_maps_{genetic_map.name}"
+        # TODO figure out how to make a link here to the corresponding detail
```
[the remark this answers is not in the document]
jeromekelleher: I'll make another pass at the catalog rendering once the PR is merged --- still figuring out how all this stuff works!

jeromekelleher:
> once we merge this we will need to do some serious follow up on the analysis repo. we might consider coordinating efforts a bit there.

I've rounded off a few corners here from your comments @andrewkern (thanks btw, very helpful!). I think it's good enough to merge now --- the public APIs should be fairly sorted. There's still a bunch of stuff to be done internally, but that's OK.

How do you want to do the coordination with the analysis repo? If you're happy to merge this, then I can make a pass through the analysis repo to update. But perhaps we want to land the large PRs outstanding there? (Or, I could update those PRs as well?)

ndukler:
I'm looking through this now and testing this out. Once I've figured out the new documentation schema I can probably address #118 as well.

andrewkern:
> 
> How do you want to do the coordination with the analysis repo? If you're happy to merge this, then I can make a pass through the analysis repo to update. But perhaps we want to land the large PRs outstanding there? (Or, I could update those PRs as well?)

I think we will want to get this PR https://github.com/popgensims/analysis/pull/39#issuecomment-537172950 in to the analysis repo before we merge... unsure

ndukler:
@jeromekelleher @andrewkern I think this may be broken on windows now due a dependency on the `resource` module. I'm not sure of where this is used but could we swap over to `psutil` instead for cross platform compatibility?

andrewkern:
huh-- what seems to be broken exactly?

ndukler:
When I tried to run `stdpopsim` from the command line I got the error `ModuleNotFoundError: No module named 'resource'` which led me to [here](https://stackoverflow.com/questions/37710848/importerror-no-module-named-resource).

ndukler:
I also just went back and installed it from the master branch with the previous CLI and I get the same problem. I don't know why I didn't bring it up before, I probably just switched over to linux and forgot about it.

andrewkern:
SWITCH BACK OVER TO LINUX!!! 

more seriously I have no clue about windows

ndukler:
> SWITCH BACK OVER TO LINUX!!!
> 
> more seriously I have no clue about windows

I know, I know :). If we're going to restrict the CLI to bash only (which probably doesn't matter) we should at least mention it so we don't get an issue for it at some point.

jeromekelleher:
> @jeromekelleher @andrewkern I think this may be broken on windows now due a dependency on the resource module. I'm not sure of where this is used but could we swap over to psutil instead for cross platform compatibility?

Good spot @ndukler, I'll change this. We hit the same issue on tsinfer, I'll port in the workaround we used here.

jeromekelleher:
> I think we will want to get this PR popgensims/analysis#39 (comment) in to the analysis repo before we merge... unsure

Unless this is ready to go straight away, I'd rather we merged this now so we can try it out and add more functionality (I want to get the cache management functionality implemented before demoing on  Sunday). I'm happy to open a PR onto the in-flight PRs to bring them up to date. The changes will be small.

jeromekelleher:
Windows problem should be fixed now. If not, let's open an issue to track @ndukler. 

Windows should be a first-class platform for us --- we need to set up Appveyor CI to make sure it's all working.

jeromekelleher:
Codecov is complaining about the patch coverage, but it's just because of the conditional import with resource. It's not worth complicating things over it, so we can ignore Codecov here.

ndukler:
> Windows problem should be fixed now. If not, let's open an issue to track @ndukler.
> 
> Windows should be a first-class platform for us --- we need to set up Appveyor CI to make sure it's all working.

Working for me now as well, thanks!

ndukler:
> > I think we will want to get this PR [popgensims/analysis#39](https://github.com/popgensims/analysis/pull/39) (comment) in to the analysis repo before we merge... unsure
> 
> Unless this is ready to go straight away, I'd rather we merged this now so we can try it out and add more functionality (I want to get the cache management functionality implemented before demoing on Sunday). I'm happy to open a PR onto the in-flight PRs to bring them up to date. The changes will be small.

It's just a small quality of life thing but did you see my comment on #118 about curtailing some of the verbosity @jeromekelleher ?

jeromekelleher:
> It's just a small quality of life thing but did you see my comment on #118 about curtailing some of the verbosity @jeromekelleher ?

I agree with you @ndukler --- let's pull some of the model help text out into an option or something. I'll follow up on the issue when I get a chance. Let's get this beast merged first though.

jeromekelleher:
OK, I'm going to merge this. I'll open some issues signposting what needs doing over in the analysis repo.

PR #115, opened 2019-09-13
further build out of generic models for non-human genomes

andrewkern:
this is the build out of the generic mixin paradigm for all of the non-human genomes that we currently have. be sure to see the new `is_generic` container that I have given to each model. that is only in there to not break the `all_models()` helper function for the test suite.

### DIFF
--- a/stdpopsim/arabidopsis_thaliana.py
+++ b/stdpopsim/arabidopsis_thaliana.py
@@ -7,7 +7,7 @@ import numpy as np
 import stdpopsim.models as models
 import stdpopsim.genomes as genomes
 import stdpopsim.genetic_maps as genetic_maps
-import stdpopsim.generic_models as generics # NOQA
+import stdpopsim.generic_models as generic_models
 
 ###########################################################
 #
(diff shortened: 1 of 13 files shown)


ckyriazis:
hey @andrewkern would it be possible to add a generic IM model to this PR?

andrewkern:
yup definitely but want to get this merged first. I think Jerome is traveling right now and so I'm waiting for him to review the code before I hit merge myself.

stdpopsim/arabidopsis_thaliana.py
```diff
@@ -77,14 +77,41 @@ class Salome2012(genetic_maps.GeneticMap):
...
+class GenericConstantSize(ArabidopsisThalianaModel, generic_models.ConstantSizeMixin):
+    def __init__(self):
+        ArabidopsisThalianaModel.__init__(self)
+        generic_models.ConstantSizeMixin.__init__(self, self.default_population_size)
+
+
+class GenericTwoEpoch(ArabidopsisThalianaModel, generic_models.TwoEpochMixin):
+    def __init__(self, n2=None, t=None):
+        ArabidopsisThalianaModel.__init__(self)
+        n1 = self.default_population_size
```
jeromekelleher: Two things I don't like about this: (a) there's a bunch of arbitrary decisions being made here and (b) we're duplicating these arbitrary decisions in different species.

I don't have a better suggestion right now though.

jeromekelleher:
Sorry for the long delay in getting to this @andrewkern. I'm getting increasingly uneasy with this way of doing things to be honest, it's starting to look like a *lot* of extra work for us, duplicated across a bunch of different modules. 

I think maybe a factory that allows users to build these models for themselves, using the basic parameters defined in the species modules would be a better way to go. I'm going to merge this and see if I can come up with a way that involves less code and that offers a simple way of building the CLI.

Issue #37, opened 2019-03-17
License is actually GPL


jeromekelleher:
The license for the package is currently marked as MIT, but after thinking about this a bit I think it's not correct. The stdpopsim package imports msprime, which is GPL, and stdpopsim is therefore also GPL. This is annoying, but can't be helped.

@molpopgen, do you agree with this?

molpopgen:
I believe this to be correct.  The "derivative product" concept seems to
apply here.  But, I am not a lawyer 😉.

grahamgower:
I think that the GPL will apply to *distribution* of stdpopsim as a whole. But this should not preclude individual components within stdpopsim being MIT licensed. In practice, someone should be permitted to take the sources and use them under the terms of the MIT license. Of course, if they use the parts with msprime imports, then they will also be bound to distribute under the terms of the GPL. I think this is more or less the same as MIT-licensed code linking against a GPLed library.

https://opensource.stackexchange.com/questions/8134/if-mit-licensed-code-links-to-gpl-licensed-code-what-should-be-the-license-of-th

jeromekelleher:
This is a tricky issue. I think since we're using the msprime DemographicEvent infrastructure quite centrally, it would be difficult to see how any code within stdpopsim doesn't depend on this. It's annoying the GPL is like this, but that's where we are and there's not much we can do about it. If you feel strongly about parts of stdpopsim not being GPL, I'm sure we can work something out though. It'd be more efficient to discuss offline, so let me know if you'd like to chat about it.

PR #163, opened 2019-11-10
Slim backend

grahamgower:
There's plenty more that needs to be done, but its functional already. Try it with:
```
stdpopsim -vv -e slim --slim-no-burnin homsap -c chr22 -l 0.001 -m ooa_3 -o /dev/null 10 10 10
```

The generated SLiM script is intended to be extensible outside of stdpopsim. To obtain the generated SLiM script:
```
stdpopsim -vv -e slim --slim-no-burnin --slim-script-file out.slim homsap -c chr22 -l 0.001 -m ooa_3 -o /dev/null 10 10 10
```

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -1,0 +1,598 @@
+import string
+import tempfile
+import subprocess
+import functools
+import itertools
+import collections
+
+import stdpopsim
+import numpy as np
+import msprime
+import pyslim
+
+_slim_upper = """
+initialize() {
+    defineConstant("verbosity", $verbosity);
+
+    // Scaling factor to speed up simulation.
+    defineConstant("Q", $Q);
+
+    defineConstant("generation_time", $generation_time);
+    defineConstant("mutation_rate", Q * $mutation_rate);
+    defineConstant("recombination_rate", Q * $recombination_rate);
+    defineConstant("chromosome_length", $chromosome_length);
+    defineConstant("trees_file", "$trees_file");
+    defineConstant("check_coalescence", $check_coalescence);
+"""
+
+_slim_lower = """
+    defineConstant("N", asInteger(_N/Q));
+
+    initializeTreeSeq(checkCoalescence=check_coalescence);
+    initializeMutationRate(mutation_rate);
+    initializeMutationType("m1", 0.5, "f", 0);
+    initializeGenomicElementType("g1", m1, 1.0);
+    initializeGenomicElement(g1, 0, chromosome_length-1);
+    initializeRecombinationRate(recombination_rate);
+}
+
+function (void)dbg(string$ s, [integer$ debug_level = 2]) {
+    if (verbosity >= debug_level)
+        catn(sim.generation + ": " + s);
+}
+
+// Return the number of generations that separate t0 and t1.
+function (integer)gdiff(numeric$ t0, numeric t1) {
+    return asInteger(round((t0-t1)/generation_time/Q));
+}
+
+// Output tree sequence file and end the simulation.
+function (void)end(void) {
+    sim.treeSeqOutput(trees_file);
+    sim.simulationFinished();
+}
+
+// Create initial populations and migration rates.
+1 {
+    // Create initial populations.
+    for (i in 0:(num_populations-1)) {
+        if (N[0,i] > 0) {
+            dbg("sim.addSubpop("+i+", "+N[0,i]+");");
+            sim.addSubpop(i, N[0,i]);
+        }
+    }
+
+    // Migration rates.
+    i = 0;
+    for (j in 0:(num_populations-1)) {
+        for (k in 0:(num_populations-1)) {
+            if (j==k | N[i,j] == 0 | N[i,k] == 0)
+                next;
+
+            m = migration_matrices[k,j,i];
+            p = sim.subpopulations[j];
+            dbg("p"+j+".setMigrationRates("+k+", "+m+");");
+            p.setMigrationRates(k, m);
+        }
+    }
+}
+
+// Check if burn-in has completed.
+1 {
+    if (!check_coalescence) {
+        setup();
+        return;
+    }
+
+    if (sim.treeSeqCoalesced()) {
+        /*
+         * All current generation individuals now have a common ancestor
+         * born after the start of our simulation.  But the genealogy at this
+         * first coalescence is not a good representative of an average
+         * genealogy, as the TMRCA is biased low.  To understand why,
+         * consider a haploid population with constant N=2.  At first
+         * coalescence, the TMRCA is always one generation, yet clearly
+         * genealogies with longer TMRCA's are possible for this population.
+         *
+         * To obtain a burn-in genealogy drawn from the full distribution
+         * of possible genealogies, we may continue the simulation forwards
+         * until this ergodic process converges.  Unfortunately, it's not
+         * clear how many more generations are required for convergence,
+         * and this is likely dependent on demography, and non-neutral
+         * processes (if any).  10*N appears to be reasonable for a single
+         * neutrally evolving population with constant N.
+         */
+        N_max = max(N[0,0:(num_populations-1)]);
+        g = sim.generation + 10 * N_max;
+        sim.registerEarlyEvent(NULL, "{setup();}", g, g);
+    } else {
+        if (sim.generation == 1)
+            dbg("Waiting for burn-in...");
+        // Reschedule the current script block 10 generations hence.
+        // XXX: find a less arbitrary generation interval.
+        g = sim.generation + 10;
+        sim.rescheduleScriptBlock(self, g, g);
+    }
+}
+
+// Register events occurring at time _T[0] or more recently.
+function (void)setup(void) {
+
+    dbg("setup()");
+
+    // Once burn-in is complete, we know the starting generation (which
+    // corresponds to T_0) and can thus calculate the generation
+    // for each remaining event.
+    G_start = sim.generation;
+    T_0 = max(_T);
+    G = G_start + gdiff(T_0, _T);
+    G_end = max(G)+1;
+
+    // Split events.
+    if (length(subpopulation_splits) > 0 ) {
+        for (i in 0:(ncol(subpopulation_splits)-1)) {
+            g = G_start + gdiff(T_0, subpopulation_splits[0,i]);
+            newpop = subpopulation_splits[1,i];
+            size = asInteger(subpopulation_splits[2,i] / Q);
+            oldpop = subpopulation_splits[3,i];
+            sim.registerLateEvent(NULL,
+                "{dbg(self.source); " +
+                "sim.addSubpopSplit("+newpop+","+size+","+oldpop+");}",
+                g, g);
+        }
+    }
+
+    // Population size changes.
+    if (num_epochs > 1) {
+        for (i in 1:(num_epochs-1)) {
+            g = G[i-1];
+            for (j in 0:(num_populations-1)) {
+                if (N[i,j] != N[i-1,j]) {
+                    sim.registerLateEvent(NULL,
+                        "{dbg(self.source); " +
+                        "p"+j+".setSubpopulationSize("+N[i,j]+");}",
+                        g, g);
+                }
+
+                if (growth_rates[i,j] != 0) {
+                    growth_phase_start = g+1;
+                    if (i == num_epochs-1)
+                        growth_phase_end = G[i];
+                    else {
+                        // We already registered a size change at generation G[i].
+                        growth_phase_end = G[i] - 1;
+                    }
+
+                    N0 = N[i,j];
+                    r = Q * growth_rates[i,j];
+
+                    sim.registerLateEvent(NULL,
+                        "{" +
+                            "dbg(self.source); " +
+                            "gx=sim.generation-"+g+"; " +
+                            "size=asInteger(round("+N0+"*exp("+r+"*gx))); " +
+                            "p"+j+".setSubpopulationSize(size);" +
+                        "}",
+                        growth_phase_start, growth_phase_end);
+                }
+            }
+        }
+
+        // Migration rates.
+        for (i in 1:(num_epochs-1)) {
+            for (j in 0:(num_populations-1)) {
+                for (k in 0:(num_populations-1)) {
+                    if (j==k | N[i,j] == 0 | N[i,k] == 0)
+                        next;
+
+                    m = migration_matrices[k,j,i];
+                    g = G[i-1];
+                    sim.registerLateEvent(NULL,
+                        "{dbg(self.source); " +
+                        "p"+j+".setMigrationRates("+k+", "+m+");}",
+                        g, g);
+                }
+            }
+        }
+    }
+
+    // Sample individuals.
+    for (i in 0:(ncol(sampling_episodes)-1)) {
+        pop = sampling_episodes[0,i];
+        n = sampling_episodes[1,i];
+        g = G_start + gdiff(T_0, sampling_episodes[2,i]);
+        sim.registerLateEvent(NULL,
+            "{dbg(self.source); " +
+            "inds=p"+pop+".sampleIndividuals("+n+"); " +
+            "sim.treeSeqRememberIndividuals(inds);}",
+            g, g);
+    }
+
+    sim.registerLateEvent(NULL, "{dbg(self.source); end();}", G_end, G_end);
+}
+"""
+
+
+def slim_makescript(
+        script_file,
+        trees_file,
+        model,
+        samples,
+        Q,
+        check_coalescence,
+        mutation_rate,
+        generation_time,
+        population_configurations,
+        recombination_map,
+        migration_matrix,
+        demographic_events,
+        verbosity):
+
+    if generation_time <= 0:
+        raise Exception(f"generation_time={generation_time} is invalid")
+
+    if len(recombination_map.get_positions()) > 2:
+        raise Exception("recombination_map not supported")
+
+    pop_names = [pc.metadata["name"] for pc in population_configurations]
+
+    # The demography debugger constructs event epochs, which we use
+    # to define the forwards-time events.
+    dd = msprime.DemographyDebugger(
+            population_configurations=population_configurations,
+            migration_matrix=migration_matrix,
+            demographic_events=demographic_events)
+
+    epochs = sorted(dd.epochs, key=lambda e: e.start_time, reverse=True)
+    T = [int(e.start_time*generation_time) for e in epochs]
+    migration_matrices = [e.migration_matrix for e in epochs]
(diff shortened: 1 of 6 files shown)


grahamgower:
Is there a recommended way to make pyslim available to the CI test framework?

jeromekelleher:
Very cool, thanks @grahamgower! This is a great start.

There's a number of sticky issues to work through though --- will you be able to make the call tomorrow (Tuesday?)

grahamgower:
Yep, I can join if a zoom link is posted to the popgen_benchmark list.

petrelharp:
Very nice. I'm also going to ping @bhaller and @gtsambos, who may have input.

Your approach of doing all the demography set-up in SLiM is very elegant. I'm curious about the rationale? I'm sure you thought about the other approach, of building up the SLiM script itself in blocks, which it seems like would end up with a more readable script?

gtsambos:
Hey; sorry, I didn't realise we'd already started on this! So speedy - and awesome work @grahamgower! I will look through this later in the week.

> Your approach of doing all the demography set-up in SLiM is very elegant. I'm curious about the rationale? I'm sure you thought about the other approach, of building up the SLiM script itself in blocks, which it seems like would end up with a more readable script?

Btw, my current `slime` setup takes the other approach -- it generates SLiM scripts in blocks. It's possibly a little ugly, but I'm happy to share this code if it might help.

Unfortunately, I don't think I can make the call (3am my time...) but v happy to help in other ways and have a bit of room for it. I'm excited that this is starting to move!

petrelharp:
> Btw, my current slime setup takes the other approach -- it generates SLiM scripts in blocks. It's possibly a little ugly, but I'm happy to share this code if it might help.

Is it on github somewhere we can see it? It'd be great to have you involved, at least a bit.

gtsambos:
Most of the relevant code for the SLiM bits of slime are [here](https://github.com/gtsambos/slime/blob/master/slime/scriptgenerator.py), btw, and there's v-rough-work-in-progress documentation [here](http://slime.readthedocs.io/). Btw, the pip installation hasn't been updated in a while, so not everything in the docs will work until I do that.

grahamgower:
@petrelharp I expect a major use case of this code will be to take a stdpopsim model, convert it to a SLiM script, then make changes that will not end up back in stdpopsim. If I'd written out the standard generation blocks in SLiM, that would require an end user to make laborious changes to all the blocks, even for simply adjusting Q or the generation_time.

I'm not hugely attached to the names or formatting of the constants, so any suggestions to improve readability here would be welcome. In particular, SLiM matrices/arrays are column major order, so using them here has the unfortunate side effect that rows on the screen correspond to columns of the matrix. Yes, this will be confusing for someone reading the code. Sorry! If there's a way to do "lists of tuples" in SLiM, that would be a preferable data representation.

@gtsambos I'll take a look through your slime code. I'm sure this PR is not the last word on how to make a bridge between stdpopsim and SLiM.

gtsambos:
Nws, ping me if there's anything else I can help with 🎆

bhaller:
Hi folks.  I've only skimmed this thread so far; I wasn't aware of any of this stuff, so I'm a bit dazzled.  :->  @grahamgower, regarding this:

> In particular, SLiM matrices/arrays are column major order, so using them here has the unfortunate side effect that rows on the screen correspond to columns of the matrix. Yes, this will be confusing for someone reading the code. Sorry! If there's a way to do "lists of tuples" in SLiM, that would be a preferable data representation.

Yes, Eidos follows R in this respect (as in many others).  Why not just use `t()` to transpose the matrices, so that you can build them in the opposite order if you wish?  Or, I haven't looked at your generated code to see how you're building things, but the `matrix()` function has an optional `byrow` parameter that you could set to T if you wanted to do it that way.  There is no list of tuples in Eidos, sorry.  :->

Let me know if y'all need anything from me.  Cool stuff.  Should I add a mention of these things in the SLiM manual?

petrelharp:
> I expect a major use case of this code will be to take a stdpopsim model, convert it to a SLiM script, then make changes that will not end up back in stdpopsim. If I'd written out the standard generation blocks in SLiM, that would require an end user to make laborious changes to all the blocks, even for simply adjusting Q or the generation_time.

Ah, I see. That makes sense. I was hoping for separate blocks because they'd be easier in many ways for this use case. Just to talk through this a bit more... I need to parse the models more carefully, but couldn't some of the redundancy be dealt with by defining some constants? Is the main problem rescheduling script blocks?  We could alternatively put separate script blocks, but still reschedule them dynamically in SLiM to deal with this?

grahamgower:
Thanks @bhaller. FYI here is an example of a generated SLiM script, corresponding to the Gutenkunst et al. 2009 three population out-of-Africa model.
[oom_3.slim.txt](https://github.com/popgensims/stdpopsim/files/3837713/oom_3.slim.txt)

@petrelharp Yeah, explicit script blocks could be produced and rescheduled as you suggest. I suppose which approach to take is just a matter of taste.

bhaller:
@grahamgower Very interesting, thanks.  It's neat to see you using features that, as far as I know, nobody else has used, like multidimensional arrays!

jeromekelleher:
Just a quick note on branch-management @grahamgower --- your life will be easier if you [rebase](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) off upstream master rather than merging.

jeromekelleher:
One thing we'll need to get sorted is running SLiM on Circle. I think the only viable option is to build it each time. But, that shouldn't be too hard: we can adapt something like [this](https://github.com/tskit-dev/msprime/blob/master/data/Makefile#L3) for downloading and compiling it. 

I guess we'll also want a command line option for ``stdpopsim`` like ``--slim-path`` which allows you to set the executable path for SLiM.  Might save a bit of messing around in situations like this.

grahamgower:
@jeromekelleher Yep, building it should be ok, I just need to first reproduce the CircleCI process offline. I guess that's easy, but I haven't tried yet. Doing all the same tests for each simulation engine is going to blow out the runtime of the tests though.

What worries me more, is that there are bugs lurking in the SLiM engine code I've written. Clearly there should be some correspondence between the trees produced with each engine (for a given model). Any ideas for how to assess this? I understand that stdpopsim has a design goal of *not* doing statistical tests on the simulation outputs...

grahamgower:
Oh, and as per my last commit, the `Model.simulate()` method probably isn't the best API to expose. It might be worth putting in the `SimulationEngine` class (without the slim_engine) and removing `Model.simulate()` before the first stdpopsim release.

bhaller:
> What worries me more, is that there are bugs lurking in the SLiM engine code I've written. Clearly there should be some correspondence between the trees produced with each engine (for a given model). Any ideas for how to assess this? I understand that stdpopsim has a design goal of _not_ doing statistical tests on the simulation outputs...

This is a problem for SLiM as a whole, too, of course – one we struggle with.  What's the rationale for not testing using statistics on simulation outputs?  That is a direction I hope to take SLiM in in future, actually.  Yes, those tests can potentially fail even when the code is correct; but it seems to me one can design tests that are extremely unlikely to fail, so this doesn't seem like a real problem to me (and I don't see any better solution).

apragsdale:
I was just in the middle of writing the same thing about burn-in: Dominic Nelson and I noticed the same thing when testing his discrete time Wright-Fisher model implemented in msprime. By stopping burn in when you find coalescence everywhere, you truncate the distribution of TMRCAs across the genome and tend to artificially shorten them (the strength of the effect changed with sample size). We ended up simulating an extra 10N generations on top of that check to be sure that we weren't distorting TMRCAs, but our tests for how long we needed weren't all that rigorous.

jeromekelleher:
> Oh, and as per my last commit, the Model.simulate() method probably isn't the best API to expose. It might be worth putting in the SimulationEngine class (without the slim_engine) and removing Model.simulate() before the first stdpopsim release.

Good call. We should have a talk about the actual design of this --- can you send me an email so we can figure out a time that'll work please?

jeromekelleher:
> What worries me more, is that there are bugs lurking in the SLiM engine code I've written. Clearly there should be some correspondence between the trees produced with each engine (for a given model). Any ideas for how to assess this? I understand that stdpopsim has a design goal of not doing statistical tests on the simulation outputs...

We're going to have to do some statistical tests to verify what we're doing is correct, I think. Probably what we'll need to do is have a separate ``verification.py`` script (or something) like we have in [msprime](https://github.com/tskit-dev/msprime/blob/master/verification.py) that checks the validity of various things statistically. These are **not** run as part of CI, as you need to eyeball a bunch of QQ plots afterwards (plus, they take a long time to run, and are just not 'unit' tests).

So, maybe we could do something similar here, where we run the msprime and SLiM versions of various models and see whether the qq-plots of various things come out looking similar?

bhaller:
> > What worries me more, is that there are bugs lurking in the SLiM engine code I've written. Clearly there should be some correspondence between the trees produced with each engine (for a given model). Any ideas for how to assess this? I understand that stdpopsim has a design goal of not doing statistical tests on the simulation outputs...
> 
> We're going to have to do some statistical tests to verify what we're doing is correct, I think. Probably what we'll need to do is have a separate `verification.py` script (or something) like we have in [msprime](https://github.com/tskit-dev/msprime/blob/master/verification.py) that checks the validity of various things statistically. These are **not** run as part of CI, as you need to eyeball a bunch of QQ plots afterwards (plus, they take a long time to run, and are just not 'unit' tests).
> 
> So, maybe we could do something similar here, where we run the msprime and SLiM versions of various models and see whether the qq-plots of various things come out looking similar?

That would provide some useful validation for SLiM itself, too.  :->

grahamgower:
Thanks @bhaller, @apragsdale. After thinking about a haploid population with constant N=2, I do see the problem. Terminating once coalescence occurs will always give TMRCA = 1 generation. Do you have a recommendation for a general stopping strategy---one that makes sense for non-neutral burn-in too?

jeromekelleher:
> Thanks @bhaller, @apragsdale. After thinking about a haploid population with constant N=2, I do see the problem. Terminating once coalescence occurs will always give TMRCA = 1 generation. Do you have a recommendation for a general stopping strategy---one that makes sense for non-neutral burn-in too?

I'm not sure we have a good answer for that. @petrelharp, you had the best handle on this --- what's a practical thing to do here in terms of neutral burn in?

stdpopsim/slim_engine.py
```diff
@@ -0,0 +1,631 @@
...
+                        verbosity=verbosity,
+                        )
+
+        script_file.flush()
+
+        if not run_slim:
+            return None
+
+        slim_cmd.append(script_file.name)
+        ret = subprocess.call(
```
jeromekelleher: ``subprocess.check_call`` will check the return value automatically

jeromekelleher:
Would you mind rebasing this please @grahamgower?  It would be good to have a reasonably up-to-date version that's compatible with the Engine infrastructure, so we can more easily see where it's going.

grahamgower:
> Would you mind rebasing this please @grahamgower? It would be good to have a reasonably up-to-date version that's compatible with the Engine infrastructure, so we can more easily see where it's going.

I had this sitting here and just forgot to push it. I haven't yet fixed the recombination map stuff like we discussed, so just ignore that.

gtsambos:
Hi @grahamgower, this is really awesome work -- I've only just gone through it in detail now (have been bogged down by some other work here at home, and 'project documentation' seemed more urgent for stdpopsim so I got on that first). This looks more useful (and a bit more elegant) from what I was making myself. A few Qs:

 - I'd be keen to do something that helps me understand your code a bit better -- is there some (relatively easy-ish) first task I could do, like testing, that would also take some of the burden off you?

 - It looks like all of the populations have to be defined at the start (maybe I'm missing something tho) -- if you wanted to simulate an admixed population, could you do this at present?

gtsambos:
> We're going to have to do some statistical tests to verify what we're doing is correct, I think. Probably what we'll need to do is have a separate verification.py script (or something) like we have in msprime that checks the validity of various things statistically. 

I'm happy to help out with this -- might need some pointers to get started tho.

Something that might play into the way you specify the different populations involved in a divergence event. You could specify an 'ancestral population', Pop0, from which migrants flow into two new populations, Pop1 and Pop2. When you do things this way, your present-day samples will be drawn from Pop1 and Pop2.

An alternative is to have a population that is 'always' present, Pop0, but for some migrants to flow out of Pop0 into a new population, Pop1. When you do things this way, your present-day samples will be drawn from pop0 and pop1.

As @agladstein discovered (see [this issue](https://github.com/tskit-dev/msprime/issues/784)), which of these you do seems to have some implications for the simulation output (though we could not quite figure out why)

It looks like your subPop.split command is doing this second thing (I think?) We'll need to check that all of the coded up msprime models are doing this also -- and think about what to do if they aren't

grahamgower:
Hey @gtsambos, it would be great to get you involved in the details! I've appended a rough TODO list, with some of the simpler tasks shown in bold. In any event, please play with the slim branch from both the CLI (e.g. `stdpopsim -vv -e slim --slim-no-burnin homsap -c chr22 -l 0.001 -m ooa_3 -o /dev/null 10 10 10`) and the python API, and report back anything that's broken. Also, it would be great if you could look at the code (in particular the generated SLiM code) and suggest changes to make it more readable and/or flexible.

Part of the current design is to store all the demographic parameters in variables within the `initialize()` block, and construct everything within SLiM from these variables. So all populations, past and present, have entries for their sizes at each epoch. Populations are not created unless they have non-zero population size, and changing a population size from non-zero to zero triggers SLiM to remove the population. As @petrelharp rightly pointed out, this makes lots of the SLiM code harder to follow. Consider it an exploration of how to specify a backwards in time simulation, within SLiM.

The way sampling is currently done may at first appear surprising. The samples requested by the user are explicitly *remembered*, which thus permits temporal sampling. This means that after the simulation, all the non-remembered individuals must be `simplify()`'d away.

With regard to admixed populations, these should be supported for most of the models in stdpopsim already. There is a corner case with MassMigrations that have `proportion` < 1, which might need careful consideration. I haven't done much testing here though---I just checked that demographic changes occur in an appropriate order for a couple of the homsap models. I would encourage you to test the two different admixture strategies which you mentioned, and confirm or deny their correctness!


TODO:
* Feature parity with msprime engine.
  - Support msprime.MassMigration with proportion < 1.
  - **Identify stdpopsim models that don't work with the SLiM engine (and why).**
* **Add option to do recap/neutral-mutation-overlay with pyslim.**
* Check trees files have similar properties to those made with msprime.
  - QQ plots for a range of summary stats.
  - See msprime/verification.py
* Test framework.
  - Duplicate existing tests for both msprime and slim engines.
  - Add new tests where relevant.
* Set population size to zero if there is no influence on the rest of the
  simulation. 
  - After ancient sampling.
  - Ancestral populations that don't/shouldn't exist at t=0.
* **More internal checks.**
  - **Fail earlier for impossible sampling. I.e. cannot ask for n samples, where n > N/Q.**
  - **Verify that the final tree sequence object has the requested number of sampled individuals for each population.**
* **Output slim script to stdout, instead of specifying another file.**
* Fix recombination map (don't cut across the stdpopsim.Contig API). Either:
  - Output the positions/rates into compactly formatted eidos arrays, or,
  - Write out a temp file which is read in to the SLiM script.
* **Tree sequence metadata.**
  - **Check and clean up any redundant provenance info.**
  - **Add population metadata.**
* **Clean up the parameter lists for slim_simulate() and slim_makescript().**

gtsambos:
Cool, thanks for this helpful explanation and list! I will see what I can pick off.

@jeromekelleher and @grahamgower: could we possibly merge this existing code into a 'slim' branch inside the main stdpopsim repo? We can then merge this branch into the master once it's ready to deploy. Atm, I will have to fork Graham's repo to get this code, and it'll be dangerously easy for me to mess something up if I have multiple clones of stdpopsim hanging around. It'll also make it easier for you to review my code -- otherwise we'll have to do this over in Graham's github

grahamgower:
Makes sense. Looks like the branch will live longer than I'd originally imagined.

grahamgower:
One important thing that hasn't been discussed yet is the license for the generated SLiM code. I'm in favour of this being MIT-licensed. Any objections?

ndukler:
See #37

grahamgower:
Thanks @ndukler, but I don't think that applies in this case. Eg. consider that gcc is GPL licensed, but you are free to distribute binaries of your own code, compiled by gcc, under any license you choose. So the choice of what license to distribute generated SLiM code does not necessarily have to follow the license under which stdpopsim is distributed. But perhaps I'm wrong, so feel free to suggest otherwise.

jeromekelleher:
Unfortunately @grahamgower, I think the only way to make this non-GPL would be to make it an external package that doesn't use anything in msprime, that we depend on here. It's annoying, but it's the only way to get a clean separation and licensing clarity. I'm happy to help with making this package and to talk about the details.

However, since the express goal is to translate msprime demographic model descriptions into SLiM code, this would necessarily involve duplicating the msprime modelling code at some point, and would need some coordination to make sure these object models don't get out of sync. Probably what we'd end up having to do is to make another third-party (non-GPL) package which basically redefines the msprime demographic modelling framework, which we then use in stdpopsim (to translate into msprime objects, which we can do since stdpopsim is GPL) and the SLiM translation package.  IMO it's probably not worth the trouble and it would be simplest to keep the license as GPL though.

jeromekelleher:
Re branching; yes, having a dev branch would be good when we have several developers. I've made a ``slim-engine-dev`` branch, where we can do this work. @grahamgower, can you change this PR so that it's pointing to this new branch please?

We can merge in master periodically so that it's kept reasonably up to date.

molpopgen:
> Eg. consider that gcc is GPL licensed, but you are free to distribute binaries of your own code, compiled by gcc, under any license you choose.

This is due to a specific exception in the GPL and a GPL-licensed product must grant is explicitly: https://en.wikipedia.org/wiki/GPL_linking_exception

grahamgower:
There are several relevant sections in the GPL FAQ. Here are two, which I think suggest that components of a GPLed project can be public domain (or analogously MIT licensed) and then taken and used by a third party under the non-GPL (but GPL-compatible) license.
https://www.gnu.org/licenses/gpl-faq.en.html#CombinePublicDomainWithGPL
https://www.gnu.org/licenses/gpl-faq.en.html#GPLModuleLicense

In any event, I don't care enough to argue my position any further, so GPL is fine.

jeromekelleher:
Looks like there's something funny going on with the branching here @grahamgower -- can you rebase off ``upstream/slim-engine-dev``, and get this ready to merge please? We can pull in the changes from master later in a separate PR.

jeromekelleher:
Shall we merge this @grahamgower? We can open some issues to track progress then.

grahamgower:
> Shall we merge this @grahamgower? We can open some issues to track progress then.
Yes, please do.

jeromekelleher:
Done!

Issue #166, opened 2019-11-11
Infrastructure for ancient samples


jeromekelleher:
Some models require ancient samples (see #146 and discussion in #140). Add infrastructure to specify a ``sampling_time``. This can replace the existing ``allow_samples`` attribute, doing something like 

```python
class Population(object):
    """
    Class recording metadata representing a population in a simulation.
    """    
    def __init__(self, name, description,sampling_time=None):
        self.name = name
        self.description = description
        self.sampling_time = sampling_time
    @property
    def allow_samples(self):
          return self.sampling_time is not None
```

So, this should give us the ability to specify sampling times or not. It'll need a little updating of existing model code, but that should be OK.

ndukler:
I'm coding this up over the next few days

jeromekelleher:
Closed in #177.

PR #169, opened 2019-11-12
Add qc for Kamm ancient Europe model

apragsdale:
QC model implementation for 9-pop model from Kamm et al (2019). See Issue #167, and note that this model includes ancient samples (#166). There were also some ambiguities (I think) in the published model description, which should be ironed out first as well - see discussion in #167.

### DIFF
--- a/qc/homo_sapiens_qc.py
+++ b/qc/homo_sapiens_qc.py
@@ -367,3 +367,157 @@ class RagsdaleArchaic(models.Model):
             msprime.MassMigration(
                 time=T_NEAN_split, source = 3, dest = 0, proportion=1.0)
         ]
+
+
+class KammAncientSamples(models.Model):
+    """
+    Demographic inferred by momi described in Kamm et al. (2019). The model is
+    illustrated in Figure 3, with parameters given in Table 2.
+    """
+    def __init__(self):
+        super().__init__()
+
+        generation_time = 25
+
+        # population sizes
+        N_Losch = 1.92e3
+        N_Mbu = 1.73e4
+        N_Mbu_Losch = 2.91e4
+        N_Han = 6.3e3
+        N_Han_Losch = 2.34e3
+        N_Nean_Losch = 1.82e4
+        N_Nean = 86.9
+        N_LBK = 75.7
+        N_Sard = 1.5e4
+        N_Sard_LBK = 1.2e4
+
+        # unknown population sizes
+        # these should be set to ancestral Eurasian population size,
+        # but at the moment it's unclear to me if that is N_Han_Losch? N_Losch?
+        N_Basal = N_Losch
+        N_Ust = N_Basal
+        N_MA1 = N_Basal
+
+        # population merge times in years, divided by generation time
+        t_Mbu_Losch = 9.58e4 / generation_time
+        t_Han_Losch = 5.04e4 / generation_time
+        t_Ust_Losch = 5.15e4 / generation_time
+        t_Nean_Losch = 6.96e5 / generation_time
+        t_MA1_Losch = 4.49e4 / generation_time
+        t_LBK_Losch = 3.77e4 / generation_time
+        t_Basal_Losch = 7.98e4 / generation_time
+        t_Sard_LBK = 7.69e3 / generation_time
+        # t_GhostWHG_Losch = 1.56e3 / generation_time
+
+        # pulse admixture times and fractions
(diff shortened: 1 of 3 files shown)


apragsdale:
There were a few tweaks I had to make to the implemented model, with the Neanderthal branch having to do with the timing of the start and stop of neanderthal exponential decay. This branch is constant in size until the same time as the split of Mbuti from Eurasians, and then it declines until the Altai sampling time, reported as 50kya. After these tweaks, the checks pass for this model.

jeromekelleher:
@ndukler, can you take a quick look here please? I think we're ready to merge and make the Kamm model officially QC'd then, right?

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -836,9 +836,9 @@ def __init__(self):
         N_Nean = 86.9
         N_LBK = 75.7
         N_Sard = 15000
-        N_Sard_LBK = 1200
-        # Nea pop size change 45,000 years ago (Jack Kamm pers comm)
-        t_NeaPopSizeChange = 45000 / self.generation_time
+        N_Sard_LBK = 12000
+        # Table A.1 has Altai at 50,000 years ago
+        t_NeaPopSizeChange = 50000 / self.generation_time
```
ndukler: Why is this different from the number given in the personal communication?

qc/homo_sapiens_qc.py
```diff
@@ -366,4 +366,149 @@ def __init__(self):
...
+
+        # sample_times (in years), divided by estimated generation time
+        # t_Mbuti = 0
+        # t_Han = 0
+        # t_Sardinian = 0
+        # t_Loschbour = 7.5e3 / generation_time
+        # t_LBK = 8e3 / generation_time
+        # t_MA1 = 24e3 / generation_time
+        # t_UstIshim = 45e3 / generation_time
+        t_Altai = 50e3 / generation_time
```
ndukler: This is okay for now but it will be a problem when I add a check for consistency of sampling times during the model comparison tests. These sampling times will have to be passed to the population configurations.

stdpopsim/catalog/homo_sapiens.py
apragsdale: That was my mistake - in the issue #167 I wrote Altai sampling of 45k, but I meant to write 50k (so I think Fernando was just following my typo). The Altai sampling was reported as 50k in the table in the ms.

qc/homo_sapiens_qc.py
apragsdale: @ndukler I'll change these now.

apragsdale:
When updating to also include the sampling time in QC model, I found a couple discrepancies in the sampling dates between implemented and QC model, double checked against published model, and fixed those as well.

jeromekelleher:
LGTM -- can you ping me when this is ready to merge please?

apragsdale:
Should be good @jeromekelleher

jeromekelleher:
Ah, looks like the committer is wrong here too. Can you rebase please @apragsdale to update the commit info?

apragsdale:
Maybe I'm being dense here, since I thought I had rebased after changing my .gitconfig file. Locally, if I run `git config --list`, I see the correct updated email. I'm not entirely sure how to check that the updates went through to what you're seeing.

apragsdale:
Oh, I might need to change my user name in `.gitconfig` and not just the email. Hang on with the merge.

jeromekelleher:
When I do git log I see:
```
commit 2093ffa20f92b0ae4acf1b1464fa5c5935b6e6c7 (HEAD -> kamm_qc, apragsdale/kamm_qc)
Author: Aaron Ragsdale <aragsdale@math.arizona.edu>
Date:   Mon Nov 11 23:37:43 2019 -0500

    Closes #167 QC for Kamm ancient Europe model
```

Is the arizona address the correct one?

jeromekelleher:
Also, crap, looks like we're hitting rate limits from the 1000G HTTP server on Appveyor.

apragsdale:
It's an old email, but it checks out.

I added that email to my github account, so commits from that email should be attributed to my account now. I wouldn't worry about it, as long as it's not a problem for you.

jeromekelleher:
OK, looks like it's all sorted out now. I'm going to merge this, and ignore the failures on Appveyor since they are caused by (hopefully) intermittent network issues.

Issue #171, opened 2019-11-12
Generic IM model


ckyriazis:
As discussed on our call today, it would be great to have a generic IM model as a way to test the two population analysis pipeline under a scenario where there is not model misspecification. Ideally, this model would have three population size parameters (Na, N1, N2) one divergence parameter (Tdiv) and two migration rate parameters (M12 and M21).

apragsdale:
@ckyriazis are you working on this? I'm happy to put this together if no one else is already.

jeromekelleher:
Go for it I'd say @apragsdale, ping me if you'd like any help!

jeromekelleher:
There you go, it's official now; you're assigned!

ckyriazis:
Thanks @apragsdale !

grahamgower:
The comment in the header for this model doesn't match the defined populations. I.e.
```
    Takes parameters N0, N1, N2, T, M12, and M21, as list of arguments:
    (NA, N1, N2, T, M12, M21).
    Sampling is disallowed in population index 0, as this is the ancestral
    population.
```
But then the populations definition has the ancestral population last.
```
populations = [_pop0, _pop1, _popAnc]
```

apragsdale:
You're right, good catch. Labels switched around after some discussion of how to label and define pops, and I overlooked the comment. I can push a fix to this this afternoon.

apragsdale:
FYI I think this can be closed, no? (I've since learned the "Closes #xxx" trick in PRs)

PR #172, opened 2019-11-13
Add IM model to models

apragsdale:
I added a generic IM model (issue #171) to models.py, below the PiecewiseConstantSize model. Possible that this model should live elsewhere?

### DIFF
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -235,8 +235,11 @@ class Model(object):
         return ts
 
 
-# Resuable generic population
+# Reusable generic populations
 _pop0 = Population(name="pop0", description="Generic population")
+_pop1 = Population(name="pop1", description="Generic population")
+_popAnc = Population(name="popAnc", description="Generic ancestral population",
+                     allow_samples=False)
 
 
 class PiecewiseConstantSize(Model):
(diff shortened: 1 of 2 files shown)


stdpopsim/models.py
```diff
@@ -259,3 +261,41 @@ def __init__(self, N0, *args):
...
+        generations ago, with migration rates M12 and M21 between the split
+        populations.
+        """
+    citations = []
+    populations = [_pop0, _pop1, _pop2]
+    author = None
+    year = None
+    doi = None
+
+    def __init__(self, *args):
```
jeromekelleher: Might as well put these into the arguments explicity, rather than unpacking (I don't think it's right atthe moment anyway)

apragsdale: Right - hurrying through and forgot the tests.

And yeah, arguments could be explicit. Will get to it shortly

andrewkern: yeah agree this is the correct place

andrewkern:
looks like this is now passing all checks. everyone good if I merge?

ndukler:
Do we want to block sampling from population 0? It doesn't really exist at the present time.

apragsdale:
> Do we want to block sampling from population 0? It doesn't really exist at the present time.

That would make sense - though I'm not entirely sure how to forbid sampling from a given population in `msprime`.

andrewkern:
maybe you could just write an `assert` statement into the `init` function?

apragsdale:
Yeah I'll play around with that. Need to familiarize myself with the usage of the `Model` class a bit, and how a user would define samples... a bit confused by it.

apragsdale:
I was just playing around with including a check that `allow_samples` is `True` when getting the list of `msprime.Samples` objects in `get_samples` in the `Model` class. Something like
```
assert np.all([args[i] == 0
            or self.populations[i].allow_samples == True
            for i in range(len(args))]), "samples given for invalid population"
```

But this causes tests for other models to fail, and ties in with the lack of infrastructure for ancient samples (see Issue #166). I don't think I'm able to check the samples inside the `init` because the model has no knowledge of the samples at that point (right?).

Since checking whether samples are allowed for given pops and allowing ancient samples is a different/broader issue, maybe we can call this good for now, and address it with #166? I'm happy to think about that issue more tomorrow.

stdpopsim/models.py
```diff
@@ -235,8 +235,12 @@ def simulate(self, contig, samples, seed=None):
         return ts
 
 
-# Resuable generic population
+# Resuable generic populations
 _pop0 = Population(name="pop0", description="Generic population")
+_popAnc = Population(name="popA", description="Generic ancestral population",
```
andrewkern: i'm a bit unsure what the scope of these objects is. @apragsdale how do you want these to be available?

apragsdale: Yeah good question - I don't know if there's a good reason for them to be defined outside the model classes. I was mainly following how the `PiecewiseConstantSize` model was set up for consistency, and I imagine `_pop0` was meant to be reused if needed (though I didn't, maybe should have). Probably more intuitive to just define the populations within each class as needed, you think?

andrewkern: I think it's okay to define these at this level but it would be best to reuse them (e.g. `_pop0`). @jeromekelleher you okay with this bit of ugliness too?

apragsdale: Alright - made a few small changes. Let me know what you think @andrewkern, @jeromekelleher , and if anything else could be tidied up

andrewkern:
I'm inclined to just merge this so we can move on. We can always come back to this very niche bit.

jeromekelleher:
Nicely done, thanks @apragsdale.

PR #176, opened 2019-11-16
update development docs installation section

agladstein:
Resolves https://github.com/popgensims/stdpopsim/issues/156


agladstein:
Now also resolves https://github.com/popgensims/stdpopsim/issues/178

But, I did not include docs for a conda env because I don't use conda and when i tried to do it, it failed after trying to install things for like 15 min.

jeromekelleher:
Great stuff, thanks @agladstein!

PR #177, opened 2019-11-17
Added ability to set sample times for populations

ndukler:
Added the ability to set a `sampling_time` attribute for a population that allows for ancient samples. A negative value for this attribute prevents sampling. Updated the models and other functions as necessary. Not sure what tests should be added.

### DIFF
--- a/stdpopsim/catalog/homo_sapiens.py
+++ b/stdpopsim/catalog/homo_sapiens.py
@@ -426,7 +426,7 @@ class _BrowningAmerica(HomoSapiensModel):
         stdpopsim.Population(name="ASIA", description="Contemporary Asian population"),
         stdpopsim.Population(
             name="ADMIX", description="Ancient admixed population",
-            allow_samples=False),
+            sampling_time=None),
     ]
 
     citations = [
@@ -538,9 +538,9 @@ class _RagsdaleArchaic(HomoSapiensModel):
         _ceu_population,
         _chb_population,
         stdpopsim.Population(
-            "Neanderthal", "Putative Neanderthals", allow_samples=False),
+            "Neanderthal", "Putative Neanderthals", sampling_time=None),
         stdpopsim.Population(
-            "ArchaicAFR", "Putative Archaic Africans", allow_samples=False),
+            "ArchaicAFR", "Putative Archaic Africans", sampling_time=None),
     ]
     citations = [
         stdpopsim.Citation(
@@ -728,7 +728,7 @@ class _SchiffelsZigzag(HomoSapiensModel):
 
         self.population_configurations = [
             msprime.PopulationConfiguration(
-                initial_size=N0)
+                initial_size=N0, metadata=self.populations[0].asdict())
         ]
 
         self.migration_matrix = [
@@ -783,9 +783,6 @@ class _KammAncientEurasia(HomoSapiensModel):
     populations = [
         stdpopsim.Population(name="Mbuti",
                              description="Present-day African Mbuti"),
-        stdpopsim.Population(name="BasalEurasian",
-                             description="Basal Eurasians",
-                             allow_samples=False),
         stdpopsim.Population(name="LBK",
                              description="Early European farmer (EEF)"),
         stdpopsim.Population(name="Sardinian",
@@ -799,7 +796,10 @@ class _KammAncientEurasia(HomoSapiensModel):
         stdpopsim.Population(name="UstIshim",
                              description="early Palaeolithic Ust'-Ishim"),
         stdpopsim.Population(name="Neanderthal",
-                             description="Altai Neanderthal from Siberia")
+                             description="Altai Neanderthal from Siberia"),
+        stdpopsim.Population(name="BasalEurasian",
+                             description="Basal Eurasians",
+                             sampling_time=None),
     ]
     citations = [
         stdpopsim.Citation(
@@ -880,94 +880,96 @@ class _KammAncientEurasia(HomoSapiensModel):
             [0, 0, 0, 0, 0, 0, 0, 0, 0],
             [0, 0, 0, 0, 0, 0, 0, 0, 0]
         ]
-        # Population IDs: Mbuti = 0; Basal Eurasian = 1;
-        # LBK = 2; Sardinian = 3; Loschbour = 4
-        # MA1 = 5; Han = 6; Ust Ishim = 7; Neanderthal = 8
+        # Population IDs: Mbuti = 0; LBK = 1;
+        # Sardinian = 2; Loschbour = 3; MA1 = 4;
+        # Han = 5; Ust Ishim = 6; Neanderthal = 7;
+        # Basal Eurasian = 8;
         self.population_configurations = [
             msprime.PopulationConfiguration(
                 initial_size=N_Mbu, metadata=self.populations[0].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_Basal, metadata=self.populations[1].asdict()),
+                initial_size=N_LBK, metadata=self.populations[1].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_LBK, metadata=self.populations[2].asdict()),
+                initial_size=N_Sard, metadata=self.populations[2].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_Sard, metadata=self.populations[3].asdict()),
+                initial_size=N_Losch, metadata=self.populations[3].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_Losch, metadata=self.populations[4].asdict()),
+                initial_size=N_MA1, metadata=self.populations[4].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_MA1, metadata=self.populations[5].asdict()),
+                initial_size=N_Han, metadata=self.populations[5].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_Han, metadata=self.populations[6].asdict()),
+                initial_size=N_Ust, metadata=self.populations[6].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_Ust, metadata=self.populations[7].asdict()),
+                initial_size=N_Nean, metadata=self.populations[7].asdict()),
             msprime.PopulationConfiguration(
-                initial_size=N_Nean, metadata=self.populations[8].asdict())
+                initial_size=N_Basal, metadata=self.populations[8].asdict())
         ]
         self.demographic_events = [
             # Sardinian receives admixture from Loschbour / WHG
             msprime.MassMigration(
-                time=t_GhostWHG_to_Sard, source=3, destination=4,
+                time=t_GhostWHG_to_Sard, source=2, destination=3,
                 proportion=p_GhostWHG_to_Sard),
             # Sardinian merges into LBK / EEF
+            # Now pop 1: Sardinian-LBK ancestral pop
             msprime.MassMigration(
-                time=t_Sard_LBK, source=3, destination=2,
+                time=t_Sard_LBK, source=2, destination=1,
                 proportion=1.0),
             # Sardinian-LBK ancestral pop size change
             msprime.PopulationParametersChange(
                 time=t_Sard_LBK, initial_size=N_Sard_LBK,
-                population_id=2),
+                population_id=1),
             # LBK / EEF receives admixture from Basal Eurasians
             msprime.MassMigration(
-                time=t_Basal_to_EEF, source=2, destination=1,
+                time=t_Basal_to_EEF, source=1, destination=8,
                 proportion=p_Basal_to_EEF),
             # LBK / EEF merges into Loschbour
             msprime.MassMigration(
-                time=t_LBK_Losch, source=2, destination=4,
+                time=t_LBK_Losch, source=1, destination=3,
                 proportion=1.0),
             # MA1 merges into Loschbour
             msprime.MassMigration(
-                time=t_MA1_Losch, source=5, destination=4,
+                time=t_MA1_Losch, source=4, destination=3,
                 proportion=1.0),
             # Han merges into Loschbour
             msprime.MassMigration(
-                time=t_Han_Losch, source=6, destination=4,
+                time=t_Han_Losch, source=5, destination=3,
                 proportion=1.0),
             # Change in population size in Han-Losch ancestral pop
             msprime.PopulationParametersChange(
                 time=t_Han_Losch, initial_size=N_Han_Losch,
-                population_id=4),
+                population_id=3),
             # UstIshim merges into Loschbour
             msprime.MassMigration(
-                time=t_Ust_Losch, source=7, destination=4,
+                time=t_Ust_Losch, source=6, destination=3,
                 proportion=1.0),
             # Loschbour / Non-Africans receive admixture from Neanderthals
             msprime.MassMigration(
-                time=t_Nean_to_Eurasian, source=4, destination=8,
+                time=t_Nean_to_Eurasian, source=3, destination=7,
                 proportion=p_Nean_to_Eurasian),
             # Basal Eurasians merge into Loschbour / Non-Africans
             msprime.MassMigration(
-                time=t_Basal_Losch, source=1, destination=4,
+                time=t_Basal_Losch, source=8, destination=3,
                 proportion=1.0),
             # Mbuti merge into Loschbour / Non-Africans
             msprime.MassMigration(
-                time=t_Mbu_Losch, source=0, destination=4,
+                time=t_Mbu_Losch, source=0, destination=3,
                 proportion=1.0),
             # Change in population size in Mbuti-Losch ancestral pop
             msprime.PopulationParametersChange(
                 time=t_Mbu_Losch, initial_size=N_Mbu_Losch,
-                population_id=4),
+                population_id=3),
             # Neanderthal change population size
             msprime.PopulationParametersChange(
                 time=t_NeaPopSizeChange, initial_size=N_Nean_Losch,
-                population_id=8),
+                population_id=7),
             # Neanderthal merge into Loschbour / modern humans
             msprime.MassMigration(
-                time=t_Nean_Losch, source=8, destination=4,
+                time=t_Nean_Losch, source=7, destination=3,
                 proportion=1.0),
             # Ancestral hominin population size change
             msprime.PopulationParametersChange(
                 time=t_Nean_Losch, initial_size=N_Nean_Losch,
-                population_id=4),
+                population_id=3),
         ]
 
 
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -112,18 +112,33 @@ def verify_demographic_events_equal(
 class Population(object):
     """
     Class recording metadata representing a population in a simulation.
+
+    :ivar name: the name of the population
+    :vartype name: str
+    :ivar description: a short description of the population
+    :vartype description: str
+    :ivar sampling_time: an integer value which indicates how many
+        generations prior to the present individuals should samples should
+        be drawn from this population. If `None`, sampling not allowed from this
+        population (default = 0).
+    :vartype sampling_time: int
     """
     # TODO change this to use the usual id, name combination
-    def __init__(self, name, description, allow_samples=True):
+    def __init__(self, name, description, sampling_time=0):
         self.name = name
         self.description = description
-        self.allow_samples = allow_samples
+        self.sampling_time = sampling_time
+
+    @property
+    def allow_samples(self):
+        return self.sampling_time is not None
 
     def asdict(self):
         """
         Returns a dictionary representing the metadata about this population.
         """
-        return {"name": self.name, "description": self.description}
+        return {"name": self.name, "description": self.description,
+                "sampling_time": self.sampling_time}
 
 
 class Model(object):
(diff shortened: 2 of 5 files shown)


ndukler:
I'm going to leave this here for now as a solution for #166. I'm not sure why the AppVeyor build is failing as the errors seem to be unrelated to the changes I made in the code. I've also had to do some cleanup on the chimp model (which generally needs an overhaul). Once this goes through I can finish up with some of the documentation and #169 can go through as well.

ndukler:
Ok, so looking into the appVeyor it looks like there is some network flakiness on their end that's causing url lookups to fail. All the errors are `urllib.error.URLError: <urlopen error [Errno 11002] getaddrinfo failed>`. We could try to make the code more robust to this sort of thing using `appveyor-retry`. What do you think @jeromekelleher ?

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -426,7 +426,7 @@ class _BrowningAmerica(HomoSapiensModel):
         stdpopsim.Population(name="ASIA", description="Contemporary Asian population"),
         stdpopsim.Population(
             name="ADMIX", description="Ancient admixed population",
-            allow_samples=False),
+            sampling_time=-1),
```
jeromekelleher: I think it'd be better to use ``None`` as the exceptional value here. Seems easier to read, somehow with ``sampling_time=None``. It's a bit more explicit, and bit less likely to happen by accident.

stdpopsim/cli.py
```diff
@@ -96,7 +96,7 @@ def get_models_help(species_id, model_id):
         models_text += indent + "Populations:\n"
 
         for population in model.populations:
-            if population.allow_samples:
```
jeromekelleher: Let's keep this as ``allow_samples``, it's more  readable and it means we don't need to know what the exceptional value is.

stdpopsim/models.py
```diff
@@ -112,18 +112,29 @@ def verify_demographic_events_equal(
...
+        population (default = 0).
+    :vartype sampling_time: int
     """
     # TODO change this to use the usual id, name combination
-    def __init__(self, name, description, allow_samples=True):
+    def __init__(self, name, description, sampling_time=0):
         self.name = name
         self.description = description
-        self.allow_samples = allow_samples
+        self.sampling_time = sampling_time
```
jeromekelleher: And add a property like
```python
@property
def sampling_time(self):
    return self.allow_samples is not None
```

stdpopsim/models.py
```diff
@@ -160,7 +171,7 @@ def num_populations(self):
 
     @property
     def num_sampling_populations(self):
-        return sum(int(pop.allow_samples) for pop in self.populations)
+        return sum(int(pop.sampling_time >= 0) for pop in self.populations)
```
jeromekelleher: Ditto keeping ``allow_samples`` here.

stdpopsim/models.py
```diff
@@ -216,7 +227,14 @@ def get_samples(self, *args):
         """
         samples = []
         for pop_index, n in enumerate(args):
-            samples.extend([msprime.Sample(pop_index, time=0)] * n)
+            # Only add to the samples vector if the sampling time is non-negative
+            if(self.populations[pop_index].sampling_time >= 0):
```
jeromekelleher: brackets are extraneous here in the ``if`` statement.

jeromekelleher: Also, use ``allow_samples``.

stdpopsim/models.py
```diff
@@ -239,7 +257,7 @@ def simulate(self, contig, samples, seed=None):
 _pop0 = Population(name="pop0", description="Generic population")
 _pop1 = Population(name="pop1", description="Generic population")
 _popAnc = Population(name="popAnc", description="Generic ancestral population",
-                     allow_samples=False)
+                     sampling_time=-1)
```
jeromekelleher: sampling_time=None

ndukler: Done

stdpopsim/models.py
ndukler: Done

stdpopsim/models.py
ndukler: Done

stdpopsim/models.py
ndukler: Done

stdpopsim/cli.py
ndukler: Done

stdpopsim/catalog/homo_sapiens.py
ndukler: Done

jeromekelleher:
Have you pushed your changes @ndukler? I'm not seeing any updates here.

stdpopsim/catalog/homo_sapiens.py
ndukler: done

ndukler:
> Have you pushed your changes @ndukler? I'm not seeing any updates here.

Not yet

ndukler:
Just working on adding some tests now

ndukler:
Ok, so I think I see a problem in the CLI. Since non-sampling populations are omitted from the CLI documentation the user will not seem them as present. However when the list of samples to generate are passed in the `run_simulation` function as follows

```
samples = model.get_samples(*args.samples)
````

So we can get a situation with three populations like this

- PopA: Sampling
- PopB: Non-sampling
- PopC: Sampling

but the user will only see PopA and PopC. So if they pass a vector indicating the number of samples they may naturally pass something like this `2, 3` for A and C respectively. However the code as it stands will try to draw 2 samples from PopA and 3 samples from B. I'm going to change the `run_simulations` function so this doesn't happen by inserting 0's in the argument vector where non-sampling populations are present.

**Comment by jeromekelleher** [comment]
Maybe we should force all the sampling populations to be at the lower indexes? Seems confusing to require a zero in the middle of the argument vector.

**Comment by ndukler** [comment]
I implemented a fix that doesn't require that 0 (for the CLI at least). Instead it just re-parses the user's input to insert a 0 at the position of the non-sampling population. That being said, I think your solution would make sampling easier for people using the python API. It would require some model refactoring and some additional checks but I can do it.

**Comment by jeromekelleher** [comment]
Would seem cleaner to do it this way all right, it'll be a consistent source of confusion I'd imagine. Do you mind doing the model refactoring?

**Comment by ndukler** [comment]
I can do that, but someone else will need to redo the QC models. I don't want to risk making a mistake in both.

**Comment by jeromekelleher** [comment]
OK, if you get the main models done I'll add a commit to your PR to finish up the changes on the QC models.

**Comment by ndukler** [comment]
How do we want to ensure that the populations are correctly sorted? I have implemented one version of that where I change `model.populations` to be a property (change the former `populations` to `_populations`) and create a setter functions that performs validity checks as follows:

```
    @property
    def populations(self):
        return(self._populations)

    @populations.setter
    def populations(self, pops):
        if(len(pops) > 0):
            allow_sample_status = [int(p.allow_samples) for p in pops]
            if(sum(allow_sample_status) < len(pops)):
                first_non_sampling = allow_sample_status.index(0)
                last_sampling = len(pops) - allow_sample_status[::-1].index(1)
                print("Test " + str(first_non_sampling) + "  " + str(last_sampling))
                order_valid = (first_non_sampling > last_sampling)
            else:
                order_valid = True
        else:
            order_valid = True
        
        if(not order_valid):
            raise Exception("""Non-sampling populations must have indexes greater than
                            populations which permit sampling""")
        self._populations = pops
```

**Comment by ndukler** [comment]
Although now I'm realizing that I probably want to impose the constraint on the `population_configurations` attribute

**Comment by apragsdale** [comment]
Just to suggest an alternative approach to this issue that wouldn't require refactoring models and might be more intuitive for a user (though I'm not sure this is the best way to go):

Instead of passing a list of sample sizes to `get_samples`, we could pass something like a dictionary that with usage like `get_samples({pop_name1: num_samples1, pop_name2: num_samples2})`, which would throw an exception if the population doesn't allow samples, and doesn't require entering sample sizes in any given order.

I think it would be more explicit and intuitive from a user's perspective, protect users against accidentally putting samples in unintended place in the list, and probably require the least amount of refactoring code in the models?

**Comment by apragsdale** [comment]
Or even something like
`samples = model.get_samples([popA, popB], [nA, nB])`
where we only need the relative order of the lists to match (population name to number of samples from that pop), then inside the function we simply grab the population index from the name?

**Comment by ndukler** [comment]
So there's a variety of issues here, some b/c of the CLI, and some having to do with the python API. I agree that this could be a solution from the python API side but certainly would be a pain for the CLI so we probably have to reorder the populations regardless. I also intend to add a `--dry-run` option once this issue is resolved that would give the user the ability to see what they're going to simulate. I do like your idea though. What do you think @jeromekelleher ?

**Comment by apragsdale** [comment]
Right.. I was just thinking from the python API perspective, and didn't think of the CLI. Though if there's a solution that doesn't require refactoring a bunch of models that have already been QC'd, it's probably worth thinking about it a bit.

**Comment by ndukler** [comment]
I have implemented a version that fixes just the CLI without requiring refactoring but it does result in some weirdness where people using the python API and the CLI have to input different parameter vectors so I don't really like that solution.

As of now I'm doing a two step solution:

1. Force `model.populations` to be ordered such that the non-sampling populations are last (this will require refactoring).
2. Force the `model.population_configurations` to use populations in the same order that they are in `model.populations`.

This makes everything very robust and provides some guarantees that will likely be useful for later development. I know it's a bit of a pain but everything else does feel a bit fragile and this only requires refactoring two models I think.

**Comment by ndukler** [comment]
The only model I had to refactor was the Kamm (2009) model so it really isn't that bad. I think it looks good now. I'm going do a bit of linting and push everything shortly. There are a bunch of changes but most of them are fairly superficial.

**Comment by ndukler** [comment]
I've added checks that enforce the two constraints I laid out about and refactored all the related code. I also added some tests for the added code although I could probably use a few more for the validity check of the `population_configurations` attribute. What do people think?

**Comment by jeromekelleher** [comment]
I agree with @ndukler, and now is surely the time to do any refactorings that are needed. I also agree it would be nice to support some sort of dict approach @apragsdale, where we don't insist on just using the ordering of population --- but, for the CLI, I don't see any other way to do this other than having some ordering or by having some mini-language to describe the samples. We will want to support the simple order based approach anyway for small models, so I think we should do this as a baseline. We can add in other methods later for sampling which will be less error-prone for complex models.

I don't think we need to check that the populations are correctly ordered at run-time @ndukler --- this seems like overkill and extra complexity. Can't we add this to our standard unit tests that we apply to all models (i.e., to the ModelTestMixin) and this will be checked when the models are created?

File: stdpopsim/cli.py [hunk 7]
```diff
@@ -365,8 +365,16 @@ def run_simulation(args):
...
                 "populations")
-        samples = model.get_samples(*args.samples)
-
+        """
+        Create vector of samples with zeros at indexes in place of non-sampling and
+        unspecified populations.
+        """
+        expanded_samples = [0] * model.num_populations
+        for pop_idx in range(model.num_populations):
+            if(args.samples and model.populations[pop_idx].allow_samples):
```
jeromekelleher: Don't need these parentetheses, should be:
```python
if args.samples and model.populations[pop_idx].allow_samples:
```

Also, it's good to avoid using the Falsey tests. What exactly are you testing about ``args.samples`` here? Is it ``args.samples is not None`` or ``len(args.samples) > 0``? It's helpful to the reader to know which you mean.

stdpopsim/models.py
```diff
@@ -162,6 +185,48 @@ def num_populations(self):
     def num_sampling_populations(self):
         return sum(int(pop.allow_samples) for pop in self.populations)
 
+    # Make populations a property to restrict the order of populations
+    @property
```
jeromekelleher: See point below --- I think we can check both of these things as a standard test case, rather than doing it at run time.

stdpopsim/models.py
```diff
@@ -216,7 +281,21 @@ def get_samples(self, *args):
...
+                # Only add to the samples vector if sampling is allowed
+                if(self.populations[pop_index].allow_samples):
+                    samples.extend(
+                        [msprime.Sample(
+                            pop_index,
+                            time=self.populations[pop_index].sampling_time
+                        )] * n
+                    )
+                else:
+                    raise ValueError("""Non-zero number of samples requested from
```
jeromekelleher: This shouldn't be necessary if all sampling populations are adjacent now, right?

tests/test_models.py
```diff
@@ -31,7 +31,14 @@ def test_simulation_runs(self):
         # With a recombination_map of None, we simulate a coalescent without
         # recombination in msprime, with no mutation.
         contig = stdpopsim.Contig()
-        samples = self.model.get_samples(*([2] * self.model.num_populations))
+        # Generate vector with 2 samples for each pop with sampling enabled
+        sample_count = []
+        for p in self.model.populations:
+            if(p.allow_samples):
```
jeromekelleher: Parentheses are unnecessary (also elsewhere)

ndukler:
> I agree with @ndukler, and now is surely the time to do any refactorings that are needed. I also agree it would be nice to support some sort of dict approach @apragsdale, where we don't insist on just using the ordering of population --- but, for the CLI, I don't see any other way to do this other than having some ordering or by having some mini-language to describe the samples. We will want to support the simple order based approach anyway for small models, so I think we should do this as a baseline. We can add in other methods later for sampling which will be less error-prone for complex models.
> 
> I don't think we need to check that the populations are correctly ordered at run-time @ndukler --- this seems like overkill and extra complexity. Can't we add this to our standard unit tests that we apply to all models (i.e., to the ModelTestMixin) and this will be checked when the models are created?

I agree that it would be useful to have another method of sampling for more complex models and I could add implementing @apragsdale 's dict suggestion to my TODO list. My rationale for having the two constraints I implemented being checked at runtime because as the resources associated with `stdpopsim` grow I imagine this package may be used by other people to implement models which they may or may not submit to the central repository. They may or may not ever run the tests so this will ensure that they are doing things correctly. They are also both fairly subtle but very important constraints for having the documentation and sampling done correctly. That being said if people strongly disagree I can move them to the unit tests.

stdpopsim/models.py
ndukler: Yeah, I can revert this to its original form or implement it with @apragsdale 's suggestion

jeromekelleher:
> My rationale for having the two constraints I implemented being checked at runtime is that as resources associated with stdpopsim grow, it may be used by other people to implement models which they may or may not submit to the central repository. They may or may not ever run the tests so this will ensure that they are doing things correctly. They are also both fairly subtle but very important constraints for having the documentation and sampling done correctly. That being said if people strongly disagree I can move them to the unit tests.

Let's keep it as simple as possible for now. We're not documenting these APIs externally anyway, so we're under no obligation to worry about how people will use them. For now, stdpopsim is just about providing a fixed catalog of models. When people are adding new models to the catalog, the tests to check that the populations are correctly ordered will be done automatically, so I'm not worried about errors creeping in there.

jeromekelleher:
> I agree that it would be useful to have another method of sampling for more complex models and I could add implementing @apragsdale 's dict suggestion to my TODO list.

Let's get the simple order-based approach merged, and we can open an issue to track the dict-based method afterwards. It's not essential for the initial release, I think, but might be easy enough to just do anyway.

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -880,94 +880,96 @@ def __init__(self):
...
             # Neanderthal merge into Loschbour / modern humans
             msprime.MassMigration(
-                time=t_Nean_Losch, source=8, destination=4,
+                time=t_Nean_Losch, source=7, destination=3,
                 proportion=1.0),
             # Ancestral hominin population size change
             msprime.PopulationParametersChange(
                 time=t_Nean_Losch, initial_size=N_Nean_Losch,
-                population_id=4),
+                population_id=3),
```
andrewkern: hey @ndukler there are a lot of tweaks to this model in this PR. i think we should probably trigger a QC step here. Rather than have an independent implementation, maybe it would be enough to compare simply summaries of variation between the old and new versions?

tests/test_models.py
andrewkern: is this going to conflict with https://github.com/popgensims/stdpopsim/pull/187?

stdpopsim/catalog/homo_sapiens.py
jeromekelleher: Also shows that hard-coding these population IDs is problematic. What if we have short variable names for these? That would make it easier to read and verify that we're doing the right thing.

ndukler:
Ok, I think I've done resolved everything but I did change a whole bunch so please take a look

stdpopsim/catalog/homo_sapiens.py
andrewkern: agreed

jeromekelleher:
OK, thanks @ndukler. I'll pick this up next Monday.

stdpopsim/catalog/homo_sapiens.py
ndukler: The QC for this model hasn't actually been completed so they could just change the population indexes for their initial run through.

stdpopsim/models.py
```diff
@@ -216,7 +231,16 @@ def get_samples(self, *args):
...
-            samples.extend([msprime.Sample(pop_index, time=0)] * n)
+            if self.populations[pop_index].allow_samples is True:
+                samples.extend(
+                               [msprime.Sample(
+                                    pop_index,
+                                    time=self.populations[pop_index].sampling_time)
+                                ] * n
+                              )
+            elif(n > 0):
+                raise ValueError("Samples requested from non-sampling population (" +
```
jeromekelleher: This is neater:
```python
   raise ValueError(
        f"Samples requested from non-sampling population {pop_index}")
```

stdpopsim/models.py
```diff
@@ -216,7 +231,16 @@ def get_samples(self, *args):
         """
         samples = []
         for pop_index, n in enumerate(args):
-            samples.extend([msprime.Sample(pop_index, time=0)] * n)
+            if self.populations[pop_index].allow_samples is True:
+                samples.extend(
```
jeromekelleher: This isn't the most readable at the moment. How about:
```python
sample = msprime.Sample(
       pop_index, time=self.populations[pop_index].sampling_time))
samples.extend([sample] * n)
```

stdpopsim/models.py
```diff
@@ -216,7 +231,16 @@ def get_samples(self, *args):
         """
         samples = []
         for pop_index, n in enumerate(args):
-            samples.extend([msprime.Sample(pop_index, time=0)] * n)
+            if self.populations[pop_index].allow_samples is True:
```
jeromekelleher: Use 
```python
if self.populations[pop_index].allow_samples:
```
When the value is a boolean. From PEP8:
>  Don't compare boolean values to True or False using ==.
> Yes:   if greeting:
> No:    if greeting == True:
> Worse: if greeting is True:

tests/test_models.py
```diff
@@ -407,3 +414,54 @@ def test_migration_rates(self):
...
+        sample_times = [i.time for i in test_samples]
+        self.assertEqual(sample_times, [0, 0, 10])
+
+    # Test that all sampling populations are specified before non-sampling populations
+    # in the model.populations list
+    def test_population_order(self):
+        for model in stdpopsim.all_models():
+            allow_sample_status = [int(p.allow_samples) for p in model.populations]
+            if sum(allow_sample_status) < len(model.populations):
+                first_non_sampling = allow_sample_status.index(0)
```
jeromekelleher: I'm not following the logic here. What about
```
num_sampling = sum(allow_sample_status)
first_non_sampling = allow_sample_status.index(0)
self.assertEqual(sum(allow_sample_status[:first_non_sampling]), num_sampling)
```

ndukler: This is just checking that the total number of sampling populations is the same as the number of sampling populations up to the index of the first non-sampling population. In other words that all sampling populations are before all non-sampling populations.

ndukler: Mine just gets the index of the last sampling and the first non-sampling populations and checks that the non-sampling comes after the sampling. I'm happy to use your test code instead if you'd prefer.

jeromekelleher: It just wasn't obvious to me what it was doing @ndukler, so it'd be nice to make it more obvious. It's even simpler, right?
```python
allow_sample_status = [int(p.allow_samples) for p in model.populations]
num_sampling = sum(allow_sample_status)
# All sampling populations must be at the start of the list
self.assertEqual(sum(allow_sample_status[num_sampling:]), 0)
```

jeromekelleher:
You need to [rebase](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) this rather than merging @ndukler.

ndukler:
> You need to [rebase](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) this rather than merging @ndukler.

I did it correctly locally, just waiting to push. Did you have a strong opinion on the last point you raised about the test logic?

ndukler:
> > You need to [rebase](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) this rather than merging @ndukler.
> 
> I did it correctly locally, just waiting to push. Did you have a strong opinion on the last point you raised about the test logic?

Ah, nvm, just refreshed and saw you answered. Didn't see it before for some reason.

tests/test_models.py
ndukler: I like this version a lot! I think the previous one may have had an issue when all the samples were sample-able depending on the behavior of `.index(0)` when there was no zero in the vector.

ndukler:
I'm ready to merge whenever. I'll probably have to rebase b/c it looks like master is getting patched. Just let me know when to do it.

jeromekelleher:
Great, thanks @ndukler. OK, one more rebase and we'll merge immediately then. The code base is moving quickly!

ndukler:
Rebased :) @jeromekelleher

jeromekelleher:
Done! Thanks @ndukler.

Issue #168, opened 2019-11-12
Add --help-species


jeromekelleher:
From @ndukler in #164:

> Once this gets pulled I will add a --help-defaults option so that people can see the default population size and generation time for the species (and citations). I realized I couldn't find an easy way for people to see it from the CLI when I was writing the tutorial.

Great idea @ndukler. Is 'defaults' the right name though? You can't override them from the CLI, so I'm not sure it is. They're the basic popgen properties of the species, so perhaps ``--help-species``? (Open to other ideas!)

ndukler:
Sounds good to me.

ndukler:
So I have it working nicely but I would like to add a bit of actual citation information for the default population sizes and generation times. Does anyone have some good references for those? @petrelharp @andrewkern @jeromekelleher

jeromekelleher:
Good question!

ndukler:
Okay, so I added the default mutation rates and recombination rates to this but then realized that there is no good way to automatically produce citations for any of this. In reality what I think would be best is the default mutation rates and recombination rates to operate on a genetic map object stored as `.default_genetic_map` instead of just being passed a string as they are right now. Then the citation could just be pulled from that for any given species. However this could require a bit of coding to set up. Is it worth it or should I just set up some properties like `.default_mutation_rate_cite` as placeholders that can be replaced under the hood later with a better solution? @jeromekelleher

ndukler:
See PR #182

ndukler:
Can we close this @jeromekelleher

jeromekelleher:
FYI: adding "Closes #XYZ" in the body of the commit message will do this automatically.

ndukler:
Ah, thanks! I was confused about how that worked :)

PR #182, opened 2019-11-18
WIP --species-help for CLI

ndukler:
Adds a `--species-help` option to the CLI which details population defaults

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -157,26 +157,10 @@ def get_species_help(species_id):
     species = stdpopsim.get_species(species_id)
     species_text = f"\nDefault population parameters for {species.name}:\n"
 
-    # Add the default generation times and citation(s)
-    indent = " " * 4
-    wrapper = textwrap.TextWrapper(initial_indent=indent, subsequent_indent=indent)
     species_text += f"Generation time: {species.generation_time}\n"
-    for gtc in species.generation_time_citations:
-        species_text += wrapper.fill(textwrap.dedent("- " + str(gtc)))
-        species_text += "\n"
-
-    # Add the default population size and citation(s)
     species_text += f"Population size: {species.population_size}\n"
-    for psc in species.population_size_citations:
-        species_text += wrapper.fill(textwrap.dedent("- " + str(psc)))
-        species_text += "\n"
-
-    # Default mutation rate
     species_text += f"Mutation rate: {species.genome.mean_mutation_rate}\n"
-
-    # Default recombination rate
     species_text += f"Recombination rate: {species.genome.mean_recombination_rate}\n"
-
     return species_text
 
 
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -149,6 +149,37 @@ class HelpGeneticMaps(argparse.Action):
         parser.exit()
 
 
+def get_species_help(species_id):
+    """
+    Generate help text for the given species with some of the species attributes
+    that are not covered by the other helps.
+    """
+    species = stdpopsim.get_species(species_id)
+    species_text = f"\nDefault population parameters for {species.name}:\n"
+
+    # Add the default generation times and citation(s)
+    indent = " " * 4
+    wrapper = textwrap.TextWrapper(initial_indent=indent, subsequent_indent=indent)
+    species_text += f"Generation time: {species.generation_time}\n"
+    for gtc in species.generation_time_citations:
+        species_text += wrapper.fill(textwrap.dedent("- " + str(gtc)))
+        species_text += "\n"
+
+    # Add the default population size and citation(s)
+    species_text += f"Population size: {species.population_size}\n"
+    for psc in species.population_size_citations:
+        species_text += wrapper.fill(textwrap.dedent("- " + str(psc)))
+        species_text += "\n"
+
+    # Default mutation rate
+    species_text += f"Mutation rate: {species.genome.mean_mutation_rate}\n"
+
+    # Default recombination rate
+    species_text += f"Recombination rate: {species.genome.mean_recombination_rate}\n"
+
+    return species_text
+
+
 def get_environment():
     """
     Returns a dictionary describing the environment in which stdpopsim
@@ -271,7 +302,7 @@ def add_simulate_species_parser(parser, species):
         "option to specify a filename."
     )
 
-    description_text = textwrap.fill(header)
+    description_text = textwrap.fill(header) + "\n" + get_species_help(species.id)
 
     species_parser = parser.add_parser(
         f"{species.id}",
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -425,6 +425,10 @@ class TestHelp(unittest.TestCase):
         for species in stdpopsim.all_species():
             self.run_stdpopsim(f"{species} --help-genetic-maps")
 
+    def test_all_species_species_help(self):
+        for species in stdpopsim.all_species():
+            self.run_stdpopsim(f"{species} --help-species")
+
 
 class TestWriteCitations(unittest.TestCase):
     """


stdpopsim/cli.py
```diff
@@ -298,6 +340,11 @@ def add_simulate_species_parser(parser, species):
                 "given as an argument, show help for this map. Otherwise show "
                 "help for all available genetic maps"))
 
+    species_parser.add_argument(
```
jeromekelleher: Probably makes more sense to put ``--help-species`` first, before ``--help-models`` ?

Or, we could structure it slightly and have this as a top-level argument, so that we type
```
$ python3 -m stdpopsim  --help-species homsap
```
rather than
```
$ python3 -m stdpopsim homsap  --help-species 
```

This would be analogous to how we're doing it with the GeneticMaps. It's easy to do, you just add it to the `top_parser` in the ``stdpopsim_cli_parser`` function.

Not a strong opinion though, just a suggestion.

ndukler: I would prefer to keep it as 
"""
python3 -m stdpopsim homsap  --help-species 
"""
because I think it better follows what the logical process of the user is. II my mind, they aren't selecting the species based on the defaults, they're selecting the species, then asking what the defaults are. They aren't going to switch to simulating with one of the other species if they really are working with human data.

ndukler: But I will swap the order of `--help-species` and `-help-models`

jeromekelleher: SGTM!

stdpopsim/cli.py
```diff
@@ -149,6 +149,48 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        species_text += "\n"
+    return species_text
+
+
+class HelpSpecies(argparse.Action):
+    """
+    Action used to produce genetic map help text.
+    """
+    def __init__(self, nargs=0, **kw):
+        if nargs != 0:
```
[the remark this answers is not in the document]
jeromekelleher: Can also use 
```python
 def __init__(self, *args,  **kwargs):
 ```
so you don't have to declare the unused nargs.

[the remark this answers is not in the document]
Comment (ndukler) [reply stdpopsim/cli.py]: Removed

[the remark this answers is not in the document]
Comment (jeromekelleher) [reply stdpopsim/cli.py]: Actually, the ``__init__`` does nothing useful here, you can just remove it entirely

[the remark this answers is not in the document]
Comment (ndukler) [reply stdpopsim/cli.py]: Ah, okay

**Comment by ndukler** [comment]
I'm not sure how to fix the codecov issues. I added a test for the new code, but I'm not sure how/why to have that recognized or if it's being run correctly.

**Comment by jeromekelleher** [comment]
Hmm, after playing with this, I wonder why we don't add this text to the output of ``python3 -m stdpopsim homsap --help``. I.e., it seems odd to have a separate help-topic for just these two small pieces of information. We can add some extra text to the header (or footer) of the rest of the help text that will contain this extra info.

**Comment by ndukler** [comment]
Sure. I just thought that eventually we might want to add additional information to it but we can always add that to a later version. So I guess I'll just add the relevant text to the general `--help` flag?

**Comment by jeromekelleher** [comment]
> Sure. I just thought that eventually we might want to add additional information to it but we can always add that to a later version. So I guess I'll just add the relevant text to the general --help flag?

Well, this will probably still be associated with the *species* so it makes sense to keep it within the species-specific help. I guess you can add the text to the current ``description_text``?

**Comment by andrewkern** [comment]
i agree- i like the idead of `--species-help` as being separate from `--help` myself

**Comment by jeromekelleher** [comment]
> i agree- i like the idead of --species-help as being separate from --help myself

It is though:
```
$ python3 -m stdpopsim homsap --help
usage: __main__.py homsap [-h] [--help-models [HELP_MODELS]]
                          [--help-genetic-maps [HELP_GENETIC_MAPS]] [-q] [-g]
                          [-c] [-l LENGTH_MULTIPLIER] [-s SEED] [-m]
                          [-o OUTPUT]
                          samples [samples ...]

Run simulations for Homo sapiens using up-to-date genome information,
genetic maps and simulation models from the literature. NOTE: By
default, the tskit '.trees' binary file is written to stdout,so you
should either redirect this to a file or use the '--output' option to
specify a filename.

positional arguments:
  samples               The number of samples to draw from each population. At
                        least two samples must be specified. The number of
                        arguments that will be accepted depends on the
                        simulation model that is specified: for a model that
                        has n populations, we can specify the number of
                        samples to draw from each of these populations.We do
                        not need to provide sample numbers of each of the
                        populations; those that are omitted are set to zero.

optional arguments:
  -h, --help            show this help message and exit
  --help-models [HELP_MODELS]
                        Print descriptions of simulation models and exit. If a
                        model ID is provided as an argument show help for this
                        model; otherwise show help for all available models
  --help-genetic-maps [HELP_GENETIC_MAPS]
                        Print list of genetic maps and exit. If a genetic map
                        ID is given as an argument, show help for this map.
                        Otherwise show help for all available genetic maps
  -q, --quiet           Do not write out citation information
  -g , --genetic-map    Specify a particular genetic map. If no genetic map is
                        specified use a flat map by default. Available maps:
                        HapmapII_GRCh37, Decode_2010_sex_averaged.
  -c , --chromosome     Simulate a specific chromosome. Options: chr1, chr2,
                        chr3, chr4, chr5, chr6, chr7, chr8, chr9, chr10,
                        chr11, chr12, chr13, chr14, chr15, chr16, chr17,
                        chr18, chr19, chr20, chr21, chr22, chrX, chrY.
                        Default=chr1.
  -l LENGTH_MULTIPLIER, --length-multiplier LENGTH_MULTIPLIER
                        Simulate a chromsome of length l times the named
                        chromosome
  -s SEED, --seed SEED  The random seed to use for simulations. If not
                        specified a high-quality random seed will be generated
                        automatically. For msprime, seeds must be > 0 and <
                        2^32.
  -m , --model          Specify a simulation model. If no model is specified,
                        a single populationconstant size model is used.
                        Available models:ooa_3, ooa_2, african, america,
                        ooa_archaic, zigzag, kamm_ancient_eurasia. Please see
                        --help-models for details of these models.
  -o OUTPUT, --output OUTPUT
                        Where to write the output tree sequence file. Defaults
                        to stdout if not specified

```

The current suggestion is to modify this help output to look something like
```
Run simulations for Homo sapiens using up-to-date genome information,
genetic maps and simulation models from the literature. NOTE: By
default, the tskit '.trees' binary file is written to stdout,so you
should either redirect this to a file or use the '--output' option to
specify a filename.

species defaults:
  generation time:      25 years
  population size:       10000 

positional arguments:
  samples               The number of samples to draw from each population. At
                        least two samples must be specified. The number of
                        arguments that will be accepted depends on the
                        simulation model that is specified: for a model that
                        has n populations, we can specify the number of
                        samples to draw from each of these populations.We do
                        not need to provide sample numbers of each of the
                        populations; those that are omitted are set to zero.
[snip]
```

There just doesn't seem much point in having an extra command just to put in two extra bits of information.

**Comment by ndukler** [comment]
> Well, this will probably still be associated with the _species_ so it makes sense to keep it within the species-specific help. I guess you can add the text to the current `description_text`?

Sorry, yeah that's what I meant. I'll get on it

**Comment by jeromekelleher** [comment]
How's this looking @ndukler? We should try to finish it up and get it merged, I think.

**Comment by ndukler** [comment]
It works fine, just needs some test coverage. I can look into it today and try to finish it up.

**Comment by ndukler** [comment]
Oops, I just realized that it also need to have information about the default mutation and recombination rate.

**Comment by ndukler** [comment]
Okay, so I added the default mutation rates and recombination rates to this but then realized that there is no good way to automatically produce citations for any of this. In reality what I think would be best is the default mutation rates and recombination rates to operate on a genetic map object stored as `.default_genetic_map` instead of just being passed a string as they are right now. Then the citation could just be pulled from that for any given species. However this could require a bit of coding to set up. Is it worth it or should I just set up some properties like `.default_mutation_rate_cite` as placeholders that can be replaced under the hood later with a better solution? @jeromekelleher

File: stdpopsim/cli.py [hunk 3]
```diff
@@ -149,6 +149,44 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+
+    # Add the default population size and citation(s)
+    species_text += f"Population size: {species.population_size}\n"
+    for psc in species.population_size_citations:
+        species_text += wrapper.fill(textwrap.dedent("- " + str(psc)))
+        species_text += "\n"
+
+    # Default mutation rate
+    species_text += f"Mutation rate: {species.genome.mean_mutation_rate}\n"
+    # for mrc in species.genome.mutation_rate_citations:
```
jeromekelleher: I don't think we need citations here in the help --- we only need to write these out afterwards so that users can cite the literature relevant to the simulations they've actually run. Citations in the help text don't help the user in any way, I think.

stdpopsim/cli.py
```diff
@@ -149,6 +149,44 @@ def __call__(self, parser, namespace, values, option_string=None):
         parser.exit()
 
 
+def get_species_help(species_id):
+    """
+    Generate help text for the given species with some of the species attributes
+    that are not covered by the other helps. Does not accept any additional
```
jeromekelleher: The last sentance is redundant.

stdpopsim/cli.py
ndukler: ok

jeromekelleher:
Merging. I've added a commit here to remove the citations entirely @ndukler. There's two reasons for this 
1. I don't think citations are actually helpful here --- this is the help text, not the postscript nagging the users to cite the relevant literature. If they're interested they can always look in the online docs.
2. There aren't currently any citations, so we never test this code.

We can follow up in issues if you want to revert back to including the citations later when they're defined.

ndukler:
Thanks! Closes #168

PR #183, opened 2019-11-18
IM model comments

apragsdale:
This had been merged (#172), but @grahamgower pointed out in #171 that the description in the comments didn't match the model after I had fiddled around with the setting up the populations. This should fix that.


apragsdale:
Looks to be superseded by #184, so going to close.

PR #187, opened 2019-11-21
Implement Engine class and remove Model.simulate().

grahamgower:
This PR is separated into ~~two~~multiple commits. The first adds the ~~`SimulationEngine`~~`Engine` class, retaining API backwards compatibility by making `Model.simulate()` use the default simulation engine. The second commit removes `Model.simulate()` and changes any users of this function so that they first obtain the default simulation engine and use its `simulate()` method. Remaining commits can be squashed in once everyone is happy.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -217,7 +217,7 @@ def write_output(ts, args):
         ts.dump(args.output)
 
 
-def write_citations(contig, model):
+def write_citations(engine, model, contig):
     """
     Write out citation information so that the user knows what papers to cite
     for the simulation engine, the model and the mutation/recombination rate
@@ -231,9 +231,10 @@ def write_citations(contig, model):
     printerr("******************")
     printerr("Simulation engine:")
     printerr("******************")
-    printerr(
-        "\tmsprime: Kelleher et al. 2016: "
-        "https://doi.org/10.1371/journal.pcbi.1004842")
+    printerr(f"{engine.name}:")
+    for citation in engine.citations:
+        printerr(f"\t{citation.author} {citation.year}")
+        printerr(f"\t{citation.doi}")
     printerr("******************")
     printerr("Genetic map:")
     printerr("******************")
@@ -370,14 +371,18 @@ def add_simulate_species_parser(parser, species):
         contig = species.get_contig(
             args.chromosome, genetic_map=args.genetic_map,
             length_multiplier=args.length_multiplier)
+        engine = stdpopsim.get_engine(args.engine)
         logger.info(
             f"Running simulation model {model.name} for {species.name} on "
-            f"{contig} with {len(samples)} samples")
-        ts = model.simulate(contig, samples, args.seed)
+            f"{contig} with {len(samples)} samples using {engine.name}.")
+
+        kwargs = vars(args)
+        kwargs.update(model=model, contig=contig, samples=samples)
+        ts = engine.simulate(**kwargs)
         summarise_usage()
         write_output(ts, args)
         if not args.quiet:
-            write_citations(contig, model)
+            write_citations(engine, model, contig)
 
     species_parser.set_defaults(runner=run_simulation)
 
@@ -417,6 +422,17 @@ def stdpopsim_cli_parser():
             "STDPOPSIM_CACHE. If both the environment variable and this "
             "option are set, the option takes precedence. "
             f"Default: {stdpopsim.get_cache_dir()}"))
+
+    top_parser.add_argument(
+        "-e", "--engine",
+        default=stdpopsim.get_default_engine().id,
+        choices=[e.id for e in stdpopsim.all_engines()],
+        help="Specify a simulation engine.")
+
+    for engine in stdpopsim.all_engines():
+        group = top_parser.add_argument_group(
+                f"{engine.name} specific parameters")
+        engine.add_arguments(group)
 
     subparsers = top_parser.add_subparsers(dest="subcommand")
     subparsers.required = True
--- a/stdpopsim/engines.py
+++ b/stdpopsim/engines.py
@@ -1,0 +1,127 @@
+import logging
+
+import attr
+import msprime
+import stdpopsim
+
+logger = logging.getLogger(__name__)
+
+_registered_engines = {}
+
+
+def register_engine(engine):
+    """
+    Registers the specified simulation engine.
+    """
+    logger.debug(f"Registering simulation engine '{engine.id}'")
+    _registered_engines[engine.id] = engine
+
+
+def get_engine(id):
+    """
+    Returns the simulation engine with the specified id.
+    """
+    if id not in _registered_engines:
+        raise ValueError(f"Simulation engine '{id}' not registered")
+    return _registered_engines[id]
+
+
+def all_engines():
+    """
+    Returns an iterator over all registered simulation engines.
+    """
+    for sim_engine in _registered_engines.values():
+        yield sim_engine
+
+
+@attr.s(frozen=True)
+class Engine(object):
+    """
+    Abstract class representing a simulation engine.
+
+    To implement a new simulation engine, one should inherit from this
+    class. At a minimum, the ``id``, ``name`` and ``citations`` attributes must
+    be set, and the :func:`simulate()` and :func:`get_version()` methods must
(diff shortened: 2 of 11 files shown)


stdpopsim/cli.py
```diff
@@ -217,7 +217,7 @@ def write_output(ts, args):
         ts.dump(args.output)
 
 
-def write_citations(contig, model):
+def write_citations(contig, model, engine):
```
jeromekelleher: Let's standardise on (engine, model, contig) for these positional args

stdpopsim/cli.py
```diff
@@ -418,6 +425,19 @@ def stdpopsim_cli_parser():
             "option are set, the option takes precedence. "
             f"Default: {stdpopsim.get_cache_dir()}"))
 
+    sim_engines = [se.id for se in stdpopsim.all_simulation_engines()]
+    top_parser.add_argument(
+        "-E", "--simulation-engine",
```
jeromekelleher: Maybe just ``--engine``? The ``simulation`` bit could be understood, and this would save us a bit of typing. (Also elsewhere in the API, e.g. ``stdpopsim.all_engines()``. It's an application for running simulations, so it seems OK that an "engine" would be interpreted as a simulation engine.

Happy to discuss other terminology though (you originally preferred 'backend').

stdpopsim/simulation_engines.py
```diff
@@ -0,0 +1,92 @@
...
+        See msprime example in `simulation_engines.py`.
+    """
+    id = attr.ib(type=str, kw_only=True)
+    name = attr.ib(type=str, kw_only=True)
+    citations = attr.ib(factory=list, kw_only=True)
+    add_arguments = attr.ib(default=None, kw_only=True)  # function
+    simulate = attr.ib(kw_only=True)  # function
+
+
+def _msprime_simulate_wrapper(model, contig, samples, **kwargs):
```
jeromekelleher: What we could do here is,
```python
def _msprime_simulate_wrapper(model=None, contig=None, samples=None, *, **kwargs):
```
This would allow us to use the initial three args as keyword args, and also make sure that all subsequence args are keywords. (Second bit isn't essential)

tests/test_simulation_engines.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_get_default_simulation_engine(self):
+        engine = stdpopsim.get_default_simulation_engine()
+        self._test_simulation_engine(engine)
+
+    def test_all_simulation_engines(self):
+        for engine in stdpopsim.all_simulation_engines():
+            self._test_simulation_engine(engine)
+
+    def test_register_engine(self):
```
jeromekelleher: This tests has a side-effect, which might bite us later. Not sure what do about it to be honest; might be OK to just drop the test.

stdpopsim/cli.py
andrewkern: reading through this PR now. I vote for `--engine`` too

stdpopsim/simulation_engines.py
andrewkern: I think we want the`SimulationEngine` class to carry around the version number of the engine too. Can we add a class variable for that? Also is that simple enough to access from SLiM programmatically? I don't know the SLiM side well

andrewkern:
Yes @grahamgower this is going to be awesome! Thanks so much

stdpopsim/simulation_engines.py
grahamgower: So, you want to call out to the slim binary each time the stdpopsim library is imported to populate this variable? Or did you mean you want a minimum supported version number?

I don't think there's a simple function to get the version number from within SLiM. If you were really keen, of course, you could do something crazy like: `system("slim -version | awk '{split($3, a, ","); print a[1]}' ")` and capture the output.

stdpopsim/cli.py
grahamgower: Yep, I'll do s/simulation_engine/simulation_engine/ and s/SimulationEngine/Engine/ too for consistency.

stdpopsim/simulation_engines.py
andrewkern: yes i was thinking the former-- calling out to `slim` to get the version, if it were not available from returned tree sequence. that simple `system()` call is absolutely fine by me, if not elegant.

tests/test_simulation_engines.py
grahamgower: This is just within the scope of the running tests, though, right? There is a (somewhat hackish) way to remove this, because the underscored dict is only private by convention.
```
import stdpopsim.simulation_engines
del stdpopsim.simulation_engines._registered_simulation_engines["test-engine"]
```

A test like this could be added for `register_species()` too.

stdpopsim/simulation_engines.py
jeromekelleher: ``get_version()`` can be a method of the engine, called on demand. I think the version information is captured in detail in the Provenance information in the returned tree sequence too though.

tests/test_simulation_engines.py
jeromekelleher: Yes, it's just in the context of running the tests. I guess we could clean it up like you say. Another thing we could do is maybe mock that variable or something?

stdpopsim/simulation_engines.py
grahamgower: I added a get_version() method (its simpler than using an attribute for calling to an external binary), and changed changed engine implementation to be via subclassing. The attr.ib approach seems inappropriate for classes containing many methods.

stdpopsim/cli.py
```diff
@@ -370,14 +371,21 @@ def run_simulation(args):
...
             length_multiplier=args.length_multiplier)
+        engine = stdpopsim.get_engine(args.engine)
         logger.info(
             f"Running simulation model {model.name} for {species.name} on "
-            f"{contig} with {len(samples)} samples")
-        ts = model.simulate(contig, samples, args.seed)
+            f"{contig} with {len(samples)} samples using {engine.name}.")
+
+        kwargs = vars(args)
+        for var in ("model", "contig", "samples"):
```
jeromekelleher: We don't need to do this now, right?

grahamgower: I'm not sure which *exact* thing you're referring to... Did I misunderstand your earlier comment about how to clean up the CLI/kwargs?

In the future, I anticipate that engines (in particular, the SLiM engine) will want their own command line arguments, and this kwargs mechanism is the cleanest way I could think of to get at those CLI arguments from the engine's `simulate()` function. Currently, the `seed` CLI argument is the only parameter in this kwargs dict that is being used. It is named explicitly in the parameter list for `_MsprimeEngine.simulate()`. I guess a seed parameter could instead be part of the API---i.e. have it as a named parameter to the abstract `Engine.simulate()`.

`vars(args)` produces a dictionary of the CLI args, which contains keys for both `model` and `samples` as these names are used in calls to `species_parser.add_arguments()`. So it's necessary to remove these from the dictionary, as the names clash with the explicit model and samples objects we wish to give `simulate()`.

jeromekelleher: The vars thing is very neat --- I just meant we don't need the for loop, because we can call 
```python
engine.simulate(**kwargs)
```
now, right?

grahamgower: Yes, good point. I've fixed this, and squashed all the commits, ready for merging.

stdpopsim/engines.py
```diff
@@ -0,0 +1,122 @@
...
+    :vartype name: str
+    :ivar citations: A list of citations for the simulation engine.
+    :vartype citations: list of :class:`.Citation`
+    """
+
+    def simulate(self, model=None, contig=None, samples=None, **kwargs):
+        """
+        Function to execute the actual simulation.
+
+        :var model
```
jeromekelleher: Sorry, didn't spot this earlier --- this should be ``:param Model model: The model to simulate``. I think the ``var`` sphinx types are for instance variables. See below also.

jeromekelleher:
Just spotted a minor issue with the sphinx documentation @grahamgower --- good to merge after that though.

grahamgower:
Ok, API docs fixed now.

jeromekelleher:
Fantastic, thanks @grahamgower! Merging.

andrewkern:
this is great @grahamgower!

PR #210, opened 2019-11-28
Fix wrong 'attr' package in setup.py

jeromekelleher:
First commit tries to check this in CI to verify.


jeromekelleher:
Hmm, looks like this is hard to trigger a problem on. Oh well, I've added a note in setup.py to make it hard for us to regress on this on.

PR #218, opened 2019-11-29
Add more tests

grahamgower:
The test_citations() change might conflict with #214.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -466,9 +466,9 @@ class TestWriteBibtex(unittest.TestCase):
 
 class TestWriteCitations(unittest.TestCase):
     """
-    Make sure all models can write citation information.
+    Make sure citation information is written.
     """
-    def test_model(self):
+    def test_model_citations(self):
         contig = stdpopsim.Contig()
         species = stdpopsim.get_species("homsap")
         model = species.get_model("ooa_3")
@@ -476,21 +476,36 @@ class TestWriteCitations(unittest.TestCase):
         stdout, stderr = capture_output(
                 cli.write_citations, engine, model, contig)
         self.assertEqual(len(stdout), 0)
-        # TODO Parse out the output for the model and check that the text is
-        # in there.
-        self.assertGreater(len(stderr), 0)
+        genetic_map = None
+        self.check_citations(engine, species, genetic_map, model, stderr)
 
-    def test_genetic_map(self):
+    def test_genetic_map_citations(self):
         species = stdpopsim.get_species("homsap")
-        contig = species.get_contig("chr22", genetic_map="HapmapII_GRCh37")
+        genetic_map = species.get_genetic_map("HapmapII_GRCh37")
+        contig = species.get_contig("chr22", genetic_map=genetic_map.name)
         model = stdpopsim.PiecewiseConstantSize(species.population_size)
         engine = stdpopsim.get_default_engine()
         stdout, stderr = capture_output(
                 cli.write_citations, engine, model, contig)
         self.assertEqual(len(stdout), 0)
-        # TODO Parse out the output for the model and check that the text is
-        # in there.
-        self.assertGreater(len(stderr), 0)
+        self.check_citations(engine, species, genetic_map, model, stderr)
+
+    def check_citations(self, engine, species, genetic_map, model, stderr):
+        if genetic_map is None:
+            genetic_map = stdpopsim.GeneticMap(species.id, citations=[])
+        for citations, assert_msg in zip(
+                (engine.citations, model.citations, genetic_map.citations,
+                    species.generation_time_citations,
+                    species.population_size_citations),
+                (f"engine citation not written for {engine.id}",
+                    f"model citation not written for {model.id}",
+                    f"genetic map citation not written for {genetic_map.name}",
+                    f"generation time citation not written for {species.id}",
+                    f"population size citation not written for {species.id}")):
+            for citation in citations:
+                self.assertTrue(citation.author in stderr, msg=assert_msg)
+                self.assertTrue(str(citation.year) in stderr, msg=assert_msg)
+                self.assertTrue(citation.doi in stderr, msg=assert_msg)
 
 
 class TestCacheDir(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -428,31 +430,37 @@ def test_all_species_genetic_maps_help(self):
...
+            genetic_maps = copy.copy(species.genetic_maps)
+            if len(genetic_maps) == 0:
+                genetic_maps = [stdpopsim.GeneticMap(species.id)]
+            for genetic_map, model in itertools.product(
+                    genetic_maps, species.models):
+                contig = species.get_contig(
+                        species.genome.chromosomes[0].id,
+                        genetic_map=genetic_map.name)
+                stdout, stderr = capture_output(
+                    cli.write_citations, engine, model, contig)
```
Chris1221: I haven't run your code here, but #214 doesn't alter the output streams other than appending a line to the end telling the user where the bibtex was written. So I think the only thing needed to make it compatible is to add in a `None` flag to the `write_citations` function. I.e. 

```py
bibtex=None
stdout, stderr = capture_output(
                    cli.write_citations, engine, model, contig, bibtex)
```

which is the same thing as in the [PR](https://github.com/popgensims/stdpopsim/blob/35b39a19f4abece25ab14093a7bb54bfba432d4c/tests/test_cli.py#L471-L472). It's still a WIP, but I don't see that part of it changing!

tests/test_cli.py
```diff
@@ -428,31 +430,37 @@ def test_all_species_genetic_maps_help(self):
...
-        model = stdpopsim.PiecewiseConstantSize(species.population_size)
-        engine = stdpopsim.get_default_engine()
-        stdout, stderr = capture_output(
-                cli.write_citations, engine, model, contig)
-        self.assertEqual(len(stdout), 0)
-        # TODO Parse out the output for the model and check that the text is
-        # in there.
-        self.assertGreater(len(stderr), 0)
+    def test_citations(self):
+        for engine, species in itertools.product(
```
jeromekelleher: Nice. I'm not sure we need to take the products here though. Perhaps we could split this into set of discrete tests, ``test_engine`` and ``test_species``, ``test_genetic_maps`` etc. There's two problems with taking products I think (1) it makes things slow, at some point, as your testing a lot of stuff that doesn't change, and (2) it can be hard to see what's after breaking when a test fails inside in nested loops. Having discrete tests (that uses common verification code) is a nice pattern for this.

tests/test_genetic_maps.py
```diff
@@ -64,7 +64,7 @@ def __init__(self):
 
 # TODO add some parameters here to check different compression options,
 # number of chromosomes etc.
-def get_genetic_map_tarball():
+def get_genetic_map_tarball(custom_file_f=None, filter=None):
```
jeromekelleher: Cunning! A quick comment here noting what the parameters are for would help the reader though.

tests/test_genetic_maps.py
```diff
@@ -150,6 +154,37 @@ def test_correct_url(self):
                 gm.download()
         mocked_get.assert_called_once_with(gm.url, filename=unittest.mock.ANY)
 
+    def test_bad_tar_members(self):
+
+        def mock_retrieve_factory():
```
jeromekelleher: This is two separate tests I think --- the code they share is pretty trivial, and it's easier to read as two tests.

tests/test_cli.py
grahamgower: I agree that there will be too many tests. I'm not sure what choices to make here regarding which engines, species, genetic_maps, and models, should be checked.

As for your point (2), I don't think this too bad, as the assertion message indicates the specific thing that fails (I changed the code slightly to make this more obvious). Eg. the test currently fails like this:

```
======================================================================
FAIL: test_citations (tests.test_cli.TestWriteCitations)
----------------------------------------------------------------------
  File "/home/grg/src/stdpopsim/tests/test_cli.py", line 449, in test_citations
    self.check_citations(engine, species, genetic_map, model, stderr)
  File "/home/grg/src/stdpopsim/tests/test_cli.py", line 461, in check_citations
    self.assertTrue(citation.author in stderr, msg=assert_msg)
AssertionError: False is not true : genetic_map: HapmapII_GRCh37
```

jeromekelleher: For unit tests I think it's reasonable to pick one thing to vary, and keep everything else constant unless there are definitely tricky interactions between the things we're testing. In this case, there aren't any interactions between, say, genetic maps and simulation models, so there's no point in doing a cross product there.

grahamgower: Ok, done. I also made the assertion failure messages more informative. Current failures are:
```
AssertionError: False is not true : genetic map citation not written for HapmapII_GRCh37
```
and
```
AssertionError: False is not true : generation time citation not written for TestSpecies
```

jeromekelleher:
Looks like this is ready to go --- can you update please @grahamgower?

tests/test_genetic_maps.py
```diff
@@ -171,6 +184,59 @@ def test_multiple_threads_downloading(self):
             gm.is_cached = saved
 
 
+class TestGeneticMapDownload_security(tests.CacheWritingTest):
```
jeromekelleher: Class names are CamelCase according to pep8. Might become something that generates pep8 failures in the future, so might as well stick to the conventions.

jeromekelleher:
So, these should pass once we've rebased, now we've merged #239 @grahamgower?

grahamgower:
I rebased again. #239 only fixes one of the tests here. Citations listed in `species.generation_time_citation` and `species.population_size_citation` are not currently being written (write_citation() doesn't have a species object with which to write this).

jeromekelleher:
> I rebased again. #239 only fixes one of the tests here. Citations listed in species.generation_time_citation and species.population_size_citation are not currently being written (write_citation() doesn't have a species object with which to write this).

Ah, OK. How do you want to move this forward? (Wait for PRs to fix this issue, or fix it here?)

grahamgower:
Maybe just wait with this PR. I'll fix the generation_time_citation and population_size_citation stuff in a separate PR.

jeromekelleher:
Awesome, thanks @grahamgower.

grahamgower:
@jeromekelleher I changed my mind. I've removed the generation_time_citation and population_size_citation bits. There are some semantic issues with these that I'll raise in a new issue. Please merge this PR.

jeromekelleher:
Done - thanks @grahamgower.

Issue #144, opened 2019-10-09
Improve citation rendering in CLI


jeromekelleher:
Outputing citation information is an important part of stdpopsim. Currently, the rendering of  citations in our output is crude, partial and ugly. We should be able to make a text rendering that's much prettier.

This would be a good first issue for someone looking to do some coding on the internals of stdpopsim. The code for rendering is in ``cli.py``.

```
$ python3 -m stdpopsim homsap -m ooa_3 -l 0.01 10 tmp.trees
If you use this simulation in published work, please cite the following papers:
******************
Simulation engine:
******************
        msprime: Kelleher et al. 2016: https://doi.org/10.1371/journal.pcbi.1004842
******************
Genetic map:
******************
        TODO
Simulation model: Three population out-of-Africa
        Gutenkunst et al., 2009: https://doi.org/10.1371/journal.pgen.1000695
```

agladstein:
How would you like it to look prettier?
If I were using it, I would love a plain text output that I could copy and also a latex citation in bib format. 
I suppose you could have a cli citation option to output an actual .bib file 😃 
Maybe that would be too much. But, if I were a user, I would love it.

jeromekelleher:
> How would you like it to look prettier?

I don't know really, I was hoping someone would be creative and come up with a nice pretty way of rendering this!

BibTex output would be awesome, but would require a fair bit more infrastructure (well, we'd need to store valid bibtex records for each paper). Not that big a deal, we could just copy in the Bibtex records from Google Scholar, but a bit of work all the same.

If this is something you're interested in doing, please open an issue to track it!

agladstein:
I see. I don't really have ideas on how to make it prettier either...

For BibTex output, I created the issue https://github.com/popgensims/stdpopsim/issues/157

Chris1221:
> BibTex output would be awesome, but would require a fair bit more infrastructure.

If it's reasonable to assume that the user has an internet connection when using the CLI, it could be done on the fly through crossref with the DOIs we already store. ([e.g.](https://tex.stackexchange.com/questions/6848/automatically-dereference-doi-to-bib))

```bash
curl -LH "Accept: text/bibliography; style=bibtex" https://doi.org/10.1371/journal.pcbi.1004842
```

returns 

```bibtex
[...]

PR #219, opened 2019-11-29
Print genetic map citations by adding a genetic_map attrib to Contig.

grahamgower:


stdpopsim/citations.py
```diff
@@ -24,4 +24,7 @@ class Citation(object):
     year = attr.ib(type=int, kw_only=True)
 
     def __str__(self):
-        return f"{self.author}, {self.year}: {self.doi}"
+        if self.author is not None:
+            return f"{self.author}, {self.year}: {self.doi}"
+        else:
```
jeromekelleher: Is there a valid reason for not having an author and year? If not, I think this will just mask mistakes. Better for something to be obviously wrong so that we can fix it I think.

stdpopsim/cli.py
```diff
@@ -224,27 +224,18 @@ def write_citations(engine, model, contig):
     information.
     """
     printerr = functools.partial(print, file=sys.stderr)
-    # TODO say this better
-    printerr(
-        "If you use this simulation in published work, please cite the following "
-        "papers:")
-    printerr("******************")
```
jeromekelleher: Deleting the banners is related to #144.

stdpopsim/citations.py
grahamgower: I didn't look to see how widespread it is, but at least the `HapmapII_GRCh37` genetic map has a url-only citation, with the url stuffed into the doi field.

grahamgower: Ok, I removed this change now.

jeromekelleher: Great --- I got rid of the URL citation in HapmapII_GRCh37

jeromekelleher:
I think this is ready to merge --- can you do a quick rebase please @grahamgower?

jeromekelleher:
Closes #144.

jeromekelleher:
Looks good, thanks @grahamgower. Can you update this to take into account the newly merged bibtex feature please? Good to merge then I think.

jeromekelleher:
Thanks @grahamgower, merging.

Issue #215, opened 2019-11-29
Improve CLI help output


gtsambos:
Some of the help output in the CLI could be clearer. I'm going to make a pull request to deal with some relatively minor things (typos etc), but have some bigger questions that I'll turn into separate issues to log here.

jeromekelleher:
I think we've done some good work here --- can we close down this meta issue?

gtsambos:
Yes -- I think we're now at the point where individual issues for specific features will suffice

Issue #118, opened 2019-10-01
Documentation for CLI


jeromekelleher:
We need some documentation for the CLI. At a minimum we want:

- CLI docs generated by [sphinx_argparse](https://sphinx-argparse.readthedocs.io/en/stable/)

- A section of the tutorial showing how to use it.

ndukler:
I can work on this.

ndukler:
@jeromekelleher This looks awesome! Can we make the output of `stdpopsim species --help` a little less verbose. Maybe add a `--show-model-details` or something so that we don't get all the model text?

ndukler:
What exactly do you have in mind for the CLI docs? As in a subheading for an example CLI use for each model under the catalog like we used to have? I'll get on the tutorial though.

jeromekelleher:
> What exactly do you have in mind for the CLI docs? As in a subheading for an example CLI use for each model under the catalog like we used to have? I'll get on the tutorial though.

Sphinx arg-parse should handle most of it -- won't really know until we try it out and see what the output looks like.

A tutorial section would be **super** helpful though!

jeromekelleher:
I think we can close this with #165 and #197.

PR #222, opened 2019-12-01
Added page showing CLI options using sphinx-argparse.

### LINKED ISSUE #197: Docs/API: List of CLI options
At the moment, the docs don't show users how the CLI works, or what options there are. I know @ndukler is working on a tutorial to show how some of these work in #165, but it would be good to have a list of all of the possible arguments (including all of the possible species, recombination maps, model arguments) somewhere in the docs too. Either on an API page, or elsewhere.

Subissue of #137

gtsambos:
Closes #197 and #118; uses a very vanilla implementation of sphinx-argparse to show command-line options in the docs. Hopefully this will make also make it easier for us to spot which parts of the CLI help could be improved (see #215).

I have to say, I don't think that the default way that sphinx has organised and rendered these is going to be very clear or accessible for users of `stdpopsim`. I don't even know what 'Positional' and 'Named' arguments are, but these get their own section heading here. All of the simulation-specific commands are hidden inside the 'Subcommands' subsection.

Nevertheless, might be useful to pull this in now so that others can see and comment on it, and we can address these issues in a later pull request.

### DIFF
--- a/docs/cli_arguments.rst
+++ b/docs/cli_arguments.rst
@@ -1,0 +1,10 @@
+.. _sec_cli_args:
+
+====================
+Command-line options
+====================
+
+.. argparse::
+	:module: stdpopsim.cli
+	:func: stdpopsim_cli_parser
+	:prog: stdpopsim
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,6 +65,7 @@ extensions = [
     'sphinx.ext.ifconfig',
     'sphinx.ext.viewcode',
     'speciescatalog',
+    'sphinxarg.ext',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,6 +14,7 @@ Welcome to stdpopsim's documentation!
    installation
    catalog
    tutorial
+   cli_arguments
    api
    development
    changelogs


andrewkern:
>  I don't even know what 'Positional' and 'Named' arguments are, but these get their own section heading here.

so IIRC python has two kinds of args: keywords and positional. keyword args are named like in `f(x=2, y=4)` x and y would be keyword args and can be set with a key, value pair dict using `**`. positional args are not keyword args, and have to be given in a particular order to a function. 

i'm guessing this is what sphinx-argparse is talking about?

gtsambos:
> so IIRC python has two kinds of args: keywords and positional. keyword args are named like in f(x=2, y=4) x and y would be keyword args and can be set with a key, value pair dict using **. positional args are not keyword args, and have to be given in a particular order to a function.

This makes sense. Still, I don't think this is going to be too helpful for users, who may not even be users of Python. I think it would be better if we could organise the headings/ titles so that they mirror those in the Catalog more closely, and perhaps put all of the 'global' options (--help, --verbosity etc) under a single heading called 'Global' or something like that.

andrewkern:
gosh that would be great. i don't find these argparse defaults helpful either

grahamgower:
The 'positional' and 'named' parameters are command line parameters, which differ slightly from positional and keyword args in python functions. There's a terrible lack of explanation regarding what these mean in the argparse documentation.
https://docs.python.org/3/library/argparse.html

Maybe this page will help (it calls positional parameters "operands" and named parameters are called "options").
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

jeromekelleher:
How CLI arguments are discussed and defined is debated. Positional arguments are things like the basic arguments to ``mv``. Named arguments are things like ``--help, -h``. Normally, we'd call these options, but some people have structured their CLIs so that these 'options' are mandatory (like, having things like ``--out``, or whatever). Hence the slightly awkward terminology.

This has nothing to do with Python's positional and keyword args though.

I'm going to merge this so that we can see what it looks like, and iterate on improving things in follow-ups.

Thanks @gtsambos!

gtsambos:
Hmm, something hasn't quite worked here. When I navigate to the page, it's empty:

Screen Shot 2019-12-04 at 8 37 36 am

FYI, the local html file looks like this in my browser:

Screen Shot 2019-12-04 at 8 38 47 am

jeromekelleher:
Fixed in #237 and #238. We needed to be able to run the cli on RTD and some requirements were missing.

PR #225, opened 2019-12-01
improving a_thal and added model

andrewkern:
doing some clean up on the existing `Durvasula2017MSMC()` model and added a simpler, two epoch model that was published in 2018 by Christan Huber et al. 

I'll start a QC issue for this now.


jeromekelleher:
LGTM. I forget; do we merge this now and then start QC?

andrewkern:
yup we merge now without CLI hooks or API docs, then start QC

Issue #226, opened 2019-12-01
QC for new Arabidopsis model (HuberTwoEpoch)


andrewkern:
Started a PR for this new model here https://github.com/popgensims/stdpopsim/pull/225

Anyone up for QC? This should be quite straight forward as the model params are plainly listed in this supplement: https://static-content.springer.com/esm/art%3A10.1038%2Fs41467-018-05281-7/MediaObjects/41467_2018_5281_MOESM1_ESM.pdf

ndukler:
I can handle this.

andrewkern:
thanks @ndukler. take a look at table S1 for the two epoch model params. once we merge this we might as well do the three epoch model too

ndukler:
@andrewkern The model got merged in without rebasing. Could you quickly set master back one commit , then rebase the `athal_huberModel` then re-merge it?

ndukler:
To clarify, not the QC model, PR #225 .

andrewkern:
sorry to be dense here-- where is the issue? looks like a single commit?

ndukler:
It is a single commit. It just wasn't rebased against the head of master before it was merged.

ndukler:
Ok, I have the QC model ready.

andrewkern:
so should I revert, rebase, and then merge for you?

ndukler:
Yeah, that should do it. Sorry for the inconvenience :(

ndukler:
Can we close this down @andrewkern

PR #233, opened 2019-12-03
QC for huber two epoch model

ndukler:


ndukler:
Ready to merge @andrewkern

ndukler:
Closes #226

PR #245, opened 2019-12-05
Small bibtex changes

Chris1221:
A few small things here:

- Fixed a typo in the Tennessen doi
- Writes genetic map citations to bibtex with a test
- Changes name of bibtex option to `bibtex-file` and adds the short `b` option

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -237,9 +237,12 @@ def write_bibtex(engine, model, contig, bibtex_file):
     Write bibtex for available citations to a file."""
     for citation in engine.citations:
         bibtex_file.write(citation.fetch_bibtex())
-
+    if contig.genetic_map is not None:
+        for citation in contig.genetic_map.citations:
+            bibtex_file.write(citation.fetch_bibtex())
     for citation in model.citations:
         bibtex_file.write(citation.fetch_bibtex())
+    bibtex_file.close()
 
 
 def write_citations(engine, model, contig):
@@ -304,7 +307,7 @@ def add_simulate_species_parser(parser, species):
             "is provided as an argument show help for this model; otherwise "
             "show help for all available models"))
     species_parser.add_argument(
-        "--bibtex_file",
+        "-b", "--bibtex-file",
         type=argparse.FileType('w'),
         help="Write citations to a given bib file. This will overwrite the file.",
         default=None,
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -171,7 +171,7 @@ class TestHomoSapiensArgumentParser(unittest.TestCase):
         bib = "tmp.bib"
 
         with mock.patch.object(argparse.FileType, '__call__') as call:
-            args = parser.parse_args([cmd, "--bibtex_file", bib, "-o", output, "2"])
+            args = parser.parse_args([cmd, "-b", bib, "-o", output, "2"])
             self.assertEqual(args.output, output)
             self.assertEqual(args.samples, [2])
             call.assert_called_with(bib)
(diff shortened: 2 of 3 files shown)


Chris1221:
Whoops sorry, one moment, I overly mocked something.

stdpopsim/cli.py
```diff
@@ -237,9 +237,12 @@ def write_bibtex(engine, model, contig, bibtex_file):
     Write bibtex for available citations to a file."""
     for citation in engine.citations:
         bibtex_file.write(citation.fetch_bibtex())
-
-    for citation in model.citations:
-        bibtex_file.write(citation.fetch_bibtex())
+    if contig.genetic_map is not None:
+        for citation in contig.genetic_map.citations:
+            bibtex_file.write(citation.fetch_bibtex())
+    if len(model.citations) > 0:
```
jeromekelleher: This check is redundant.

stdpopsim/cli.py
```diff
@@ -237,9 +237,12 @@ def write_bibtex(engine, model, contig, bibtex_file):
...
+        for citation in model.citations:
+            bibtex_file.write(citation.fetch_bibtex())
```
jeromekelleher: I forgot to check this --- are we closing the bibtex file after we finish writing to it? We could just put a ``bibtex_file.close()`` at the end of the method.

tests/test_cli.py
```diff
@@ -463,6 +463,25 @@ def test_whole_bibex(self):
...
+        # Test that genetic map citations are converted.
+        species = stdpopsim.get_species("homsap")
+        genetic_map = species.get_genetic_map("HapmapII_GRCh37")
+        contig = species.get_contig("chr22", genetic_map=genetic_map.name)
+        model = stdpopsim.PiecewiseConstantSize(species.population_size)
+        engine = stdpopsim.get_default_engine()
+
+        with mock.patch("builtins.open", mock.mock_open()):
+            with open('tmp.bib', 'w') as bib:
+                with mock.patch.object(
```
jeromekelleher: Why do we need two levels of patches here? Doesn't the first one catch them all?

Chris1221: My thought was that I wanted to check to ensure that specifically the genetic map `fetch_bibtex` was being called, not just the general one (which we're already testing). But since I don't think its possible to check if a specific instantiation of the class, I separately patched out the genetic map `fetch_bibtex` and ensured that it gets called (while the previous ones still need to be patched to avoid actually downloading anything). We could alternatively just patch out everything and ensure that it gets called the correct number of times, but I thought this was more explicit. Up to you!

stdpopsim/cli.py
Chris1221: Right, sorry I had forgotten about this. I'll close it.

tests/test_cli.py
jeromekelleher: Ah, I see. I think asserting it gets called the expected number of times would be good, or else add in a comment describing the strategy. These mocking tests can be  a bit opaque at times.

Chris1221:
So I removed the redundant check, closed the connection, and changed the test to be more transparent (removed the second mock, and simply asserted that `fetch_bibtex` was being called the right number of times). Rebased and squashed.

tests/test_cli.py
```diff
@@ -463,6 +463,27 @@ def test_whole_bibex(self):
...
+        # Patch out writing to a file, then
+        # ensure that the method is called
+        # the correct number of times.
+        with mock.patch("builtins.open", mock.mock_open()):
+            with open('tmp.bib', 'w') as bib:
+                with mock.patch.object(
+                        stdpopsim.citations.Citation,
+                        "fetch_bibtex") as mock_bib:
+                    cli.write_bibtex(engine, model, contig, bib)
+                    assert mock_bib.call_count == ncite
```
jeromekelleher: This should be ``self.assertEqual()`` as we're using nose rather than pytest.

jeromekelleher:
Looks great, thanks @Chris1221. One minor issue above.

Chris1221:
Whoops, my bad @jeromekelleher. Should be fixed now.

Issue #243, opened 2019-12-05
semantics of Species attributes and their use in (non)specific models


grahamgower:
Currently, the `population_size_citations` and `generation_time_citations` attributes of a given species are not being written with the other citations. Ideally they should be, but this raises the question of when these citations are relevant to a given simulation.

Consider two homsap models. `ooa_3` uses the species defined generation_time of 25. Whereas `ooa_archaic` defines generation_time to be 29. In the former case, the species generation_time_citations are relevant, in the latter they are not (ignoring of course that generation_time_citations is not yet defined for homsap). Ideally, a specific model implementation would copy the species `generation_time_citations` into their own citations field. And analogously for the `population_size_citations`, as relevant.

It might be useful to have some explicit mechanism for marrying species attributes with a given model (specific, or not), which doesn't rely on class inheritance in the specific case.

Regarding non-specific models, `PiecewiseConstantSize` is used by the CLI when no model is specified, and the `population_size_citations` are correctly copied into the model citations in this case. Probably the `generation_time_citations` are also relevant, but are not currently copied.

jeromekelleher:
Very good points @grahamgower. Ideally these things should be declared explicitly when defining the models, so we have attributes inside in the models that must be filled. (If not, CI fails when you try to add the model). So, a model author can either choose to use the default species generation time (and copy the citations) or they declare their own and must put in a citation for it.

It's a bit less clear in terms of the population size, since that's never really used in 'real' models, just the generic ones. So, perhaps we could manually copy the species' ``population_size_citations`` into the model's ``citations`` list? I'm not overly worried about making this work transparently in the Python API (that seems quite hard without some elaborate multiple inheritance model, which would be quite tricky and probably opaque), it's just something we need to make work for the CLI.

What do you think?

grahamgower:
Yeah, its probably best to enforce it at model implementation time, for the specific models. The canonical way its done should also enforce coping both `param, param_citations` simultaneously (as a tuple, or its own attrib class?), where param is e.g. generation_time.

jeromekelleher:
I think we can probably check that these are filled in as part of the automatic CI checks that should be performed when the model is added, so I don't think we need to do anything fancy. We just check to see that ``generation_time is not None`` and ``generation_time_citations is not None`` as one of the checks in the [ModelTestMixin](https://github.com/popgensims/stdpopsim/blob/master/tests/test_models.py#L15). That should do it, right?

grahamgower:
[...]

PR #246, opened 2019-12-05
Require Model.generation_time, and related cleanups.

grahamgower:
Species specific subclasses of stdpopsim.Model have been removed. Addresses #243.

### DIFF
--- a/stdpopsim/catalog/arabidopsis_thaliana.py
+++ b/stdpopsim/catalog/arabidopsis_thaliana.py
@@ -80,15 +80,8 @@ _species.add_genetic_map(_gm)
 ###########################################################
 
 
-class ArabidopsisThalianaModel(stdpopsim.Model):
-    """
-    TODO: documentation
-    """
-    species = _species
-
-
 # FIXME this documentation needs to be filled out.
-class _Durvasula2017MSMC(ArabidopsisThalianaModel):
+class _Durvasula2017MSMC(stdpopsim.Model):
     id = "SMA_1pop"
     name = "South Middle Atlas population size history"
     description = """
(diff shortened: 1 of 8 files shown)


stdpopsim/catalog/arabidopsis_thaliana.py
```diff
@@ -80,21 +80,15 @@
...
 # FIXME this documentation needs to be filled out.
-class _Durvasula2017MSMC(ArabidopsisThalianaModel):
+class _Durvasula2017MSMC(stdpopsim.Model):
     id = "SMA_1pop"
     name = "South Middle Atlas population size history"
     description = """
         Model estimated from two homozygous individuals from the South Middle Atlas
         using MSMC (TODO: more detail).
     """
+    species = _species
```
jeromekelleher: Are we actually using the species attribute in client code? If not, I think we should drop it. We can use the ``_species`` variable to get access to the generation times, etc.

stdpopsim/cli.py
```diff
@@ -381,7 +381,9 @@ def add_simulate_species_parser(parser, species):
     def run_simulation(args):
         if args.model is None:
             model = stdpopsim.PiecewiseConstantSize(species.population_size)
-            model.citations = species.population_size_citations
+            model.generation_time = species.generation_time
+            model.citations = species.population_size_citations + \
```
jeromekelleher: Safer to ``extend`` this list here. Perhaps we might have a citation in there all ready for the generic model.

jeromekelleher:
Ah, excellent, thanks @grahamgower, looks like species is extraneous. OK, one more rebase and lets merge (I don't want to do it myself and force push to your master, seems like bad etiquette; probably not a great idea to use master as your feature branch though)

grahamgower:
Ok, rebased again.

Issue #241, opened 2019-12-04
Output ACGT mutations instead of 0/1


jeromekelleher:
Msprime defaults to generating infinite sites with the classical 0/1 haplotype values (making it easy to see which is the ancestral and derived states). However, it also supports generating infinite sites mutations from the ACGT alphabet. The tree sequence still has complete information about which state is ancestral.

I'm guessing most users would prefer/expect ACGT mutations?

apragsdale:
Do you envision having a catalog of mutation models here, or just randomly assigning mutations from ACGT (which I guess is a very simple mutation model)? That could be quite useful, but possibly a lot of work to compile various mutation models.

jeromekelleher:
Ultimately we'd like to have a catalog of mutation models as well, yeah. These'll need to be implemented first though. For now, we just randomly assign an ancestral state from ACGT and a derived state from the remainder.

andrewkern:
having the nucleotide output as an option would be good. yeah.

jeromekelleher:
> having the nucleotide output as an option would be good. yeah.

Option or default though? I don't see much reason for having the old-school 0/1 mutations other than that's the way it's always been done.

dschride:
I would vote for 'option', since some users might want to plug their output directly into summary stat calculations, and many tools for this expect 0/1.

jeromekelleher:
OK, cool. Any ideas on the interface? ``--mutation-model/-u [model_id]`` maybe? Defaults to ``infinite_sites_nucleotides`` and also available is ``infinite_sites_01``?

dschride:
Yeah I think that would be fine.

ndukler:
Can we separate the mutational model and the output form? That way any model can have 0/1 or ACGT based output.

jeromekelleher:
I think 0/1 really only makes sense in the most naive infinite sites case. Any other model depends on the actual nucleotide values, so wouldn't really make sense. As I say, there's no loss of information here: the ancestral/derived states are stored in the tree sequence and it's trivial to get 0/1 numeric values out using the genotype_matrix method. We're not offering any direct textual output anyway, and conversion to VCF will have the REF allele set to the ancestral state. 

So, my take would be to just offer ``infinite_sites_nucleotides`` as the default (and only option) for the initial release, and we can add the 0/1 model later if someone actually asks for it.

How does this sync with how SLiM does things @grahamgower?

grahamgower:
[...]

PR #247, opened 2019-12-05
Change mutations to ACGT

jeromekelleher:
For discussion; see #241 for more details.

This is tricky, because it cuts across our nice clean "Engine" interface. Presumably, we'll want to put mutations down on tree sequences output by SLiM as well, and therefore we probably want to restructure the API a bit for dealing with mutations in a forward compatible way. The question is, do we want to do this before the first release?

Also, do we want to default to 0/1 mutations or ACGT? @petrelharp, @andrewkern, what's your take on this?

### DIFF
--- a/stdpopsim/engines.py
+++ b/stdpopsim/engines.py
@@ -104,14 +104,17 @@ class _MsprimeEngine(Engine):
 
     def simulate(self, model=None, contig=None, samples=None, seed=None,
                  **kwargs):
-        return msprime.simulate(
+        ts = msprime.simulate(
                 samples=samples,
                 recombination_map=contig.recombination_map,
-                mutation_rate=contig.mutation_rate,
                 population_configurations=model.population_configurations,
                 migration_matrix=model.migration_matrix,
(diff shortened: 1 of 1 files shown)


andrewkern:
my take is we default to 0/1 but put in the option for ACGT as the engine allows

grahamgower:
I imagine we'll have a whole slew of things like this: e.g. gene conversion, selective sweeps, DFE, sequencing errors. Mostly things that apply independently of the demographic model, and are readily implemented using specific features of one or more simulation engines (or are a postprocessing step on tree sequences, that is independent of the simulation engine). So we probably need a catalog for these `Feature`s, and a simple way to parameterise them. `Engine.simulate()` could take an extra `features` param, which is a list of `Feature`s. (And feel free to suggest a better name than 'feature').

jeromekelleher:
Good ideas @grahamgower, but I think we need to shelve this for now in the interest of getting v 0.1 out the door. Since basically everyone seems to be happy with 0/1 mutations as the default, we can just leave things as they are for now.

PR #250, opened 2019-12-05
Change licence to GPLv3.

### LINKED ISSUE #37: License is actually GPL

jeromekelleher:
Closes #37

See #37 for discussion --- we basically have no choice here as msprime is GPLv3.


jeromekelleher:
Unless someone shouts I'll assume this is OK with everyone.

andrewkern:
looks like it's fine with everyone. once you rebase let's merge

Issue #199, opened 2019-11-26
Add --dry-run option to CLI


ndukler:
Add flag to allow users to see exactly what they're going to run before they run it. Output includes the species and model names, the numbers of samples from each population, and the sampling times. This will probably be done by adding a `dry_run` parameter to the `simulate()` function in each engine and update the abstract class appropriately.

jeromekelleher:
I think we can do this without involving the Engine --- we just write out a report about the ``engine``, ``contig``, ``model`` and ``samples`` objects that we would have passed to the simulate method.

OK if I assign this one to you?

ndukler:
Sure. I think that having the dry run being a property of the engine might be better in the long run though if the way different engines produce the dry run output differs. I suppose I sort of think how the dry run goes is a property of the engine. What do you think?

jeromekelleher:
Well, I guess it might need to become a property of the engine at some point, but I think it's a good idea to keep all the interface code within the ``cli`` module. Let's just do it the easy way for now, and avoid the complications of formatting output text from outside the CLI module.

ndukler:
Do we want the python API to have the ability to do a dry run as well?

jeromekelleher:
No --- this is just for the CLI.

ndukler:
Ok, makes more sense to keep it in the `cli` module then. I'll get on it

jeromekelleher:
If the API is well designed, a dry-run text output will be no more informative than the code for creating the simulation. Reading the code should be like a dry run.

ndukler:
I think once we have the key-value pairs for the sampling set up that will definitely be true, so I'll throw up an issue for that as well. Feel free to assign it to me :)

jeromekelleher:
Thinking about it this morning, perhaps it would be good to write this report out by default before running the simulation? We're already writing out the citation information, so perhaps a concise report of what the simulation did would be a good idea as well. We can turn this off using the ``--quiet`` option.

The ``--dry-run`` option then would do everything except run the simulation. What do you think?

ndukler:
Sounds good to me!

jeromekelleher:
Any updates on this @ndukler? I'd like to get this feature in for 0.1.0, if possible.

ndukler:
Sorry, I've been busy with some of my other projects. I'll try to get it done today.

PR #268, opened 2019-12-06
Added sphinxcontrib-programoutput directives in tutorials

### LINKED ISSUE #266: Docs: investigate using sphinxcontrib-programoutput for docs
There is a sphinx plugin for including program output into sphinx docs. We could use this for thinks like the tutorial for including the output of ``stdpopsim --help``.  It has a reasonable chance of working I think, and would be pretty handy.

Anyone want to try it out?

https://pythonhosted.org/sphinxcontrib-programoutput/

gtsambos:
Closes #266 

This works locally, but to get it running on RTD I suspect you'll also need to tell it that we are using the `sphinxcontrib.programoutput` extension, and to make `stdpopsim` available to it so that the commands can be compiled remotely.

A few thoughts I had while doing this:

 - This will contribute to massive bloat if we use it indiscriminately. I propose that *only help output* should be displayed in this way, and this should be truncated to 20 lines unless abs necessary.

I don't think it's a good idea to be running `stdpopsim` simulations themselves within the program-output blocks -- some of the ones in the existing tutes take some time to run, and I'm not sure what RTD would do with the output files that these commands are supposed to create.

 - Presumably, we wanted to do this to save us from having to continually edit docs as we also edit the help output (#215). It's not totally clear to me that RTD is going to just automatically update these docs, though -- I'm guessing it would need to be (manually) pointed to an up-to-date stdpopsim installation for this to happen? 

More generally, I don't know much about how ``stdpopsim`` will work on whatever RTD uses to execute these commands...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,4 +56,5 @@ jobs:
       - run:
           name: Build docs
           command: |
+            source venv/bin/activate
             cd docs && make && make singlehtml
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -66,6 +66,7 @@ extensions = [
     'sphinx.ext.viewcode',
     'speciescatalog',
     'sphinxarg.ext',
+    'sphinxcontrib.programoutput'
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -48,7 +48,7 @@ help+wanted%22>`_
 To get started helping with ``stdpopsim`` development, please read the
 following sections to learn how to contribute.
 
-
+.. sec_development_installation:
 
 ************
 Installation
@@ -63,7 +63,9 @@ For ``pip`` users, install the packages required for development using::
 
     $ python3 -m pip install -r requirements/development.txt
 
+You can then install the development version of ``stdpopsim`` like this::
 
+    $ python3 setup.py install
 
 For ``conda`` users, you will need to add the conda-forge channel to your conda
 environment and then should be able to install the development requirements using::
@@ -569,4 +571,6 @@ It is defined in the ``docs`` directory.
 To build the documentation type ``make`` in the ``docs`` directory. This should build
 HTML output in the ``_build/html/`` directory.
 
+.. note::
 
+    You will need ``stdpopsim`` to be installed for the build to work.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -28,20 +28,17 @@ by ``stdpopsim``. The first step for using the CLI is to select the species that
 you are interested in simulating data for. In order to see which species are
 available run
 
-.. code-block:: console
-
-    $ stdpopsim --help
+.. command-output:: stdpopsim --help
 
 This shows the species currently supported by ``stdpopsim``. This means that
 ``stdpopsim`` knows various traits of these species including chromosome size
 and recombination rates. Once we've selected a species, in this case humans, we
 can look at the help again as follows.
 
-.. code-block:: console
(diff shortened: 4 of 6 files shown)


gtsambos:
Btw, I think the CI has failed just because it doesn't know about sphinxcontrib-programoutput

agladstein:
I agree with you.

You should also add `sphinxcontrib-programoutput` to the appropriate requirements file (requirements/development.txt or requirements/rtd.txt?)

jeromekelleher:
Excellent, thanks @gtsambos! @agladstein is right, we just need to add the package to requirements/development.txt and this should all work.

Totally agree on using this sparingly in the docs --- what you've done looks ideal.

Re running on RTD, I think this much should work fine (it clones the git repo, so it has the full code). If we want to run actual simulations, we'll need to rejig the RTD setup a little and install msprime using conda. (Now, we're mocking out msprime so that we don't need to install it; it won't install because there's no GSL available on RTD.) There might be some teething issues, but I'd say let's merge this once CI passes and I'll fix up the RTD setup then.

When we get the ``--dry-run`` option working (#199) we can also show the citations output.

gtsambos:
Thanks @agladstein  and @jeromekelleher! the CI is giving this error now --
```
Warning, treated as error:
/home/circleci/stdpopsim/docs/tutorial.rst:31:Command 'stdpopsim --help' failed: [Errno 2] No such file or directory: 'stdpopsim': 'stdpopsim'
Makefile:12: recipe for target 'html' failed
```

jeromekelleher:
This is because the ``stdpopsim`` binary isn't in your path. I see two ways to fix this:

1. We use ``python3 -m stdpopsim`` in the tutorial. This has the advantage of being a bit more robust to people's weird setups, and means that we don't have to tell them about setting their PATH. It's a bit less pleasing to the eye though.
2. We build the docs in a virtualenv where stdpopsim has been installed. This is easy enough to do, but makes things a little bit harder for developers when they are building the docs.

I guess (2) is the best for now, since it's only for CI purposes and we can probably assume that developers will have stdpopsim in their PATH. You need to edit ``.circleci/config.yml`` and insert a line ``source venv/bin/activate" immediately before the build docs line at the bottom.

gtsambos:
Thanks @jeromekelleher - I think it's all good now 😅 

> We build the docs in a virtualenv where stdpopsim has been installed. This is easy enough to do, but makes things a little bit harder for developers when they are building the docs.

I don't think this should be a big issue for developers, since they can still compile the files locally without having a virtual environment as long as stdpopsim is also installed locally (which I have done in this case, and yeah, I imagine most of us are doing this). I might add a note or two to the docs to mention this.

jeromekelleher:
Thanks @gtsambos, merging.

jeromekelleher:
This is all working on RTD now @gtsambos. I ended up having to install msprime using conda to get it working, so we should even be able to run (small) simulations on rtd now!

gtsambos:
great, thanks @jeromekelleher, sounds like this was more work than anticipated...

jeromekelleher:
> great, thanks @jeromekelleher, sounds like this was more work than anticipated...

It did, a bit, but it's worth it!

agladstein:
@gtsambos can make an issue about adding dev install instructions you are talking about?
There is already a in-depth section about it on RTD. So, it'd fit in fine there. 
https://stdpopsim.readthedocs.io/en/latest/development.html#installation

gtsambos:
@agladstein I did it as part of this pull request already? And I added a line about `python setup.py install` so that they know to do that

agladstein:
Ah okay. Sorry, I didn't notice that. You got it under control!

PR #270, opened 2019-12-07
Added Huber three epoch model

noscode:
I have added three epoch model from Huber et al. 2018.

Issue #122, opened 2019-10-03
ID scheme for population models


jeromekelleher:
After #110 was merged we have a ID scheme for population models. The idea is to have a unique (within species), human-memorable string that is used as the shorthand way of finding models in the CLI and from ``species.get_model()`` (we can always add other ways of querying based on other properties later).

This is leading on from #85.

What we have at the moment is:
```
homsap
    ooa_3
    ooa_2
    african   
    america
    ooa_archaic
    zigzag
dromel
    afr_3epoch
    ooa_2
aratha
    fixme
```

These are not consistent, well thought through, or anything like that; it's just a starting point.

The catalog and online CLI help should be enough to explain what these mean, so I think it's OK to have reasonably terse IDs.

Any ideas on how to do this well would be great!

andrewkern:
each of the population models will be nested within an organism (with the exception of the Generics), so perhaps all we need is information about the number of populations and some ID for the model itself. Using first author names might work for the latter. e.g.

```
homsap
    3_Gut
    2_Fu
    1_Ten   
    4_Bro
    5_Rag
   ...
```

jeromekelleher:
It's a good idea @andrewkern, but it doesn't give any hint about **what** the model is. I don't find the names of authors very helpful here in making the models IDs memorable. At least OOA is 'out-of-Africa'.

@dschride, @reedacartwright, you've had good ideas on this type of thing before? How do we come up with short, unique, memorable tags for population models?

dschride:
I still like the idea of a prefix like "2pop_" or "1pop_", and agree that this should be followed by a name that is descriptive of the model rather than the author name. Author name could be a suffix, but then things are ballooning a bit.

reedacartwright:
I think model names sound be treated like variable names: make them as long as necessary to make it clear what they represent.

human:out_of_africa_3pop
human:out_of_africa_2pop
human:africa
human:america

If we want brevity, we can index the models like BLAST indexes genetic codes.

human:1 # Human model 1 is the out_of_africa_3pop model erc
human:2
human:3

dschride:
> make them as long as necessary to make it clear what they represent

Sounds right to me.

petrelharp:
Proposal (from the call): 

 - Request that people make up a 6-9 letter descriptive string, and rely on the help and docs for more info.
 - Perhaps we can do long and short versions? Longer than 6-9 letters would be quite long.
 - desire to uniformize these

andrewkern:
[...]

PR #275, opened 2019-12-08
Turn subclasses of stdpopsim.Model into functions.

### LINKED ISSUE #269: Model subclasses are just functions
`stdpopsim/catalog/homsap.py` has the following comment regarding Model subclasses:
```
# TODO we want to move away from defining these as classes which
# can be instantiated and rather to creating *instances* of the Model
# class which has this behaviour. However, it's not clear what form
# the refactored versions would have. Marking these classes with a
# __ to emphasise that they're not supposed to be used externally
# like this, but should be found via the species catalog.
```
The Model subclasses just define a few attributes and an `__init__` function. So they're not really classes. Using `PiecewiseConstantSize` and `GenericIM` as examples, here's what the refactored versions could look like (untested!):
```
diff --git a/stdpopsim/models.py b/stdpopsim/models.py
index 1dff2df..9489624 100644
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -3,6 +3,7 @@ Common infrastructure for specifying demographic models.
 """
 import sys
 
+import attr
 import msprime
 import numpy as np
 
@@ -141,6 +142,7 @@ class Population(object):
                 "sampling_time": self.sampling_time}
 
 
+@attr.s
 class Model(object):
     """
     Class representing a simulation model that can be run to output a tree sequence.
@@ -155,19 +157,16 @@ class Model(object):
         written text, e.g., "Three population Out-of-Africa"
     :vartype informal_name: str
     """
-    # TODO the infrastructure here is left over from a structure that
-    # rigidly used class definitions as a way to define population
-    # models. This contructor should take all the instance variables
-    # as parameteters, and we should use factory functions to define
-    # the model instances that are added to the catalog rather than
-    # subclasses.
-
-    def __init__(self):
-        self.population_configurations = []
-        self.demographic_events = []
-        # Defaults to a single population
-        self.migration_matrix = [[0]]
-        self.generation_time = None
+
+    id = attr.ib(type=str, kw_only=True)
+    name = attr.ib(type=str, kw_only=True)
+    description = attr.ib(type=str, kw_only=True)
+    generation_time = attr.ib(type=int, kw_only=True)
+    populations = attr.ib(type=list, kw_only=True)
+    population_configurations = attr.ib(default=[], kw_only=True)
+    demographic_events = attr.ib(default=[], kw_only=True)
+    migration_matrix = attr.ib(default=[[0]], kw_only=True)
+    citations = attr.ib(default=[], kw_only=True)
 
     @property
     def num_populations(self):
@@ -249,10 +248,9 @@ _popAnc = Population(name="popAnc", description="Generic ancestral population",
                      sampling_time=None)
 
 
-class PiecewiseConstantSize(Model):
+def PiecewiseConstantSize(N0, *args):
     """
-    Class representing a generic simulation model that can be run to output a
-    tree sequence. This is a piecewise constant size model, which allows for
+    Returns a piecewise constant size model, which allows for
     instantaneous population size change over multiple epochs in a single population.
 
     :ivar N0: The initial effective population size
@@ -267,32 +265,29 @@ class PiecewiseConstantSize(Model):
         model1 = stdpopsim.PiecewiseConstantSize(N0, (t1, N1)) # One change
         model2 = stdpopsim.PiecewiseConstantSize(N0, (t1, N1), (t2, N2)) # Two changes
     """
-
-    id = "constant"
-    name = "Piecewise constant size"
-    description = "Piecewise constant size population model over multiple epochs."
-    citations = []
     populations = [_pop0]
-    author = None
-    year = None
-    doi = None
-
-    def __init__(self, N0, *args):
-        self.population_configurations = [
-            msprime.PopulationConfiguration(
-                initial_size=N0, metadata=self.populations[0].asdict())
-        ]
-        self.migration_matrix = [[0]]
-        self.demographic_events = []
-        for t, N in args:
-            self.demographic_events.append(msprime.PopulationParametersChange(
-                time=t, initial_size=N, growth_rate=0, population_id=0))
-
-
-class GenericIM(Model):
+    population_configurations = [
+        msprime.PopulationConfiguration(
+            initial_size=N0, metadata=populations[0].asdict())
+    ]
+    demographic_events = []
+    for t, N in args:
+        demographic_events.append(msprime.PopulationParametersChange(
+            time=t, initial_size=N, growth_rate=0, population_id=0))
+
+    return Model(
+            id="constant",
+            name="Piecewise constant size",
+            description="Piecewise constant size population model over "
+                        "multiple epochs.",
+            populations=populations,
+            population_configurations=population_configurations,
+            demographic_events=demographic_events)
+
+
+def GenericIM(NA, N1, N2, T, M12, M21):
     """
-    Class representing a generic simulation model that can be run to output a tree
-    sequence. A generic isolation with migration model where a single ancestral
+    Returns a generic isolation with migration model where a single ancestral
     population of size NA splits into two populations of constant size N1
     and N2 time T generations ago, with migration rates M12 and M21 between
     the split populations. Sampling is disallowed in population index 0,
@@ -319,33 +314,28 @@ class GenericIM(Model):
         model1 = stdpopsim.GenericIM(NA, N1, N2, T, M12, M21)
 
     """
-    id = "IM"
-    name = "Isolation with migration"
-    description = """
-        A generic isolation with migration model where a single ancestral
-        population of size NA splits into two populations of constant size N1
-        and N2 time T generations ago, with migration rates M12 and M21 between
-        the split populations.
-        """
-    citations = []
     populations = [_pop0, _pop1, _popAnc]
-    author = None
-    year = None
-    doi = None
-
-    def __init__(self, NA, N1, N2, T, M12, M21):
-        self.population_configurations = [
-            msprime.PopulationConfiguration(
-                initial_size=N1, metadata=self.populations[0].asdict()),
-            msprime.PopulationConfiguration(
-                initial_size=N2, metadata=self.populations[1].asdict()),
-            msprime.PopulationConfiguration(
-                initial_size=NA, metadata=self.populations[2].asdict())
-        ]
-        self.migration_matrix = [[0, M12, 0], [M21, 0, 0], [0, 0, 0]]
-        self.demographic_events = [
-            msprime.MassMigration(
-                time=T, source=0, destination=2, proportion=1),
-            msprime.MassMigration(
-                time=T, source=1, destination=2, proportion=1)
-        ]
+    return Model(
+            id="IM",
+            name="Isolation with migration",
+            description="""
+                A generic isolation with migration model where a single ancestral
+                population of size NA splits into two populations of constant size N1
+                and N2 time T generations ago, with migration rates M12 and M21 between
+                the split populations.
+                """,
+            population_configurations=[
+                msprime.PopulationConfiguration(
+                    initial_size=N1, metadata=populations[0].asdict()),
+                msprime.PopulationConfiguration(
+                    initial_size=N2, metadata=populations[1].asdict()),
+                msprime.PopulationConfiguration(
+                    initial_size=NA, metadata=populations[2].asdict())
+            ],
+            migration_matrix=[[0, M12, 0], [M21, 0, 0], [0, 0, 0]],
+            demographic_events=[
+                msprime.MassMigration(
+                    time=T, source=0, destination=2, proportion=1),
+                msprime.MassMigration(
+                    time=T, source=1, destination=2, proportion=1)
+            ])
```

Obviously, all existing model definitions would need changing, plus documentation, ...

grahamgower:
Closes #269.

The Model class is nicer, but the rest is just code churn. There are a couple of things in here that aren't intended to be merged. E.g. the Final metaclass was just to find where the subclasses were hiding. Also, there are two files with dos end of line characters that got converted to unix end of line characters (`qc/arabidopsis_thaliana_qc.py` and `qc/drosophlia_melanogaster_qc.py`), which makes the diffs harder to read.

### DIFF
--- a/qc/arabidopsis_thaliana_qc.py
+++ b/qc/arabidopsis_thaliana_qc.py
@@ -5,7 +5,6 @@ import stdpopsim.models as models
 
 class Durvasula2017MSMC(models.Model):
     def __init__(self):
-        super().__init__()
 
         # Both of the following are directly
         # converted from MSMC output scaled by A.Thaliana
@@ -68,7 +67,6 @@ class HuberTwoEpoch(models.Model):
     ]
 
     def __init__(self):
-        super().__init__()
 
         # Time of second epoch
         T_2 = 568344
--- a/stdpopsim/catalog/arabidopsis_thaliana.py
+++ b/stdpopsim/catalog/arabidopsis_thaliana.py
@@ -85,139 +85,138 @@ _species.add_genetic_map(_gm)
 
 
 # FIXME this documentation needs to be filled out.
-class _Durvasula2017MSMC(stdpopsim.Model):
-    id = "SMA_1pop"
-    name = "South Middle Atlas population size history"
-    description = """
-        Model estimated from two homozygous individuals from the South Middle Atlas
-        using MSMC (TODO: more detail).
-    """
-    populations = [
-        stdpopsim.Population(
-            name="a_thaliana", description="""
-            Arabidopsis Thaliana South Middle Atlas population
-        """)
-    ]
-    citations = [stdpopsim.Citation(
-        author="Durvasula et al.",
-        year=2017,
-        doi="https://doi.org/10.1073/pnas.1616736114")
-    ]
+def _sma_1pop():
+    # the size during the interval times[k] to times[k+1] = sizes[k]
+    times = np.array([
+        699, 2796, 6068, 9894, 14370, 19606, 25730, 32894, 41275,
+        51077, 62544, 75958, 91648, 110001, 131471, 156584, 185960, 220324,
+        260520, 307540, 362541, 426879, 502139, 590173, 693151, 813610,
+        954517, 1119341, 1312147, 1537686, 1801500, 2110100])
+    sizes = np.array([
+        42252426, 42252426, 60323, 72174, 40591, 21158, 21442,
+        39942, 78908, 111132, 110745, 96283, 87661, 83932, 83829, 91813,
+        111644, 143456, 181571, 217331, 241400, 246984, 238593, 228222,
+        217752, 198019, 165210, 121796, 121796, 73989, 73989, 73989])
 
-    def __init__(self):
-        # the size during the interval times[k] to times[k+1] = sizes[k]
-        self.times = np.array([
-            699, 2796, 6068, 9894, 14370, 19606, 25730, 32894, 41275,
-            51077, 62544, 75958, 91648, 110001, 131471, 156584, 185960, 220324,
-            260520, 307540, 362541, 426879, 502139, 590173, 693151, 813610,
-            954517, 1119341, 1312147, 1537686, 1801500, 2110100])
-        self.sizes = np.array([
-            42252426, 42252426, 60323, 72174, 40591, 21158, 21442,
-            39942, 78908, 111132, 110745, 96283, 87661, 83932, 83829, 91813,
-            111644, 143456, 181571, 217331, 241400, 246984, 238593, 228222,
-            217752, 198019, 165210, 121796, 121796, 73989, 73989, 73989])
+    # MSMC is accurate from 40Kya-1.6Mya for A.thaliana (Durvasula et al 2017)
+    # set the first 7 sizes
+    # equal to the size at 8 (~40Kya)
+    sizes[:8] = sizes[8]
+    # set the last 2 entries equal
+    # to the size at 30 (~1.6Mya)
+    sizes[30:32] = sizes[30]
 
-        # MSMC is accurate from 40Kya-1.6Mya for A.thaliana(Durvasula et al 2017)
-        # set the first 7 sizes
-        # equal to the size at 8 (~40Kya)
-        self.sizes[:8] = self.sizes[8]
-        # set the last 2 entries equal
-        # to the size at 30 (~1.6Mya)
-        self.sizes[30:32] = self.sizes[30]
-        # generation time is 1 year
-        self.generation_time = 1
-        self.demographic_events = []
-        for idx, t in enumerate(self.times):
-            self.demographic_events.append(
-                msprime.PopulationParametersChange(
-                    time=t, initial_size=self.sizes[idx], population_id=0))
-
-        self.migration_matrix = [[0]]
-
-        self.population_configurations = [
-           msprime.PopulationConfiguration(initial_size=self.sizes[0],
-                                           metadata=self.populations[0].asdict())
-        ]
-
-
-_species.add_model(_Durvasula2017MSMC())
-
-
-class _HuberTwoEpoch(stdpopsim.Model):
-    id = "Afr_2epoch"
-    name = "African two epoch model"
-    description = """
-        Model estimated from site frequency spectrum of synonymous
-        SNPs from African samples using Williamson et al. (2005)
-        methodology.
-    """
-    populations = [
-        stdpopsim.Population(
-            name="a_thaliana", description="Arabidopsis thaliana African population")
-    ]
-    citations = [stdpopsim.Citation(
-        author="Huber et al.",
-        year=2018,
-        doi="https://doi.org/10.1038/s41467-018-05281-7")
-    ]
-
-    def __init__(self):
-        N_A = 746148
-        N_0 = 100218
-        t_1 = 568344
-        self.generation_time = 1
-        self.demographic_events = []
-        self.population_configurations = [
-            msprime.PopulationConfiguration(initial_size=N_0,
-                                            metadata=self.populations[0].asdict())]
-        self.demographic_events = [
+    demographic_events = []
+    for sz, t in zip(sizes, times):
+        demographic_events.append(
             msprime.PopulationParametersChange(
-                time=t_1, initial_size=N_A, population_id=0)]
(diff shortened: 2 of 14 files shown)


jeromekelleher:
Thanks a million for all this work @grahamgower!

Part of me wants to put this off until after the 0.1 release, as it's a big diff and touches a lot of code. Most of it is pretty simple changes though, so perhaps it's better to grasp this nettle now.

There's a few things I'd like to change though:

1. I think we should keep the Generic models as they are: a function that looks like a class that returns an instance of a class, is really a class (If you see what I mean!). The difference between the generics and the models in the catalog is that they are intended to be instantiated directly, whereas catalog models are retrieved from the catalog.

2. Functions should be ``camel_cased``. It's confusing to have functions that looks like classes, and we don't want to confuse incoming model contributors. 

What do you think?

qc/arabidopsis_thaliana_qc.py
```diff
@@ -1,92 +1,83 @@
-import msprime
-import numpy as np
-import stdpopsim.models as models
-
-
-class Durvasula2017MSMC(models.Model):
```
jeromekelleher: To avoid changing too much code at once, we could leave the QC module as it is and refactor it as a follow up. This would give us some confidence that we're not making any model-breaking changes in the diff.

stdpopsim/catalog/arabidopsis_thaliana.py
```diff
@@ -81,95 +81,96 @@
...
 
 # FIXME this documentation needs to be filled out.
-class _Durvasula2017MSMC(stdpopsim.Model):
-    id = "SMA_1pop"
-    name = "South Middle Atlas population size history"
-    description = """
-        Model estimated from two homozygous individuals from the South Middle Atlas
-        using MSMC (TODO: more detail).
-    """
+def _Durvasula2017MSMC():
```
jeromekelleher: See earlier point about function names. Maybe change this to "_sma_1pop" (i.e., the ID of the model with a leading underscore)?

grahamgower:
I'm happy to pull the  QC changes out and do that as a separate commit. It certainly makes sense to satisfy ourselves that these (invasive) changes aren't breaking things. And I'll revert GenericIM and PiecewiseConstantSize, that's easy to change later if you change your mind.

The name changes for the species specific model functions are a bit of a pain though.
 - If you want function names to be snake_case (I agree, this is better), then the test class for `TestDurvasula2017MSMC` will become `Test_sma_1pop`.
 - The TestFoo class for every model 'foo' contains the name Foo multiple times.
 - These will all get touched again because of #122.

I think both these problems could be solved by removing the need for an explicit test class for every new model. E.g. have the model QC process do `from qc.arabidopsis_thaliana_qc import qcmodels`, where models is a dict with the same keys as `Species.models`, and values are the qc models. Then the model tests just iterate on a zip of the species model and the species qc model, etc... Something like this will also reduce the burden on model implementors.

jeromekelleher:
Good, thanks @grahamgower. Perhaps we could also change the tests minimally also, ie., we leave the name of the test class as it is, and just change the model, so we'd have
```python
# This goes before the first model test
species = stdpopsim.get_species("homsap")

class TestTennessenTwoPopOutOfAfrica(unittest.TestCase, test_models.QcdModelTestMixin):
    model = species.get_model("ooa3")
    qc_model = homo_sapiens_qc.TennessenTwoPopOutOfAfrica()
```

This would then be step towards other refactoring we'll want to do later (we shouldn't be importing directly from the module anyway, this is another artefact of an older design).

> I think both these problems could be solved by removing the need for an explicit test class for every new model. E.g. have the model QC process do from qc.arabidopsis_thaliana_qc import qcmodels, where models is a dict with the same keys as Species.models, and values are the qc models. Then the model tests just iterate on a zip of the species model and the species qc model, etc... Something like this will also reduce the burden on model implementors.

Yes, good ideas. Let's get these lined up for post 0.1.0 so we can really streamline the model addition/qc process as much as possible.

jeromekelleher:
I'd like to get the refactoring done on the production stuff ASAP so that we can tackle #122. I'm not so worried about getting all the test and QC code fully refactored, as it's not strictly necessary before a release of the software, and we can tidy this up after release.

grahamgower:
I think that should be everything now. It's just missing documentation for all the Model attributes.

jeromekelleher:
Fantastic, thanks so much for all this work @grahamgower.

I think this is a big simplification, and we can make some issues on how to follow up on refactoring the QC code and how we move towards a more streamlined model addition/QC process.

In the mean time, we have a conflict with #270. It would be nice to get that merged first since it's from a first-time contributor. I'm happy to sort out the merge issues in here after that's in.

jeromekelleher:
OK, #270 is in, so we're ready to go on this. Would you like me to finish this up @grahamgower, or are you happy sorting out the conflicts?

grahamgower:
I can fix the conflicts.

jeromekelleher:
Thanks @grahamgower, merging. I'll open up some follow-on issues.

andrewkern:
hey guys-- this is great, but i fear all this churn is going to hang up the analysis side of things. Can we hold off bug refactors until after our initial release / pre-print?

jeromekelleher:
Sure, this is the last one pre 0.1.0 @andrewkern. There shouldn't be any changes here that affect the external API though.

andrewkern:
okay phew! I love seeing all this development, but I've got to keep the analysis side up-to-date!

PR #284, opened 2019-12-08
Removed daiquiri requirement.

### LINKED ISSUE #161: Drop daquiri or make conda package?
The daiquiri package is used to make setting up logging easier and get coloured output for messages. It's not essential, but it is useful. Unfortunately, it's not packaged on conda.

Therefore, we either need to drop daiquiri, or create a package for it on conda-forge.

jeromekelleher:
Closes #161

We lose the nice coloured output for logs, but it's not worth the packaging hassle of dealing with daiquiri.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,10 +15,6 @@ install:
   - cmd: conda install --yes --file=requirements\conda-minimal.txt
   - cmd: conda info -a
   - cmd: python --version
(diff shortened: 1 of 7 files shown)


jeromekelleher:
I think this is a fairly uncontroversial change, so I'm going to merge without review. Easy enough to reinstate daiquiri later (or do log message colouring ourselves) if we want to.

PR #297, opened 2019-12-09
Modified wording about default recombination rates.

### LINKED ISSUE #217: Default 'flat' recombination maps
Most of the API/help for the recombination inputs say something like this (which I took from the CLI help)

```
  -g , --genetic-map    Specify a particular genetic map. If no genetic map is
                        specified use a flat map by default. Available maps:
```

A flat map is a uniform recombination rate, right? It's not clear where users can find out what this uniform rate is.

Subissue of #215 and #137

gtsambos:
closes #217


jeromekelleher:
LGTM, thanks @gtsambos.

PR #310, opened 2019-12-09
Standardise demographic model ids

jeromekelleher:
Here's my effort to standardise the names of things following the conventions described in #122 

It's **really important** we get this right, so it would be great to get lots of eyes on this. The scheme is up for discussion, but I thought I'd open this PR so we can see what things will look like.

I've not touched the tutorials yet --- we can follow up on those once the dust has settled on all these final naming-things changes.


jeromekelleher:
Thanks @andrewkern. I've updated this to include a regex to define and test the form of the ID string (possibly overly strict, but we can relax as and when) and updated various stuff in the docs.

jeromekelleher:
Probably shouldn't merge this yet though, let's get the explicit go-ahead from the group tomorrow.

gtsambos:
Looks good to me!

jeromekelleher:
Nobody objected to this new scheme on the call, so I'm going to merge.

Issue #309, opened 2019-12-09
Rename --model/-m to --demographic-model -d in CLI


jeromekelleher:
Left over from #303 and #307

jeromekelleher:
Closed in #327

PR #311, opened 2019-12-09
Provide information about simulations run from CLI to user

### LINKED ISSUE #199: Add --dry-run option to CLI

ndukler:
Adds a printout of the simulation details whenever one is run from the CLI

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -329,6 +329,9 @@ def add_simulate_species_parser(parser, species):
     species_parser.add_argument(
         "-q", "--quiet", action='store_true',
         help="Do not write out citation information")
+    species_parser.add_argument(
+        "-D", "--dry-run", action='store_true', default=False,
+        help="Do not run actual simulation")
 
     if len(species.genetic_maps) > 0:
         species_parser.add_argument(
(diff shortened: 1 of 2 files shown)


ndukler:
@jeromekelleher I noticed a bit of code left in the `test_cli.py` file that I should have removed when we decided that a separate option for help species was not necessary. Should I take care of that here or make a separate PR?

ndukler:
Also, here's some sample output from this command:

```
Simulation information:
    Engine: msprime (0.7.1)
    Model id: kamm_ancient_eurasia
    Model name: Multi-population model of ancient Eurasia (Kamm et al. 2019)
        Contig length: 249250621.0
        Mean recombination rate: 1.1485597641285933e-08
        Mean mutation rate: 1e-08
        Genetic map: None
    Population: number_samples (sampling_time):
        Mbuti: 2 (0)
        LBK: 1 (320)
        Sardinian: 3 (0)
        Loschbour: 2 (300)
        MA1: 4 (960)
        Han: 2 (0)
        UstIshim: 1 (1800)
        Neanderthal: 0 (2000)
```

ndukler:
The decrease in coverage doesn't seem to be lines I added.

jeromekelleher:
> @jeromekelleher I noticed a bit of code left in the test_cli.py file that I should have removed when we decided that a separate option for help species was not necessary. Should I take care of that here or make a separate PR?

Well spotted. Please tack it on as an extra commit (or just bundle it in with the same commit, doesn't really matter)

stdpopsim/cli.py
```diff
@@ -420,6 +423,42 @@ def run_simulation(args):
...
+    dry_run_text += f"{indent}Engine: {engine.id} ({engine.get_version()})\n"
+    # Get information about model
+    dry_run_text += f"{indent}Model id: {model.id}\n"
+    dry_run_text += f"{indent}Model name: {model.name}\n"
+    # Get information about relevant contig
+    gmap = "None" if contig.genetic_map is None else contig.genetic_map.name
+    mean_recomb_rate = contig.recombination_map.mean_recombination_rate
+    mut_rate = contig.mutation_rate
+    contig_len = contig.recombination_map.get_length()
+    dry_run_text += f"{indent * 2}Contig length: {contig_len}\n"
```
jeromekelleher: Why is the contig information indented here?  It makes it look like this is sub-information of the demographic model, but it's really top-level information, I think.

stdpopsim/cli.py
```diff
@@ -420,6 +423,42 @@ def run_simulation(args):
     species_parser.set_defaults(runner=run_simulation)
 
 
+def dry_run(engine, model, contig, samples, seed=None):
```
jeromekelleher: This isn't a great name --- ``write_simulation_summary``?

jeromekelleher:
Not sure why codecov is moaning. Let's see what happens when we rebase.

ndukler:
> Looks good @ndukler. This doesn't really close #199 though, as there's no `--dry-run` option to prevent running the simulation. We don't necessarily need this now, but we shouldn't close the issue until we've added the functionality.

Why don't I just add it now then. I just wasn't sure whether we were going to implement it as a separate option.

stdpopsim/cli.py
```diff
@@ -329,6 +329,9 @@ def add_simulate_species_parser(parser, species):
     species_parser.add_argument(
         "-q", "--quiet", action='store_true',
         help="Do not write out citation information")
+    species_parser.add_argument(
+        "-d", "--dry-run", action='store_true', default=False,
+        help="Print out summary of simulation configurations without running simulation")
```
jeromekelleher: More accurate would be "Do not run the actual simulation". It's possible to have ``--dry-run`` ``--quiet`` and print out nothing (although, I'm not sure how helpful that is)

stdpopsim/cli.py
```diff
@@ -329,6 +329,9 @@ def add_simulate_species_parser(parser, species):
     species_parser.add_argument(
         "-q", "--quiet", action='store_true',
         help="Do not write out citation information")
+    species_parser.add_argument(
+        "-d", "--dry-run", action='store_true', default=False,
```
jeromekelleher: I think we want to keep ``-d`` for ``--demographic-model`` (see #309). Maybe "-D"?

ndukler: Sounds good

ndukler:
> Looks good, thanks @ndukler. We'll need to test the dry run functionality. Should be easy enough to do: run a simulation like we do in TestEndToEnd, and assert that stdout is empty (default) or the file specified by "-o" doesn't exist. (Might be need to change how the "--output-file" option is implemented to ensure that; currently we automatically open the file I think. Leaving around empty files is bad manners).

If the `--dry-run` option is enabled `write_output()` is never called so I think that fixes the problem unless there's something I'm missing?

jeromekelleher:
> If the --dry-run option is enabled write_output() is never called so I think that fixes the problem unless there's something I'm missing?

Try it out --- you'll need to write the test case anyway!

ndukler:
Think this is good to go @jeromekelleher

jeromekelleher:
Nice, thanks @ndukler. Merging.

PR #312, opened 2019-12-09
Huber three epoch model QC

ndukler:


andrewkern:
lgtm @ndukler. thanks!

Issue #17, opened 2019-02-04
Mutation rates/recombination rates required for humans


jeromekelleher:
In homo_sapiens.py we have the length of each chromosome and placeholders for mutation and recombination rate data. Ideally, we'd like to have a mean rate that has been estimated for each chromosome. I'm not sure where we'd get this information for mutation rates.

carjed:
I can take care of the mutation rates with data I generated for [this paper](https://www.nature.com/articles/s41467-018-05936-5)—rather than per chromosome, would it be better to do it per 1Mbp window?

jeromekelleher:
That sounds great @carjed! I think we want to have both ultimately --- support for mutation maps in msprime is in the pipeline, so having a map would be extremely useful. But we'd also want simple summaries too for people who don't need this level of detail.

carjed:
cool, do you have a planned format for the mutation map implementation in msprime? A bed file (or files) with columns CHR START END RATE makes the most sense to me, but I can tweak it if needed.

ndukler:
What license are we planning on releasing this under? If we're planning on using the GPL should we use pybedtools BedTool class which already probably has a bunch of useful functionality and optimizations?

andrewkern:
I don't know that package but I'm fine with the GPL. Opinions on future license @jeromekelleher @petrelharp ?

jeromekelleher:
> What license are we planning on releasing this under? If we're planning on using the GPL should we use pybedtools BedTool class which already probably has a bunch of useful functionality and optimizations?

We're currently under MIT, but I really don't mind. GPL can be annoying in that it prevents some people from using your code, so in the interest of this being as widely used as possible I think a permissive license makes sense. But I don't feel strongly enough about it to argue the point at any more length than this.

I don't think we need a BED file dependency here for stdpopsim, the mutation rate map parsing will be done in msprime.

jeromekelleher:
> cool, do you have a planned format for the mutation map implementation in msprime? A bed file (or files) with columns CHR START END RATE makes the most sense to me, but I can tweak it if needed.

Anything like that is fine @carjed, there will be some sort of text input that will map to this scheme.

agladstein:
We should resolve this before submitting the manuscript. If we don't have mutation maps by the publication, at least use the latest published overall mutation rate.

petrelharp:
ping @carjed

jeromekelleher:
[...]

PR #316, opened 2019-12-10
add citations for homsap attributes

### LINKED ISSUE #289: homsap generation_time/population_size values and citations
We need citations for the generation time and population size in humans. These are defined [here](https://github.com/popgensims/stdpopsim/blob/master/stdpopsim/catalog/homo_sapiens.py#L71)

Can someone who knows about these things propose the correct values and citations for these please?

apragsdale:
This adds citations for generation time (default to 29) and Ne (default to 10**4) in humans.

One thing I'm wondering is where we should add citations for the default mutation rate. I have a recent mutation rate estimate citation included here, and updated the mutation rate, but not sure where the citation should go for it, or how (or if) it is displayed when used in a simulation.

Closes #289

### DIFF
--- a/stdpopsim/catalog/homo_sapiens.py
+++ b/stdpopsim/catalog/homo_sapiens.py
@@ -53,24 +53,52 @@ chrX 	 155270560 	 1.164662223273842e-08
 chrY 	 59373566 	 0.0
 """
 
+_tian2019 = stdpopsim.Citation(
+    doi="https://doi.org/10.1016/j.ajhg.2019.09.012",
+    year="2019",
+    author="Tian, Browning, and Browning",
+    reasons={stdpopsim.CiteReason.MUT_RATE}
+)
+
+_tremblay2000 = stdpopsim.Citation(
+    doi="https://doi.org/10.1086/302770",
+    year="2000",
+    author="Tremblay and Vezina",
+    reasons={stdpopsim.CiteReason.GEN_TIME}
+)
+
+_takahata1993 = stdpopsim.Citation(
+    doi="https://doi.org/10.1093/oxfordjournals.molbev.a039995",
+    year="1993",
+    author="Takahata",
+    reasons={stdpopsim.CiteReason.POP_SIZE}
+)
+
 _chromosomes = []
 for line in _chromosome_data.splitlines():
     name, length, mean_rr = line.split()[:3]
     _chromosomes.append(stdpopsim.Chromosome(
         id=name, length=int(length),
-        mutation_rate=1e-8,  # WRONG!,
+        mutation_rate=1.29e-8,
         recombination_rate=float(mean_rr)))
 
-_genome = stdpopsim.Genome(chromosomes=_chromosomes)
+_genome = stdpopsim.Genome(
+        chromosomes=_chromosomes,
+        mutation_rate_citations=[
+            _tian2019.because(stdpopsim.CiteReason.MUT_RATE)])
 
 _species = stdpopsim.Species(
     id="HomSap",
     name="Homo sapiens",
     common_name="Human",
     genome=_genome,
-    # TODO reference for these
-    generation_time=25,
-    population_size=10**4)
+    generation_time=30,
+    generation_time_citations=[
+        _tremblay2000.because(stdpopsim.CiteReason.GEN_TIME)],
+    population_size=10**4,
+    population_size_citations=[
+        _takahata1993.because(stdpopsim.CiteReason.POP_SIZE)]
+    )
 
 stdpopsim.register_species(_species)
 
@@ -176,8 +204,7 @@ def _ooa_3():
         reasons={stdpopsim.CiteReason.DEM_MODEL})
     ]
 
-    generation_time = _species.generation_time
-    citations.extend(_species.generation_time_citations)
+    generation_time = 25
 
     # First we set out the maximum likelihood values of the various parameters
     # given in Table 1.
@@ -276,8 +303,7 @@ def _ooa_2():
             reasons={stdpopsim.CiteReason.DEM_MODEL})
     ]
 
-    generation_time = _species.generation_time
-    citations.extend(_species.generation_time_citations)
+    generation_time = 25
 
     T_AF = 148e3 / generation_time
     T_OOA = 51e3 / generation_time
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Great, thanks @apragsdale. Re the mutation rate, I think we were hoping to get per-chromosome estimates in there (see #17). You're right though, we have no hooks for citations on that. I'll open an issue to track.

apragsdale:
Ok, that sounds fine. I'll just clean this up and re-push, and it should be good to go.

Chris1221:
I'm so sorry if this is an incredibly naive question @apragsdale, I just wanted to quickly clarify something. I could very easily be reading this wrong, and it's not my area of expertise, but when I read the Trembley paper cited for the generation time of 29 years, it was my impression that they were advocating a mean value of 30 years (though they do say 29 is a good value for mitochondrial analysis). I'm just wondering if you could possibly clarify where you've found the 30 year value. Again, very sorry if I've read something incorrectly!

apragsdale:
That's a good point @Chris1221 - going back and reading it, you're right. I hadn't double checked that in a while. I will set it to 30, or if someone has a generation time reference that is different that they prefer, I'm happy using that instead.

apragsdale:
I feel like I'm kind of mangling the generation time/citations functionality. In many (most) of the inferred models, the original publications say "we assumed a generation time of xx" without citation. The ooa_3 model doesn't give a citation for 25 years, for example. In this case, there shouldn't be a citation for the generation time.

If I change the default generation time to something other than 25, and set `generation_time = 25` instead of `generation_time = _species.generation_time` in the ooa_3 model, I fail some of the tests, such as `test_model_citations` (`AssertionError: False is not true : generation time citation not written for homsap`).

Not really sure what the best solution here is.

jeromekelleher:
> If I change the default generation time to something other than 25, and set generation_time = 25 instead of generation_time = _species.generation_time in the ooa_3 model, I fail some of the tests, such as test_model_citations (AssertionError: False is not true : generation time citation not written for homsap).

If 25 is the value that is assumed by the model authors then we should be able to do the same things as the ``_ooa_archaic()`` function, i.e., not specify any extra citations. If not, the we should probably use the species value, and update the QC code to do so also?

jeromekelleher:
For the ``test_model_citations``, we'd need to change the model involved to one that does use the species generation time to make this test pass.

grahamgower:
This looks like a bug in the tests too me. Certainly you should be able to have a model with a generation_time set, but without using the species.generation_time_citation.

apragsdale:
From the call today, most implemented published models won't be using the default generation time or Ne, so tests for those citations being written should be made on some simple generic model, like a single-pop constant size model that grabs default species and chromosome attributes. Does that make sense?

jeromekelleher:
That's a good idea.

apragsdale:
I made some changes here, including a constant size model in the homo sapiens catalog, which uses the default generation time and effective size. The cli test now uses this model to check if citations are written. What do you think?

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -149,6 +169,42 @@
     doi="https://doi.org/10.1126/science.1219240")
 
 
+def _constant():
```
jeromekelleher: This is unnecessary I think --- we'll get this just by running ``stdpopsim homsap`` without specifying a model.

tests/test_cli.py
```diff
@@ -495,7 +495,7 @@ class TestWriteCitations(unittest.TestCase):
     def test_model_citations(self):
         contig = stdpopsim.Contig()
         species = stdpopsim.get_species("homsap")
-        model = species.get_demographic_model("OutOfAfrica_3G09")
+        model = species.get_demographic_model("Constant")
```
jeromekelleher: We need to fix this test so that it's not so strict --- @grahamgower, what can we do here?

stdpopsim/catalog/homo_sapiens.py
apragsdale: Ah, you're right! This model is unnecessary

apragsdale:
The cli does print the citations for the effective size and generation time (for example by running 'stdpopsim homsap 2 -c chr22 -o test.ts'). It doesn't say what those citations are referring to though, so the output looks like
```
If you use this simulation in published work, please cite:
Simulation engine: msprime
  Kelleher et al., 2016: https://doi.org/10.1371/journal.pcbi.1004842
Simulation model: Piecewise constant size
  Takahata, 1993: https://doi.org/10.1093/oxfordjournals.molbev.a039995
  Tremblay and Vezina, 2000: https://doi.org/10.1086/302770
```
when it would be nice to way what the citations are for exactly:
```
If you use this simulation in published work, please cite:
Simulation engine: msprime
  Kelleher et al., 2016: https://doi.org/10.1371/journal.pcbi.1004842
Simulation model: Piecewise constant size
Ne: 1e4
  Takahata, 1993: https://doi.org/10.1093/oxfordjournals.molbev.a039995
Generation time: 30 years
  Tremblay and Vezina, 2000: https://doi.org/10.1086/302770
```

tests/test_cli.py
grahamgower: I think the test is just bogus. It makes no sense to have to change the test code (potentially) every time the model code is changed. I'll send a PR removing the test.

grahamgower:
@apragsdale That's a really good point. How about if we add an optional string field to the `Citation` class, which indicates why something is being cited?

jeromekelleher:
@apragsdale, @grahamgower, can you coordinate on fixing this up to use the new ''reason" introduced in #323 please?

apragsdale:
I'll work on fixing up this PR with the citation updates, in the next day or two.

jeromekelleher:
What's the ETA on this @apragsdale? We'd really like to get this sorted out before release.

apragsdale:
Sorry for the delay. I'm working on this this morning and will push changes today

apragsdale:
I think everything here is looking right after fiddling through the rebase for the new citation scheme. It could use a careful review to make sure I'm calling the citation reasons properly.

jeromekelleher:
OK, thanks @apragsdale.

PR #321, opened 2019-12-10
added citations for drosophila attributes

andrewkern:
conveniently the Li and Stephan paper from 2006 publishes Ne estimates from African flies and a generation time (10 gens/yr). I've used that here. I've also included the Schrider et al mutation rate estimate.

### DIFF
--- a/stdpopsim/catalog/drosophila_melanogaster.py
+++ b/stdpopsim/catalog/drosophila_melanogaster.py
@@ -26,12 +26,23 @@ chrY   3667352
 chrM   19524
 """
 
+# citations
+_LiAndStephan = stdpopsim.Citation(
+    author="Li et al.",
+    year=2006,
+    doi="https://doi.org/10.1371/journal.pgen.0020166")
+
+_SchriderEtAl = stdpopsim.Citation(
+    author="Schrider et al.",
+    year=2013,
+    doi="https://doi.org/10.1534/genetics.113.151670")
+
 _chromosomes = []
 for line in _chromosome_data.splitlines():
     name, length = line.split()[:2]
     _chromosomes.append(stdpopsim.Chromosome(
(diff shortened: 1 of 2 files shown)


dschride:
Sounds good.

stdpopsim/catalog/drosophila_melanogaster.py
```diff
@@ -45,9 +56,10 @@
     id="dromel",
     name="Drosophila melanogaster",
     genome=_genome,
-    # TODO reference for these
     generation_time=0.1,
-    population_size=10**4)  # TODO: This is wildly wrong, right?
+    generation_time_citations=[_LiAndStephan],
+    population_size=8.603e06,
```
apragsdale: This is the current size of the African population, and the ancestral size was inferred to be 1/5 the current size. Do we want to use the current Ne from this paper, or the ancestral size (1.72e6)?

andrewkern: i think we'd want to use the current size?

apragsdale: My thought was that the ancestral size probably better reflects average heterozygosity (similar to using Ne=1e4 in humans even though it's much smaller than current sizes). Practically, the smaller size probably is easier to simulate with a recombination map (are those large rho values still bogging down dmel simulations)? But I'm fine with either choice, and things look good otherwise :)

andrewkern: yeah rho still bogs things down. happy to go with the smaller estimate, as I think you are right-- it would be ideal if the defaults ran relatively fast

andrewkern: okay made that change @apragsdale

Issue #317, opened 2019-12-10
Add mutation_rate_citation and recombination_rate_citation to Genome


jeromekelleher:
As discussed in #316, we have no way to register citations for mutation rate and recombination rate estimates. 

Adding attributes ``mutation_rate_citation`` and ``recombination_rate_citation`` to the Genome object seems like the logical thing to do. We would need to think about how to use these citations when generating the list of citations as well.

grahamgower:
I'll take care of this.

jeromekelleher:
Was this closed in #323?

grahamgower:
> Was this closed in #323?

Yep.

PR #323, opened 2019-12-11
Add a `Citation.reason` attribute.

grahamgower:
Modify citation writing as suggested in #316.

This overloads the addition operator of `Citation` to augment the citation with a reason. An explicit method call might make the code more obvious.

### DIFF
--- a/stdpopsim/catalog/homo_sapiens.py
+++ b/stdpopsim/catalog/homo_sapiens.py
@@ -96,7 +96,8 @@ _gm = stdpopsim.GeneticMap(
         stdpopsim.Citation(
             doi="https://doi.org/10.1038/nature06258",
             year=2007,
-            author="1000 Genomes Project consortium"),
+            author="1000 Genomes Project consortium",
+            reasons={stdpopsim.CiteReason.GEN_MAP}),
         ]
     )
 _species.add_genetic_map(_gm)
@@ -116,7 +117,8 @@ _gm = stdpopsim.GeneticMap(
         stdpopsim.Citation(
             year=2010,
             author="Kong et al",
-            doi="https://doi.org/10.1038/nature09525")]
+            doi="https://doi.org/10.1038/nature09525",
+            reasons={stdpopsim.CiteReason.GEN_MAP})]
     )
 # TODO add a URL citation (see above) here for this:
 # "Please see https://www.decode.com/addendum/ for more details."),
@@ -168,7 +170,8 @@ def _ooa_3():
     citations = [stdpopsim.Citation(
         author="Gutenkunst et al.",
         year=2009,
-        doi="https://doi.org/10.1371/journal.pgen.1000695")
+        doi="https://doi.org/10.1371/journal.pgen.1000695",
+        reasons={stdpopsim.CiteReason.DEM_MODEL})
     ]
 
     generation_time = _species.generation_time
@@ -263,11 +266,12 @@ def _ooa_2():
         stdpopsim.Population(name="EUR", description="European Americans")
     ]
     citations = [
-        _tennessen_et_al,
+        _tennessen_et_al.because(stdpopsim.CiteReason.DEM_MODEL),
         stdpopsim.Citation(
             author="Fu et al.",
             year=2013,
-            doi="https://doi.org/10.1038/nature11690")
+            doi="https://doi.org/10.1038/nature11690",
+            reasons={stdpopsim.CiteReason.DEM_MODEL})
     ]
 
     generation_time = _species.generation_time
@@ -354,7 +358,7 @@ def _african():
     populations = [
         stdpopsim.Population(name="AFR", description="African"),
     ]
-    citations = [_tennessen_et_al]
+    citations = [_tennessen_et_al.because(stdpopsim.CiteReason.DEM_MODEL)]
 
     generation_time = _species.generation_time
     citations.extend(_species.generation_time_citations)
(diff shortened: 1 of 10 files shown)


jeromekelleher:
Thanks for this @grahamgower. Sorry I don't have time to review this properly. A thought that occured to me would be to have subclasses of Citation for these purposes, so we could have PopulationSizeCitation, GenerationTimeCitation, and these classes would then generate the appropriate text like "generation time: <doi>" etc in their ``__str__`` methods (or something).

We'd lose the values, but these are written out already anyway, I think?

grahamgower:
Ne and generation time are not currently being written out with the simulation information. Example output from `stdpopsim dromel -l0.001 -o /dev/null 10` (with this pr applied).
```
Simulation information:
    Engine: msprime (0.7.1)
    Model id: constant
    Model name: Piecewise constant size
    Population: number_samples (sampling_time_generations):
        pop0: 10 (0)
Contig Description:
    Contig length: 23542.271
    Mean recombination rate: 8.4e-09
    Mean mutation rate: 5.49e-09
    Genetic map: None

If you use this simulation in published work, please cite:
Simulation engine: msprime
  Kelleher et al., 2016: https://doi.org/10.1371/journal.pcbi.1004842
Simulation model: Piecewise constant size
Ne: 1720600
  Li et al., 2006: https://doi.org/10.1371/journal.pgen.0020166
Generation time: 10/year
  Li et al., 2006: https://doi.org/10.1371/journal.pgen.0020166
```

andrewkern:
partially related-- we don't have a separate citation slot for default mutation rates. that could be good to output as well.

jeromekelleher:
Seems like the Ne and Generation time information belong more in the first part though @grahamgower? I think it's good that we're separating the information (before) and citation output (after). Perhaps the piecewise constant size model needs and ``__str__`` or something, which outputs the sizes and generation times? Then citations of difference classes are "for" different things?

I'm not keen on hardcoding the reason strings in the model definitions: we'll never get them uniform otherwise and it's another burden for reviewers to have to check that they make sense and are correct.

grahamgower:
I've updated the PR to not include the Ne and generation time values in the citation output. These should be added to the simulation information section with a separate PR. I didn't make subclasses for each of the citation reasons, as there would end up being lots of them and they'd all be almost empty.

grahamgower:
And the second commit here should close #317. I haven't added all the necessary citation entries in the various species' genomes. `stdpopsim DroMel -l0.001 -m OutOfAfrica_2L06 -o /dev/null 10` looks like this:
```
Simulation information:
    Engine: msprime (0.7.1)
    Model id: OutOfAfrica_2L06
    Model name: Three epoch model for African and European populations
    Population: number_samples (sampling_time_generations):
        AFR: 10 (0)
        EUR: 0 (0)
Contig Description:
    Contig length: 23542.271
    Mean recombination rate: 8.4e-09
    Mean mutation rate: 5.49e-09
    Genetic map: None

If you use this simulation in published work, please cite:
[simulation engine]
  Kelleher et al., 2016: https://doi.org/10.1371/journal.pcbi.1004842
[mutation rate]
  Schrider et al., 2013: https://doi.org/10.1534/genetics.113.151670
[generation time, demographic model]
  Li et al., 2006: https://doi.org/10.1371/journal.pgen.0020166
```

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -263,11 +266,12 @@ def _ooa_2():
         stdpopsim.Population(name="EUR", description="European Americans")
     ]
     citations = [
-        _tennessen_et_al,
+        _tennessen_et_al.because(stdpopsim.CiteReason.DEM_MODEL),
```
andrewkern: seems like using `.because` here is a break in the style. we should set the reason when `_tennessen_et_al` is defined IMO

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -354,7 +358,7 @@ def _african():
     populations = [
         stdpopsim.Population(name="AFR", description="African"),
     ]
-    citations = [_tennessen_et_al]
+    citations = [_tennessen_et_al.because(stdpopsim.CiteReason.DEM_MODEL)]
```
andrewkern: same as above-- no reason to use this code here if we set the reason originally as with the other citations

stdpopsim/citations.py
```diff
@@ -23,10 +36,37 @@ class Citation(object):
...
 
+    def print(self, file=sys.stdout):
+        print("[", end="", file=file)
+        for i, reason in enumerate(self.reasons):
+            end = "" if i == len(self.reasons)-1 else ", "
+            print(reason, end=end, file=file)
+        print("]", file=file)
+        print("  "+str(self), file=file)
+
+    def because(self, reasons):
```
andrewkern: I'm not convinced we need the `because` function. I see that it is taking a logical 'or' of `self.reasons` and `reasons` --- will there be cases where we need to add new reasons to a citation once that citation has been defined?

grahamgower: Why yes there is, I'm glad you asked. :)
Its mostly when using DroMel right now, because `Li et al. 2016` can be used for three different reasons, and there are (at least) two reasonable combinations of two of these reasons.

E.g. `stdpopsim DroMel -l0.001 -o /dev/null 10` gives
```
[generation time, population size]
  Li et al., 2006: https://doi.org/10.1371/journal.pgen.0020166
```
While `stdpopsim DroMel -l0.001 -m OutOfAfrica_2L06 -o /dev/null 10` gives
```
[generation time, demographic model]
  Li et al., 2006: https://doi.org/10.1371/journal.pgen.0020166
```

stdpopsim/catalog/homo_sapiens.py
grahamgower: Yep, thanks. Updated.

stdpopsim/citations.py
andrewkern: ah very nice! I didn't look there, and this offers great flexibility then. I like it.

PR #325, opened 2019-12-12
Changed description of length multiplier.

### LINKED ISSUE #216: Length multiplier -- ambiguous descriptions
after reading through all the relevant bits of docs/help and the `get_contig` API, I'm still a little confused about how the `length-multiplier` option is actually changing the simulation. If I specify an msprime simulation with length multiplier 0.5 and some pre-specified chromosome and recombination rate, am I just simulating the first half of this chromosome? Or am I scaling all of the entries in the recombination map by 0.5 so that there are ~half as many mutations between any given positions? ( like, simulating every second SNP in the chromosome)?

In other words, does this parameter affect the number of mutations, or the number of trees in the simulation? Most of the current wording leaves both interpretations open -- for example, the CLI help entry for the -l options is:

```
  -l LENGTH_MULTIPLIER, --length-multiplier LENGTH_MULTIPLIER
                        Simulate a chromsome of length l times the named
                        chromosome
```

This could be documented more clearly in the `get_contig` API and CLI help (see #215 ), especially if this functionality is eventually to be extended to non-uniform recombination rates

gtsambos:
closes #216 -- small changes to length multiplier descriptions

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -354,7 +354,8 @@ def add_simulate_species_parser(parser, species):
                 f"Default={choices[0]}."))
     species_parser.add_argument(
         "-l", "--length-multiplier", default=1, type=float,
-        help="Simulate a chromsome of length l times the named chromosome")
+        help="Simulate a sequence of length l times the named chromosome's length, "
+             "using the named chromosome's mutation and recombination rates.")
     species_parser.add_argument(
         "-s", "--seed", default=None, type=int,
         help=(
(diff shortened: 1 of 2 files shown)


stdpopsim/cli.py
```diff
@@ -354,7 +354,8 @@ def add_simulate_species_parser(parser, species):
                 f"Default={choices[0]}."))
     species_parser.add_argument(
         "-l", "--length-multiplier", default=1, type=float,
-        help="Simulate a chromsome of length l times the named chromosome")
+        help="Simulate a sequence of length l times the named chromosome, using "
```
andrewkern: should this be `"Simulate a chromsome of length l times the named chromosome's length"`?

gtsambos:
thanks @andrewkern! this new commit should do it

PR #326, opened 2019-12-12
updated docs with a new section on Naming conventions

andrewkern:
closes https://github.com/popgensims/stdpopsim/issues/122

...i think...


andrewkern:
@jeromekelleher leaving this for you to peak at before merging so that we can close the issue referred to above

PR #327, opened 2019-12-13
swapped out -m, --model for -d, --demographic-model in cli

andrewkern:
i tried to change as little as possible in the cli, so i left function names like `_model_` alone rather than change them to `_demographic_model_`.  can change them if you think we should.


jeromekelleher:
LGTM, thanks @andrewkern.

PR #328, opened 2019-12-13
updated genetic map IDs to meet schema

andrewkern:
this PR will close https://github.com/popgensims/stdpopsim/issues/123


andrewkern:
that might be really hard as the assembly names don't conform to anything. here is what i added to the docs on naming https://stdpopsim.readthedocs.io/en/latest/development.html#naming-conventions

jeromekelleher:
Thanks @andrewkern. I might follow up with another proposal, as it's a bit loose at the moment.

andrewkern:
sounds good @jeromekelleher

PR #333, opened 2019-12-15
Parameter tables for the remaining models and warning when they are missing

noscode:
I have added parameters for models:
- Zigzag_1S14
- AncientEurasia_9K19
- SouthMiddleAtlas_1D17
- African2Epoch_1H18

Also I have fixed `speciescatalog.py` to emit warning.

I am not sure for SouthMiddleAtlas_1D17 as I have not found table in the paper, it is based on the code. Also there is no common notation among tables, for example in my tables I used `Time`, but I saw something like `Epoch time` and `ADMIX time` in other places. May be we want to discuss it?

### DIFF
--- a/docs/_ext/speciescatalog.py
+++ b/docs/_ext/speciescatalog.py
@@ -142,10 +142,8 @@ class SpeciesCatalogDirective(SphinxDirective):
         path = pathlib.Path(f"parameter_tables/{species.id}/{model.id}.csv")
         if not path.exists():
             message = f"Skipping model parameters for {model.id} due to missing table"
-            # TODO change this to emit a warning once we have the model tables filled
-            # in. See issue #272.
             print(message)
-            # logger.warn(message)
+            logger.warn(message)
             return None
 
         with open(path) as csv_file:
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Fantastic, thanks @noscode! You're right we should standardise the notation in the tables --- let's open an issue to discuss once we've merged this PR.

The SouthMiddleAtlas_1D17 is problematic, because we're not actually using all the values in the arrays I think (just the first 8 I think? It's in the code anyway). Can you update the table to just output the values we're using please? (We should also just remove the stuff we're not using, it's confusing to have it in there, but that should probably be done in another PR where we document the model properly also.)

Can we get some reviewers to check the values here also please?

noscode:
I have looked to the code for SouthMiddleAtlas_1D17. As I see from lines 105-117 all values are used. There are some comments like that:
    # MSMC is accurate from 40Kya-1.6Mya for A.thaliana (Durvasula et al 2017)
    # set the first 7 sizes
    # equal to the size at 8 (~40Kya)

    # set the last 2 entries equal
    # to the size at 30 (~1.6Mya)

I have take them into account while I was filling the table, so it seems there are less values but still a lot.

noscode:
Reviews will be appreciated!

jeromekelleher:
> these look good to me, one question though--- why did you end up uncommenting the logger.warning call in speciescatalog.py?

This is so we raise a warning when generating the docs if the model table is missing, and CI will therefore fail.

jeromekelleher:
Can you rebase this please @noscode? I think there's some updates needed also.

noscode:
I have rebased, please, check it.

jeromekelleher:
Perfect, thanks @noscode!

PR #337, opened 2019-12-17
Tidy up genetic map IDs and docs.

jeromekelleher:
This should finalise the changes needed in the genetic maps before release. Main changes:

- Tried to add some descriptive documentation to the maps. Users who don't know anything about the maps shouldn't have to read the paper. The long_descriptions attempt to give a summary of how the maps were made. The ``description`` is a shorter one-line description of what the map is.

- Made an attempt to make the IDs more descriptive, plus trying to avoid future name clashes (we could have the male and female maps for deCode later). I'm not sure there's much point in adding a year to the IDs since the combination of descriptive name and assembly is likely to be unique.

### DIFF
--- a/stdpopsim/catalog/homo_sapiens.py
+++ b/stdpopsim/catalog/homo_sapiens.py
@@ -112,21 +112,28 @@ stdpopsim.register_species(_species)
 
 _gm = stdpopsim.GeneticMap(
     species=_species,
-    id="HapmapII_GRCh37",
-    description="Hapmap Phase II",
-    long_description=(
-        "The Phase II HapMap Genetic map (lifted over to GRCh37) used in "
-        "1000 Genomes. Please see the README for more details."),
+    id="HapMapII_GRCh37",
+    description="HapMap Phase II lifted over to GRCh37",
+    long_description="""
+        This genetic map is from the Phase II Hapmap project
+        and based on 3.1 million genotyped SNPs
+        from 270 individuals across four populations (YRI, CEU, CHB and JPT).
+        Genome wide recombination rates were estimated using LDHat.
+        This version of the HapMap genetic map was lifted over to GRCh37
+        (and adjusted in regions where the genome assembly had rearranged)
+        for use in the 1000 Genomes project. Please see the README file on
+        the 1000 Genomes download site for details of these adjustments.
+        """,
     url=(
         "https://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/working/"
         "20110106_recombination_hotspots/"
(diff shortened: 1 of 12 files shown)


andrewkern:
this LGTM @jeromekelleher.

jeromekelleher:
Great, thanks @andrewkern. 

@reedacartwright, @dschride, quick opinion here: do genetic map IDs like ``HapMapII_GRCh37``,
``DeCodeSexAveraged_GRCh36``, ``ComeronCrossover_dm6`` and  ``SalomeAveraged_TAIR7`` look OK to you?

reedacartwright:
On first pass they look good. However, the style is inconsistent with the model id scheme.  An alternative scheme is `TAIR7_Averaged_S12`, `dm6_Crossover_C12`, `GRCh37_HapMapII_H07`, and `GRCh36_DecodeSexAveraged_K10`.  I like the refernce name first because if we did a liftover to a new reference, we would be changing the first element.  Of course I could see reasons for putting it last as well.

jeromekelleher:
> On first pass they look good. However, the style is inconsistent with the model id scheme. An alternative scheme is TAIR7_Averaged_S12, dm6_Crossover_C12, GRCh37_HapMapII_H07, and GRCh36_DecodeSexAveraged_K10. I like the refernce name first because if we did a liftover to a new reference, we would be changing the first element. Of course I could see reasons for putting it last as well.

Thanks @reedacartwright, great to get your input. Thoughts anyone else? I see the arguments here, but the current scheme feels more memorable and we're not expecting to have as many genetic maps as demographic models so we don't need to worry about collisions so much. 

I'm happy to change if there's consensus, but I'd rather not otherwise I think. Opinions either way appreciated.

andrewkern:
I'd rather not change it at this point as we sought consensus on this issue over the past weeks and arrived at the current strategy.

reedacartwright:
That makes sense.

stdpopsim/catalog/homo_sapiens.py
```diff
@@ -84,38 +84,49 @@
...
-    description="Hapmap Phase II",
-    long_description=(
-        "The Phase II HapMap Genetic map (lifted over to GRCh37) used in "
-        "1000 Genomes. Please see the README for more details."),
+    id="HapMapII_GRCh37",
+    description="HapMap Phase II lifted over to GRCh37",
+    long_description="""
+        This genetic map is from the Phase II Hapmap project
+        and based on 3.1 million genotyped SNPs
+        from 270 individuals across four populations (YRI, CEY, CHB and JPT).
```
gtsambos: Tiny typo -- this should be CEU

PR #351, opened 2019-12-19
Added a mention of generic models into the catalog.

### LINKED ISSUE #191: Docs revamp: "Catalog" and "Tutorial" sections
I think these are the two sections that users will jump to to make a start on using `stdpopsim`, so it's particularly important that these sections are clear and well-structured. At present I think they're both a bit confusing:

 1. It's not clear what 'Catalog' means, as @agladstein has noted in #148. Also, although there is some repeated structure to this section (Genome, Genetic Maps, Models), it's not obvious why these particular features are the things the user needs to choose. The stdpopsim ingredients that they correspond to have different names ('Genome' information is part of the `Species` object and 'Genetic Maps' are part of the the `Contig` object, right)? I have some specific suggestions for what to do about this that I'll put in #148.

 2. The 'generic models' should all be included in a subsection of 'Catalog' as well -- at the moment, you can only see that these models exist in `stdpopsim` because of the API docs (and the single example in the 'Tutorials' section).

3. The 'Tutorials' section should have some examples of different demographic models (generic and specific) and different species, to highlight the fact that there are different default parameter values in different models and species (BTW: I think this is going to be a big gotcha for users, and will mention in a separate issue somewhere). I think it would be good if the 'Tutorials' section was split into 'generic' and 'specific' subsections also.

I plan to get a start on all of these, but wanted to open up discussion here in case there's disagreement.

gtsambos:
Closes #191.

> How about we add a "Generic models" section to the end of the catalog page, and put a manual table in there linking to the generic models?

I played around with this, but ended up deciding against it because it was just repeating a lot of the info in the API. There's also no natural way to structure this table so that it looks like the rest of the catalog, given can be used with any of the species.

The main point was that the catalog should make mention of *all* of the simulation options available -- so I've added a small bit to the intro that explains where the generic models are listed.

### DIFF
--- a/docs/catalog.rst
+++ b/docs/catalog.rst
@@ -31,8 +31,14 @@ and saves the output to a file called ``test.trees``:
 
     $ stdpopsim HomSap -c chr22 -o test.trees -g HapMapII_GRCh37 --model OutOfAfrica_3G09 2 2 2
 
-(To learn more about using ``stdpopsim`` via the command-line, read our :ref:`sec_tutorial`
-about it.)
+(To learn more about using ``stdpopsim`` via the command-line, read our 
+:ref:`tutorial <sec_cli_tute>` about it.)
+
+In addition to the species-specific models listed in this catalog, ``stdpopsim`` offers
+a number of generic demographic models that can be run with any species.
+These are described and listed in the :ref:`Generic models API <sec_api_generic_models>`.
+Simulations using these generic models must be run via the Python interface; see our
+:ref:`Python tutorial <sec_python_tute>` to learn how to do this.
 
 Are there other well-known organisms, genetic maps or models that
 you'd like to see in ``stdpopsim``? Head to our :ref:`sec_development`
@@ -50,3 +56,4 @@ Then, if you feel ready, make an issue on our
 .. speciescatalog:: AraTha
 
 .. speciescatalog:: EscCol
+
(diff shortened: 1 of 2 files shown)


gtsambos:
actually, don't pull this in yet!! One small thing I'll add when I get back from lunch 🍔

gtsambos:
I was naive to think I could do this after the office Xmas party 🍻  ... It's not too urgent -- will let you know when it's done

agladstein:
ping me when you're ready for me to have a look at it.

gtsambos:
image
Thanks @agladstein -- done now.

gtsambos:
Hmmm, actually, there's still something missing here -- we will also need to add information about the *populations* that are being sampled from. This isn't always obvious -- for example, in the `IsolationWithMigration` model, there are 3 populations, but only 2 of them are sampled from and you need to know their IDs to work with them.

andrewkern:
yes this goes back to that discussion we had in one of the last calls before break-- the issue of "sampling populations" vs populations that are implemented in the model.

jeromekelleher:
Hey @gtsambos, do you think we can pick this up again and get it merged? Needs a rebase now.

jeromekelleher:
I think we should keep this on ice until after 0.1.2 is shipped. Particularly as there's some issues with the API for IM model (#533)

andrewkern:
just bumping this PR cause it would be great to have mentions of the Generic models in the catalog

andrewkern:
i pulled this PR locally and was able to manually resolve the conflicts-- they were easy-- i think i'll push these few edits and then merge

Issue #359, opened 2019-12-22
add citations for genome assemblies


andrewkern:
Deanna Church noted on twitter (https://twitter.com/deannachurch/status/1208752315483574272?s=20) that we are not outputting assembly citations. We should definitely add that. We will need to make a new slot in the Species class I figure?

andrewkern:
actually i take that back. the new `assembly_citation` should probably go here https://github.com/popsim-consortium/stdpopsim/blob/e2e76fb1f6987233e1f4c1d6eedb0ae0bcdb1f41/stdpopsim/genomes.py#L29

PR #360, opened 2019-12-22
adding citation information to genome assembly

andrewkern:
closes https://github.com/popsim-consortium/stdpopsim/issues/359

### DIFF
--- a/stdpopsim/catalog/arabidopsis_thaliana.py
+++ b/stdpopsim/catalog/arabidopsis_thaliana.py
@@ -36,7 +36,17 @@ for line in _chromosome_data.splitlines():
         mutation_rate=7e-9,
         recombination_rate=8.1e-9))
 
-_genome = stdpopsim.Genome(chromosomes=_chromosomes)
+_SwarbreckEtAl = stdpopsim.Citation(
+    doi="https://doi.org/10.1093/nar/gkm965",
+    year="2007",
+    author="Swarbreck et al.",
+    reasons={stdpopsim.CiteReason.ASSEMBLY}
+)
+
+_genome = stdpopsim.Genome(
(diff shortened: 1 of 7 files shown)


stdpopsim/catalog/arabidopsis_thaliana.py
```diff
@@ -36,7 +36,17 @@
...
+    doi="https://doi.org/10.1093/nar/gkm965",
+    year="2007",
+    author="Swarbreck et al.",
+    reasons={stdpopsim.CiteReason.ASSEMBLY}
+)
+
+_genome = stdpopsim.Genome(
+        chromosomes=_chromosomes,
+        assembly_citations=[
+            _SwarbreckEtAl.because(stdpopsim.CiteReason.ASSEMBLY)])
```
grahamgower: If the citation has `reasons` specified, then you don't need the `because()` call later. You could probably just put the ASSEMBLY citation inline here (with `reasons` given), instead of storing it in _SwarbreckEtAl, as it probably wont be used elsewhere.

andrewkern: can do

stdpopsim/cli.py
```diff
@@ -239,6 +239,7 @@ def get_citations(engine, model, contig, species):
     Return a list of all the citations.
     """
     citations = engine.citations[:]
+    citations.extend(species.genome.assembly_citations)
```
andrewkern: @grahamgower isn't this all I need? seems to print fine

grahamgower: Oh, yep. Sorry. Not sure how I missed that!

andrewkern:
okay this should be all set @grahamgower. if you give approval I'll merge.

Issue #354, opened 2019-12-19
PonPyg is wrong ID


andrewkern:
Genome we have is actually PonAbe. This is a bug in first release, but only the identifier / name is affected.

jeromekelleher:
Ideal thing to do would be to fix PonPyg and add another species for PonAbe

jeromekelleher:
Simplest fix would be to temporarily remove Pongo and push out a quick release. We can figure out the right thing to do afterwards. Having a multispecies model does break our assumptions about how the whole thing is set up, so we should consider it carefully before committing to supporting something.

andrewkern:
agreed

dschride:
> Having a multispecies model does break our assumptions about how the whole thing is set up

Well, there are species, and then there are _species_.

apragsdale:
I agree with Jerome that we should have two different species defined, in that each has their own genome attributes. Will need to think about how to share demographic models between them.

Sorry that I made a mess of the pongo species.. I kind of arbitrarily picked which to use for chromosome lengths, and I'm waiting still to hear back about whether we can include the two recombination maps for them.

I'm still mostly away from the computer this week, but I'll start trying to clean some of this up over the next few days and play with some ideas of sharing the same demographic model across the two species.

andrewkern:
no worries @apragsdale-- easy enough to hint at Pongo being next in the manuscript.

jeromekelleher:
Yeah, no worries @apragsdale, this is a small detail easily fixed.

apragsdale:
Fixed with #363.

Issue #342, opened 2019-12-18
Pongo recombination map


apragsdale:
As of #338, pongo is more or less up and running. It is only missing a recombination map. In looking for a published map, I contacted the first author of a study that inferred maps using LDhat for both P. pygmaeus and P. abelii (doi: 10.1016/j.cub.2017.09.047), and he sent me a link to download them.

So we could include them if we want. Is there an issue with us making these maps publicly available, when I had to ask the authors if they were available since they weren't included in the supp material or an easy-to-find repo?

andrewkern:
Would you mind asking the authors what they think? I'm actually just working on the changes to push the PonPyg documentation live

apragsdale:
Ok, I'll do that. It would be very nice to include them.

andrewkern:
and I'm going to push ahead with the documentation. should be ready for a PR in a minute and maybe you could review it for me?

apragsdale:
Picking this back up -- I'd like to get the Pongo species(es) sorted out (and finally resolve PR #363), including getting the two recombination maps cached in our map repository.

Do we now have an independent location to store recombination maps (did this get resolved in #335), and is there information about what format these need to be in? Even if the storage isn't sorted out yet, I can reformat the Pongo maps into whatever we need for them.

grahamgower:
I don't know about the repository, but the file format is described here: https://msprime.readthedocs.io/en/stable/api.html#msprime.RecombinationMap.read_hapmap

apragsdale:
Maps now included from PR #363

Issue #335, opened 2019-12-16
collect all maps at one URL?


andrewkern:
i think it's time to think about a more permanent home for all of our maps. does anyone have any good suggestions as to where we might home these data?

jeromekelleher:
How about [here](https://aws.amazon.com/opendata/public-datasets/)? I emailed one of the people in charge of it a while back and he said we'd be a pretty easy sell.

andrewkern:
yeah that looks great to me. they have a deadline coming up on 12/31. Should we go for it?

ndukler:
Started work on a document that fills out some of the fields that require a paragraph form response here: https://docs.google.com/document/d/1JVUDVb39jKEoX80Hjpl7XEtN6j6rWuwylE9K3s0cQeY/edit?usp=sharing

ndukler:
Application here: https://application.opendata.aws/

ndukler:
I'll try to fill it out more tomorrow

jeromekelleher:
I'm checking out for the holidays today, so I won't have time to do this. Very happy for others to follow it up though.

ndukler:
I can handle most of this but I think it makes sense for either @andrewkern or @jeromekelleher to be the actual corresponding individual. What do you think of the text so far @andrewkern ?

andrewkern:
happy to work on it today with you @ndukler

ndukler:
I think I've filled out the entire form on the googledoc.

ndukler:
Any idea what the data-set license should be? @andrewkern @petrelharp ?

andrewkern:
this is all not our data so unclear to me? the only thing with possible restrictions I imagine is decode

ndukler:
They seem to provide it all here (https://www.decode.com/addendum/) with no limitations?

andrewkern:
great so let's proceed with a completely open license

ndukler:
@andrewkern it's complete if you'd like to make some suggestions on the googledoc

andrewkern:
@ndukler this looks great! i've made a pass and made a few quick edits. If you can submit it for me that would be amazing!!

ndukler:
So I submitted it awhile ago. Did you get an email?

andrewkern:
nope

ndukler:
Hmm, I'll resubmit

ndukler:
[...]

PR #363, opened 2020-01-02
Pongo update

apragsdale:
So I got around to doing some more digging on the orangutan genome assemblies. Even though there are multiple Pongo species, it looks like all genomic studies use a Pongo abelii (Sumatran) reference genome (typically Susie, i.e. PonAbe2) to map reads, even for Pongo pygmaeus. So I think it makes sense to keep a single species, and use PonAbe2 as the genome build. I've gone through and changed names (e.g. PonPyg to PonAbe), and updated some citation and genome information. (see #354, #342 for related issues)

I also have permission to upload the two genetic maps, one each for Pongo abelii and pygmaeus. The map info is included in this PR, but I'm not exactly sure what needs to be done to get these maps in the cache (esp. in light of #335).


jeromekelleher:
Thanks @apragsdale, looks good. As a temporary workaround I've pushed a change in #367 which removes Pongo from the catalog so we can take our time fixing this.

The updates here are excellent and should be used to define the PonAbe species. However, it's not clear how we should deal with the multispecies model here. Whose genome should this be simulated with?

This might require a redesign and refactor.

apragsdale:
> Thanks @apragsdale, looks good. As a temporary workaround I've pushed a change in #367 which removes Pongo from the catalog so we can take our time fixing this.

Makes sense, especially if a refactor is necessary.

> The updates here are excellent and should be used to define the PonAbe species. However, it's not clear how we should deal with the multispecies model here. Whose genome should this be simulated with?

I'm not sure what the right general solution for this problem would be. But for orangutan, even though there are two or three species, genomic studies map reads from each species to a common reference genome, either ponAbe2 or ponAbe3, which comes from a Sumatran orangutan, and all the genome attributes, such as the recombination map, chromosome lengths, etc, correspond to PonAbe.

So if we think of our "species" as "genome assemblies + attributes", I don't think there is an inconsistency in this case. If there also happened to be a genome assembly from a Bornean orangutan, then I think it would warrant its own "species", and we could duplicate the demographic model there, so you would be able to run the simulation using either assembly. Maybe a solution to avoid duplicated code would be to have a "shared_models.py" file, and each genome assembly/species that uses one of those shared models would import the model it needs and add it to the species using `_species.add_demographic_model`?

jeromekelleher:
All sounds sensible to me, thanks @apragsdale. A pragmatic definition of what we mean by species which we document clearly would go a long way to clearing things up.

andrewkern:
hey @apragsdale lets bash this out while you are visiting UO

apragsdale:
I don't know if it would be possible, but getting the Pongo species and demographic model wrapped up before we resubmit would be really great. As it stands now, I have two recombination maps for the two Orang species in hand, both mapped to the standard Pongo abelii genome build (see #342).

I'd propose we file both maps under this species/genome build, which avoids dealing with any multi-species issues for now. I think it's becoming clear in any case that we think of our "species" as linked to a specific genome build, so this would fit under that interpretation. I'll need to update a bit and rebase, but I think it's really close.

apragsdale:
Not sure why the travis CI is failing.. but this is almost ready to go. The recombination maps are also in the HapMap format. I might need some help figuring out where to get those uploaded. Am I forgetting anything else we need to clear up for the Orangs?

jeromekelleher:
LGTM, thanks @apragsdale. @petrelharp, @grahamgower anyone want to take a look here to sort out our Orangs?

(Travis fail just needs a rebase)

stdpopsim/catalog/pongo_abelii.py
```diff
@@ -20,36 +20,36 @@
 # List of chromosomes.
 
 # length information can be found here
-# <http://https://www.ncbi.nlm.nih.gov/genome/325>
-# Orangutan Jan. 2018 (Susie_PABv2/ponAbe3) Browser Sequences
-# Note: no chrY
-# Average recombination rate estimated in Locke et al (doi:10.1038/nature09687)
-# which reports 0.95 +/- 0.72 cM/Mb
+# <http://https://www7.ncbi.nlm.nih.gov/genome/325>
```
grahamgower, line 23: "http://" is duplicated here, and I had to remove the 7 from `www7` to get the orangutan page.

apragsdale: Thanks Graham, not sure how that got mangled. Will fix.

apragsdale:
Thanks @ndukler for uploading the maps. I'm running into some issues passing the tests for downloading, getting

`ValueError: Tarball format error: member rec_map.PA not a file`.

Is there a trick to getting this to work, or did I perhaps mess something up with the recombination map files I shared with you?

ndukler:
I don't know. It worked fine for the other species and it looks like the URL is correct.

apragsdale:
There's probably some issue with how I compressed the files? I was kind of guessing at exactly how it needs to be formatted for the upload. I don't think that information is documented currently.

ndukler:
I'll look into it

ndukler:
I've got a suspicion it's the `.` in the filenname. Let me rename it to `rec_map_PA.tar.gz` and  `rec_mapPP_.tar.gz`. Please change the urls to reflect this

apragsdale:
Yeah, I think I see the issue - the recombination maps are within a directory in the tar.gz. Let me re-compress this and reshare it with you

apragsdale:
Ok @ndukler I just replaced the maps in that Dropbox I shared with you (they are named NaterPA_PonAbe2.tar.gz and NaterPP_PonAbe2.tar.gz). I think those should work. Let me know if something seems off. And thanks for helping out with this.

ndukler:
And they're up

apragsdale:
Yessss all checks have passed. Thanks a bunch Noah!

ndukler:
I literally just clicked upload a few time :) Thanks for everything you did!

grahamgower:
Awesome, thanks folks! @apragsdale, could you rebase and squash please?

apragsdale:
Yep, all set @grahamgower. Thanks all, really happy to get our Orangutans fixed up before the resubmission.

grahamgower:
Merged! Thanks @apragsdale. There doesn't seem to be an Orangutan-relevant reaction emoticon... Might need to send a github a feature request about that!

PR #367, opened 2020-01-02
Remove PonPyg from the catalog temporarily.

### LINKED ISSUE #365: Remove PonPyg from catalog temporarily
To quickly resolve the issues around how we manage the PonPyg/PonAbe issue (#354) and how we manage multi species models, I think we should remove PonPyg from the catalog temporarily.

With this, and having sorted the packaging issues in #364 and #357 we can push out a quick release and then take our time dealing with this tricky problem.

jeromekelleher:
Closes #365

### DIFF
--- a/docs/catalog.rst
+++ b/docs/catalog.rst
@@ -43,8 +43,6 @@ Then, if you feel ready, make an issue on our
 
 .. speciescatalog:: HomSap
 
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Ignoring the codecov complaint as it's trivial.

PR #377, opened 2020-01-08
Support pulses of admixture in the SLiM engine.

### LINKED ISSUE #258: [slim-engine-dev] Support msprime.MassMigration with proportion < 1
E.g. try supporting the Browning et al. 2011 homsap/america model.

grahamgower:
Closes #258.

This works for regular pulses of admixture, like in `HomSap/AncientEurasia_9K19`, by turning off migrations one generation after the admixture pulse is requested. However, due to a quirk of the model implementation, `HomSap/AmericanAdmixture_4B11` will silently produce the wrong admixture proportions.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -17,7 +17,7 @@ How backwards-time demographic events are mapped to forwards-time SLiM code:
  * `msprime.DemographyDebugger()` does much of the hard work by extracting
    epochs from the given model's `demographic_events`, and calculating a
    migration_matrix for each epoch from the `msprime.MigrationRateChange`
-   events. The epochs boundaries defined here are indirectly translated into
+   events. The epoch boundaries defined here are indirectly translated into
    "late events" in SLiM.
 
  * `msprime.PopulationParametersChange` events are translated into SLiM as
@@ -30,6 +30,9 @@ How backwards-time demographic events are mapped to forwards-time SLiM code:
  * `msprime.MassMigration` events with proportion<1 indicate an admixture
    pulse at a single point in time. In SLiM, we call `pop.setMigrationRates()`
    in the relevant generation, and turn off migrations in the next generation.
+   When multiple MassMigration events correspond to a single SLiM generation,
+   the migration proportions multiply, following the msprime behaviour and
+   event ordering.
 
  * The migration_matrix for each epoch describes continuous migrations that
    occur over long time periods. In SLiM, we call `pop.setMigrationRates()`.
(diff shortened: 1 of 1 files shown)


jeromekelleher:
LGTM, I think. It could do with some more explanation of how such admixture pulses are defined though --- I worry that this could be slightly over generalised.

Also, can you rebase on slim-engine-dev please?

grahamgower:
Where would you expect to see this explanation? Just in the commit message, or somewhere in the code? (If so, where?)

"""
An admixture pulse is an `msprime.MassMigration` event with `proportion<1`, and indicates admixture at a single point in time. This is as opposed to `msprime.MassMigration` events with proportion=1 (forwards in time, these represent population splits), or continuous migrations that occur over longer time periods (defined with a migration_matrix for each epoch).
""""

Probably there will need to be some more extensive documentation describing the internal mapping of msprime-style backwards-time API to the slim code. I'm open to suggestions as to where this should go. I'll open a ticket.

jeromekelleher:
You're right, this would probably need to done in the developer documentation. Could just add a section in there and add bits as you go? We can easily move it somewhere else later if we want.

grahamgower:
Ok, I added some notes at the top of the slim_engine.py file.

grahamgower:
@jeromekelleher If you're happy with this now, I'll squash the commits ready for merging.

It closes #258 and #378.

jeromekelleher:
Test coverage is pretty low @grahamgower --- anyway we can bring this up this a bit? Even just parsing the Eidos code to make sure it compiles would be good, if running small simulations isn't feasible.

We can skipIf(Windows), and use the conda package on OSX for travis.

grahamgower:
I'm not aware of a `--parse-only` type flag for slim, but I'm working on putting in some tests today. I have a few other cleanups pending too. Do you mind if I just stack a bunch of not-really-related slim commits under one PR?

jeromekelleher:
> I'm not aware of a --parse-only type flag for slim, but I'm working on putting in some tests today.

What does the ``eidos`` command that comes with SLiM do? I had assumed it was a standalone interpreter which could be used for something like this (it's not important though --- actually running sims would be much better).

> I have a few other cleanups pending too. Do you mind if I just stack a bunch of not-really-related slim commits under one PR?

Sure, go for it.

grahamgower:
> What does the `eidos` command that comes with SLiM do?

It's an interpreter for the eidos language, and yes it stands alone from the slim binary. But it doesn't include the class definitions and `initialize()` block stuff. Those are slim-specific extensions to the eidos language.

Issue #373, opened 2020-01-07
QC for PapuansOutOfAfrica_10J19 model


andrewkern:
opening this issue from merge of https://github.com/popsim-consortium/stdpopsim/pull/372

Jacobs et al doi is https://doi.org/10.1016/j.cell.2019.02.035

grahamgower:
Thanks @andrewkern.

The population indices for QCing can be found here:
https://stdpopsim.readthedocs.io/en/latest/catalog.html#sec_catalog_homsap_models_papuansoutofafrica_10j19

Most model parameters are taken from Jacobs et al., Table S5 and Figure S5, but a few were found in the supplement text (section S10.i and S10.h, pgs. 30/45 & 31/45).

grahamgower:
I occurs to me that maybe no one wants to QC this because its big and will be time consuming. So I'll offer an incentive: if you QC this model, I promise to QC a future model of your choice.

ndukler:
I'll do it this weekend. I've just been a bit bogged down with my other work recently.

noscode:
I have almost done it. I have some questions though:
1) Time of Neanderthal Altai and Neanderthal Introgressing split event - I found in the Table S5A 3375 generations, but in implemented model there is 2058. It seems to be sampling time of DenA population.
2) 

`    # Fig. 4B, and S10.h p. 30/45
    T_Den1_Papuan_mig = 29.8e3 / generation_time
    T_Den2_Papuan_mig = 45.7e3 / generation_time`

I have not found that parameters but I think that they should be both equal to 1353. In section S10.h I found the following:

> 3. Instead of a single Denisovan introgression into Australians 1353 generations ago of 4%, there are two introgressions 1353 generations ago into Papuans, one from D1 and one from D2

grahamgower:
> I have almost done it. I have some questions though:

Awesome, thanks @noscode!

> 
> 1. Time of Neanderthal Altai and Neanderthal Introgressing split event - I found in the Table S5A 3375 generations, but in implemented model there is 2058. It seems to be sampling time of DenA population.

Yep, that looks wrong. I'll send a pull request changing that.

> 
> ` # Fig. 4B, and S10.h p. 30/45`
> ` T_Den1_Papuan_mig = 29.8e3 / generation_time`
> ` T_Den2_Papuan_mig = 45.7e3 / generation_time`
> 
> I have not found that parameters but I think that they should be both equal to 1353. In section S10.h I found the following:
> 
[...]

PR #386, opened 2020-01-16
Use 3375 generations for Nea/Nea split time in PapuansOutOfAfrica_10J19.

grahamgower:
Found by @noscode during QC, #373.

### DIFF
--- a/docs/parameter_tables/HomSap/PapuansOutOfAfrica_10J19.csv
+++ b/docs/parameter_tables/HomSap/PapuansOutOfAfrica_10J19.csv
@@ -1,3 +1,4 @@
+Generation time (yrs.),29,Generation time
 Population size,"48,433",African pop. size
 Population size,"6,962",European pop. size
 Population size,"9,025",East Asian pop. size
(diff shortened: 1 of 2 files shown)


ndukler:
@grahamgower Make sure you propagate any changes in parameter values to the documentation, in particular the parameter table.

grahamgower:
Ah, yes. Thanks @ndukler

jeromekelleher:
LGTM, happy to merge this if I get thumbs up from @ndukler.

ndukler:
This looks mostly good. Just should probably have some indication in the table of what the generation time used to translate between the years on the parameter table and the generations in the model is. I know there is a default but I think it is better to be explicit.

noscode:
I found another thing that looks like a mixed up populations. It is in the same issue. So may be @grahamgower will check and add it here too?

grahamgower:
@ndukler I moved the generation time to the top of the parameter table, where it should be more obvious.

@noscode Thanks for tracking down the mixed up popuations. I've fixed that too.

noscode:
That's great @grahamgower! You are welcome. I will make QC PR after this will be merged.

jeromekelleher:
OK, I'm going to merge this to @noscode's PR can be updated. @ndukler, please shout out if there was anything else needed here so we can follow up.

PR #387, opened 2020-01-17
QC for PapuansOutOfAfrica_10J19 model

noscode:
I have implemented QC model. (Issue [373](https://github.com/popsim-consortium/stdpopsim/issues/373))

However tests that models are equal failed locally. If I comment all admixtures in both model they are equal, but if I add any one admixture, even the same lines in both models then test failed.  May be somebody look at tests and check it?

### DIFF
--- a/qc/homo_sapiens_qc.py
+++ b/qc/homo_sapiens_qc.py
@@ -526,3 +526,248 @@ class KammAncientSamples(models.DemographicModel):
                 time=t_Nean_Losch, initial_size=N_Nean_Losch,
                 population_id=3)
         ]
+
+
+class DenisovanAncestryInPapuans(models.DemographicModel):
+    """
+    Demographic model from Jacobs et al (2019). The model is
+    illustrated on Figure S5, parameters are in Table S5
+    """
+    def __init__(self):
+        self.generation_time = 29  # Just information
+
+        # sizes of populations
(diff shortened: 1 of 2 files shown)


noscode:
Also while I was looking for the errors I found that test is sensitive to the order of migration changes. So if I have two models where at the same time two migrations are set to zero but in different order, these models are not equal, but they are originally the same.

I think that because this model is very big it is very difficult to make tests work. I could try to find out why all of this happens but not right now.

ndukler:
I'n order to pass the QC the migration events must be in the same order even if they are simultaneous and the ordering has no actual effect on the model. I'm 99.99% sure you can always rearrange simultaneous migration rate changes without consequence and I've had to do that for other QC models as well to get them to match production code. That is not true for mass migrations so be more careful with those but just skimming the code I didn't see that happening. @noscode

noscode:
@ndukler what do you mean saying to be careful with MassMigration? I have found my problem, I did not noticed sort of demographic events. Including it solve everything, so it seems qc check is also sensitive to MassMigration order.

ndukler:
In your case I don't think it is a problem as all mass migrations have different times. It's just a general caveat from an experience I had with another model.

jeromekelleher:
Looks like we're good to go? @andrewkern, if you agree, can you merge please?

PR #393, opened 2020-01-30
Add Msprime DTWF simulation engine.

grahamgower:


jeromekelleher:
Looks good, but let's think through the idea of making the ``model`` a parameter for the msprime engine also. Looks like a fair bit of duplication if we make a separate engine for each model in msprime.

grahamgower:
Superceded by #398.

PR #398, opened 2020-02-07
Support non-hudson msprime simulation models.

### LINKED ISSUE #361: Make it clear that random seeds are available in Python API
Can you specify a random seed for simulations that are run with the Python API?

If not, we should add this!!
If so, we should document it better.

### LINKED ISSUE #392: Use msprime's wirght-fisher model (model='dtwf') for simulation
Based on this preprint, https://www.biorxiv.org/content/10.1101/674440v1.full.pdf, it is better to simulate large sample/large chromosome region using the wright-fisher model in msprime. Could such functionality be provided in stdpopsim as well? (Correct me if it's already there, maybe I just missed it from the API). Thx!

grahamgower:
Closes #392. Includes related doc changes which closes #361.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -509,10 +509,32 @@ def stdpopsim_cli_parser():
         choices=[e.id for e in stdpopsim.all_engines()],
         help="Specify a simulation engine.")
 
-    for engine in stdpopsim.all_engines():
-        group = top_parser.add_argument_group(
-                f"{engine.id} specific parameters")
-        engine.add_arguments(group)
+    supported_models = stdpopsim.get_engine("msprime").supported_models
+    msprime_parser = top_parser.add_argument_group("msprime specific parameters")
+    msprime_parser.add_argument(
+            "--msprime-model",
+            default=supported_models[0],
+            choices=supported_models,
+            help="Specify the simulation model used by msprime. "
+                 "See msprime API documentation for details.")
+
+    def time_or_model(arg, _arg_is_time=[True, ], parser=top_parser):
+        if _arg_is_time[0]:
+            try:
+                arg = float(arg)
+            except ValueError:
+                parser.error(f"`{arg}' is not a number")
+        else:
+            if arg not in supported_models:
+                parser.error(f"`{arg}' is not a supported model")
+        _arg_is_time[0] = not _arg_is_time[0]
+        return arg
+    msprime_parser.add_argument(
+            "--msprime-change-model",
+            metavar=("T", "MODEL"), type=time_or_model,
+            default=[], action="append", nargs=2,
+            help="Change to the specified simulation MODEL at generation T. "
+                 "This option may provided multiple times.")
 
     subparsers = top_parser.add_subparsers(dest="subcommand")
     subparsers.required = True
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I can see a few issues with this design @grahamgower ---- let's discuss tomorrow.

grahamgower:
I'd be happy to collapse the `--msprime-model` and ``--msprime-change-model`` into one parameter and/or change the parameter names.

stdpopsim/cli.py
```diff
@@ -509,10 +509,27 @@ def stdpopsim_cli_parser():
...
+    msprime_parser.add_argument(
+                "--msprime-model",
+                default=msprime.supported_models[0],
+                choices=msprime.supported_models,
+                help="Specify the simulation model used by msprime. "
+                     "See msprime API documentation for details.")
+    msprime_parser.add_argument(
+                "--msprime-change-model",
+                default=[],
+                choices=msprime.supported_models, action="append",
```
jeromekelleher: I thought it would be possible to set ``nargs=2`` here and so something like 
```
--msprime-change-model 10 dtwf --msprime-change-model 100 hudson
```
would give an args list [("10", "dtwf"), ("100", "hudson")]. Is it possible to mix nargs with ``action="append"``?

grahamgower: Yeah, even the metavar works with nargs:
```
parser.add_argument('--msprime-change-model',
        metavar=("T","MODEL"),
        default=[], action="append", nargs=2)
```
But `type=some_func` doesn't, because argparse wants to call the converting function separately for each of the nargs. So we'll still need to have something like the following in stdpopsim_main() after calling parse_args() to check the types:

```
change_model = []
for t, model in args.msprime_change_model:
    try:
        t = float(t)
    except ValueError:
        parser.error(f"`{t}' is not a number")
    if model not in supported_models:
        parser.error(f"`{model}' is not a supported model")
    change_model.append((t, model))
args.msprime_change_model = change_model
```

grahamgower: Actually, `type=some_func` can be made to work through the magic of mutable default function parameters...
```
def tuple_of_time_model(arg, _expect_time=[True,]):
    if _expect_time[0]:
        try:
            arg = float(arg)
        except ValueError:
            parser.error(f"`{arg}' is not a number")
    else:
        if arg not in supported_models:
            parser.error(f"`{arg}' is not a supported model")
    _expect_time[0] = not _expect_time[0]
    return arg

parser.add_argument('--msprime-change-model',
        metavar=("T","MODEL"),
        type=tuple_of_time_model,
        default=[], action="append", nargs=2)
```

tests/test_engines.py
```diff
@@ -49,3 +51,52 @@ def test_abstract_base_class(self):
         e = stdpopsim.Engine()
         self.assertRaises(NotImplementedError, e.simulate)
         self.assertRaises(NotImplementedError, e.get_version)
+
+
+class TestMsprimeEngine(unittest.TestCase):
```
jeromekelleher: Nice tests, let's just move them into ``tests/test_cli.py`` since they are CLI tests.

Issue #397, opened 2020-02-03
[slim-engine-dev] Use conda package on OSX


jeromekelleher:
SLiM now has a [conda package](https://anaconda.org/conda-forge/slim). Update the TravisCI config to install it from conda-forge.

There won't be a Windows package for SLiM for a good long while, so let's ``unittest.skip`` the SLiM engine unit tests on windows.

jeromekelleher:
I don't think it's worth changing the CircleCI config to use the conda package, as this would get quite complicated and would basically require a rewrite of the Circle setup.

grahamgower:
Closed by #400.

PR #400, opened 2020-02-10
various slim-engine-dev improvements

grahamgower:


stdpopsim/slim_engine.py
```diff
@@ -214,34 +286,31 @@
...
-        mutation_rate,
-        generation_time,
-        population_configurations,
-        recombination_map,
-        migration_matrix,
-        demographic_events,
-        verbosity):
+        script_file, trees_file,
+        demographic_model, contig, samples,
+        Q, check_coalescence, verbosity):
```
jeromekelleher: Q isn't the most self explanatory parameter name, any chance of something more descriptive?

stdpopsim/slim_engine.py
```diff
@@ -214,34 +286,31 @@
...
     if len(recombination_map.get_positions()) > 2:
         raise Exception("recombination_map not supported")
 
-    pop_names = [pc.metadata["id"] for pc in population_configurations]
+    pop_names = [pc.metadata["id"] for pc in demographic_model.population_configurations]
+
+    # Reassign event times according to integral SLiM generations.
+    # This collapses the time deltas used in HomSap/AmericanAdmixture_4B11.
+    for event in demographic_model.demographic_events:
+        event.time = int(event.time / Q) * Q
```
jeromekelleher: Why not ``event.time = round(event.time)``? That seems fairly well specified and easy to document

jeromekelleher: Although I'd prefer if we errored out with non-integral times --- seems like we're storing up gotchas for later here, and the original source of the problem could be resolved.

tests/test_slim_engine.py
```diff
@@ -0,0 +1,89 @@
...
+import sys
+import unittest
+import tempfile
+
+import tskit
+
+import stdpopsim
+import stdpopsim.cli
+from . test_cli import capture_output
+
```
jeromekelleher: Probably simplest to skip the entire module if we're on Windows. I think unittest will allow that. If not, we can skip the classes.

tests/test_slim_engine.py
```diff
@@ -0,0 +1,89 @@
...
+
+class TestSLiM_API(unittest.TestCase):
```
jeromekelleher: The ``SLiM`` bits in these class names is redundant -- we're in the test_slim module.

stdpopsim/slim_engine.py
grahamgower: Simply rounding event.time doesn't quite cut it. This ``Q`` parameter rescales several things, one of which is the generation times (So 1 generation in SLiM will correspond to Q actual generations). I've rounded the value that actually needs to be an integer (``event.time/Q``)

tests/test_slim_engine.py
grahamgower: I kind of expected some tests to fail on windows because `_SLiMEngine.get_version()` calls out to the slim binary. Have you disabled the travisCI windows build? I didn't manage to find the logs for that.

In princial, it should be possible to generate a slim script without having a slim binary available. I think this should work fine, if get_version() is fixed to return `unknown` or something.

jeromekelleher: I think we should fail early on Windows, and let the user know it's just not supported because SLiM doesn't build on Windows (yet). So, there's no point in finessing the tests here IMO.

It's only Appveyor is on Windows. TravisCI should be doing OSX and LInux, but we haven't pulled in the conda package yet AFAIK

jeromekelleher: See #397 for discussion on conda package.

grahamgower: Ok. To be honest, I'm a little confused with all the different CI stuff. Is there something I need to include in the test_slim_engine.py file for the skipping or erroring out that you're suggesting? (Eg. test for windows at the top of the file and raise X). Or does this go somewhere else, like in the appveyor.yml?

test_engines.py currently calls get_engine() for all registered engines. So this test will need to be changed, or the get_version() should be amended to return something sensible if the slim binary can't be found.

jeromekelleher: Yes, it's a bit confusing all right. Might be easier to have a chat about this: you free tomorrow at any point?

jeromekelleher:
LGTM, thanks @grahamgower. A quick scan through the codecov report and it looks like you just need to cover some error conditions (specifying a recombination map, failing to find the slim exe, etc) and you're done.

grahamgower:
Yep, working on it now. :)

jeromekelleher:
Very nice, thanks @grahamgower. Merging.

PR #402, opened 2020-02-14
Add get_chunk(), to get a subset of the genome.

### LINKED ISSUE #401: Add option to use specific chrom region, and random region with given length
For CLI parameters, I think the length should be `-L` and the chromosomal region should should be `-r chr:from-to` to match samtools, etc. It should be possible to specify, e.g `-L 1000 -r chr1`.

grahamgower:
Should eventually close #401.

I\m opening this PR early to verify that the basic approach is sound.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -12,6 +12,7 @@ import tempfile
 import pathlib
 import shutil
 import functools
+import random
 
 import msprime
 import tskit
@@ -348,6 +349,7 @@ def add_simulate_species_parser(parser, species):
 
     if len(species.genome.chromosomes) > 1:
         choices = [chrom.id for chrom in species.genome.chromosomes]
+        choices.append("any")
         species_parser.add_argument(
             "-c", "--chromosome", choices=choices, metavar="", default=choices[0],
             help=(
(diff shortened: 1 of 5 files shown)


grahamgower:
Added CLI parameters and docstrings. I decided against using `chr:from-to` syntax on the CLI.

stdpopsim/cli.py
```diff
@@ -348,26 +349,52 @@ def add_simulate_species_parser(parser, species):
...
+        "-L", "--length", type=length, metavar="L", default=None,
+        help="Simulate a sequence of length L, "
+             "using the named chromosome's mutation and recombination rates. "
+             "The suffixes `k` or `m` are permitted, which indicate kilobases "
+             "and megabases respectively. "
+             "If no genomic coordinate is specified, it will be drawn randomly. "
+             "If the special chromosome `any` is requested, a chromosome will "
+             "also be drawn at random, with probability proportional to the "
+             "chromosome length.")
+    species_parser.add_argument(
```
jeromekelleher: Hmmm. There's a lot of good stuff in here, but I wonder if having a ``--start`` and ``--end`` coordinate as an argument would be simpler. Since people are going to be using this to slice up the genome into N bits, specifying a single slice is the most basic way of doing this, which someone can build more sophisticated stuff on top of later. The user already needs to know how long the chromosome is.

I would vote for just having a simple ``start`` and ``end`` coordinates specified as the first pass, and we could think about having things like ``--end=+1M`` later to make specifying less precise slices a bit easier.

jeromekelleher: We could also add things like ``--start=uniform`` or something later too, if we like.

stdpopsim/species.py
```diff
@@ -113,6 +115,77 @@ class Species(object):
     demographic_models = attr.ib(factory=list, kw_only=True)
     genetic_maps = attr.ib(factory=list, kw_only=True)
 
+    def get_chunk(self, length, chromosome=None, position=None, genetic_map=None):
```
jeromekelleher: I'm not sure we need this: why not add ``start`` and ``end`` parameters to the ``get_contig`` method?

stdpopsim/cli.py
```diff
@@ -405,9 +433,20 @@ def run_simulation(args):
                 "populations")
         samples = model.get_samples(*args.samples)
 
-        contig = species.get_contig(
-            args.chromosome, genetic_map=args.genetic_map,
-            length_multiplier=args.length_multiplier)
+        if args.chromosome == "any":
```
jeromekelleher: I don't see the need for randomly selecting a chromosome, seems like a complicated thing to implement. Are we choosing a chromosome randomly from those available? Or is it weighted by sequence lenght? Or is it weighted by overall recombination rate?

stdpopsim/species.py
```diff
@@ -113,6 +115,77 @@ class Species(object):
...
+
+        if position is None:
+            position = random.randrange(chrom.length - length)
+
+        if genetic_map is None:
+            gm = None
+            recomb_map = msprime.RecombinationMap.uniform_map(
+                    length, chrom.recombination_rate)
+        else:
+            # Extract a subset of the full genetic map.
```
jeromekelleher: Let's move this basic functionality to msprime: https://github.com/tskit-dev/msprime/issues/901

stdpopsim/cli.py
grahamgower: It's weighted by sequence length. If a genetic map is provided, and the user wants just a specific length, then sampling a chromosome at random makes sense to me. I prefer this behaviour to using `--length-multiplier`, partly because it's just easier to choose the length.

stdpopsim/species.py
grahamgower: Ok.

stdpopsim/cli.py
jeromekelleher: Yeah... it's not really coupled to the start, end functionality though, so I'd vote on pulling it out and doing in a follow up.

jeromekelleher:
I converted this to a draft to make it easy see what PRs are live for the next release @grahamgower - feel free to change back if you disagree.

jeromekelleher:
Is it simplest to close this @grahamgower or would you like to keep it open and update to use the msprime 1.0 apis?

grahamgower:
I think I'd prefer to keep this open for now, so I don't forget about it. As you say, it should be possible now to use the msprime 1.0 api.

andrewkern:
hey @grahamgower -- it would be great to get this feature in to stdpopsim at this point. want to revisit it? if you're busy i'm sure someone else could pick it up

grahamgower:
It does seem like all the pieces are now in place for this get finished up. I don't have the time/motivation to work through this right now, but I'd be happy to advise if someone else wanted to pick it up. The pr needs to be split into two:
- The ability to specify a genomic interval, so that the recombination map and annotations correspond to that interval.
- Randomly choosing a genomic interval of a given length. This is a little trickier because we now support masking of recombination maps, and we definitely want to respect those masks.

nspope:
I can take this over, if that's OK with you @grahamgower.

grahamgower:
That would be great, thanks @nspope!

PR #403, opened 2020-02-14
Add recapitation, and make it the default.

### LINKED ISSUE #259: [slim-engine-dev] Support recap/neutral-mutation-overlay with pyslim
There is some commented out code for this already, but some details need to be figured out. I think this should be the default, with a command line parameter to turn it off and do within-slim burn-in.

grahamgower:
Closes #259.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -47,6 +47,7 @@ import functools
 import itertools
 import collections
 import contextlib
+import random
 
 import stdpopsim
 import numpy as np
@@ -122,7 +123,7 @@ function (void)end(void) {
     }
 }
 
-// Check if burn-in has completed.
+// Check if burn in has completed.
 1 {
     if (!check_coalescence) {
         setup();
@@ -152,7 +153,7 @@ function (void)end(void) {
         sim.registerEarlyEvent(NULL, "{setup();}", g, g);
     } else {
         if (sim.generation == 1)
-            dbg("Waiting for burn-in...");
+            dbg("Waiting for burn in...");
         // Reschedule the current script block 10 generations hence.
         // XXX: find a less arbitrary generation interval.
         g = sim.generation + 10;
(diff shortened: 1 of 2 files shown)


grahamgower:
I don't see any RNG comments. Did this get lost somehow?

stdpopsim/slim_engine.py
```diff
@@ -625,41 +638,28 @@ def script_file_f():
 
             ts = pyslim.load(ts_file.name)
 
-        # random.seed(seed)
-        # s1, s2 = random.randint(1,2**32-1), random.randint(1,2**32-1)
+        if do_recap:
+            random.seed(seed)
```
jeromekelleher: Should create our own instance of RNG here rather than modifying the global state, ``rng = random.Random(seed)``.

jeromekelleher:
> I don't see any RNG comments. Did this get lost somehow?

Hmm, did get lost, ta.

stdpopsim/slim_engine.py
grahamgower: Done. I also changed `randint(1, 2**32-1)` to the much clearer `randrange(2**32)`.

stdpopsim/slim_engine.py
```diff
@@ -625,41 +638,28 @@ def script_file_f():
 
             ts = pyslim.load(ts_file.name)
 
-        # random.seed(seed)
-        # s1, s2 = random.randint(1,2**32-1), random.randint(1,2**32-1)
+        if do_recap:
+            rng = random.Random(seed)
+            s1, s2 = rng.randrange(2**32), rng.randrange(2**32)
```
jeromekelleher: randrange doesn't work I'm afraid, as 0 isn't a valid seed for msprime.

jeromekelleher: Well, randrange does work, you just need to put in ``rng.randrange(1, 2**32)``

PR #407, opened 2020-02-19
Warn when the user runs simulations of non-autosomes.

### LINKED ISSUE #384: Warn when user runs simulations of x chromosome
When the user runs a simulation of the X chromosome, we should emit a warning. Probably the best thing to do is to have an ``autosome`` boolean attribute we associate with each chromosome in the genome definitions, and then we ``warnings.warn()`` when a simulation is attempted where ``autosome`` is False.

This should be done reasonably quickly and be pushed into a release so we are not misleading users about the quality of our non-autosomal simulations.

See #383 for background.

grahamgower:
Currently affects:
CanFam chrX
DroMel chrX chrY chrM
PonPyg chrX
HomSap chrX chrY

Closes #384.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -404,7 +404,6 @@ def add_simulate_species_parser(parser, species):
                 f"Cannot sample from more than {model.num_sampling_populations} "
                 "populations")
         samples = model.get_samples(*args.samples)
-
         contig = species.get_contig(
             args.chromosome, genetic_map=args.genetic_map,
             length_multiplier=args.length_multiplier)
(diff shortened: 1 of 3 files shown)


grahamgower:
I didn't add a boolean field for each chromosome entry like suggested in #384. Given that this "fix" is going to be yanked out anyway, and we only have chromosomes with well defined names, this seemed far less invasive.

jeromekelleher:
Fair enough -- this works! One thing though, shouldn't we also worry about people using the Python API?

grahamgower:
Yep, good point. I moved the chomosome name check into Species.get_contig().

jeromekelleher:
Nice, good pragmatic fix. Merging.

PR #408, opened 2020-02-19
Update slim-engine-dev branch to master

grahamgower:


grahamgower:
Hmm... sphinx is trying to cross reference the Species.id attribute with the engines._Msprime.id field (and now complaining here because there is also a _SLiMEngine.id). Definitely not correct.

grahamgower:
I think this is good to go @jeromekelleher, unless you have an objection to the sphinx workaround (which would be very reasonable, because its horrible).

jeromekelleher:
LGTM, thanks @grahamgower. Merging.

Issue #369, opened 2020-01-02
Fix Appveyor builds


jeromekelleher:
Since the organisation name changed the Appveyor builds are broken. Will probably need to set them up again. Check the URL for badge in the README also.

PR #409, opened 2020-02-19
Add SLiM engine

grahamgower:


grahamgower:
You already opened a ticket about appveyor #369.

grahamgower:
(and this code will probably fail on windows, because there's still an outstanding use of NamedTempFile which should be changed to TempDir)

jeromekelleher:
> You already opened a ticket about appveyor #369.

Aha, so I did, thanks. Jeez, I've got too many things going on at once!

jeromekelleher:
Would you mind having a quick look here @andrewkern/@petrelharp please? I don't want to merge something this big without other approvals.

(@grahamgower, would you mind rebasing once more please? It's behind the tip of master again)

petrelharp:
Will do, but will be a couple of days.

jeromekelleher:
Bumping this for @petrelharp and @andrewkern

stdpopsim/slim_engine.py
```diff
@@ -0,0 +1,663 @@
...
+         * To obtain a burn-in genealogy drawn from the full distribution
+         * of possible genealogies, we may continue the simulation forwards
+         * until this ergodic process converges.  Unfortunately, it's not
+         * clear how many more generations are required for convergence,
+         * and this is likely dependent on demography, and non-neutral
+         * processes (if any).  10*N appears to be reasonable for a single
+         * neutrally evolving population with constant N.
+         */
+        N_max = max(N[0,0:(num_populations-1)]);
+        g = sim.generation + 10 * N_max;
```
andrewkern: I think we are going to want to have the burn-in period as a parameter available through the API, rather than be fixed in this way

stdpopsim/slim_engine.py
```diff
@@ -0,0 +1,663 @@
...
+                        "{dbg(self.source); " +
+                        "p"+j+".setMigrationRates("+k+", "+m+");}",
+                        g, g);
+                }
+            }
+        }
+    }
+
+    // Admixture pulses.
+    if (length(admixture_pulses) > 0 ) {
```
andrewkern: not sure that there _is_ reason for concern, but does the ordering of these event types (splits, then size changes, then admixture events) bias things in the simulation? are there edge cases we might worry about?

stdpopsim/slim_engine.py
```diff
@@ -0,0 +1,663 @@
...
+
+            slim_cmd.append(script_file.name)
+            stdout = subprocess.DEVNULL if verbosity == 0 else None
+            subprocess.check_call(slim_cmd, stdout=stdout)
+
+            ts = pyslim.load(ts_file.name)
+
+        if do_recap:
+            rng = random.Random(seed)
+            s1, s2 = rng.randrange(1, 2**32), rng.randrange(1, 2**32)
```
andrewkern: we are going to want to be able to set these seeds externally

stdpopsim/cli.py
```diff
@@ -535,6 +537,36 @@ def time_or_model(arg, _arg_is_time=[True, ], parser=top_parser):
...
+    slim_parser.add_argument(
+            "--slim-script", action="store_true", default=False,
+            help="Write script to stdout and exit without running SLiM.")
+    slim_parser.add_argument(
+            "--slim-scaling-factor", metavar="Q", default=10, type=float,
+            help="Rescale model parameters by Q to speed up simulation. "
+                 "See SLiM manual: `5.5 Rescaling population sizes to "
+                 "improve simulation performance`. "
+                 "[default=%(default)s].")
+    slim_parser.add_argument(
```
andrewkern: is it worth having both this option and the "no-burn-in" option available at the CLI? I think it might make for a simpler UX if we ditch the latter option.

andrewkern:
a couple bits I think should be polished here but looks great @grahamgower

stdpopsim/slim_engine.py
grahamgower: Yeah, that's a good idea. At the moment this code waits for all individuals to have a common ancestor after the start of the simulation (via `sim.treeSeqCoalesced()`), and *then* waits `10*N` more generations. So should we:
1. Wait until coalescence and have the burn-in period parameter change the `10*N`, or
2. remove the coalescence check altogether, potentially leaving genealogies without a single root. (Note this burn-in will only be used when recapitation is disabled).
3. Is it ok to have the parameter in units of `N` generations? (Where N is the size of the largest population that exists at the start of the simulation---no model currently has more than one population).

stdpopsim/slim_engine.py
grahamgower: I don't think the ordering currently matters. This just sets callbacks, which run in the order in which the callbacks are set up. But the effect of the callbacks are determined by SLiM's generation cycle, and this is not affected by the order of callbacks here.

The order of callbacks can and does matter in more complex cases involving saving/reloading simulation state (used with selective sweeps). We can certainly discuss event ordering again down the road when I submit a PR with selection bits.

grahamgower: Oh, there are some limitations on event ordering here. Subpopulation splits create new populations, and the new population must exist before they can be manipulated in other events (if any). But no bias here, I don't think.

stdpopsim/slim_engine.py
grahamgower: I'm not sure exactly what you mean here. The seeds are inherited from the `seed` parameter passed to `engine.simulate()`, so simulations should be deterministic when that parameter is used.

stdpopsim/cli.py
grahamgower: I guess this ties in with having a parameter for the burn-in time. How about changing `--slim-no-burnin` to be `--slim-burnin X`, where X is the length of the burn-in period in units of N generations? The default could be 10, and the user can set it to 0 to disable burn in. And then this is still separated from whether the coalescence check is done.

In the future case of having background selection, probably we'll want to disable the coalescence check and fix a longer burn in time. (I tried some simulations with a distribution of fitness effects, which didn't complete coalesce before I killed the jobs after a couple of hours). Does this kind of use case warrant leaving in both options?

grahamgower:
Thanks @andrewkern! I can work on the changes once we decide how to resolve the questions you raised about burn in.

stdpopsim/slim_engine.py
jeromekelleher: Yes, this is fine, resolving.

jeromekelleher:
Thanks for the review @andrewkern, some great points raised. My feeling is that we should push the burn-in etc question to an issue, discuss in detail there, and then follow up with a PR. This is a big patch we'd like to land so that development can go ahead on other parts of the code.

@grahamgower, I think we can make an executive decision here and merge. Are you happy to make issues for any of the unresolved comments so that we can follow them up?

grahamgower:
Sounds good @jeromekelleher. I'll open a follow up issue about burn in bits.

jeromekelleher:
> Sounds good @jeromekelleher. I'll open a follow up issue about burn in bits.

Shall we merge then?

grahamgower:
Yes please!

jeromekelleher:
Done! Congratulations and well done @grahamgower, this is a massive step forward and you've done an amazing job!

grahamgower:
Thanks @jeromekelleher! I couldn't have done it without your constant feedback!

PR #412, opened 2020-02-27
Add recombination map support in SLiM engine.

### LINKED ISSUE #262: [slim-engine-dev] Support recombination maps
Either:
- Output the positions/rates into compactly formatted eidos arrays, or,
- Write out a temp file which is read in to the SLiM script.

grahamgower:
Closes #262.

Recombination map positions and rates are stored as arrays in the
SLiM script. As SLiM includes each script in the provenance table,
whole chromosome recombination maps result in very long arrays in
the provenance table. Note also that SLiM versions prior to 3.3
print the entire recombination map when initializeRecombinationRate()
is called.

I'll rebase onto master once #409 is merged.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -48,6 +48,7 @@ import itertools
 import collections
 import contextlib
 import random
+import textwrap
 
 import stdpopsim
 import numpy as np
@@ -65,10 +66,14 @@ initialize() {
 
     defineConstant("generation_time", $generation_time);
     defineConstant("mutation_rate", Q * $mutation_rate);
-    defineConstant("recombination_rate", (1-(1-2*$recombination_rate)^Q)/2);
     defineConstant("chromosome_length", $chromosome_length);
     defineConstant("trees_file", "$trees_file");
     defineConstant("check_coalescence", $check_coalescence);
+
+    _recombination_rates = $recombination_rates;
+    _recombination_ends = $recombination_ends;
+    defineConstant("recombination_rates", (1-(1-2*_recombination_rates)^Q)/2);
+    defineConstant("recombination_ends", _recombination_ends);
 """
 
 _slim_lower = """
@@ -79,7 +84,7 @@ _slim_lower = """
     initializeMutationType("m1", 0.5, "f", 0);
     initializeGenomicElementType("g1", m1, 1.0);
     initializeGenomicElement(g1, 0, chromosome_length-1);
-    initializeRecombinationRate(recombination_rate);
+    initializeRecombinationRate(recombination_rates, recombination_ends);
 }
 
 function (void)dbg(string$ s, [integer$ debug_level = 2]) {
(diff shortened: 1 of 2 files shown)


stdpopsim/slim_engine.py
```diff
@@ -288,15 +292,44 @@
 """
 
 
+def join_wrap(a, delim, wrap=80, newline="\n"):
```
jeromekelleher: Can't we do this with [textwrap](https://docs.python.org/3.6/library/textwrap.html)?

grahamgower: Oh, cool. I never knew that existed.

jeromekelleher:
Looks good, thanks @grahamgower.

Re the provenance, I think we should drop the intermediate provenances in general, as they are pointless. They're supposed to allow us reproduce the simulation, which the top-level stdpopsim provenance record will do quite nicely. There's no point in storing these large recombination maps as JSON in the provenance record (which would also be a problem in msprime, if it actually stored provenance properly).

My take would be that the simulators should have an option "store_provenance=False" which prevents them from writing provenance records. However, the infrastructure isn't in place for this at the moment, so unless the records are very big I wouldn't worry too much about it.

grahamgower:
I seem to have messed something up here and dropped the changes that use textwrap. Will let you know once I figure it out.

grahamgower:
Ok, the diff looks right now. Sqaushed/rebased and ready for merging now @jeromekelleher.

jeromekelleher:
I've requested a review from @petrelharp here @grahamgower. I know he's under the weather right now though, so let's just merge in tomorrow if he doesn't shout out. We can always change stuff later.

stdpopsim/slim_engine.py
```diff
@@ -65,10 +66,14 @@
...
     defineConstant("generation_time", $generation_time);
     defineConstant("mutation_rate", Q * $mutation_rate);
-    defineConstant("recombination_rate", (1-(1-2*$recombination_rate)^Q)/2);
     defineConstant("chromosome_length", $chromosome_length);
     defineConstant("trees_file", "$trees_file");
     defineConstant("check_coalescence", $check_coalescence);
+
+    _recombination_rates = $recombination_rates;
+    _recombination_ends = $recombination_ends;
+    defineConstant("recombination_rates", (1-(1-2*_recombination_rates)^Q)/2);
```
petrelharp: We're introducing some floating-point error here (since x != (1 - (1 - x))). I suppose this is harmless but a little annoying.

petrelharp:
Looks good to me!

stdpopsim/slim_engine.py
petrelharp: I also don't see what the factor of 2 is doing?

grahamgower: In ignorance, I copied this from the SLiM manual section 5.5 (p. 103 in the version I have), where one Peter Ralph is credited. :P

jeromekelleher:
Looks like this is good to merge, but I'm not sure if the conversation above about the 2 was resolved. @petrelharp, can you merge this if you're happy with it please? @grahamgower, can you rebase once we get the other PR merged please, so it's ready for Peter to push the button?

stdpopsim/slim_engine.py
petrelharp: > where one Peter Ralph is credited. :P
Hm, dubious. =)
> This formula is based upon the probability that a binomial draw will be odd
Oh, right.  Ok, all good.

petrelharp:
Great, thanks!!!

PR #421, opened 2020-03-18
Update appveyor config

### LINKED ISSUE #369: Fix Appveyor builds

jeromekelleher:
Closes #369

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -660,4 +660,6 @@ class _SLiMEngine(stdpopsim.Engine):
         return ts
 
 
-stdpopsim.register_engine(_SLiMEngine())
+# SLiM does not currently work on Windows.
+if sys.platform != "win32":
+    stdpopsim.register_engine(_SLiMEngine())
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,9 @@ install:
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
   - source activate test-environment
   - conda install --yes --file=requirements/conda-minimal.txt
+  # SLiM is on conda-forge, but it's not available for windows so we don't
+  # include it in the 'minimal' requirements.
+  - conda install --yes slim
   # pyslim is not yet on conda-forge.
   - pip install pyslim
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # stdpopsim
 
-[![Docs](https://readthedocs.org/projects/stdpopsim/badge/?version=latest&style=flat)](https://stdpopsim.readthedocs.io/en/latest/)[![CircleCI](https://circleci.com/gh/popgensims/stdpopsim.svg?style=svg)](https://circleci.com/gh/popgensims/stdpopsim)[![codecov](https://codecov.io/gh/popgensims/stdpopsim/branch/master/graph/badge.svg)](https://codecov.io/gh/popgensims/stdpopsim)[![AppVeyor](https://ci.appveyor.com/api/projects/status/3968mxfylqyaad4p/branch/master?svg=true)](https://ci.appveyor.com/project/jeromekelleher/stdpopsim/branch/master)[![TravisCI](https://travis-ci.org/popgensims/stdpopsim.svg?branch=master)](https://travis-ci.org/popgensims/stdpopsim)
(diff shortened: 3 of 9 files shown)


jeromekelleher:
Looks like some transient problem with appveyor/conda. Will try again later today.

jeromekelleher:
Still not working, but it definitely looks like a problem on the appveyor side. Will try again tomorrow.

grahamgower:
I think the appveyor problem is described in the conda troubleshooting guide.
https://conda.io/projects/conda/en/latest/user-guide/troubleshooting.html#ssl-connection-errors

jeromekelleher:
> I think the appveyor problem is described in the conda troubleshooting guide.
> https://conda.io/projects/conda/en/latest/user-guide/troubleshooting.html#ssl-connection-errors

Thanks @grahamgower, good spot. I tried adding `` conda config --set ssl_verify false`` but it didn't help. The weird thing here is that basically exactly the same Appveyor config is working fine on tskit...

grahamgower:
I don't think its a lack of appropriate certificates (which might be fixed by setting ssl_verify=false), it looks more like the openssl libs are not being found. It might need some %PATH% frobbing or a magic `conda activate` command.

I suppose there's some version differences between the windows CI here and for tskit?

jeromekelleher:
> I don't think its a lack of appropriate certificates (which might be fixed by setting ssl_verify=false), it looks more like the openssl libs are not being found. It might need some %PATH% frobbing or a magic conda activate command.

Yes, you're right. I'll have another look.

> I suppose there's some version differences between the windows CI here and for tskit?

Must be, I guess.

jeromekelleher:
It worked! Thanks @grahamgower, you saved me a lot of time there.

Now we just need to fixup the tests broken on windows. I'll just skip any that seem tricky and fix easy ones.

jeromekelleher:
@grahamgower, I made some minimal changes to get this to work on Windows. Only potentially controversial thing is to not add slim as an engine on Windows. Do you see any potential issues with this?

grahamgower:
The command line options for slim will still be shown. And all the slim tests need to be skipped.

I think getting the slim engine to work on windows will not be a lot of work though, it probably just needs (1) to use a temp directory instead of temp files, and maybe (2) get_version() should not fail if the slim binary is not available. Of course, developing for windows without a windows machine is a pain.

jeromekelleher:
> The command line options for slim will still be shown. And all the slim tests need to be skipped.

Good call. How does the latest update look?

jeromekelleher:
> I think getting the slim engine to work on windows will not be a lot of work though, it probably just needs (1) to use a temp directory instead of temp files, and maybe (2) get_version() should not fail if the slim binary is not available. Of course, developing for windows without a windows machine is a pain.

We'd need SLiM to work on windows first though. Until this is supported upstream there's no point in us putting effort in. (I've been nagging Ben about this, but I think he needs actual windows users to start making their voices heard. No reason SLiM couldn't work natively on windows.)

jeromekelleher:
Codecov reports are misleading here, can be ignored.

stdpopsim/slim_engine.py
```diff
@@ -660,4 +660,6 @@ def script_file_f():
         return ts
 
 
-stdpopsim.register_engine(_SLiMEngine())
+# SLiM does not currently work on Windows.
+if sys.platform != "win32":
```
grahamgower: I suppose this should be `sys.platform.startswith("win")` to catch 64 bit windows too?

grahamgower:
LGTM. Just a query about checking for "win32". I assume "win64" exists also?

grahamgower:
It seems I'm fretting needlessly. sys.platform is always "win32" on windows, even on non-intel architecture ports.
https://github.com/python/cpython/blob/9052f7a41b90f2d34011c8da68f9a4facebc8a97/PC/pyconfig.h#L282

jeromekelleher:
Yeah, it's always win32 according to the docs and [here](https://stackoverflow.com/questions/2144748/is-it-safe-to-use-sys-platform-win32-check-on-64-bit-python)

andrewkern:
i'm running into a similar issue on #560.... why are we trying to support windows at all?

jeromekelleher:
> i'm running into a similar issue on #560.... why are we trying to support windows at all?

Because most computers run Windows. It's elitist and exclusive to assume that people either have enough money to own a mac or are tech savvy enough to run linux. Windows support isn't that hard, and it hugely opens up the potential user base.

Issue #304, opened 2019-12-09
Refactor QC module


jeromekelleher:
The QC module currently uses subclasses of Model to define the relevant demographic models. This is quite boilerplatey. Refactor to use something similar to the function approach in #275.

We should wait on the model ID scheme being finalised first in #122 

Not necessary for shipping 0.1.0 since it's not part of the external API or behaviour.

PR #429, opened 2020-03-19
Added function to check sampling time equality

ndukler:
Applies test for sampling times when doing model QC and updates models to be in compliance.

I'm not sure if setting the default value for the list of populations is working as I don't really have a great grasp of decorators. Could someone look at the `DemographicModels` class?

### DIFF
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -69,6 +69,43 @@ def verify_population_configurations_equal(
         raise UnequalModelsError("Growth rates differ")
 
 
+def sampling_times_equal(
+        populations1, populations2, rtol=DEFAULT_RTOL, atol=DEFAULT_ATOL):
+    """
+    Returns True if the sampling times for the specified lists of Populations
+    objects are equal to the specified tolerances.
+    """
+    try:
+        verify_sampling_times_equal(
+            populations1, populations2, rtol=rtol, atol=atol)
+        return True
+    except UnequalModelsError:
+        return False
+
+
+def verify_sampling_times_equal(
+        populations1, populations2, rtol=DEFAULT_RTOL, atol=DEFAULT_ATOL):
+    """
+    Check the the sampling times for lists of population objects are the same
+    """
+    # Retrieve sampling times
+    sampling_times1 = np.array([p.sampling_time for p in populations1])
+    sampling_times2 = np.array([p.sampling_time for p in populations2])
+    # Check for equal vector length
+    if len(sampling_times1) != len(sampling_times2):
+        raise UnequalModelsError("Different numbers of populations")
+    # Get indicies where None values present and compare
+    s1_none = np.where(sampling_times1 == None) # noqa
+    s2_none = np.where(sampling_times2 == None) # noqa
+    if not np.array_equal(s1_none, s2_none):
+        raise UnequalModelsError("None-valued sampling times differ")
+    # Subset out only non-None values and cast to float so they can be compared
+    s1_subset = sampling_times1[sampling_times1 != np.array(None)].astype(float)
+    s2_subset = sampling_times2[sampling_times2 != np.array(None)].astype(float)
+    if not np.allclose(s1_subset, s2_subset, rtol=rtol, atol=atol):
+        raise UnequalModelsError("Positive real-valued sampling times differ")
+
+
 def demographic_events_equal(
         events1, events2, num_populations, rtol=DEFAULT_RTOL, atol=DEFAULT_ATOL):
     """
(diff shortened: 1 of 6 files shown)


grahamgower:
Hey @ndukler. I'm a little worried that population ids aren't unique here. (It seems none of the tests complained, so maybe a new one should be added). In any event, could you do something like the following instead?
```
populations = [Population(f"pop{i}", f"Population {i}") for i in range(npops)]
```

FYI, the decorator usage looks right to me. The attr.ib function also provides a "factory" parameter for this use case, but I don't have a strong opinion about which approach should be used.

ndukler:
@grahamgower I suppose I could do that, but the sampling time, needs to be set separately for each population (at least if it's non-zero). I also don't want to over-test and we've already tested a bunch of things that guarantee that the models are exactly equal even if a few cosmetic strings are not. As such I don't think it's necessary to name the populations accurately as we already know they behave and are sampled in equivalently to the production models.

ndukler:
Also, just a thought, but the accuracy of the strings is really addressed by the human part of this equation which I don't really see as being testable.

grahamgower:
In the QC models, I tend to agree. However for default attributes, I think it makes sense to keep IDs unique.

ndukler:
Ah, I misunderstood. Sure!

stdpopsim/models.py
```diff
@@ -69,6 +69,43 @@ def verify_population_configurations_equal(
...
+    Check the the sampling times for lists of population objects are the same
+    """
+    # Retrieve sampling times
+    sampling_times1 = np.array([p.sampling_time for p in populations1])
+    sampling_times2 = np.array([p.sampling_time for p in populations2])
+    # Check for equal vector length
+    if len(sampling_times1) != len(sampling_times2):
+        raise UnequalModelsError("Different numbers of populations")
+    # Get indicies where None values present and compare
+    s1_none = np.where(sampling_times1 == None) # noqa
```
jeromekelleher: I'm surprised numpy arrays can store None, but it turns out it's using dtype=object arrays. Learn something new every day!

stdpopsim/models.py
```diff
@@ -189,7 +226,13 @@ class DemographicModel(object):
     @migration_matrix.default
     def _migration_matrix_default(self):
         npops = len(self.population_configurations)
-        return [[0]*npops]*npops
+        populations = [Population(f"pop{i}", f"Population {i}") for i in range(npops)]
+        return populations
```
grahamgower, line 230: This bit needs to go under _populations_default(), no?

grahamgower:
Hey @ndukler, any chance you can get that tiny last thing fixed so this can get merged? I'd like to do some more extensive refactoring of the QC code (#304), and don't want to make you deal with the conflicts. Cheers!

ndukler:
Yea, I'll do it by end of day tomorrow if that's fast enough.

ndukler:
Just short on test coverage now ... I'll add some in a bit

ndukler:
@jeromekelleher  @grahamgower Honestly I have no idea how to test the remaining two lines it's complaining about or when they'd ever be called. It's just the default value for the `populations` property of the `DemographicModel` class. Basically this line:
```
populations = [Population(f"pop{i}", f"Population {i}") for i in range(npops)]
``` 

Any ideas?

ndukler:
Nevermind, figured it out along with how `attrs` works. It's still pretty awkward to test as `populations` is supposed to be required and `population_configurations` is not so basing the default on something that is not required is going to screw up all manner of things as it will then become required. Honestly I think we just have to scrap the default for `populations` as it seems like it will cause problems. Although it is still required by the `migration_matrix` default so ¯\_(ツ)_/¯ .  Also after experimentation I don't really understand how `attrs` determines required keywords. I'm going on too long now but this is just to say I could use some input.

grahamgower:
Looks right to me, @ndukler. The `population_configurations = attr.ib(factory=list)` line means that if no population_configurations parameter is provided when making a `DemographicModel(...)`, then a default value will be constructed by calling the factory function (in this case, the default will be an empty list). So the functions providing default values for `populations` and `migration_matrix` should work as intended, even when no population_configurations are specified. The `_populations_default()` method can be tested by making a DemographicModel instance with no populations specified --- we probably want two of these, one with a population_configuration and one without.

ndukler:
It actually won't because of the order of `attrs.ib()` declarations (`population_configurations` needs to be before `populations`). When I fix that however I end up with an error that `population_configurations` is required by `__init__`. I appreciate that `attrs` is supposed to remove a lot of class boilerplate but it's a little too magical for me as a novice :)

ndukler:
Ok, so I've set it all up so it inherits the population configuration metadata if it's there. The code is a little messier but I think it has to be that way for now. Just going to add the tests and push.

ndukler:
Ready to merge @jeromekelleher if you approve of the changes

jeromekelleher:
LGTM. @grahamgower, can you merge if you're happy please?

grahamgower:
Thanks @ndukler! Sorry for the attrs run around.

PR #431, opened 2020-03-19
Validation script.

grahamgower:
Supports multiprocessing, because some SLiM simulations will be slow. Running the simulations can be easily done on a multicore server and the stats plotted on a workstation. Now it just needs more stats and simulation scenarios.

Output from `./validation.py -j 8 msprime_basic1 slim_basic1` is attached (which took about 40s to run on my laptop).
[msprime_basic1__slim_basic1.pdf](https://github.com/popsim-consortium/stdpopsim/files/4356655/msprime_basic1__slim_basic1.pdf)

### DIFF
--- a/validation.py
+++ b/validation.py
@@ -1,0 +1,517 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import pathlib
+import argparse
+import random
+import functools
+import itertools
+import concurrent.futures
+import time
+
+import numpy as np
+import tskit
+# Work around an issue on systems with large numbers of cores.
+# https://github.com/cggh/scikit-allel/issues/285
+os.environ["NUMEXPR_MAX_THREADS"] = f"{os.cpu_count()}"  # NOQA
+import allel
+
+import matplotlib
+matplotlib.use('Agg')  # NOQA # don't try to use $DISPLAY
+import matplotlib.pyplot as plt
+from matplotlib.backends.backend_pdf import PdfPages
+
+import stdpopsim
+import stdpopsim.cli
+
+
+def warning(msg):
+    """
+    Print a warning, with output less ugly than that of warnings.warn().
+    """
+    print(f"WARNING: {msg}", file=sys.stderr)
+
+
+#
+# Simulation functions.
+#
+
+
+def do_cmd(cmd, out_dir, seed):
+    cmd = cmd.split()
+    assert "-o" not in cmd and "--output" not in cmd
+    assert "-s" not in cmd and "--seed" not in cmd
+    out_file = out_dir / f"{seed}.trees"
+    full_cmd = cmd + f" -q -o {out_file} -s {seed}".split()
+    t0 = time.perf_counter()
+    stdpopsim.cli.stdpopsim_main(full_cmd)
+    t1 = time.perf_counter()
+    assert os.path.exists(out_file)
+    return out_file, t1 - t0
+
+
+def _onepop_PC(engine_id, out_dir, seed, N0=1000, *size_changes, **sim_kwargs):
+    species = stdpopsim.get_species("CanFam")
+    contig = species.get_contig("chr35", length_multiplier=0.01)  # ~265 kb
+    model = stdpopsim.PiecewiseConstantSize(N0=N0, *size_changes)
+    model.generation_time = species.generation_time
+    samples = model.get_samples(100)
+    engine = stdpopsim.get_engine(engine_id)
+    t0 = time.perf_counter()
+    ts = engine.simulate(model, contig, samples, seed=seed, **sim_kwargs)
+    t1 = time.perf_counter()
+    out_file = out_dir / f"{seed}.trees"
+    ts.dump(out_file)
+    return out_file, t1 - t0
+
+
+def onepop_constantN_msprime1(out_dir, seed):
+    """
+    Single population with constant population size.
+    """
+    return _onepop_PC("msprime", out_dir, seed)
+
+
+def onepop_constantN_slim1(out_dir, seed):
+    """
+    Single population with constant population size.
+    There are no demographic_events, so SLiM exits immediately, and
+    tree sequences are constructed via recapitation.
+    """
+    return _onepop_PC("slim", out_dir, seed)
+
+
+def onepop_constantN_slim2(out_dir, seed):
+    """
+    Single population with constant population size.
+    No recapitation. No scaling.
+    """
+    return _onepop_PC(
+            "slim", out_dir, seed,
+            slim_no_recapitation=True, slim_scaling_factor=1)
+
+
+def onepop_constantN_slim3(out_dir, seed):
+    """
+    Single population with constant population size.
+    No recapitation. Time is rescaled by a factor of 10.
+    """
+    return _onepop_PC(
+            "slim", out_dir, seed,
+            slim_no_recapitation=True, slim_scaling_factor=10)
+
+
+def _twopop_IM(
+        engine_id, out_dir, seed,
+        NA=1000, N1=500, N2=5000, T=1000, M12=0, M21=0,
+        **sim_kwargs):
+    species = stdpopsim.get_species("AraTha")
+    contig = species.get_contig("chr5", length_multiplier=0.01)  # ~270 kb
+    model = stdpopsim.IsolationWithMigration(
+            NA=NA, N1=N1, N2=N2, T=T, M12=M12, M21=M21)
+    model.generation_time = species.generation_time
+    samples = model.get_samples(50, 50, 0)
+    engine = stdpopsim.get_engine(engine_id)
+    t0 = time.perf_counter()
+    ts = engine.simulate(model, contig, samples, seed=seed, **sim_kwargs)
+    t1 = time.perf_counter()
+    out_file = out_dir / f"{seed}.trees"
+    ts.dump(out_file)
+    return out_file, t1 - t0
+
+
+def twopop_no_migration_msprime1(out_dir, seed):
+    """
+    Two populations with different sizes and no migrations.
+    """
+    return _twopop_IM("msprime", out_dir, seed)
+
+
+def twopop_no_migration_slim1(out_dir, seed):
+    """
+    Two populations with different sizes and no migrations.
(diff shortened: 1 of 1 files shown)


grahamgower:
And we have a problem. :(
The attached comparison is between msprime and slim without recapitation. I.e.
```
def msprime_basic1(args):
    cmd = "-e msprime HomSap -c chr1 -l 0.001 100"
    return do_cmd(cmd, *args)

def slim_basic2(args):
    cmd = "-e slim --slim-no-recapitation HomSap -c chr1 -l 0.001 100"
    return do_cmd(cmd, *args)
```

[msprime_basic1__slim_basic2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4358771/msprime_basic1__slim_basic2.pdf)

grahamgower:
Times are all out by a factor of 10, which is the default scaling factor. Makes sense I suppose, given times are in units of generations.

andrewkern:
okay this is super interesting @grahamgower and exactly what we should try to present in figures in the paper. pinging @jradrion here, as he is thinking about such validation figs. I think the immediate thing is to change the scalars here compare stats as you have already started to.

I vote that we merge this PR and start to trouble shoot the validation. @jeromekelleher that ok?

grahamgower:
Thanks for the vote of confidence @andrewkern! I've squashed and rebased to make it less ugly.

grahamgower:
Comparisons involving msprime, and slim with no recapitation and no rescaling of time, etc. In the last file, the density of segregating sites is too few to make sensible LD plots.
[onepop_constantN_msprime1__onepop_constantN_slim2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4368026/onepop_constantN_msprime1__onepop_constantN_slim2.pdf)
[twopop_no_migration_msprime1__twopop_no_migration_slim2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4368028/twopop_no_migration_msprime1__twopop_no_migration_slim2.pdf)
[twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4368027/twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim2.pdf)

validation.py
```diff
@@ -0,0 +1,492 @@
...
+    out_file = os.path.join(out_dir, f"{seed}.trees")
+    full_cmd = cmd + f" -q -o {out_file} -s {seed}".split()
+    t0 = time.time()
+    stdpopsim.cli.stdpopsim_main(full_cmd)
+    t1 = time.time()
+    assert os.path.exists(out_file)
+    return out_file, t1 - t0
+
+
+"""
```
jeromekelleher: Slightly odd to use a docstring here, and it might have unintended consequences (I'm not sure what this docstring gets attached to). Best use a comment (banner lines are fine IMO, if you want to split up the file visually)

validation.py
```diff
@@ -0,0 +1,492 @@
...
+    print(f"WARNING: {msg}", file=sys.stderr)
+
+
+def do_cmd(cmd, out_dir, seed):
+    cmd = cmd.split()
+    assert "-o" not in cmd and "--output" not in cmd
+    assert "-s" not in cmd and "--seed" not in cmd
+    out_file = os.path.join(out_dir, f"{seed}.trees")
+    full_cmd = cmd + f" -q -o {out_file} -s {seed}".split()
+    t0 = time.time()
```
jeromekelleher: This is wall time which isn't the right thing I think. We either want to use ``perf_counter`` or to wrap the command we're calling using ``/usr/bin/time`` to do the work for us. See [here](https://github.com/mcveanlab/treeseq-inference/blob/master/src/evaluation.py#L377) for an example of using time.

validation.py
```diff
@@ -0,0 +1,492 @@
...
+        return None
+    stats = dict()
+    for j, k in itertools.combinations(pops, 2):
+        sample_sets = [ts.samples(j), ts.samples(k)]
+        stats[f"$f_2$[{j},{k}]"] = ts.f2(sample_sets)
+        stats[f"$Y_2$[{j},{k}]"] = ts.Y2(sample_sets)
+    return stats
+
+
+def linkage_disequilibrium(ts, span=int(1e5), bins=50, min_num_sites=20):
```
jeromekelleher: ``10**5`` rather than ``int(1e5)``?

validation.py
```diff
@@ -0,0 +1,492 @@
...
+stats_functions = {f.__name__: f for f in _stats_functions}
+simulation_functions = {f.__name__: f for f in _simulation_functions}
+default_comparisons = [(t[0].__name__, t[1].__name__)
+                       for t in _default_comparisons]
+
+
+def do_simulations(rng, output_folder, num_replicates, num_procs, sim_keys):
+    files = dict()
+    times = dict()
+    with multiprocessing.Pool(num_procs) as pool:
```
jeromekelleher: Just FYI, the cool kids are using https://docs.python.org/3.5/library/concurrent.futures.html now rather than multiprocessing.

validation.py
```diff
@@ -0,0 +1,492 @@
...
+            # dump timing info to a file
+            np.savetxt(os.path.join(out_dir, "times.txt"), times[key])
+    return files, times
+
+
+def find_simulations(output_folder, sim_keys):
+    files = dict()
+    times = dict()
+    for key in sim_keys:
+        out_dir = os.path.join(output_folder, "trees", key)
```
jeromekelleher: Pathlib would make this (and the previous) function a bit shorter and easier to read.

validation.py
```diff
@@ -0,0 +1,492 @@
...
+                # using arbitrary folders that are in the right place.
+                warning(f"unknown scenario key ``{key}''")
+            else:
+                parser.error(f"unknown scenario key ``{key}''")
+
+    return args
+
+
+if __name__ == "__main__":
+    args = parse_args()
```
jeromekelleher: I think it's better to split the job up explicitly into "run" and "plot" (or some other better names). The running bit will end up taking a lot of time, and you want to be able to do the two things independently. It's especially useful to be able to run plots for simulations that are ongoing (so it's good to always flush out any results that have been calculated when you can).

validation.py
grahamgower: I'll change it to use perf_counter.

YFI, I got bitten by this (supposedly fixed) GNU time bug last year. https://groups.google.com/forum/#!topic/gnu.utils.help/u1MOsHL4bhg

The fix wasn't included in a release until late 2017 (according to http://ftp.gnu.org/gnu/time/), so there are likely many systems still affected. Unhelpfully, `/usr/bin/time --version` does not report a proper version either, so the only way to check if the bug is present is to allocate a specific amount of memory and see if the /usr/bin/time output is sensible.

jeromekelleher: > YFI, I got bitten by this (supposedly fixed) GNU time bug last year. https://groups.google.com/forum/#!topic/gnu.utils.help/u1MOsHL4bhg
> 
> The fix wasn't included in a release until late 2017 (according to http://ftp.gnu.org/gnu/time/), so there are likely many systems still affected. Unhelpfully, /usr/bin/time --version does not report a proper version either, so the only way to check if the bug is present is to allocate a specific amount of memory and see if the /usr/bin/time output is sensible.

Huh, that is interesting. Looks like a bunch of memory usages I've reported will be off by a factor of 4 then. Oh well.

ps. I looked up [YFI](https://www.internetslang.com/YFI-meaning-definition.asp) to see if this was a new one I hadn't heard of. Interesting slip of the keyboard if it was unintended!

grahamgower: RE: YFI; definitely unintended! :)

grahamgower:
I've made lots more changes, mostly related to your review, @jeromekelleher. I changed to concurrent.futures, and have interleaved plotting with simulating, so plots are produced when the data is ready. It still uses `map`, rather than submit/futures, so there's probably room for more efficiency by avoiding the current pool synchronisation due to `list(executor.map(...))` each time in the loop. I'm not inclined to try and fix that though.

I dropped the population size for the constantN model comparisons, from human values to 1000. This makes slim without recapitation and without scaling finish in a reasonable amount of time (30 seconds, compared to 6000 seconds before).

Once again, plots are attached. I'm considering generating a simple report that can be uploaded to the wiki. Such a report would include docstrings for the models and stats being plotted as a figure caption.
[onepop_constantN_msprime1__onepop_constantN_slim1.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369629/onepop_constantN_msprime1__onepop_constantN_slim1.pdf)
[onepop_constantN_msprime1__onepop_constantN_slim2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369630/onepop_constantN_msprime1__onepop_constantN_slim2.pdf)
[onepop_constantN_msprime1__onepop_constantN_slim3.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369631/onepop_constantN_msprime1__onepop_constantN_slim3.pdf)
[twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim1.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369633/twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim1.pdf)
[twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369634/twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim2.pdf)
[twopop_no_migration_msprime1__twopop_no_migration_slim1.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369635/twopop_no_migration_msprime1__twopop_no_migration_slim1.pdf)
[twopop_no_migration_msprime1__twopop_no_migration_slim2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4369636/twopop_no_migration_msprime1__twopop_no_migration_slim2.pdf)

jeromekelleher:
Looks great @grahamgower, shall we merge?

grahamgower:
Yes please!

jeromekelleher:
Awesome, can you squash the commits please?

grahamgower:
Yep, done.

andrewkern:
awesome

PR #432, opened 2020-03-20
Adjust ts node times when using SLiM scaling factor.

grahamgower:
Fixes the first problem found in validation. See #431.

It seems SLiM doesn't use the migration table (yet?), so this just updates node times.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -638,6 +638,14 @@ class _SLiMEngine(stdpopsim.Engine):
 
             ts = pyslim.load(ts_file.name)
 
+        # Node times come from SLiM generation numbers, which may have been
+        # divided by a scaling factor for computational tractibility.
+        tables = ts.dump_tables()
+        for table in (tables.nodes, tables.migrations):
+            table.time *= slim_scaling_factor
+        ts = pyslim.SlimTreeSequence.load_tables(tables)
+        ts.slim_generation *= slim_scaling_factor
+
         if do_recap:
             rng = random.Random(seed)
             s1, s2 = rng.randrange(1, 2**32), rng.randrange(1, 2**32)


jeromekelleher:
LGTM, will need some people who know about forward simulation to confirm this is the right approach.

In the mean time, a unit test would be nice.

jeromekelleher:
Not sure exactly how to unit test it though. Might need some creative mocking.

grahamgower:
Yeah, the unit test seemed awkward so I'm trying not to think about it. :)

grahamgower:
The previous fix was wrong. I've changed it to multiply node times before recapitation, as doing it afterwards makes all recapitated node times wrong. An alternative would be to scale everything during recapitation and mutation overlay, but I think this is simpler.

And I've avoided the unit test by removing the conditional branch. This is the default code path anyway, and if `slim_scaling_factor == 1`, substantially more time will be spent within slim anyway, so there's negligible benefit to avoiding this code path.

Two validation comparisons are attached, for a single population simulation with 100 individuals. The first shows msprime versus slim with recapitation (`slim_basic1`). There are no demographic events here, so slim exits immediately and everything is driven by the recapitation/mutation-overlay. The second comparison, msprime versus `slim_basic2` is with no recapitation, and a slim scaling factor of 10. There's something funny going on with `num_sites` in slim_basic2, but ts.segregating_sites() looks normal, so I suspect it's unrelated to the present issue.
[msprime_basic1__slim_basic1.pdf](https://github.com/popsim-consortium/stdpopsim/files/4362603/msprime_basic1__slim_basic1.pdf)
[msprime_basic1__slim_basic2.pdf](https://github.com/popsim-consortium/stdpopsim/files/4362604/msprime_basic1__slim_basic2.pdf)

jeromekelleher:
Great stuff, thanks @grahamgower. I love the comparison plots. I'm going to merge this, but let's follow up with an issue to track the num_sites thing.

jeromekelleher:
Also, @petrelharp, note the tweaking of the ``ts.slim_generation *= slim_scaling_factor`` after it's been read in by pyslim.

petrelharp:
Yeah, could someone explain? This seems wierd but I am sure you have your reasons?

petrelharp:
Never mind, I see.

PR #436, opened 2020-03-23
Multiply migration rates by the scaling factor.

### LINKED ISSUE #435: SLiM simulations with migration and time rescaling don't match msprime
See validation plots: https://github.com/popsim-consortium/stdpopsim/pull/431#issuecomment-602589125

`twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim1.pdf` is a slim simulation with recapitation and time scaling by a factor of 10.
`twopop_asymmetric_migration_msprime1__twopop_asymmetric_migration_slim2.pdf` is a slim simulation without recaptitation and without time scaling.

slim2 looks good, slim1 looks off.

grahamgower:
Note that only the rate for continuous migrations is multiplied,
whereas the pulse migrations are left unscaled.

Closes #435.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -115,7 +115,7 @@ function (void)end(void) {
             if (j==k | N[i,j] == 0 | N[i,k] == 0)
                 next;
 
-            m = migration_matrices[k,j,i];
+            m = Q * migration_matrices[k,j,i];
             p = sim.subpopulations[j];
             dbg("p"+j+".setMigrationRates("+k+", "+m+");");
             p.setMigrationRates(k, m);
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Changes look good. Let's get some expert opinions on whether this is correct.

petrelharp:
Gee, there's different ways to do it... talking this through... I am not a fan of rescaling, but if you're doing it, you want to keep these things the same:
- theta = 4 N mu
- rho = 4 N r
So, we're making N smaller and compensating by making mu and r bigger.  Here 4N is a time and mu and r are rates -  so we're keeping constant the number of mutations and recombinations that you see on the tree between two typical samples (4N).  We should do the same thing for migrations, thus scaling up migration rate, as you've done.

ok! I agree!

jeromekelleher:
Good enough for me; merging!

PR #438, opened 2020-03-24
Use population configurations from oldest epoch for recapition.

### LINKED ISSUE #437: SLiM engine doesn't pass Ne or migration_rates to `ts.recapitate()`.
It's not immediately clear to me what the call to `ts.recapitate()` should look like in the slim engine, but I'm pretty sure its wrong right now.

* What should `Ne` be in our case, where we likely have a single ancestral population at time `ts.slim_generation`? Should it be the size of that ancestral population at time `ts.slim_generation`? How are the values in `population_configurations` used, if at all?
* At present, we only have demographic models with a single ancestral population. I suppose this won't always be the case. Under the assumption that we have multiple populations in existence for recapitation, how does the call to `ts.recapitate()` look then? Do we pass a modified population_configurations with the relevant population sizes for each of the remaining populations at time `ts.slim_generation`? Presumably `migration_rates` should be set following 
[the docs](https://msprime.readthedocs.io/en/latest/tutorial.html#sec-tutorial-simulate-from).

@petrelharp, it would be great to get your input here!
The relevant code is in `_SLiMEngine.simulate()`: https://github.com/popsim-consortium/stdpopsim/blob/e1d4125c64fad0bbb6bb3fd8b45f283cbbd2d6b2/stdpopsim/slim_engine.py#L557-L668

grahamgower:
Fixes #437.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -524,6 +524,8 @@ def slim_makescript(
 
     printsc(_slim_lower)
 
+    return epochs[0]
+
 
 def simplify_remembered(ts):
     """
@@ -622,7 +624,7 @@ class _SLiMEngine(stdpopsim.Engine):
 
         with script_file_f() as script_file, mktemp(suffix=".ts") as ts_file:
 
-            slim_makescript(
+            recap_epoch = slim_makescript(
                     script_file, ts_file.name,
                     demographic_model, contig, samples,
                     slim_scaling_factor, check_coalescence, verbosity)
@@ -650,12 +652,25 @@ class _SLiMEngine(stdpopsim.Engine):
             rng = random.Random(seed)
             s1, s2 = rng.randrange(1, 2**32), rng.randrange(1, 2**32)
 
-            # Recapitation.
-            r_map = contig.recombination_map
-            pop_configs = demographic_model.population_configurations
+            recombination_map = contig.recombination_map
+            positions = recombination_map.get_positions()
+            if positions[-1] != ts.sequence_length:
+                # The recombination map passed in can have non-integer length
+                # (e.g. for contig length multiplier < 1), but SLiM trees will
+                # have integer sequence_length.
+                positions[-1] = ts.sequence_length
+                rates = recombination_map.get_rates()
+                recombination_map = msprime.RecombinationMap(positions, rates)
+
+            population_configurations = [
+                    msprime.PopulationConfiguration(
+                        initial_size=pop.start_size,
+                        growth_rate=pop.growth_rate)
+                    for pop in recap_epoch.populations]
             ts = ts.recapitate(
-                    recombination_rate=r_map.mean_recombination_rate,
-                    population_configurations=pop_configs,
+                    recombination_map=recombination_map,
+                    population_configurations=population_configurations,
+                    migration_matrix=recap_epoch.migration_matrix,
                     random_seed=s1)
 
         ts = simplify_remembered(ts)
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -67,7 +67,8 @@ class TestAPI(unittest.TestCase):
     def test_simulate(self):
         engine = stdpopsim.get_engine("slim")
         species = stdpopsim.get_species("AraTha")
-        contig = species.get_contig("chr5", length_multiplier=0.0001)
+        # Integral length (10kb) to hit specific code path in slim_engine.py
+        contig = species.get_contig("chr5", length_multiplier=10000/26975502)
         model = stdpopsim.PiecewiseConstantSize(species.population_size)
         model.generation_time = species.generation_time
         samples = model.get_samples(10)


grahamgower:
Can we similarly pass the population metadata to `recapitate()` in the population_configurations (it just wraps `msprime.simulate()` after all)? I got a bit lost reading through `msprime/simulate.py` trying to figure it out.

petrelharp:
Yes!

petrelharp:
Do we not have access to an msprime recomb map here?

grahamgower:
Yes, we do. But the pyslim docs gave me the impression that it couldn't be used.
https://pyslim.readthedocs.io/en/latest/tutorial.html#things-you-cannot-do

jeromekelleher:
> Yes, we do. But the pyslim docs gave me the impression that it couldn't be used.
> https://pyslim.readthedocs.io/en/latest/tutorial.html#things-you-cannot-do

This won't be an issue after msprime 1.0 is released (will also work on msprime git master, but health warnings very much apply)

grahamgower:
Ok, great. Are you happy for me to just pass in the recombination map then, or should this wait until some future time when we explicitly depend upon msprime >= 1.0?

petrelharp:
I need to update that warning. Pass it in. The breakpoints won't be at precisely integer locations, like in slim, which is what that bit is saying, but no big deal.

grahamgower:
Maybe it will have to be considered separately. The recombination map passed in by stdpopsim can have non-integer length (`recomb_map.get_positions()[-1]` can be non-integer if we use, eg. `length_multiplier=0.01`), but slim always has integer sequence_length.
```
  File "/home/grg/.local/bin/stdpopsim", line 11, in <module>
    load_entry_point('stdpopsim', 'console_scripts', 'stdpopsim')()
  File "/home/grg/src/stdpopsim/stdpopsim/cli.py", line 623, in stdpopsim_main
    run(args)
  File "/home/grg/src/stdpopsim/stdpopsim/cli.py", line 614, in run
    args.runner(args)
  File "/home/grg/src/stdpopsim/stdpopsim/cli.py", line 431, in run_simulation
  File "/home/grg/src/stdpopsim/stdpopsim/slim_engine.py", line 664, in simulate
    ts = ts.recapitate(
  File "/home/grg/src/pyslim/pyslim/slim_tree_sequence.py", line 364, in recapitate
    recap = msprime.simulate(
  File "/home/grg/.local/lib/python3.8/site-packages/msprime/simulations.py", line 434, in simulate
    sim = simulator_factory(
  File "/home/grg/.local/lib/python3.8/site-packages/msprime/simulations.py", line 275, in simulator_factory
    raise ValueError(
ValueError: The simulated sequence length must be the same as from_ts.sequence_length
```

grahamgower:
I also tried putting in the population metadata. But when looking at the final ts file, the population metadata is from slim. I guess this looks no further than the first provenance table? I.e. I get
```
>>> tskit.load("/tmp/c.trees").population(0)
{'id': 0, 'metadata': b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00'}
```
Rather than what I'd hoped,
```
>>> tskit.load("/tmp/b.trees").population(0)
{'id': 0, 'metadata': b'{"id": "pop0", "description": "Generic population", "sampling_time": 0}'}
```

jeromekelleher:
Looks like SLiM is overwriting the population metadata... Probably we'd need an option to pyslim to avoid this, or I guess we could work around by saving the population table and writing over it after SLiM is done. Probably the work going on for metadata schemas in tskit (https://github.com/tskit-dev/tskit/issues/57 and https://github.com/tskit-dev/tskit/pull/491) will need to go through before we can fully figure out the responsibility of programs when it comes to respecting the metadata of a tree sequence that's being operated on. 

But, it's **great** that we're getting into these gritty interoperability details!

jeromekelleher:
My take, FWIW, is that stdpopsim should ensure that the output of the tree sequence metadata is standardised and makes sense. We just need to figure out what the right protocol is for libraries that produce intermediate tree sequences.

grahamgower:
But this is after SLiM has completed. So, I have a tree sequence and (effectively) call `ts = msprime.simulate(from_ts=ts, start_time=ts.slim_generation, population_configurations=popu...)`, and then ts.populations(0) has the metadata from the original tree sequence object, not that of the passed population configurations.

Anyway, I'm happy for this to get merged in without the metadata thing being resolved. I'll squash/rebase.

jeromekelleher:
> But this is after SLiM has completed. So, I have a tree sequence and (effectively) call ts = msprime.simulate(from_ts=ts, start_time=ts.slim_generation, population_configurations=popu...), and then ts.populations(0) has the metadata from the original tree sequence object, not that of the passed population configurations.

Oh, OK, so msprime isn't overwriting the metadata when it gets a from_ts argument. Interesting one. I think in the short term we'll need to just rewrite the population metadata. Let's push this to another issue though.

jeromekelleher:
To keep codecov happy I think we need an example where  ``positions[-1] != ts.sequence_length``.

petrelharp:
BTW, the right argument to pass now to the msprime.RecombinationMap is `discrete=True` to get it to do integer breakpoints (which we should).

grahamgower:
Discrete maps sound very nice. But it's not yet in a released version of msprime, so I think we should wait before using it in stdpopsim.

jeromekelleher:
Discrete maps are one to follow up on after msprime 1.0 is shipped.

Can we get a quick rebase please @grahamgower? Ready to merge then.

grahamgower:
Yep, rebased.

PR #440, opened 2020-03-25
Just use the mean recombination rate for recapitation.

### LINKED ISSUE #439: Too many roots after #438 when recapitating a truncated recombination map.
```
$ stdpopsim -e slim HomSap 100 -l 0.001 -o /tmp/a.trees --seed 1 -q
$ python
Python 3.8.2 (default, Feb 26 2020, 22:21:03)
[GCC 9.2.1 20200130] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tskit
>>> sum(1 for t in tskit.load("/tmp/a.trees").trees() if t.num_roots > 1)
2
```

I'm not entirely sure what's going on here, but it seems to be caused by using the truncated recombination map from #438. Setting `recombination_rate=recomb_map.mean_recombination_rate` for recapitation does not exhibit this problem.

grahamgower:
There's something funny going on here when using the truncated
recombination map. Some of the trees still have more than 1 root.
Msprime 1.0 will bring discrete recombination maps, which is a
better way to do recapitation with recombination maps.

Closes #439.

### DIFF
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -65,14 +65,14 @@ class TestAPI(unittest.TestCase):
     def test_simulate(self):
         engine = stdpopsim.get_engine("slim")
         species = stdpopsim.get_species("AraTha")
-        # Integral length (10kb) to hit specific code path in slim_engine.py
-        contig = species.get_contig("chr5", length_multiplier=10000/26975502)
+        contig = species.get_contig("chr5", length_multiplier=0.001)
         model = stdpopsim.PiecewiseConstantSize(species.population_size)
         model.generation_time = species.generation_time
         samples = model.get_samples(10)
         ts = engine.simulate(
                 demographic_model=model, contig=contig, samples=samples)
         self.assertEqual(ts.num_samples, 10)
+        self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees()))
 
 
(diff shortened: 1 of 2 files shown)


tests/test_slim_engine.py
```diff
@@ -67,14 +67,14 @@ def test_recombination_map(self):
...
-        # Integral length (10kb) to hit specific code path in slim_engine.py
-        contig = species.get_contig("chr5", length_multiplier=10000/26975502)
+        contig = species.get_contig("chr5", length_multiplier=0.001)
         model = stdpopsim.PiecewiseConstantSize(species.population_size)
         model.generation_time = species.generation_time
         samples = model.get_samples(10)
         ts = engine.simulate(
                 demographic_model=model, contig=contig, samples=samples)
         self.assertEqual(ts.num_samples, 10)
+        self.assertEqual(sum(1 for t in ts.trees() if t.num_roots != 1), 0)
```
jeromekelleher: ``self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees())``  possibly clearer?

petrelharp:
Ok!

tests/test_slim_engine.py
grahamgower: Yep, agreed.

jeromekelleher:
Merged; thanks @grahamgower.

PR #447, opened 2020-03-26
Validation

grahamgower:

PR #448, opened 2020-03-26
Validate damnit

grahamgower:
Same as #447, but with a different branch name.


jeromekelleher:
Weird...

grahamgower:
The circleci url is the same as for the other PR, so if there's a problem with reusing state, this didn't escape it.

jeromekelleher:
Squash the commits?

Issue #449, opened 2020-03-27
SLiM engine ends the simulation one generation later than needed


grahamgower:
I did this deliberately, but now I can't remember why. Probably it should be fixed.

grahamgower:
Fixed with #450.

Issue #445, opened 2020-03-25
Choose sane defaults for SLiM engine


grahamgower:
The SLiM engine supports enabling/disabling recapitation. Time can be scaled, or not. At the moment, the defaults are to do recaptitation and scale time by a factor of 10. Are these defaults appropriate?

Also see #420, #444.

It would be great to get some wider feedback on this and #420 from @popsim-consortium/all

petrelharp:
I strongly vote for a default scaling factor of 1.0. One of the biggest points of SLiM is doing selection, and rescaling does not give equivalent results for selection. We don't want to confuse people here.

I do think doing recapitation by default is a good idea!

grahamgower:
Just so we're all informed, it takes 3 minutes to run a 250 kb OutOfAfrica_3G09 simulation with a scaling factor of 1.0, and under 2 seconds with a scaling factor of 10.0.
```
$ time stdpopsim -e slim --slim-scaling-factor 10 HomSap -o /dev/null -l 0.001 -q -d OutOfAfrica_3G09 50 50 50

real    0m1,835s
user    0m1,659s
sys     0m0,174s
$ time stdpopsim -e slim --slim-scaling-factor 1 HomSap -o /dev/null -l 0.001 -q -d OutOfAfrica_3G09 50 50 50

real    2m58,202s
user    2m56,219s
sys     0m1,821s
```

grahamgower:
@petrelharp Can you elaborate on the differences with regard to selection? Is this all kinds of selection (positive, negative, single locus, QTL, etc)? What kind of non-equivalence? Do you have any references you can point me at?

RyanGutenkunst:
I strongly vote that the default be 1.0, but that it be made **very** easily changeable and that extensively documented. The appropriate scaling is going to vary dramatically depending on organism and computational resources. For example, humans can probably be done scaling by 1.0, but nothing at scale could be done with Drosophila without major scaling.

In our experience, even if scaling isn't perfect, many many useful simulation results do scale appropriately, particularly for modest sample sizes. For example, I find it pretty embarrassing that we failed at doing neutral genome-scale simulations for Drosophila in the paper with msprime. In our experience, SLiM simulations at a scaling of 1000 would complete quite quickly, and at the least the SFS-based analyses would give identical results to simulations with less aggressive scaling. We find similar results for models with background selection, at least with the summaries we're looking at.

petrelharp:
[...]

PR #450, opened 2020-03-27
Change SLiM engine parameters and default behaviour.

### LINKED ISSUE #420: Decide how we should handle `burn in` in the SLiM engine
The SLiM engine currently supports using recapitation, or not. When recapitation is disabled, we wait until all individuals have a common ancestor after the start of the simulation, and then we wait an additional `10*N` generations. N is taken to be the maximum population size of the populations that exist at the start of the simulation (all current models only have one). We currently have separate CLI options to disable recaptitation (`--slim-no-recapitation`) and to disable burn in (`--slim-no-burnin`).

In reviewing #409, @andrewkern suggested that the burn in time should be configurable, at least from the python API; and that we probably don't need both CLI options.

I propose that:
1. We change the `--slim-no-burnin` parameter to be `--slim-burnin X`, where `X` specifies the length of the burn in period, in units of N generations. This can be set to zero to disable burn in.
2. We add an option to disable checking for coalescence, separately to the burn in. This might be desirable in the future for non-neutral simulations that use a distribution of fitness effects (DFE). In such cases, we probably want to: disable checking for coalescence (because it will be really slow), specify a fixed burn in time with the DFE, and then optionally do recapitation to fix up genealogies without a single root.

We don't have to do (2) now. Another option is to get rid of checking for coalescence entirely.

grahamgower:
* Always do recapitation.
 * --slim-scaling-factor defaults to 1.
 * --slim-burn-in defaults to 10*N generations.

Closes #420, #445, and #449.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -561,22 +561,15 @@ def stdpopsim_cli_parser():
                 "--slim-script", action="store_true", default=False,
                 help="Write script to stdout and exit without running SLiM.")
         slim_parser.add_argument(
-                "--slim-scaling-factor", metavar="Q", default=10, type=float,
+                "--slim-scaling-factor", metavar="Q", default=1, type=float,
                 help="Rescale model parameters by Q to speed up simulation. "
                      "See SLiM manual: `5.5 Rescaling population sizes to "
                      "improve simulation performance`. "
                      "[default=%(default)s].")
         slim_parser.add_argument(
-                "--slim-no-recapitation", action="store_true", default=False,
-                help="Explicitly wait for coalescence, and add "
-                     "mutations, within the SLiM simulation. This may be much "
-                     "slower than the defaults (recapitation and neutral mutation "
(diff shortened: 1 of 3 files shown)


grahamgower:
validation.py will also need updating, but I'll tackle that separately.

grahamgower:
All the tests pass for me locally. Maybe this is a version specific failure, for some dependency (e.g. circleci runs python 3.6, I'm running 3.8).

@jeromekelleher Are you able to poke circleci to build git master again? I really want to know if something has changed since the last pr was merged.

grahamgower:
Nevermind about poking circleci. The dependency whose version changed is pyslim. I can reproduce this error of too many roots after updating to `pyslim 0.400`

CC @petrelharp

petrelharp:
Ok, I've done this and updated things on pip. Using whatever version of msprime plus pyslim 0.401 should work now.

petrelharp:
Looks good to me! Thanks a bunch!!

PR #452, opened 2020-03-28
Pin pyslim>=0.401.

grahamgower:
This avoids an issue with recapitation, arising due to changes between
msprime-0.7.4 and msprime-current.
https://github.com/tskit-dev/pyslim/issues/71

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -13,4 +13,4 @@ kastore
 attrs
 appdirs
 humanize
-pyslim
+pyslim>=0.401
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -15,7 +15,7 @@ install:
   - cmd: conda info -a
   - cmd: python --version
   # pyslim is not yet on conda-forge.
-  - cmd: python -m pip install pyslim
+  - cmd: python -m pip install "pyslim>=0.401"
```
jeromekelleher: I don't think this is needed, is it? We'll always install the latest version on appveyor, and we only need to pin when we're worried about people having older versions lying around.

grahamgower:
I removed the appveyor.yml change and rebased.

Issue #201, opened 2019-11-26
basic binder example


agladstein:
Set up a basic binder so that new users can play with stdpopsim without hassle of downloading and setting up their environment. We can point them to try out the tutorials in the binder.

To do this, we need to create a Jupyter Notebook in the master github repo and link binder to it.
https://mybinder.org/

This is for later (after pub?).
You can assign me to do this. I can do this later.

jeromekelleher:
This should  extend @carjed's PR #49 -- on yesterday's call we decided to wait until we're close to release, then update the notebook/environment.

jeromekelleher:
We definitely want this now, as stdpopsim has a conda package (hooray!). Should be easy enough to update @carjed's initial notebook to do a simple simulation and (say) compute a site frequency spectrum using tskit, and plot with matplotlib/seaborn (say)?

agladstein:
Yes, no problem. I will get on it.

agladstein:
Logistics question - should I just create a new branch for this and let https://github.com/popsim-consortium/stdpopsim/pull/49 die?

andrewkern:
sounds reasonable to me @agladstein

jeromekelleher:
Well I'd like to see @carjed get some credit for this, so it would be nice to keep his commit in the history. Logistically, I guess we'll need to let PR #49 die since @agladstein can't write to it. The simplest thing would be to create a new branch on your fork @agladstein, then to cherry-pick @carjed's commit onto that branch. Then you can add your own commits on top of that.

How does that sound @agladstein?

agladstein:
I'm just about done with this, but I've encountered something confusing with the sfs. 
Could someone please take a look at https://mybinder.org/v2/gh/agladstein/stdpopsim/binder?filepath=stdpopsim_api_example.ipynb
 (the active notebook in binder)

basically, if I follow @gtsambos way with simplifying I get one thing, and if I do what @petrelharp suggested in https://github.com/popsim-consortium/stdpopsim/pull/340 I get something else.

petrelharp:
They are both right! The only differences are in the fixed/absent categories. For instance:
```
>>> np.column_stack([
            ts_YRI.allele_frequency_spectrum(polarised=True,span_normalise=False),
            ts.allele_frequency_spectrum([YRI_samples], polarised=True,span_normalise=False)])
array([[   0., 2705.],
       [2907., 2907.],
       [1213., 1213.],
       [ 693.,  693.],
[...]

PR #453, opened 2020-03-28
add config for Binder and example notebook

agladstein:
I have the binder linked to popsim-consortium/stdpopsim:master in the README. So, that won't work until this PR is resolved. In the meantime, you can check the binder from my branch by going to
https://mybinder.org/v2/gh/agladstein/stdpopsim/binder?filepath=stdpopsim_example.ipynb

closes https://github.com/popsim-consortium/stdpopsim/issues/201

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,5 @@
 
 A community-maintained library of standard simulation population genetic simulation models.
 Please see the [documentation](https://stdpopsim.readthedocs.io/en/latest/) for further details.
+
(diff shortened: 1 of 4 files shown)


agladstein:
Somebody check that this doesn't break some RTD env setup? I'm not sure. I tried to build RTD locally while on my binder branch and it seemed broken.

jeromekelleher:
Let's just try it out. I think the rtd builds have been failing for a while because of some conda thing, so we should try and sort that out soon if it's doesn't resolve itself.

This is great, thanks @agladstein!

Issue #458, opened 2020-04-02
Write tutorials for a "pre-existing model" and the SLiM engine


petrelharp:
I'm on this, and doing a general pass through the tutorial: will post a PR later today. I hope no-one else is working on this at the same time, except maybe #340?

grahamgower:
I had a couple of thoughts that I wanted to mention here...
 * Using `-vv` on the CLI will increase verbosity to level 2, triggering the slim script to print out each registered event. This might be useful to users when scaling, to confirm that the coarsness of the simulation is not obviously crazy. Perhaps unintuitively, `-vv` can be combined with `--quiet`.
* When I started writing the slim engine, I'd imagined that advanced users might want to print out the slim script with `--slim-script`, and make some customisations before running it with slim manually. It turns out the post-slim recap bits can be quite subtle to get right and this process will be difficult for users in practice. If you have any thoughts about how/whether this sort of thing could/should be made easier, I'd love to hear them.

petrelharp:
Ah, I haven't paid attention to the CLI portion of the tutorial. I'll have a go at a SLiM/CLI portion next.

> It turns out the post-slim recap bits can be quite subtle to get right and this process will be difficult for users in practice.

What are the biggest gotchas? I was imagining that people would have to read the pyslim documentation (which I recently improved, and should link to here), but if there's only one population you're trying to recapitate it's pretty straight forward, I thought?

grahamgower:
Two additional things come to mind:
 * It requires non-standard simplification. To permit temporal sampling of individuals, I chose to make all sampling done by "remembering" individuals.
 * Node times need to be scaled to match the value of Q within the SLiMulation, so that the pre- and post- recapitation parts of the trees have the same units of time.

petrelharp:
> It requires non-standard simplification. 

That seems like the right use of "remembering"!  Well, just plain simplify won't get rid of any remembered individuals, at least, and the [`simplify_remembered`](https://github.com/popsim-consortium/stdpopsim/blob/ba517491e341fb307d423a8ecdbae0ca33c22ad7/stdpopsim/slim_engine.py#L533) would be easy enough to put in an example somewhere.

> Node times need to be scaled to match the value of Q

[...]

PR #460, opened 2020-04-03
More tutorials

petrelharp:
Here's some additions to the tutorial, some minor edits, and a bit more docs in one other place, closing #458.

I've started out with the commits for #340, rebased to the current HEAD; hopefully this doesn't cause too much confusion. We could either finish out #340 here, or folks can do it there and I can sort out the mess.

Also: I've tried out a different format for the example code. Rather than writing
```
  >>> python.code( )
  output
```
I'm doing
```
   python.code()
  # output
```
since that produces code that can be more easily cut-and-pasted. What do you all think? If you like it, I can switch the rest of the example code over to this style. Or, I can put it all back to the way it was before.


jeromekelleher:
I don't like the ">>>" prefix either @petrelharp, I prefer the way you've done it.

grahamgower:
LGTM. Agreed wrt `>>>`---being able to copy/paste from the tutorial is very useful.

agladstein:
Sounds good. Hopefully I'll have a chance today to give your PR a quick run through. But, if you don't hear from me by the end of the day and you want to merge, I can check it after the merge.

andrewkern:
one more voice of agreement in the format switch @petrelharp

petrelharp:
Question: what's something you can't do in the CLI that you can do in the python interface?

grahamgower:
That sounds like a riddle. I'm pretty sure you can poke any of the `simulate()` paramters from the CLI, one way or another.

I don't think you can get an `IsolationWithMigration` model from the CLI, but you ought to be able to simulate that with the slim engine from the Python API. Likewise for any custom demographic model. I'm sure there'll be folks who want to copy/paste/modify one of the catalog models and run it with the Python API.

petrelharp:
Ok, I'm done with my edits, and would like someone else to take a look.

Two notes:

- I removed the numbered list from the "generic model" example. I kinda liked the numbering, but it wasn't paralleled in the other examples, so it seemed out of place. Maybe we should put some more structure back in there for this and other examples?

- The bit on "simulating the dataset" in "Example analyses" is somewhat redundant with the CLI section. Should we shorten it?

agladstein:
> I removed the numbered list from the "generic model" example. I kinda liked the numbering, but it wasn't paralleled in the other examples, so it seemed out of place. Maybe we should put some more structure back in there for this and other examples?

I like the numbering (of course I do, I put it in), and I'd rather add in numbering/structure to all the examples. I feel like having numbered instructions helps for following examples.

> The bit on "simulating the dataset" in "Example analyses" is somewhat redundant with the CLI section. Should we shorten it?

I think it can be assumed that the CLI and API tutorials are somewhat separate, because someone who wants to use the CLI won't bother going through API examples and vise versa. So, I think it's okay to have some repetition.

petrelharp:
> I like the numbering (of course I do, I put it in), and I'd rather add in numbering/structure to all the examples. I feel like having numbered instructions helps for following examples.

You're right, this would be better.

petrelharp:
Whoops, I forgot about @grahamgower's comments above; that's also on the to-do list:

```
  -  Using -vv on the CLI will increase verbosity to level 2, triggering the slim script to print out each registered event. This might be useful to users when scaling, to confirm that the coarsness of the simulation is not obviously crazy. Perhaps unintuitively, -vv can be combined with --quiet.
  -  When I started writing the slim engine, I'd imagined that advanced users might want to print out the slim script with --slim-script, and make some customisations before running it with slim manually. It turns out the post-slim recap bits can be quite subtle to get right and this process will be difficult for users in practice. If you have any thoughts about how/whether this sort of thing could/should be made easier, I'd love to hear them.
```

petrelharp:
>  I like the numbering (of course I do, I put it in), and I'd rather add in numbering/structure to all the examples. I feel like having numbered instructions helps for following examples.

I've added subheaders in 2c01f40 (but not numbers, as the steps don't necessarily come in the same order) - feedback welcome.

petrelharp:
... and, I've put in sections on SLiM command line, and said something about @grahamgower's comments above. It's ready for someone else to have a read through - volunteers?

docs/tutorial.rst
```diff
@@ -100,122 +135,422 @@ the ``-l`` option. (NOTE: this may a minute or so to run).
...
+The scaling factor
+------------------
+
+However, even with only 5% of a chromosome,
+that is a pretty big simulation, due to the large number of individuals
+(unlike msprime, SLiM must actually simulate all the individuals in the population
+even just to get a few samples).
+To make it run fast enough for a tutorial,
+we can specify a *scaling factor*,
+described in more detail below (`sec_slim_scaling_factor`),
```
grahamgower, line 231: This doesn't seem to link when pulled the branch and generated the docs. Should this use double backticks?

grahamgower:
It looks great @petrelharp! I've left some more specific comments throughout.

docs/tutorial.rst
```diff
@@ -100,122 +135,422 @@ the ``-l`` option. (NOTE: this may a minute or so to run).
...
+using the ``--slim-scaling-factor/-Q`` option.
```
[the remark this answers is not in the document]
petrelharp: oh, whoops! no need, I think - long options are better, really.

petrelharp:
Thanks a bunch! I fixed those things and some other spelling errors.

petrelharp:
Whoops - I'd forgotten the note about rescaling if you use the slim script - but, once we do #458 we should update that bit of the docs to give an example of post-processing the SLiM output.

petrelharp:
Does someone else want to read through this? @agladstein? I will squash & rebase when folks think it's good.

petrelharp:
Perhaps jumping ahead, I've changed the text to use #465.

docs/tutorial.rst
```diff
@@ -294,60 +861,73 @@ the ``tskit`` package to load and analyse this simulated tree sequence file.
...
 
 .. code-block:: python
 
-    >>> import itertools
-    >>> inds = itertools.combinations_with_replacement(range(0, ts.num_populations), 2)
-    >>> inds = list(inds)
-    >>> print(inds)
-    [(0, 0), (0, 1), (0, 2), (0, 3), (1, 1), (1, 2), (1, 3), (2, 2), (2, 3),
-     (3, 3)]
+   inds = []
```
agladstein, line 908: Spacing is messed up here and will error for copy paste.
Try
```
inds = []
for i in range(0, ts.num_populations):
    for j in range(i, ts.num_populations):
      inds.append((i, j))

print(inds)
```

docs/tutorial.rst
```diff
@@ -388,9 +967,179 @@ It relies on the ``numpy``, ``seaborn`` and ``matplotlib`` packages.
...
+:meth:`tskit.TreeSequence.allele_frequency_spectrum` method.
+
+Since we wish to find the AFS separately for each of our two populations, we will
+first need to know which samples correspond to each population.
+The :meth:`tskit.TreeSequence.samples`
+method in tskit allows us to find the IDs of samples from each population:
+
+.. code-block:: python
+
+    pop_samples = [ts.samples(0), ts.samples(1)]
```
agladstein, line 1057: add print statement
```
print(pop_samples)
```

petrelharp:
> Still waiting on numbering throughout.

Sorry - I thought I explained above - I put in sub-headings, but when I came to labelling them with numbers I got uncertain, because the steps, and the order of the steps, is not the same in all the examples. And, some of the bits are number-ey ("do this; now do this") but some are more discussion-ey. So, it seemed confusing to have them. So, I'm in favor of them in principle, but I'm not sure where I'd put them in, and maybe just having subheadings (like now) serves the same purpose?

petrelharp:
> Can't currently run slim examples. #467
> Will check those later once env issue is worked out.

FWIW, I have checked the slim examples. It would be nice to throw a more informative error if SLiM is not installed/in the PATH, as seems to be the problem in #467.

agladstein:
> Sorry - I thought I explained above - I put in sub-headings, but when I came to labelling them with numbers I got uncertain, because the steps, and the order of the steps, is not the same in all the examples. And, some of the bits are number-ey ("do this; now do this") but some are more discussion-ey. So, it seemed confusing to have them. So, I'm in favor of them in principle, but I'm not sure where I'd put them in, and maybe just having subheadings (like now) serves the same purpose?

Point taken. In that case, I'd say go ahead with the merge without changing the numbering and I can potentially take a stab at numbering in a separate PR.

docs/tutorial.rst
```diff
@@ -100,122 +135,425 @@ the ``-l`` option. (NOTE: this may a minute or so to run).
...
+for more discussion.
+
+
+Another SLiM model, and debugging output
+========================================
+
+Next we'll look at running a different model with SLiM,
+but getting some sanity check output along the way.
+
+DroMel
```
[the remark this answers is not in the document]
petrelharp: Good idea - I wanted to tag this as being a different species in the header; maybe `Choose a species: Drosophila melanogaster`

petrelharp:
Thanks a bunch! I fixed up all those things. Perhaps I should squash and rebase?

agladstein:
Sounds good.

petrelharp:
Done; maybe this is ready to merge?

andrewkern:
looks good to me. merging.

Issue #444, opened 2020-03-25
Time-scaled SLiM simulations have systematically fewer num_nodes and num_edges


grahamgower:
See #442.

This is probably expected for time-scaled discrete-generation simulations, but we should satisfy ourselves that there isn't another underlying problem here. And is this problematic in general?

petrelharp:
Are these simplified?

petrelharp:
By 'time-scaled', you mean Q != 1?

grahamgower:
Simplified, yes: https://github.com/popsim-consortium/stdpopsim/blob/master/stdpopsim/slim_engine.py#L533-L541

And yes, I mean Q>1 (Q=10 in the cases I looked at).

petrelharp:
Well, fewer nodes would imply fewer edges, since each node has to be connected to the other ones. And I think fewer nodes makes sense - with fewer individuals, there will be more re-use of nodes, even with the same amount of coalescence. Imagine taking the set of marginal trees; with no rescaling and a very big population imagine that every node in every marginal tree is distinct; but as the total population size goes down (and since time changes also, the number of individuals-alive-ever goes down with Q^2) those trees get squooshed onto fewer distinct ancestral genomes.

Come to think of it, this gives us fodder for the "rescaling is not actually equivalent" bit in the manuscript, so maybe `num_nodes` *should* be a plot in the comparison?

jeromekelleher:
> Come to think of it, this gives us fodder for the "rescaling is not actually equivalent" bit in the manuscript, so maybe num_nodes should be a plot in the comparison?

Makes sense to me - the trees are what matter at the end of the day.

petrelharp:
> Makes sense to me - the trees are what matter at the end of the day.

And this is kinda tricky - because presumably the *trees* are good (marginally), but there's more correlation between trees along the sequence than you'd expect, somehow.

petrelharp:
More re-using of nodes between non-adjacent trees was what I was thinking. Since density of recombinations on the trees is scaled to be the same, mean edge length and number of trees should be (almost?) the same.  But this is hard to think about.

jeromekelleher:
That's interesting, I'm curious to take a look now. Can you point me to a set of plots from a SLiM simulation with Q=1 and Q>1 for comparison please @grahamgower? I've gotten a bit lost with all the ones that are going around.

grahamgower:
[...]

PR #461, opened 2020-04-03
More validation script changes.

grahamgower:
* Show stats and model docstrings on the figures.
* Rotate "run time" plot to fit model descriptions.
* Add node_arity statistic (disabled). See #444.
* Put allele frequency spectrum on log scale.
* LD calculation can use lots of memory, so look at only the first 1 mb
  of the tree sequence.

### DIFF
--- a/validation.py
+++ b/validation.py
@@ -90,7 +90,7 @@ def onepop_constantN_slim2(out_dir, seed):
 def onepop_constantN_slim3(out_dir, seed):
     """
     Single population with constant population size.
-    Time is rescaled by a factor of 10.
+    Time and Ne are rescaled by a factor of 10.
     """
     return _onepop_PC("slim", out_dir, seed, slim_scaling_factor=10)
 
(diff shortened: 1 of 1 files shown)


grahamgower:
Looks like this now:
[onepop_bottleneck_msprime1__onepop_bottleneck_slim3.pdf](https://github.com/popsim-consortium/stdpopsim/files/4426006/onepop_bottleneck_msprime1__onepop_bottleneck_slim3.pdf)

validation.py
```diff
@@ -90,7 +90,7 @@ def onepop_constantN_slim2(out_dir, seed):
 def onepop_constantN_slim3(out_dir, seed):
     """
     Single population with constant population size.
-    Time is rescaled by a factor of 10.
+    Time/Ne are rescaled by a factor of 10.
```
jeromekelleher, line 93: I read this as "time divided by Ne" initially. An "and" seems less ambiguous.

grahamgower: Yeah, good call. Fixed.

PR #465, opened 2020-04-04
Recapitate and rescale function for advanced SLiM users.

grahamgower:
See #458.

What do you think @petrelharp? Maybe it should be a classmethod instead?

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -530,17 +530,6 @@ def slim_makescript(
     return epochs[0]
 
 
-def simplify_remembered(ts):
-    """
-    Remove all samples except those individuals that were explicity
-    sampled in SLiM with sim.treeSeqRememberIndividuals().
-    """
-    nodes = itertools.chain.from_iterable(
-                i.nodes for i in ts.individuals()
-                if i.flags & pyslim.INDIVIDUAL_REMEMBERED)
-    return ts.simplify(samples=list(nodes), filter_populations=False)
-
-
 class _SLiMEngine(stdpopsim.Engine):
     id = "slim"  #:
     description = "SLiM forward-time Wright-Fisher simulator"  #:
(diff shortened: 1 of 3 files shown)


stdpopsim/slim_engine.py
```diff
@@ -659,14 +668,42 @@ def script_file_f():
...
             ts, rate=mutation_rate, keep=True, random_seed=s2))
 
         return ts
 
+    def recap_and_rescale(
+            self, ts, demographic_model=None, contig=None, samples=None, seed=None,
+            verbosity=0, slim_path=None, slim_script=False, slim_scaling_factor=1.0,
+            slim_burn_in=10.0, **kwargs):
+        """
+        Apply post-SLiM transformations to ``ts``. This includes rescaling node
```
petrelharp, line 684: Instead of "includes", should say exactly what the things it does are.

stdpopsim/slim_engine.py
```diff
@@ -659,14 +668,42 @@ def script_file_f():
...
+        times, recapitation, simplification, and adding neutral mutations.
+
+        If the SLiM engine was used to output a SLiM script, and the script was
+        run outside of stdpopsim, this function can be used to transform the
+        SLiM tree sequence following the procedure that would have been used
+        if stdpopsim had run SLiM itself.
+        The parameters supplied after ``ts`` must match those that were used
+        to construct the SLiM script with :func:`.simulate`.
```
petrelharp, line 692: All of them? Seed at least doesn't have to match, right? Should say which ones.

grahamgower:
Upon thinking about this further, I can imagine changes to `simulate()` that will mean `recap_and_rescale()` can't garuantee the same behaviour for SLiM scripts generated with earlier versions of stdpopsim (e.g. probably the selection PR I have open will have change the behaviour here). I've added a warning to this effect. After all, this function is provided in the hope that it will be useful (particularly for developers working on new features!), and spending development effort trying to track and support different stdpopsim versions is not feasible.

petrelharp:
Good point, but I don't know if we need a warning for possible future non-backcompatibility.

grahamgower:
I think backwards incompatibility is quite likely. Maybe a different warning is in order though. Something like: ```This function is provided in the hope that it will be useful. But as we can't anticipate what changes you'll make to the slim code before using recap_and_rescale(), the source code should be consulted to determine if it's behaviour is appropriate for your case.```

petrelharp:
Ah, good idea. I agree.

grahamgower:
Ok, all done! Rebased/squashed.

jeromekelleher:
I'm not really following what's happening here, but merging since @petrelharp has approved.

PR #475, opened 2020-04-17
Use the logging module for all terminal output.

grahamgower:
See #473.

This PR makes all terminal output go through the logging module. The `logging` module provides DEBUG, INFO, WARN, ERROR and CRITICAL logging levels. This PR sets the default to show all messages at level INFO and above. DEBUG level is set with the `-v`  CLI parameter, and WARN level is set with `-q`. I've assumed that we'll use exceptions to report errors, so the ERROR and CRITICAL levels are unused.

Terminal output is very similar to before. INFO level messages have no prefix, but DEBUG messages have a simple `DEBUG: ` prefix and WARN messages have a simple `WARNING: ` prefix. The `warnings.warn()` messages are redirected to the default logger.

Example output below.

```
$ stdpopsim HomSap 10 -l 0.01 -o /dev/null -q
```

```
$ stdpopsim HomSap 10 -l 0.01 -c chrX -o /dev/null -q
WARNING: Non-autosomal simulations are not yet supported. See https://github.com/popsim-consortium/stdpopsim/issues/383 and https://github.com/popsim-consortium/stdpopsim/issues/406

```

```
t490:stdpopsim $ stdpopsim HomSap 10 -l 0.01 -c chrX -o /dev/null
WARNING: Non-autosomal simulations are not yet supported. See https://github.com/popsim-consortium/stdpopsim/issues/383 and https://github.com/popsim-consortium/stdpopsim/issues/406

Simulation information:
    Engine: msprime (0.7.4)
    Model id: PiecewiseConstant
    Model desciption: Piecewise constant size population model over multiple epochs.
    Population: number_samples (sampling_time_generations):
        pop0: 10 (0)
Contig Description:
    Contig length: 1552705.6
    Mean recombination rate: 1.164662223273842e-08
    Mean mutation rate: 1.29e-08
    Genetic map: None

If you use this simulation in published work, please cite:
[simulation engine]
Kelleher et al., 2016: https://doi.org/10.1371/journal.pcbi.1004842
[genome assembly]
The Genome Sequencing Consortium, 2001: http://dx.doi.org/10.1038/35057062
[mutation rate]
Tian, Browning, and Browning, 2019: https://doi.org/10.1016/j.ajhg.2019.09.012
[recombination rate]
The International HapMap Consortium, 2007: https://doi.org/10.1038/nature06258
[population size]
Takahata, 1993: https://doi.org/10.1093/oxfordjournals.molbev.a039995
[generation time]
[...]

PR #483, opened 2020-04-20
Make mock unit tests less brittle, by using autospec.

grahamgower:
For motivation, see passing tests that are obviously wrong:
https://github.com/popsim-consortium/stdpopsim/pull/475#discussion_r411214337

The problem is that mock.MagicMock() objects ensure all accessed
attributes and methods are successful. This means that unit tests
can do erroneous things, e.g. due to a typo or an API change, but the
test will pass regardless. A simple solution is to provide a `spec`
for the mocked object. In the case of mocking a function, this is
most easily done by passing `autospec=True` to the mock.patch()
or mock.patch.object() call. If a `new` parameter is provided for the
mock, the spec of the new object will be used, so these cases do not
suffer from the problem just described.

Note that autospec=True probably should not be applied to non-function
objects---at least not without first consulting the autospec docs.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -214,13 +214,6 @@ def get_provenance_dict():
     return document
 
 
-def write_to_stdout(fileobj):
-    """
-    Write the contents of the specified file to stdout.
-    """
-    shutil.copyfileobj(fileobj, sys.stdout.buffer)
-
-
 def write_output(ts, args):
     """
     Adds provenance information to the specified tree sequence (ensuring that the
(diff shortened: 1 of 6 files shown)


jeromekelleher:
LGTM, thanks @grahamgower. Looks like we've dropped coverage on a critical line though - any idea what's happened here?

grahamgower:
Yep. When I added autospec to  test_canis_familiaris.py, the test failed, so I made a change that parallels what you did in #430. So probably this should have been changed in #430, and the coverage would have dropped in that PR instead.

grahamgower:
I guess we could try mocking sys.stdout with an `io.TextIOWrapper` instead?

grahamgower:
OK, I think that's better now. I guess the coverage drop is now just due to code removal.

jeromekelleher:
Very nice, thanks @grahamgower. Some great moves in there that I'll be copying!

Issue #468, opened 2020-04-06
readthedocs build failing due to prerequisites


petrelharp:
Here's what's going on: does anyone know how to fix this?
![Screenshot from 2020-04-06 16-35-38](https://user-images.githubusercontent.com/1046249/78614899-cb1c3280-7824-11ea-805a-eaea0a44cfc8.png)

It looks like it's [been failing for a while](https://readthedocs.org/projects/stdpopsim/builds/). The last passing commit was [this one](https://readthedocs.org/projects/stdpopsim/builds/), and the first failing commit after that was [this one](https://readthedocs.org/projects/stdpopsim/builds/), but there were no changes to the requirements in the intervening commits, I think?

petrelharp:
[The docs](https://docs.readthedocs.io/en/stable/guides/build-using-too-many-resources.html#use-pip-when-possible) suggest using `pip` instead of `conda`, but give no guidance on how to do this?

petrelharp:
Well, some flailing around over in #469 didn't get me anywhere. Switching to pip failed to install msprime because of not being able to find `gsl`. And we need to actually install stdpopsim for autodoc to work, and for stdpopsim we need msprime.

Not sure what to do: maybe whatever `msprime` does, but what's different about that is not clear to me.

jeromekelleher:
It used to work before, and I've been hoping it would sort itself out (you get a fair amount of transient failures with conda, unfortunately). We have to use conda because we want to run msprime. This is different to the docs for msprime because we don't actually need to run msprime there, and the low-level module is never built.

I don't really know how to fix this, other than try to figure out what's using too much memory.

jeromekelleher:
I guess a place to start is to try and figure out why our conda env is using so much memory. One possibility is that we're not being specific enough with our versioning and the solver is doing lots of work because of this.

Would it be possible for a conda user to try out this command:
```
conda env create --quiet --name latest --file requirements/rtd-conda-environment.yml
```

and report back on how much memory it uses please? We could then try pinning all the package versions and see if this helps.

grahamgower:
```
$ /usr/bin/time conda env create --quiet --name stdpopsim-latest --file requirements/rtd-conda-environment.yml
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... done
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
[...]

PR #485, opened 2020-04-22
Reduce readthedocs memory footprint.

grahamgower:
Installing a conda environment on the readthdocs host often fails due
to memory exhaustion. But the readthedocs docker images don't include
gsl, which is needed to build msprime when installing with pip.
So this change uses a minimal conda environment, containing only gsl,
and installs the remaining dependencies using pip.

See #468.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,5 +16,6 @@ conda:
 python:
   # Install the local package using pip so that stdpopsim is in the path.
   install:
+    - requirements: requirements/development.txt
     - method: pip 
       path: .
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I think this will probably fail because  
```
  - sphinx-argparse
￼  - sphinx-issues
￼  - sphinxcontrib-programoutput
``` 
are missing. But I say we merge and see what happens, it's easy enough to follow up.

jeromekelleher:
Yep, failed, **but** it installed msprime OK. So, there's hope.

PR #486, opened 2020-04-22
Move sphinx packages back into conda deps.

jeromekelleher:
Deleted the pip requirements bit since it didn't do anything. Pip installing the package should pull in all the important deps though, so hopefully this will do the trick.

Follow on to #485

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,6 +16,5 @@ conda:
 python:
   # Install the local package using pip so that stdpopsim is in the path.
   install:
-    - requirements: requirements/development.txt
-    - method: pip 
+    - method: pip
       path: .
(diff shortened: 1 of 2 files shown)


grahamgower:
Hmm, shouldn't these go into `requirements/developent.txt` instead? This will just blow out the conda memory usage again, because it has to resolve dependencies for python + packages. I'm pretty sure its the SAT solver used by conda's dependency resolution that is the non-deterministic memory hog.

jeromekelleher:
Worked, woot! https://readthedocs.org/projects/stdpopsim/builds/10890845/

There's no point in putting things into ``requirements/developent.txt`` @grahamgower, RTD was just ignoring it. It's conda or nothin (except for the package itself) it seems. Anyway, this is working for now.

Thanks for your help!

Issue #454, opened 2020-03-28
add binder link to docs


agladstein:
related to https://github.com/popsim-consortium/stdpopsim/issues/201

Add binder link to RTD. Probably very beginning of tutorial section. Maybe also end of intro section.

PR #489, opened 2020-04-23
added binder links to docs

agladstein:
closes https://github.com/popsim-consortium/stdpopsim/issues/454

added links to binder in beginning of intro section and tutorials.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -4,6 +4,13 @@
 Installation
 ============
 
+You can get started right away, without installing ``stdpopsim`` locally, by using the
+`Jupyter Binder <https://mybinder.org/v2/gh/popsim-consortium/stdpopsim/master?filepath=stdpopsim_example.ipynb>`_. |binder|
+
+.. |binder| image:: https://mybinder.org/badge_logo.svg
+ :target: https://mybinder.org/v2/gh/popsim-consortium/stdpopsim/master?filepath=stdpopsim_example.ipynb
+
+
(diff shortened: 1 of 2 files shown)


docs/tutorial.rst
```diff
@@ -37,7 +37,10 @@ Running ``stdpopsim`` with the command-line interface (CLI)
 ***********************************************************
 
 In order to use the ``stdpopsim`` CLI the ``stdpopsim`` package must be
-installed (see :ref:`Installation <sec_installation>`). The CLI provides access
+installed (see :ref:`Installation <sec_installation>`).
```
jeromekelleher, line 40: Maybe better to put a quick note at the top of the Installation section, so that we don't have to say the same thing lots of times? Less likely to go out of date then.

agladstein: Do you think add it as a note to installation section instead of putting in the intro and tutorial?

jeromekelleher: Yes - that way when someone click on "Installation" the first thing they see is "Oh, I don't actually need to install at all; cool!"

So, we don't change the intro and tutorial at all, just the installation section. Or the intro and tutorial link to a section somewhere else which tells them about the binder. I just don't want to put the same link into the docs n times - links go out of date and it's much easier to update them if there's only one place!

agladstein:
Okay. Just added a link to the installation docs.

jeromekelleher:
Great, thanks @agladstein!

Issue #491, opened 2020-04-24
questions on adding new model


agladstein:
Partly because I want to check that the adding new model docs make sense, and partly because I want to add a model, I'm trying to add a model.

One thing that I'm confused about is once I've written a demographic model function, how is it incorporated into `stdpopsim`?  Do I need to build something or??

I've added a function to the `stdpopsim/catalog/homo_sapiens.py`, but the new model I've added doesn't show up in `demographic_models`.

```
import stdpopsim
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("chr22")
for x in species.demographic_models:
   print(x.id)
   
OutOfAfrica_3G09
OutOfAfrica_2T12
Africa_1T12
AmericanAdmixture_4B11
OutOfAfricaArchaicAdmixture_5R19
Zigzag_1S14
AncientEurasia_9K19
PapuansOutOfAfrica_10J19
```

agladstein:
I also don't understand how to write the unit test. Where does it go? What does it entail?

grahamgower:
> One thing that I'm confused about is once I've written a demographic model function, how is it incorporated into `stdpopsim`? Do I need to build something or??

After defining your model function, there needs to be a call like `_species.add_demographic_model(_your_model_func_here())`. This registers your model in the catalog for the `_species` defined in that catalog file. The `_species` is defined towards the top of the file. In case you've already done this, and it's still not showing up, then maybe it's easiest if you just open a PR with your current changes so we can see what's gone wrong.

To give a concrete example, the last model entry in `stdpopsim/catalog/homo_sapiens.py` looks like this (ellipses inserted here for brevity).
```
def _papuans_10j19():
    id = "PapuansOutOfAfrica_10J19"
    ...
    return stdpopsim.DemographicModel(...)

_species.add_demographic_model(_papuans_10j19())
```

> I also don't understand how to write the unit test. Where does it go? What does it entail?

[...]

Issue #305, opened 2019-12-09
Automated model testing and QC status


jeromekelleher:
We should make it simpler for users to add new models, that automatically get put through basic tests.

Probably, each Model object should have a ``qc_complete`` attribute that defaults to False. If  ``qc_complete`` is True, the tests will ask the ``qc`` module for a model with the same ``species/id`` key, and check if it's identical (as we do now). Thus, all the model implementer needs to do is write a function that returns the model and add it to the species.

Should be done after #122, #303 (if it's done), and possibly as part of #304. Not necessary for 0.1.0

agladstein:
Any idea on what the timeframe of this is? I was confused about the model testing - not much is currently written in the docs about it (from https://github.com/popsim-consortium/stdpopsim/pull/482)

I'm just wondering if I should elaborate on what the current process is for https://github.com/popsim-consortium/stdpopsim/issues/495, or if it makes more sense to wait for this issue to be sorted out.

grahamgower:
I don't think it will be too much work, so I'll try to get a PR submitted for this issue by the end of the week.

PR #494, opened 2020-04-26
add AJ model

agladstein:
This PR is in response to https://github.com/popsim-consortium/stdpopsim/issues/491

Not ready to be merged.

### DIFF
--- a/docs/parameter_tables/HomSap/AshkSub_7G19.csv
+++ b/docs/parameter_tables/HomSap/AshkSub_7G19.csv
@@ -1,0 +1,19 @@
+Population size,"7,300",Ancestral African pop. size
+Population size,"18,197",YRI pop. size
+Population size,"4,073",CHB pop. size
+Population size,"33,113",CEU pop. size
+Population size,"436,515",Middle Eastern pop. size
+Population size,"354,813",non-AJ Jewish pop. size
+Population size,"6,606", Western AJ pop. size
+Population size,"1,949,844", Eastern AJ founder pop. size
+Epoch Time (gen.),"8,800",Expansion time of YRI ancestral pop.
+Epoch Time (gen.),"2,105",Time of OOA event
+Epoch Time (gen.),"850",Time of CEU-CHB split
+Epoch Time (gen.),"481",Time of Middle Eastern-CEU split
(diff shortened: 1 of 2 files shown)


apragsdale:
Hi @agladstein, that looks good so far. There are a few other components to include.. I think the main one is adding a parameter table in docs/parameter_tables/ which ensures that the parameters are displayed in the documentation.

I think at this point, you don't have to worry about setting up the equivalency tests with a QC model, because whoever does the QCing will add those tests. I don't think anything else is missing, and the model looks well documented.

agladstein:
I'm still confused about how it gets incorporated. It does show up in `species.demographic_models()`

grahamgower:
That looks like it should be pulled in to the catalog just fine. Maybe you're 'import'ing a stdpopsim that isn't your development folder?

It looks like one of the unit tests in the CI is complaining about "Demographic events must be time sorted.". So I'm pretty sure your model is getting pulled into the catalog. (The time sorting won't prevent you from seeing the model in the catalog, but it will prevent you from simulating it.

agladstein:
I have not added unit tests yet. Waiting on https://github.com/popsim-consortium/stdpopsim/issues/305

There are two funny things about this model. 
1. The model parameters were inferred with ABC, and because there were so many and we didn't care about all of them, we left a few parameters as "free" parameters and did not infer them. I've noted those parameters in the model function as comments and how I'm handling them here.
2. Simulations were originally done with 2.5e-8 mutation rate. For now, I've noted that in the long description.

Dan and I also talked about adding priors as a feature to stdpopsim - and this would be a good model to use for that. So, maybe it's better to implement this model once we have a priors feature and we don't deal with the issue of including parameters that weren't inferred in the paper.

grahamgower:
Hey @agladstein, this looks like it's in pretty good shape. CircleCI is just complaining about some spaces at the ends of lines in your `long_description`. If you fix that, I think this will be good to merge.

agladstein:
fixed it.

jeromekelleher:
Can you merge if you're happy please @grahamgower?

grahamgower:
Merged in. Thanks @agladstein!

PR #496, opened 2020-04-27
Add Gutunkunst OOA QC

### LINKED ISSUE #492: HomSap/OutOfAfrica_3G09 needs QCing
Someone this model has not yet been QCed!

apragsdale:
Implemented qc version of the 2009 OOA model (to close #492). I'm having issues getting the tests to catch if I made a mistake in implementation, and I think the development docs are out of date (specifically, item 5 under Demographic model review process, perhaps)? Do I need to add anything beyond `qc_model = homo_sapiens_qc.GutenkunstOOA()` to the `TestGutenkunstThreePopOutOfAfrica` class in `test_homo_sapiens.py`?

### DIFF
--- a/tests/test_homo_sapiens.py
+++ b/tests/test_homo_sapiens.py
@@ -109,11 +109,13 @@ class TestJacobsPapuansOutOfAfrica(
     qc_model = homo_sapiens_qc.DenisovanAncestryInPapuans()
 
 
-# Models that have not been QC'd:
-
 class TestGutenkunstThreePopOutOfAfrica(
-        unittest.TestCase, test_models.CatalogDemographicModelTestMixin):
+        unittest.TestCase, test_models.QcdCatalogDemographicModelTestMixin):
     model = species.get_demographic_model("OutOfAfrica_3G09")
+    qc_model = homo_sapiens_qc.GutenkunstOOA()
+
+
+# Models that have not been QC'd:
 
 
 class TestSchiffelsZigzag(
--- a/qc/homo_sapiens_qc.py
+++ b/qc/homo_sapiens_qc.py
@@ -771,3 +771,70 @@ class DenisovanAncestryInPapuans(models.DemographicModel):
                 destination=8),
         ]
         self.demographic_events.sort(key=lambda x: x.time)
+
+
+class GutenkunstOOA(models.DemographicModel):
+    """
+    From Gutenkunst et al (2009). Parameters taken from maximum likelihood values
+    in Table 1 in that paper.
+    """
+    def __init__(self):
+        generation_time = 25
+
+        # Population sizes
+        N_A = 7300
+        N_AF = 12300
+        N_B = 2100
+        N_EU0 = 1000
+        N_AS0 = 510
+
+        # Growth rates per generation
+        r_EU = 0.4e-2
+        r_AS = 0.55e-2
+
+        # Migration rates
+        m_AF_B = 25e-5
+        m_AF_EU = 3e-5
+        m_AF_AS = 1.9e-5
+        m_EU_AS = 9.6e-5
+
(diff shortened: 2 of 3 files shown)


tests/test_homo_sapiens.py
```diff
@@ -109,11 +109,13 @@ class TestJacobsPapuansOutOfAfrica(
     qc_model = homo_sapiens_qc.DenisovanAncestryInPapuans()
 
 
-# Models that have not been QC'd:
-
 class TestGutenkunstThreePopOutOfAfrica(
         unittest.TestCase, test_models.CatalogDemographicModelTestMixin):
     model = species.get_demographic_model("OutOfAfrica_3G09")
```
grahamgower, line 114: Same issue here as for the ZigZag model, regarding the need to instead inherit from `QcdCatalogDemographicModelTestMixin`.

grahamgower:
Thanks for taking on these missing QC models @apragsdale! Should be a simple change to get them doing the checks. And when #304 and #305 are resolved, no one else will get caught on changing which superclass to inherit from!

apragsdale:
Thanks, I had forgotten that needed to be changed. Coming back to implement models a few months later really does make it clear where the docs could be filled in :) I can do some updating of the QC process documentation after these are finished up.

apragsdale:
I did find a mistake in the OOA model currently in the catalog: in demographic events, migration was not turned off between the African and Bottleneck populations after they are merged, so that migration between those populations continues into the distant past.

I don't know if this will have affected any of the analyses in the paper, for example Figure 3 from the manuscript... But if so, we should probably discuss on the call tomorrow.

grahamgower:
That's a little worrying! @jeromekelleher can you comment on the likely behaviour of msprime here? Does a `migration_matrix` with non-zero entries have any effect for lineages already "joined" by a MassMigration(proportion=1) event?

apragsdale:
It definitely will, because migration will carry lineages back into the merged population, so you'll end up with longer expected Tmrcas for lineages still uncoalesced beyond that merger event. Since it's in the distant past, the effect will likely be small, but it will still be there.

jeromekelleher:
> That's a little worrying! @jeromekelleher can you comment on the likely behaviour of msprime here? Does a migration_matrix with non-zero entries have any effect for lineages already "joined" by a MassMigration(proportion=1) event?

We don't touch the migration matrix when there's MassMigrations - decided events having side effects was confusing.

@apragsdale, does this mean that the OOA model that's in the [msprime docs](https://msprime.readthedocs.io/en/latest/tutorial.html#id1) is wrong too? (Ouch!)

apragsdale:
> @apragsdale, does this mean that the OOA model that's in the [msprime docs](https://msprime.readthedocs.io/en/latest/tutorial.html#id1) is wrong too? (Ouch!)

Oh no!! It certainly looks like it.. That's surprising that hasn't been caught before, really. I've used that example in the msprime docs a bunch of times and never noticed it either.

apragsdale:
I think this (and the Zigzag model) are ready to merge. Let me know if you need me to rebase at all again for one or both.

petrelharp:
I have no idea why Travis is unhappy here - it is something to do with conda. Help, @grahamgower? @jradrion?

grahamgower:
It's an unrelated problem, and it looks like its been failing for two weeks now. Travis is running fine for linux on python 3.6. The failing builds are for osx --- maybe somethign changed at the travis end, and the `.travis.yml` file needs to have a language set explicitly for the osx case?

jeromekelleher:
Travis build should be fixed now if you rebase @apragsdale.

apragsdale:
This and #498 are now both rebased.

petrelharp:
Yay, thanks!!!

PR #498, opened 2020-04-27
Add zigzag QC

apragsdale:
QC model for Schiffels ZigZag. However, I know there are discrepancies between the model in the catalog and the model I've implemented. For example, generation time of 29 vs 30 years (the Schiffels paper used a 30 year generation time, but the model in the catalog has a generation time of 29). But it doesn't appear that the tests have caught that difference. Do I need to do something beyond adding the `qc_model` to the `TestSchiffelsZigzag` class in `test_homo_sapiens.py` to get the comparison tests to work as expected?

### DIFF
--- a/qc/homo_sapiens_qc.py
+++ b/qc/homo_sapiens_qc.py
@@ -838,3 +838,55 @@ class GutenkunstOOA(models.DemographicModel):
             msprime.PopulationParametersChange(time=T_AF, initial_size=N_A,
                                                population_id=0)
         ]
+
+
+class ZigZag(models.DemographicModel):
+    """
+    Model from Schiffels et al (2014) used to test inference methods on a "zigzag"
+    demography. Single population, repeated growth and decay of pop size.
+    """
+    def __init__(self):
+        # from section 7 of the supplement
+        self.generation_time = 30
+        mu = 1.25e-8
+        L = 10000000
+        theta = 7156
+        Ne = theta / 4 / mu / L
+
+        self.population_configurations = [
+            msprime.PopulationConfiguration(
+                initial_size=Ne, growth_rate=0,
+                metadata={"sampling_time": 0})
+        ]
+
+        rate_changes = {
+            0.000582262: 1318.18,
+            0.00232905: -329.546,
+            0.00931619: 82.3865,
+            0.0372648: -20.5966,
+            0.149059: 5.14916,
+            # 0.596236: 0
+        }
+
+        de = []
+        for time, rate in rate_changes.items():
+            de.append(
+                msprime.PopulationParametersChange(
+                    time=time * 4 * Ne,
+                    growth_rate=rate / 4 / Ne,
+                    population_id=0
+                )
+            )
+
+        de.append(
+            msprime.PopulationParametersChange(
+                time=0.596236 * 4 * Ne,
+                growth_rate=0,
+                population_id=0,
+                initial_size=0.1 * Ne
+                )
+            )
+
+        self.demographic_events = de
+
+        self.migration_matrix = [[0]]
--- a/tests/test_homo_sapiens.py
+++ b/tests/test_homo_sapiens.py
@@ -119,5 +119,6 @@ class TestGutenkunstThreePopOutOfAfrica(
 
 
 class TestSchiffelsZigzag(
-        unittest.TestCase, test_models.CatalogDemographicModelTestMixin):
+        unittest.TestCase, test_models.QcdCatalogDemographicModelTestMixin):
     model = species.get_demographic_model("Zigzag_1S14")
+    qc_model = homo_sapiens_qc.ZigZag()
--- a/stdpopsim/catalog/homo_sapiens.py
+++ b/stdpopsim/catalog/homo_sapiens.py
@@ -765,31 +765,26 @@ def _zigzag():
             reasons={stdpopsim.CiteReason.DEM_MODEL})
     ]
 
-    generation_time = 29
+    generation_time = 30
     N0 = 14312
 
-    g_1 = 0.023025
-    t_1 = 33.333
-    n_1 = N0
+    g_1 = 0.02302578256
+    t_1 = 33.333334976
 
-    g_2 = -0.005756
-    t_2 = 133.33
-    n_2 = N0/10
+    g_2 = -0.0057564631
+    t_2 = 133.3334544
 
-    g_3 = 0.0014391
-    t_3 = 533.33
-    n_3 = N0
+    g_3 = 0.00143911578
+    t_3 = 533.33324512
 
-    g_4 = -0.00035977
-    t_4 = 2133.33
-    n_4 = N0/10
+    g_4 = -0.0003597785
+    t_4 = 2133.3352704
 
-    g_5 = 8.99448e-5
-    t_5 = 8533.33
-    n_5 = N0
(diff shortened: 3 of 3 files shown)


tests/test_homo_sapiens.py
```diff
@@ -119,3 +119,4 @@ class TestGutenkunstThreePopOutOfAfrica(
 class TestSchiffelsZigzag(
         unittest.TestCase, test_models.CatalogDemographicModelTestMixin):
     model = species.get_demographic_model("Zigzag_1S14")
+    qc_model = homo_sapiens_qc.ZigZag()
```
grahamgower, line 122: This needs to be changed to instead inherit from `QcdCatalogDemographicModelTestMixin`, which should enable the QC checking unit tests for the model.

grahamgower:
> However, I know there are discrepancies between the model in the catalog and the model I've implemented. For example, generation time of 29 vs 30 years (the Schiffels paper used a 30 year generation time, but the model in the catalog has a generation time of 29).

Is the generation time used for anything here? In other models it's just used to scale the times if they're given in years. I think that's sufficient. But maybe we want to be QCing the generation time in general, because some inferences are explicitly tied to a specific generation time assumption?

tests/test_homo_sapiens.py
apragsdale: Ok, perfect! Easy fix. Thanks.

apragsdale:
Hm, not sure why that build failed.. doesn't look like it has to do with the ZigZag model. Is that due to failed test download of a recombination map?

grahamgower:
Yeah, looks like an unrelated transient failure.

qc/homo_sapiens_qc.py
```diff
@@ -771,3 +771,89 @@ def __init__(self):
...
+
+        self.population_configurations = [
+            msprime.PopulationConfiguration(
+                initial_size=Ne, growth_rate=0,
+                metadata={"sampling_time": 0})
+        ]
+
+        # I checked these values against the catalog model by hand, and then switched
+        # to match exactly the values in the catalog model
+        # rates are set at various times
```
grahamgower, line 797: I don't quite understand the motivation here. Did your commented out code not pass QC for some reason?

apragsdale: The catalog model has sizes and times that were rounded when they were transcribed, but in my first qc approach I converted from the ms commands to times and growth rates internally. So there are small fractions of difference between the models due to math on rounded floats.

jeromekelleher:
How's this looking @apragsdale @grahamgower, ready to merge?

jeromekelleher:
(Also @apragsdale, I think you need to update your email address on your gitconfig and edit the commits here. We don't want your contributions being lost to history)

grahamgower:
It's fine with me for it to be merged, but I have one question first.

> The catalog model has sizes and times that were rounded when they were transcribed, but in my first qc approach I converted from the ms commands to times and growth rates internally. So there are small fractions of difference between the models due to math on rounded floats.

Does this, or does this not, constitute a bug to be fixed in the model? Should the model equivalence checks be relaxed?

apragsdale:
I don't think it's a bug. For example, the demographic event times were different by `33.333` vs `33.333334975999996`. I just checked my QC implementation by eye for all demographic events, before changing it. I left that bit commented in the qc model so that it can be double checked, if needed.

This model is a little bit contrived, and it's been reused a handful of times in different papers, which might explain the slight differences in implementations. I think relaxing the equivalence checks (for example using `np.isclose` instead of `==`) could lead to issues down the road, so value comparisons are probably fine as is in those tests.

There is the issue in the equivalency tests that the original and qc implementers must both make the same choices for which population index is assigned to each population going back in time. Different choices can result in equivalent models that don't pass the test, but that's a harder thing to check for I think.

apragsdale:
> (Also @apragsdale, I think you need to update your email address on your gitconfig and edit the commits here. We don't want your contributions being lost to history)

Ah - I hadn't done that on my home computer. Let me do that before you merge. I think a few other commits I made had that same issue in the past, but those probably can't be fixed?

Chris1221:
@apragsdale You're correct, I also wasn't entirely sure what to do when implementing the model. I had an initial version ([diff](https://github.com/popsim-consortium/stdpopsim/pull/105/commits/977dd26cb5531e6b83d6beebd7dccd73c0fd9469)) that looks more like a less efficient version of your commented out portion, but I convinced myself that it was difficult to read and that it was better to just be explicit with the parameters. The rounding is pretty innocuous I think, but I can see how it is an issue for the QC procedure. Perhaps the catalog model should be changed to reflect the way that you have done it, I think the issue here is pretty small and we're both transcribing from the same ms code.  Also, I believe I got the g=29 from the Terhorst 2017 (SMC++) paper, though I do agree that Schiffels uses 30. At the time that this was implemented, I believe I could only add one citation to the model object, and I'm not sure if that's changed but I put the Schiffels instead of Terhorst.

qc/homo_sapiens_qc.py
jeromekelleher: Hmm, this shouldn't be necessary. We compare the values to within a given tolerance, which should mop up the floating point errors. Perhaps the values in the real model are just put in at too low a precision? I don't think we need to worry about blinding etc here. 

Your commented out QC model is much better, I think it would be better to update the values in the main model.

jeromekelleher:
>  I think relaxing the equivalence checks (for example using np.isclose instead of ==) could lead to issues down the road, so value comparisons are probably fine as is in those tests.

We're using np.isclose() in these comparisons, so I think the simplest approach is to just increase the number of digits that we include in the production model.

apragsdale:
> We're using np.isclose() in these comparisons, so I think the simplest approach is to just increase the number of digits that we include in the production model.

Ah, whoops. Should have checked that more a bit more closely. Ok, I'll do a bit more fiddling with the catalog model to be a bit more precise. And even though the generation time isn't used in that model, I'm going to also set it to 29 years, since that was the original value used in the Schiffels citation.

apragsdale:
I have a git rebasing issue that I can't figure out. I'm trying to rebase some changes against upstream/master, which recently merged the OOA model fix. When I rebase, however, it insists on overwriting my most recent commits to the zigzag model on this branch, even though I committed those changes just moments ago. Is there a trick to pulling in the OOA changes from master but not messing up my changes on this branch to the zigzag QC?

grahamgower:
Hmm... I think I had that happen once when my tracking branch was set to track a local branch by accident and then running git pull. Maybe try changing the tracking branch? So, with your local feature branch checked out: `git branch -u upstream/master`, then to rebase do `git fetch upstream` and `git rebase -i upstream/master`.

apragsdale:
Thanks Graham.. I think it was just because there were multiple PRs affecting the same branch, and they weren't edited chronologically. No worries though, got it all merged manually, and I think it's fine now.

I added some extra digits to the catalog model so that it passes equality tests, and the generation time was changed to the original publication. Think it's pretty much ready now, unless there are other issues someone notices.

apragsdale:
Not sure why that codecov test failed.. deciphering those reports is a bit unclear to me

petrelharp:
> Not sure why that codecov test failed.. deciphering those reports is a bit unclear to me
Me, too, but it sure looks like what you did is covered.

petrelharp:
Looks good to me! Thanks!

Issue #497, opened 2020-04-27
Only run slim tests if installed


apragsdale:
What do people think about setting a flag for running the tests to _not_ test the slim engine if slim is not installed? Maybe something in `test_slim_engine.py` that tries to import slim/pyslim and skips those tests if failed, or alternatively have a flag that runs tests as normal using `python3 -m nose tests` but skips slim tests if we run `python3 -m nose tests --no_slim`.

grahamgower:
Hmm... I'm not sure about that. Is it a problem to have slim as a developer requirement? And the failing tests just add noise, right?

apragsdale:
I suppose that's not a problem. It should probably be in the requirements.txt file in that case. If you don't have slim installed though, there is a lot of scrolling and flipping through pages of failed tests to check by eye that there are no non-slim failures.

apragsdale:
Is there a problem with adding slim to the required packages in `development.txt`, as in here: #500?

jeromekelleher:
Slim is awkward here @apragsdale as it's not a Python package. We can't really automate installing it in general, I'm afraid (or at least, we most definitely don't want to try!).

I'm not overly keen on making the unit tests conditional on SLiM being installed. It would be very easy to make a mistake in CI configuration and then to discover that the SLiM tests hadn't been running for 6 months. 

It's annoying, but I think the least-bad option is assume that devs have SLiM installed. (We should document this, though, I guess)

apragsdale:
Understood. Yeah, it's not ideal and a bit annoying, since it's in external package that has to be installed and maintained separately (and might not even ever be used by some developer/user, except for running the tests), but that's a good point about the CI configuration.

I think I had just tested that pull request using `conda`, so didn't realize it wouldn't be on pypi. Gonna close this then.

PR #500, opened 2020-04-28
Add slim to required packages in development

apragsdale:
From #497.


jeromekelleher:
This won't do anything @apragsdale I'm afraid - this is a list of package on pypi, and SLiM isn't a Python package.

Issue #495, opened 2020-04-26
Clarify/varify adding model docs


agladstein:
https://github.com/popsim-consortium/stdpopsim/pull/482 made good progress on Adding Model docs. I'm working on adding a new model now and adding to the docs what I think needs to be clarified.

Related to https://github.com/popsim-consortium/stdpopsim/issues/491

PR #502, opened 2020-04-28
update adding model docs

agladstein:
I tried out adding a model for the first time following the docs. Here are some changes that I think would be helpful for a new person.

- I've mostly reorganized
- I also added in some more specific examples of where to find things and what to write

What do you think about the example function?

I've left out details on adding in a unit test until https://github.com/popsim-consortium/stdpopsim/issues/305 is complete.

closes https://github.com/popsim-consortium/stdpopsim/issues/495

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -51,7 +51,7 @@ help+wanted%22>`_
 To get started helping with ``stdpopsim`` development, please read the
 following sections to learn how to contribute.
 
-.. sec_development_installation:
+.. _sec_development_installation:
 
 ************
 Installation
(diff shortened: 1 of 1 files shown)


docs/development.rst
```diff
@@ -493,48 +581,123 @@ The demographic model should include the following:
...
+        print(x.id)
+
+    # OutOfAfrica_3G09
+    # OutOfAfrica_2T12
+    # Africa_1T12
+    # AmericanAdmixture_4B11
+    # OutOfAfricaArchaicAdmixture_5R19
+    # Zigzag_1S14
+    # AncientEurasia_9K19
+    # PapuansOutOfAfrica_10J19
```
grahamgower, line 642: It can also be checked by listing the species' demographic models with the CLI. E.g. for humans:
`stdpopsim HomSap --help-models`

docs/development.rst
```diff
@@ -493,48 +581,123 @@ The demographic model should include the following:
...
+    import stdpopsim
+    species = stdpopsim.get_species("HomSap")
+    contig = species.get_contig("chr22", length_multiplier=0.01)
+    model = species.get_demographic_model('OutOfAfrica_3G09')
+    samples = model.get_samples(10, 10, 10)
+    engine = stdpopsim.get_engine('msprime')
+    ts = engine.simulate(model, contig, samples)
+    print(model.num_populations)
+    print([pop.id for pop in model.populations])
+    print(ts.num_sites)
```
grahamgower, line 675: Maybe this could just refer to other parts of the documentation, to avoid duplication?

docs/development.rst
agladstein: with all my investigation in how set the development environment up, I could not get the CLI to update the models... so I left that out.
What do you think? I can add that in and hopefully it won't confuse someone if it doesn't work...

docs/development.rst
```diff
@@ -493,48 +581,123 @@ The demographic model should include the following:
...
+``make`` in the docs directory and opening the Catalog page from the ``docs/_build/``
+directory. See `Documentation`_ for more details.
+
+
+----------------------
+Test the model locally
+----------------------
+
+Once you have written the demographic model function, you should test the model locally
+with ``stdpopsim``. Follow the development `Installation`_ instructions to install the
```
jeromekelleher, line 623: I think it's a bit safer/more stable is we use the explicit references, like ``:ref:`sec_installation`` rather than depending on the header titles matchin.

docs/development.rst
```diff
@@ -493,48 +581,123 @@ The demographic model should include the following:
...
+You should substitute ``"HomSap"`` for which ever species you added your model to.
+Your new model should be printed along with currently available demographic models.
+
+.. note::
+
+    If your demographic model does not print, check:
+
+    - After defining your model function, did you include the call ``_species.add_demographic_model(_model_func_name())``,
+      where ``_model_func_name()`` is your model function name?
+    - Did you install ``stdpopsim`` in your development environment?
```
jeromekelleher, line 655: Not sure this is helpful - why would you install stdpopsim in your dev environment? I think it's much better for development to just run everything in the project root and avoid using ``pip install -e`` etc.

docs/development.rst
agladstein: I'm leaving this as is

agladstein:
Okay. ready :)

jeromekelleher:
Not sure why the build failed on Travis, but it's nothing to do with this, so merging. Thanks @agladstein!

PR #504, opened 2020-04-30
Update miniconda url for travis.

grahamgower:
Trying to fix travis CI build, which has been failing for two weeks.


grahamgower:
Looks like curl doesn't follow redirects by default, and the url changed. I've updated the url and asked curl to follow redircts (`-L` flag), so hopefully the problem doesn't reoccur.

jeromekelleher:
Thanks @grahamgower.

PR #509, opened 2020-05-01
more issue templates

petrelharp:
Right now the "new issue" page makes it look like the only way to submit an issue is as a QC issue; this will fix that (and give a bit more guidance).
Screenshot from 2020-05-01 08-58-24

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/bug-report-issue-template.md
@@ -1,0 +1,21 @@
+---
+name: Bug report
+about: A problem with stdpopsim
+title: 
+labels: 
+assignees: ''
+
+---
+
+*This template is a suggestion that may not fit your issue: delete/modify anything!*
+
+**Short description of the problem:**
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM. Can we get the "Bug" label to automatically populate for the "Bug Report" type @agladstein?

petrelharp:
I think we can, but I thought we should let a developer decide if it's actually a bug.

agladstein:
Yes, we can automatically label it as bug. But, I think @petrelharp is right that a developer could add that label.

Looks good to me.

grahamgower:
Looks good to me too! I tend to agree that a developer should decide if a report is an actual bug. But the "feature request" template could certainly use the `enhancement` tag.

petrelharp:
Ok, I've  added the `enhancement` tag, and sounds like I can merge this - hopefully it works as we expect... the github docs are not exhaustive.

petrelharp:
Seems to work!
Screenshot from 2020-05-02 12-01-22

PR #513, opened 2020-05-02
Automate the model testing and QC process.

### LINKED ISSUE #305: Automated model testing and QC status

grahamgower:
Model implementers and QC folks no longer need to touch unit tests.

Closes #305.

### DIFF
--- a/qc/__init__.py
+++ b/qc/__init__.py
@@ -1,5 +1,0 @@
-# Main entry point for stdpopsim_qc
-
-# Species definitions.
-from . import homo_sapiens_qc  # NOQA
-from . import drosophlia_melanogaster_qc # NOQA
--- a/stdpopsim/qc/arabidopsis_thaliana_qc.py
+++ b/stdpopsim/qc/arabidopsis_thaliana_qc.py
@@ -1,6 +1,11 @@
 import msprime
 import numpy as np
+
+import stdpopsim
 import stdpopsim.models as models
+
+
+_species = stdpopsim.get_species("AraTha")
 
 # Some generic populations to use for qc
 population_sample_0 = models.Population("sampling_0",
@@ -67,6 +72,10 @@ class Durvasula2017MSMC(models.DemographicModel):
         self.migration_matrix = [[0]]
 
 
+_species.get_demographic_model(
+        "SouthMiddleAtlas_1D17").register_qc(Durvasula2017MSMC())
+
+
 class HuberTwoEpoch(models.DemographicModel):
     populations = [
         models.Population(id="ATL", description="A. thalina"),
@@ -94,6 +103,9 @@ class HuberTwoEpoch(models.DemographicModel):
             msprime.PopulationParametersChange(
                 time=T_2, initial_size=N_ANC, population_id=0),
         ]
+
+
+_species.get_demographic_model("African2Epoch_1H18").register_qc(HuberTwoEpoch())
 
 
 class HuberThreeEpoch(models.DemographicModel):
@@ -127,3 +139,6 @@ class HuberThreeEpoch(models.DemographicModel):
             msprime.PopulationParametersChange(
                 time=T_2 + T_3, initial_size=N_ANC, population_id=0),
         ]
+
+
+_species.get_demographic_model("African3Epoch_1H18").register_qc(HuberThreeEpoch())
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -4,6 +4,7 @@ Tests for simulation model infrastructure.
 import unittest
 import itertools
 import io
+import sys
 
 import numpy as np
 import msprime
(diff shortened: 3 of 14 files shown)


qc/__init__.py
```diff
@@ -1,5 +1,8 @@
 # Main entry point for stdpopsim_qc
 
+from . qc import *  # NOQA
+
 # Species definitions.
 from . import homo_sapiens_qc  # NOQA
-from . import drosophlia_melanogaster_qc # NOQA
+from . import drosophlia_melanogaster_qc  # NOQA
+from . import arabidopsis_thaliana_qc  # NOQA
```
petrelharp, line 8: do we need to also import Pongo here?

grahamgower: Yep. But Pongo is currently disabled, and adding it here will fail some tests. If we get #363 merged first, I'll add it in. Otherwise the Pongo import will need to be added in #363.

petrelharp:
This is fantastic. Someone else should look at the details, though.

grahamgower:
In thinking this through a bit more, I figured it would be best to share what I see as two downsides of this PR (both independently fixable).

1. When running the unit tests, there's no longer an explicit mention for each of the demographic models in the catalog. If a QC fails, this is a failure in the generic `TestCatalogModels` tests, and  the ID of the failing model will appear in the assertion failure. But one could imagine a bug in this PR (or introduced subsequently) that means the model equivalence checks are not run, and this could go unnoticed for some time. To fix this gotcha, it would be possible to get something like the old behaviour by auto generating a separate test function for each model, within a single unit test class. (Or auto generate a unit test class for each model).

2. The person doing the QC needs to add a `qc_complete=True` line to the catalog model implementation. This potentially breaks the intended blinding in the QC process. Moreover, this step is potentially avoidable. Here's an alternative approach: we replace the `qc_complete` attribute with a `qc_model` attribute that defaults to None; we add a `DemographicModel.register_qc_model()` method which just checks there isn't already a qc_model, and then assigns it; the qc module gets imported in `stdpopsim/__init__.py`, so that each qc model can be registered with the appropriate catalog model. Under this alternative approach, the QC person registers the QC model with `_species.get_model("Zigzag_1S14").register_qc_model(ZigZag())` and doesn't need to look at the catalog model. Currently the qc module is not imported when stdpopsim gets imported by a user, and this approach would need to change that (adding to the import time).

Thoughts?

grahamgower:
The last two commits address (1) and (2) in my previous comment.

apragsdale:
Wow, this is really great stuff! It looks good to me, and I think it will make it a lot smoother to qc models, as long as the docs can walk you through the process (which they look to do a good job at now).

I think as long as you can tell which model is failing from the assertions in the unit tests, that all is perfectly fine. Is there an easy way to ensure that every species/model is tested, so that a test would fail if a model in the catalog that has been QC'd hasn't been tested for equivalency?

grahamgower:
> Is there an easy way to ensure that every species/model is tested, so that a test would fail if a model in the catalog that has been QC'd hasn't been tested for equivalency?

Nothing comes to mind. (At least, nothing *sane*). But I think that's slightly less likely with this PR, because it's all or nothing with model QC checks, so no model QC tests will appear in the nosetests output.

stdpopsim/qc/arabidopsis_thaliana_qc.py
```diff
@@ -61,6 +66,10 @@ def __init__(self):
         self.migration_matrix = [[0]]
 
 
+_species.get_demographic_model(
```
jeromekelleher, line 69: How about 
```python
_species.get_demographic_model( "SouthMiddleAtlas_1D17").register_qc(Durvasula2017MSMC())
```
There's a tiny bit less typing and it seems redundant to type ``qc_model`` when it's a method of the DemographicModel class, and we're associating the QC with it.

tests/test_models.py
```diff
@@ -59,11 +60,22 @@ class QcdCatalogDemographicModelTestMixin(CatalogDemographicModelTestMixin):
...
+
+# Add model specific test classes, derived from one of the above.
+for species in stdpopsim.all_species():
+    for model in species.demographic_models:
+        superclasses = [unittest.TestCase]
+        if model.qc_model is not None:
+            superclasses.append(QcdCatalogDemographicModelTestMixin)
+        else:
+            superclasses.append(CatalogDemographicModelTestMixin)
+        classname = f"Test{species.id}{model.id}"
```
jeromekelleher, line 76: Nice! Generating classes at run-time like a pro.

jeromekelleher: Could do some basic sanity checks here, like this
```python
qc_test_classes = []
for species in stdpopsim.all_species():
￼    for model in species.demographic_models:
￼        superclasses = [unittest.TestCase]
￼        if model.qc_model is not None:
￼            superclasses.append(QcdCatalogDemographicModelTestMixin)
￼        else:
￼            superclasses.append(CatalogDemographicModelTestMixin)
￼        classname = f"Test{species.id}{model.id}"
        cls = type(classname, tuple(superclasses), dict(model=model))

# Do some basic sanity checks here to double check no errors are introduced 
# leading to tests not being run silently
assert len(qc_test_classes) > 0
for cls in qc_test_classes:
    assert issubclass(cls, unittest.TestCase)
    setattr(sys.modules[__name__], cls.__name__, cls)
```

stdpopsim/qc/arabidopsis_thaliana_qc.py
grahamgower: Done!

tests/test_models.py
grahamgower: I agree this could catch some kinds of bugs, so I've added it in. I don't think we can check that tests are being run in a more general case though, because it's not clear that `nose` and/or `unittest` wont change how they find tests. In particular, they could decide to register classes upon creation using a metaclass, and I'm not sure that dynamic creation using `type` would catch that (but then, I know almost nothing about metaclasses).

To give an example that it's not necessarily obvious how tests are identified... I first tried programmatically adding `test_{species.id}_{model.id}` methods to a `TestQCModels(unittest.TestCase)` class. I did this using `setattr(TestQCModels, ...)` after the class definition. Tests defined manually within the TestQCModels class ran as normal, but none of those added with setattr ran. I briefly looked at the code for unittest.TestCase, and didn't see anything obviously tricky there, so gave up and did the dynamic-class version instead (which is probably nicer anyway).

grahamgower:
I'll wait until #429 is merged before merging this one. I don't think any conflicts will be hard to deal with, but still best avoided.

tests/test_models.py
jeromekelleher: Using metaclasses is nearly always a bad idea...

PR #520, opened 2020-05-08
Set migrations rate to zero after full population mass migration

ndukler:
Addressing issue #516

PR #521, opened 2020-05-08
Fix Tennessen two-pop migration rates in QC implementation

apragsdale:
Along with #520

### DIFF
--- a/stdpopsim/qc/homo_sapiens_qc.py
+++ b/stdpopsim/qc/homo_sapiens_qc.py
@@ -163,6 +163,8 @@ class TennessenTwoPopOutOfAfrica(models.DemographicModel):
             # Coalescence between the OOA and YRI pops
             msprime.MassMigration(
                 time=T_B, source=1, destination=0, proportion=1.0),
+            msprime.MigrationRateChange(
+                time=T_B, rate=0),
             # Change to ancestral population size pre OOA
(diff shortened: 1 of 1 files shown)


ndukler:
Looks like we both saw the same thing. @jeromekelleher can you pull in #520  then #521? Or do we have to do it simultaineously to prevent the tests from failing?

petrelharp:
Hm, it would be nice to get checks passing before merging; do you think that if these changes are made on top of #520 then checks will pass? If so, the right thing to do is probably to put this commit over at #520 also?

ndukler:
Ok, @apragsdale can you make this PR to my repo then I guess? Then I can merge it into my branch?

apragsdale:
Ok, I'll do that and close this PR

petrelharp:
Hm, I think that @ndukler meant for you to change the "base" branch of this PR to his branch, so that if he clicks "merge" it'd go into his. But, @ndukler - if `origin` is the popsim-consortium repo, and you're on your `mass_migration_bugfix` branch you can do something like:
```
git fetch origin pull/521/head:pr521
git cherry-pick pr521
```
and you should have @apragsdale's commit on top of your branch?

apragsdale:
Oh, I didn't realize I could do that. I re-opened a new PR to his branch instead, which should do the same thing?

PR #525, opened 2020-05-10
Warn when using a non-QCed model.

### LINKED ISSUE #499: Add warning when using a non-QCed model
#496 has shown that we really can't trust non-QCed models. So we definitely want a *caveat emptor*-type warning if a user runs a model before it get's QCed. Related to #304 and #305.

grahamgower:
Closes #499.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -5,6 +5,7 @@ at the command line and methods to manage resources used by stdpopsim.
 import argparse
 import json
 import logging
+import warnings
 import platform
 import sys
 import textwrap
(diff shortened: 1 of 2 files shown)


stdpopsim/cli.py
```diff
@@ -436,11 +439,22 @@ def run_simulation(args):
...
+                    "More information about the QC process can be found in "
+                    "the developer documentation. "
+                    "https://stdpopsim.readthedocs.io/en/latest/development.html"
+                    "#demographic-model-review-process")
         ts = engine.simulate(**kwargs)
         summarise_usage()
         if ts is not None:
             write_output(ts, args)
-        if not args.quiet:
+        if qc_complete and not args.quiet:
```
jeromekelleher, line 457: Not sure why you're tying QC status to citation output here - we'll still have citations for other stuff that has been QC'd like the species definition. Are you thinking that "you shouldn't be using this in a publication anyway", so there's no point in writing out the citation info? Or maybe the warning gets lost in the output with all the cites?

A comment would be helpful I think.

grahamgower: Immediately following the warning, the citation information would print, which starts with `If you use this simulation in published work, please cite:`. Which seems a little odd, given that we just warned against using the model. So, yeah, I kinda figured the user shouldn't be using non-QCed models in a publication. Happy to add a comment in the code to this effect.

Issue #517, opened 2020-05-08
Update devdocs to detail data upload to AWS


ndukler:
What the title says. (Reminder to self to do this)

ndukler:
@grahamgower @jeromekelleher Is there somewhere in particular I should put this info? I don't see a natural spot in the documentation as it doesn't seem like there's a place where adding a new species/genetic map is documented.

grahamgower:
The "how to add a species" docs don't yet exist. Maybe you could add a placeholder title, with a "TODO", and add your bits as a subsection of this?

ndukler:
Sounds good!

ndukler:
I actually just decided to wrote the docs for implementing a new species. That being said as I've never actually done it they might need some refining.

ndukler:
I also really hate the sphinx `unexpected indentation` error message in the python code blocks. I cannot for the life of me figure out why it's unhappy and sometimes it just goes away after I do random stuff.

grahamgower:
Closed by #534.

PR #534, opened 2020-05-23
Documentation for adding a species to stdpopsim

ndukler:
Added documentation for adding a new species. Can someone who had actually done this take a look to make sure the details are correct and laid out in sufficient depth? @agladstein @apragsdale @grahamgower . Addresses #517 as well.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -448,6 +448,166 @@ finally force-push to the remote topic branch on your fork::
 Hard resetting and force pushing are not reversible operations, so please
 beware!
 
+********************
+Adding a new species
+********************
+To add a new species to `stdpopsim` several things are required:
+1. The genome definition
+2. Default species parameters
+3. A genetic map with local recombination rates (optional)
+
+Once you have these things the first step is to create a new file in the `catalog`
+directory named for the species (see `Naming conventions`_ for more details). All
+code described below should go in this file unless explicitly specified otherwise.
+
+--------------------------
+Default species parameters
+--------------------------
+
+Four default parameters are required to create a new species:
+1. Generation time estimate
+2. Mutation rate
+3. Recombination rate
+4. Characteristic population size
+
+These parameters should be based on what values might be drawn from a typical population
+as represented in the literature for that species. Consequently one or more citations for
+each value are expected and will be required for constructing the species object detailed
+below.
+
+--------------------------
+Adding a genome definition
+--------------------------
+A genome definition is created with a call to `stdpopsim.Genome()`  which requires a list
+of chromosomes and a citation for the assembly. This list is typically created as follows
+using genome-wide recombination and mutation rates.
+
+.. code-block:: python
+
+    # Create a string of chromosome lengths for easy parsing
+    _chromosome_data = """\
+    chr1 CHROMOSOME1_LENGTH
+    chr2 CHROMOSOME2_LENGTH
+    """
+    # Parse list of chromosomes into a list of Chromosome objects which contain the
+    # chromosome name, length, mutation rate, and recombination rate
+    _chromosomes = []
+
+    for line in _chromosome_data.splitlines():
+        name, length = line.split()[:2]
+        _chromosomes.append(stdpopsim.Chromosome(
+        id=name, length=int(length),
+        mutation_rate=FILL_ME,
+        recombination_rate=FILL_ME))
+
+    # A citation for the chromosome parameters. Additional citations may be needed if
+    # the mutation or recombination rates come from other sources. In that case create
+    # additional citations with the appropriate reasons specified (see API documentation
+    # for stdpopsim.citations)
+
+    _assembly_citation = stdpopsim.Citation(
+        doi="FILL ME",
+        year="FILL ME",
+        author="Author et al.",
+        reasons={stdpopsim.CiteReason.ASSEMBLY})
+
+    # Create a genome object
+
+    _genome = stdpopsim.Genome(
+        chromosomes=_chromosomes,
+        assembly_citations=[_assembly_citation])
+
+Once you have a genome object you can create a new `Species` object which contains
+species identifiers, the genome, and default generation time and population size settings
+along with the relevant citation(s). Below is an example species definition for
+Arabidopsis thaliana and a final line of code that registers the species in the catalog.
+
+.. code-block:: python
+
+    _gen_time_citation = stdpopsim.Citation(
+        doi="https://doi.org/10.1890/0012-9658(2002)083[1006:GTINSO]2.0.CO;2",
+        year="2002",
+        author="Donohue",
+        reasons={stdpopsim.CiteReason.GEN_TIME})
+
+    _pop_size_citation = stdpopsim.Citation(
+            doi="https://doi.org/10.1016/j.cell.2016.05.063",
+            year="2016",
+            author="1001GenomesConsortium",
+            reasons={stdpopsim.CiteReason.POP_SIZE})
+
+    _species = stdpopsim.Species(
+        id="AraTha",
+        name="Arabidopsis thaliana",
+        common_name="A. thaliana",
+        genome=_genome,
+        generation_time=1.0,
+        generation_time_citations=[_gen_time_citation],
+        population_size=10**4,
+        population_size_citations=[_pop_size_citation]
+        )
+
+    stdpopsim.register_species(_species)
+
+Once all of this is done, go to the `catalog/__init__.py` file and add a line like the
+one below using the six-letter species identifier. Make sure to keep the comment to
+prevent linting issues.
+
+.. code-block:: python
+
+    from .catalog import PonAbe  # NOQA
+
+--------------------
+Adding a genetic map
+--------------------
+Some species have sub-chromosomal recombination maps available. They can be added to
+`stdpopsim` by creating a new `GeneticMap` object and providing a formatted file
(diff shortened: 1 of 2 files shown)


docs/development.rst
```diff
@@ -448,6 +448,145 @@ finally force-push to the remote topic branch on your fork::
...
+Some species have sub-chromosomal recombination maps available. If there is one for your
+species that you wish to include it is specified in a space delimited file with four
+columns: Chromosome, Position(bp), Rate(cM/Mb), and Map(cM). Each chromosome should be
+placed in a seperate file and with the chromosome id in the file name in such a way that
+it can be programatically parsed out. IMPORTANT: chromosome ids must match those provided
+in the genome definition exactly! Below is an example start to a recombination map file::
+
+    Chromosome Position(bp) Rate(cM/Mb) Map(cM)
+    chr1 32807 5.016134 0
+    chr1 488426 4.579949 0
```
apragsdale, line 559: The map might still work (if msprime just uses the Rate instead of Map column, but the Map should be increasing above 0 with positive rate?

docs/development.rst
```diff
@@ -448,6 +448,145 @@ finally force-push to the remote topic branch on your fork::
...
+
+Once you have the recombination map files formatted, tar and gzip them into a single
+compressed archive. This file will be sent to one of the `stdpopsim` uploaders for
```
apragsdale, line 562: Might be good to say how the uploader would be contacted and how to coordinate that? Perhaps, opening an issue would work?

apragsdale:
This looks good to me. Nice work @ndukler

agladstein:
Nice work!
I'm not sure on what content is correct since I haven't added a species before. I do have a few minor cosmetic comments. If you want, I can do a review with the cosmetic comments, or we can merge and I can just do a second PR with the cosmetic changes.

One thing that I was confused about when reading the directions for the first time was the genetic map instructions. Since so far, all changes to stdpopsim are done via github it wasn't immediately obvious that the genetic map isn't done though GitHub. I think you should say in the first sentence that genetic maps are uploaded to AWS and there are people that will give you instructions on what to do to get the genetic map in. And then describe how the files should be. Because I was initially looking around in the repo for where the genetic map files go.

Another thing that was a little weird to me was that you list out the things required to add a species, but then only 2 of them have headers. Maybe add a third section called "Species Parameters", and in the initial list you can say: 
>...several things are required: 1. The genome definition 2. Species parameters (including: generation time estimate, a mutation rate, recombination rate, and a typical population size) with citations 3. A genetic map with local recombination rates (optional)

docs/development.rst
grahamgower: It might be simplest to point at the [msprime docs](https://msprime.readthedocs.io/en/stable/api.html#msprime.RecombinationMap.read_hapmap) here instead of providing such detail. Stdpopsim does just call `msprime.RecombinationMap.read_hapmap()` after all.

docs/development.rst
```diff
@@ -448,6 +448,145 @@ finally force-push to the remote topic branch on your fork::
...
+            year=9999,
+            reasons={stdpopsim.CiteReason.GEN_MAP})
+
+    _gm = stdpopsim.GeneticMap(
+        species=_species,
+        id="FILL_ME", # ID for genetic map, see naming conventions
+        description="FILL_ME",
+        long_description="FILL_ME",
+        url=("https://stdpopsim.s3-us-west-2.amazonaws.com/genetic_maps/dir/filename"),
+        file_pattern="name_{id}_more_name.txt", # a pattern the extracts the chomosome id
```
grahamgower, line 581: The comment here is a little confusing, I think. (It looks like the GeneticMap docstring for this field is somewhat awkwardly worded too.) How about `recombination map filename, where '{id}' will be replaced with the 'id' field of a given chromosome.`

grahamgower:
This looks really good, thanks @ndukler. I don't think we need to worried if there's something small missing, as we'll certainly poke the next species implementer about potential docs additions.

docs/development.rst
ndukler: I'll do that

docs/development.rst
```diff
@@ -448,6 +448,145 @@ finally force-push to the remote topic branch on your fork::
...
+
+    stdpopsim.register_species(_species)
+
+Once all of this is done, go to the `catalog/__init__.py` file and add a line like the
+one below using the six-letter species identifier. Make sure to keep the comment to
+prevent linting issues.
+
+.. code-block:: python
+
+    from .catalog import PonAbe  #NOQA
```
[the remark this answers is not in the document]
ndukler: Fixed

docs/development.rst
ndukler: Ok, after reading the `msprime` documentation it doesn't actually help explain much. I don't understand this bit: "The first column in this file is ignored, as are additional columns after the third (Position is assumed to be the second column, and Rate is assumed to be the third)". If that's the case why have the 4th column in the example? Also how is the first column ignored, it's essential for creating the per chromosome maps, right? I'll link to it but I think it needs some clarification.

ndukler:
> Nice work!
> I'm not sure on what content is correct since I haven't added a species before. I do have a few minor cosmetic comments. If you want, I can do a review with the cosmetic comments, or we can merge and I can just do a second PR with the cosmetic changes.
> 
> One thing that I was confused about when reading the directions for the first time was the genetic map instructions. Since so far, all changes to stdpopsim are done via github it wasn't immediately obvious that the genetic map isn't done though GitHub. I think you should say in the first sentence that genetic maps are uploaded to AWS and there are people that will give you instructions on what to do to get the genetic map in. And then describe how the files should be. Because I was initially looking around in the repo for where the genetic map files go.
> 
> Another thing that was a little weird to me was that you list out the things required to add a species, but then only 2 of them have headers. Maybe add a third section called "Species Parameters", and in the initial list you can say:
> 
> > ...several things are required: 1. The genome definition 2. Species parameters (including: generation time estimate, a mutation rate, recombination rate, and a typical population size) with citations 3. A genetic map with local recombination rates (optional)

Edits made, thanks!

docs/development.rst
grahamgower: I think this file format already existed before msprime started using it. So the unused columns might be useful in some contexts, but they just aren't needed in msprime (or stdopsim). In the unlikely event that one file contains recombination maps for all chromosomes, then this can be split using the first column of the file. But that would need to be done outside of msprime/stdpopsim.

grahamgower:
Hey @ndukler, I'm not sure we resolved what to do about the recombination map description during the call today. Do you want to update the PR to point at the msprime docs, under the assumption that the description there will be clarified? Or would you prefer to leave it as is? Certainly I agree with @apragsdale that its a little funny to have a column full of zeros for the map units, even if the column is ignored by msprime/stdpopsim.

ndukler:
> Hey @ndukler, I'm not sure we resolved what to do about the recombination map description during the call today. Do you want to update the PR to point at the msprime docs, under the assumption that the description there will be clarified? Or would you prefer to leave it as is? Certainly I agree with @apragsdale that its a little funny to have a column full of zeros for the map units, even if the column is ignored by msprime/stdpopsim.

Let's just leave it as is for now. I mean, I have described the two essential columns so I think it's ok. That being said I don't understand how msprime ignores the first column. How does it handle chromosomes otherwise?

apragsdale:
The way I understand it currently, msprime only allows you to pass a single recombination map, so it ignores the first column (which would normally tell you which chromosome the map is for). It gets the simulation sequence length from the length of the map, so you wouldn't need to pass the length that you want to simulate if you've given a map. The last column is some cumulative distribution of expected number of recombination events up to that base pair (x100, since cM), and it's true that the information is redundant to the 3rd column - i.e. you could figure out what the last column should be from columns 2 and 3.

I was just suggesting filling in that column because as it there is a mismatch between columns 3 and 4. But as Graham suggested, it's probably best to just link here (https://msprime.readthedocs.io/en/stable/api.html?highlight=recombination%20map#msprime.RecombinationMap.read_hapmap) and not go into too much detail here since it's redundant to the msprime docs?

ndukler:
Think this is ready to merge though. We can always make slight adjustments if need be. Closes #517

ndukler:
> The way I understand it currently, msprime only allows you to pass a single recombination map, so it ignores the first column (which would normally tell you which chromosome the map is for). It gets the simulation sequence length from the length of the map, so you wouldn't need to pass the length that you want to simulate if you've given a map. The last column is some cumulative distribution of expected number of recombination events up to that base pair (x100, since cM), and it's true that the information is redundant to the 3rd column - i.e. you could figure out what the last column should be from columns 2 and 3.
> 
> I was just suggesting filling in that column because as it there is a mismatch between columns 3 and 4. But as Graham suggested, it's probably best to just link here (https://msprime.readthedocs.io/en/stable/api.html?highlight=recombination%20map#msprime.RecombinationMap.read_hapmap) and not go into too much detail here since it's redundant to the msprime docs?

Sure, I think already linked to it so I think we're good.  Thanks for clarifying!

grahamgower:
> That being said I don't understand how msprime ignores the first column. How does it handle chromosomes otherwise?

Msprime doesn't handle chromosomes. Typically with msprime, you simulate just one genomic interval, which might implicitly correspond to a chromosome. So in stdpopsim you load up a hapmap file corresponding to the chromosome you want (based on its filename), and simulate with that.

Anyway, I'm going to merge this so the bikeshedding can stop!

PR #547, opened 2020-05-27
Change --quiet to a top-level CLI option

### LINKED ISSUE #515: ``download-genetic-maps`` cli broken
```
$ python3 -m stdpopsim -vv download-genetic-maps
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/jk/work/github/stdpopsim/stdpopsim/__main__.py", line 4, in <module>
    cli.stdpopsim_main()
  File "/home/jk/work/github/stdpopsim/stdpopsim/cli.py", line 619, in stdpopsim_main
    setup_logging(args)
  File "/home/jk/work/github/stdpopsim/stdpopsim/cli.py", line 63, in setup_logging
    if args.quiet:
```

jeromekelleher:
Fixes #515

### DIFF
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -144,8 +144,8 @@ class TestAPI(unittest.TestCase):
 class TestCLI(unittest.TestCase):
 
     def docmd(self, _cmd):
-        cmd = ("-e slim --slim-scaling-factor 20 --slim-burn-in 0 "
-               f"{_cmd} -l 0.001 -c chr1 -s 1234 -q 10").split()
+        cmd = ("-q -e slim --slim-scaling-factor 20 --slim-burn-in 0 "
+               f"{_cmd} -l 0.001 -c chr1 -s 1234 10").split()
         return capture_output(stdpopsim.cli.stdpopsim_main, cmd)
 
     def test_script_generation(self):
@@ -180,8 +180,8 @@ class TestCLI(unittest.TestCase):
 
         # verify sample counts for a multipopulation demographic model
         with tempfile.NamedTemporaryFile(mode="w") as f:
-            cmd = ("-e slim --slim-scaling-factor 20 --slim-burn-in 0 "
-                   f"HomSap -o {f.name} -l 0.001 -c chr1 -s 1234 -q "
+            cmd = ("-q -e slim --slim-scaling-factor 20 --slim-burn-in 0 "
+                   f"HomSap -o {f.name} -l 0.001 -c chr1 -s 1234 "
                    "-d OutOfAfrica_3G09 0 0 8").split()
             capture_output(stdpopsim.cli.stdpopsim_main, cmd)
             ts = tskit.load(f.name)
@@ -194,7 +194,8 @@ class TestCLI(unittest.TestCase):
         self.assertEqual(observed_counts[2], 8)
         self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees()))
 
-    def test_dry_run(self):
+    @mock.patch("stdpopsim.slim_engine._SLiMEngine.get_version", return_value="64")
+    def test_dry_run(self, _mocked_get_version):
         # --dry-run should run slim, but not create an output file.
         with mock.patch("subprocess.Popen", autospec=True) as mocked_popen:
             # Popen is used as a context manager, so we frob the return value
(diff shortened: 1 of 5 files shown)


tests/test_slim_engine.py
```diff
@@ -194,6 +194,7 @@ def test_simulate(self):
         self.assertEqual(observed_counts[2], 8)
         self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees()))
 
+    @unittest.skip("Test broken: something wrong with popen mock?")
```
jeromekelleher, line 197: Not sure what happened here @grahamgower - I couldn't figure out why this test was after breaking.

grahamgower: Looking at it now...

grahamgower: There is something funny going on with the mock.patch, and I'm not sure what that is. I guess autospeccing here is insufficient. In any event, the mocked `subprocess.Popen` is also used internally by `subprocess.check_output()` (called in the `SLiMEngine.version()` method). But the engine version should not be requested when running with --quiet.

It seems that with this PR, `args.quiet` is always None, regardless of the --quiet flag. So the existing checks for args.quiet will need to be changed, or args.quiet restored.

jeromekelleher: Ah! OK, I'll nuke any references to ``quiet`` - there should only be ``verbose`` now.

jeromekelleher: Should be fixed now - I changed to mocking subprocess.check_call. Seems a lot simpler?

jeromekelleher:
Ready for review now @grahamgower

tests/test_slim_engine.py
grahamgower: I don't think that's testing the right thing. The slim engine calls `Popen` to nominally run the slim executable when doing a simulation (or a dry run). The engine only uses `check_output` to get the version string.

grahamgower:
I guess I don't have access to push to your branch. Here's a diff against it that should do the trick though. It just mocks out the get_version call too.
```
diff --git a/tests/test_slim_engine.py b/tests/test_slim_engine.py
index 1246689..4bb2e54 100644
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -3,6 +3,7 @@ Tests for SLiM simulation engine.
 """
 import os
 import re
+import io
 import sys
 import unittest
 import tempfile
@@ -193,14 +194,20 @@ class TestCLI(unittest.TestCase):
         self.assertEqual(observed_counts[2], 8)
         self.assertTrue(all(tree.num_roots == 1 for tree in ts.trees()))
 
-    def test_dry_run(self):
+    @mock.patch("stdpopsim.slim_engine._SLiMEngine.get_version", return_value="64")
+    def test_dry_run(self, _mocked_get_version):
         # --dry-run should run slim, but not create an output file.
-        with mock.patch(
-                "subprocess.check_output", autospec=True) as mocked_check_output:
+        with mock.patch("subprocess.Popen", autospec=True) as mocked_popen:
+            # Popen is used as a context manager, so we frob the return value
+            # of the returned context manager, rather than the Popen mock itself.
+            proc = mocked_popen.return_value.__enter__.return_value
+            proc.returncode = 0
+            proc.stdout = io.StringIO()
+            proc.stderr = io.StringIO()
             with tempfile.NamedTemporaryFile(mode="w") as f:
                 self.docmd(f"HomSap --dry-run -o {f.name}")
-        mocked_check_output.assert_called_once()
-        self.assertTrue("slim" in mocked_check_output.call_args[0][0])
+        mocked_popen.assert_called_once()
+        self.assertTrue("slim" in mocked_popen.call_args[0][0])
         with tempfile.NamedTemporaryFile(mode="w") as f:
             self.docmd(f"HomSap --dry-run -o {f.name}")
             self.assertEqual(os.stat(f.name).st_size, 0)
```

jeromekelleher:
> I guess I don't have access to push to your branch. Here's a diff against it that should do the trick though. It just mocks out the get_version call too.

I don't get it - how is this better then mocking ``check_output``?

grahamgower:
> > I guess I don't have access to push to your branch. Here's a diff against it that should do the trick though. It just mocks out the get_version call too.
> 
> I don't get it - how is this better then mocking `check_output`?

The test needs to mock `Popen` in order to check that the slim binary is being called in the slim engine's `simulate()` method. Mocking `check_output` doesn't do that, because `check_output` isn't used to run slim simulations. `check_output` is only used when calling "slim -v" to get the version.

jeromekelleher:
Ahhh, very good. Thanks, patch applied and ready to merge when CI passes.

PR #549, opened 2020-05-27
Warn when using --slim-scaling-factor with Q!=1.

### LINKED ISSUE #455: add warning when using Q != 1
Proposed rough text:

> You are using rescaled populations (setting Q not equal to 1). This should give similar results for many situations, but is not equivalent, especially in the presence of lots of selection. To use rescaling, you should be careful - do checks and compare results across different values of Q.

grahamgower:
Closes #455.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -444,7 +444,7 @@ def add_simulate_species_parser(parser, species):
         write_simulation_summary(engine=engine, model=model, contig=contig,
                                  samples=samples, seed=args.seed)
         if not qc_complete:
-            warnings.warn(
+            warnings.warn(stdpopsim.QCMissingWarning(
                     f"{model.id} has not been QCed. Use at your own risk! "
                     "Demographic models that have not undergone stdpopsim's "
                     "Quality Control procedure may contain implementation "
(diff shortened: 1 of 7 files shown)


grahamgower:
Oh, you beat me @jeromekelleher with your other PR. I'll update this one to move the -q flag around in the new unit test if yours gets merged first.

stdpopsim/cli.py
```diff
@@ -450,6 +450,14 @@ def run_simulation(args):
                     "the developer documentation. "
                     "https://stdpopsim.readthedocs.io/en/latest/development.html"
                     "#demographic-model-review-process")
+        if args.slim_scaling_factor != 1 and not args.quiet:
```
jeromekelleher, line 453: Shouldn't we be doing this in the ``slim_engine.py`` file? Although I see your point that it would be irritating if this raised a warning any time someone uses Q != 1. Presumably we'd want to warn when the scaling factor is extreme (but I've no idea how you'd determine this).

grahamgower: I considered putting it in the slim engine simulate() method, but I'd rather not get the warning from the Python API as well as the CLI. In particular, I don't want to silence all the other potential warnings which I think are far more informative.

grahamgower:
I changed it to so the warning is in slim_engine.py. I also made specific warning categories for several of our warnings. This makes the unit tests less fragile for when we add new warnings, and also means that specific warnings can be filtered/ignored by users of the Python API.

grahamgower:
Feel free to merge this one if/when ci finishes @jeromekelleher. I'm off for zzzz.

Issue #459, opened 2020-04-03
generic models have no `generation_time`, so can't be run with the slim engine


petrelharp:
For instance:
```
import stdpopsim
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("chr22")
model = stdpopsim.PiecewiseConstantSize(1e4)
samples = model.get_samples(200)
engine = stdpopsim.get_engine("slim")
ts = engine.simulate(model, contig, samples, slim_burn_in=0.1)
```
produces
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/peter/.local/lib/python3.7/site-packages/stdpopsim-0.1.2.dev72+ng59af5ea-py3.7.egg/stdpopsim/slim_engine.py", line 627, in simulate
  File "/home/peter/.local/lib/python3.7/site-packages/stdpopsim-0.1.2.dev72+ng59af5ea-py3.7.egg/stdpopsim/slim_engine.py", line 296, in slim_makescript
  File "/home/peter/.local/lib/python3.7/site-packages/stdpopsim-0.1.2.dev72+ng59af5ea-py3.7.egg/stdpopsim/slim_engine.py", line 296, in <listcomp>
AttributeError: 'PiecewiseConstantSize' object has no attribute 'generation_time'
```
Maybe the call to `PiecewiseConstantSize( )` needs another argument, `species`, so it can inherit properties like `generation_time`? Or, maybe we just need a `generation_time` argument to either `PiecewiseConstantSize( )` or `_SLiMEngine.simulate( )`?

grahamgower:
I've just been setting that property of the model manually (like `cli.py` does when no demographic model is requested). Not sure what the best solution here is.

jeromekelleher:
What do you set the value to @grahamgower? Perhaps we should make ``generation_time`` an attribute that is shared by all models, and is None by default. Therefore, if ``generation_time`` doesn't make sense but the engine has to set it, the engine can make the choice?

grahamgower:
I set `model.generation_time = species.generation_time`. Certainly, we could move that into the slim engine's `simulate()` method.

grahamgower:
Oh, actually, we can't. The engine doesn't get the species object.

jeromekelleher:
I guess having the species as an optional second argument to ``PiecewiseConstantSize`` and other generic models is the best thing to do. Can you see any drawbacks?

grahamgower:
[...]

PR #553, opened 2020-05-28
Changelog for 0.1.2.

### LINKED ISSUE #551: Write changelog
Need to write a changelog for 0.1.2.

jeromekelleher:
Closes #551

I'll release tomorrow assuming there's no hiccups.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,12 +1,67 @@
 --------------------
-[0.1.2] - 2020-XX-XX
+[0.1.2] - 2020-05-29
 --------------------
 
+Important bugfix and feature release, recommended for all users.
+
+Significant errors in the HomSap/OutOfAfrica_3G09 and HomSap/OutOfAfrica_2T12
+models have been fixed. **We recommend repeating any analyses performed using
+these models**. See `here
+<https://github.com/jeromekelleher/msprime-model-errors>`__ for more details on
+the error in the three population Out of Africa model and analysis of the
+differences from the correct model.
+
+The recombination rate for AraTha was also off by a factor of 10.
+**We recommend repeating any analyses performed using this species**.
 
 **Bug fixes**:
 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Quite a lot in here, so it would be good to get some eyes on the changelog to make sure I didn't miss anything (apologies if I did!)

We should get into the habit of updating the changelog as we go from here I think - it's quite error prone (not to mention tedious) going back through the PR list.

grahamgower:
The PonAbe addition seems to be missing. Also, perhaps we could have separate headings for "new species", "new demographic models", and "new features"?

jeromekelleher:
> The PonAbe addition seems to be missing. Also, perhaps we could have separate headings for "new species", "new demographic models", and "new features"?

Good call on both counts, should be fixed now.

ndukler:
Dunno if we mention it because it's internal but we also added a QC check for sampling times in PR #429

ndukler:
Also I noticed that generation time was set to 1 for generics as per #459 although this was largely for the SLiM engine so again I don't know it it should be added

agladstein:
What belongs in a changelog? Because I'm seeing a lot of little commits that don't seem to fit into what you have already in the changelog.

jeromekelleher:
> What belongs in a changelog? Because I'm seeing a lot of little commits that don't seem to fit into what you have already in the changelog.

I think a changelog should contain the things that a user needs to know about that are *different to the last release*. So, all the stuff that happened on the slim engine isn't relevant because it didn't exist in the last release. It's a quick summary that a user can look at and know (a) is there cool new stuff that I should be using and (b) is this release going to break my code.

jeromekelleher:
> Dunno if we mention it because it's internal but we also added a QC check for sampling times in PR #429

This is more internal development infrastructure to me @ndukler, and not something that an end-user needs to know about.

PR #560, opened 2020-06-11
Annotation WIP

andrewkern:
Hey all-

I've sketched out what i think are the main bits of the infrastructure for dealing with annotation. I've aimed to mirror what we do with genetic maps, although I figure if we do this right we don't have to host the data at all, and instead we can rely on Ensembl (or similar) as I'm doing now. 

The basic idea is that annotations will be downloaded to the cache as a GFF file. Annotations will have citations, ids, etc, just as do genetic_maps. Handling of the actual annotation files is currently done using [pybedtools](https://daler.github.io/pybedtools/), although I haven't build anything in other than a function that gets the annotations from a chromosome (we will need to make this useful with SLiM-- tagging @mufernando to help me).

Currently I only have annotations defined for HomSap. I will build this out further for the rest of the catalog once people have a look over this code. 

thanks.

### DIFF
--- a/maintenance/annotation_maint.py
+++ b/maintenance/annotation_maint.py
@@ -1,0 +1,38 @@
+import allel
+import zarr
+import numpy as np
+import stdpopsim as stp
+import logging
+import warnings
+import urllib.request
+import os
+
+logger = logging.getLogger(__name__)
+# make root directory for zarr annotations
+annot_path = "annotations"
+os.mkdir(annot_path)
+# loop through species and download
+for spc in stp.all_species():
+    if spc.annotations:
+        address = spc.annotations[0].url
+        genome_version = os.path.basename(address).split(".")[1]
+        logger.info(f"Downloading GFF file {spc.id}")
+        tmp_path = f"{spc.id}.tmp.gff.gz"
+        try:
+            x, y = urllib.request.urlretrieve(address, tmp_path)
+        except FileNotFoundError:
+            warnings.warn("can't connnect to url")
+        logger.info(f"creating zarr arrays {spc.id}")
+        # create zarr store and zarr root
+        spc_path = os.path.join(annot_path, spc.id+"."+genome_version+".zip")
+        store = zarr.ZipStore(spc_path)
+        root = zarr.group(store=store, overwrite=True)
+        x = allel.gff3_to_dataframe(tmp_path)
+        for col_name in x.columns:
+            if x[col_name].dtype == 'O':
+                tmp = root.array(col_name, np.array(x[col_name], dtype=str))
+            else:
+                tmp = root.array(col_name, np.array(x[col_name]))
+        # cleanup
+        os.unlink(tmp_path)
+        store.close()
--- a/requirements/conda-minimal.txt
+++ b/requirements/conda-minimal.txt
@@ -5,3 +5,5 @@ tskit
 humanize
 attrs
 appdirs
+pandas
+zarr
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -14,3 +14,8 @@ attrs
 appdirs
 humanize
 pyslim>=0.401
+pandas
+numpy
+cython
+scikit-allel[full]
+zarr>=2.4
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,8 @@ setup(
     },
     # NOTE: make sure this is the 'attrs' package, not 'attr'!
     install_requires=["msprime>=0.7.1", "attrs>=19.1.0", "appdirs", "humanize",
-                      "pyslim>=0.401"],
+                      "pyslim>=0.401", "pandas", "scikit-allel", "zarr",
+                      "numpy"],
     url='https://github.com/popsim-consortium/stdpopsim',
     project_urls={
         'Bug Reports': 'https://github.com/popsim-consortium/stdpopsim/issues',
--- a/stdpopsim/annotations.py
+++ b/stdpopsim/annotations.py
@@ -1,0 +1,164 @@
+"""
+Infrastructure for defining information about genome annotation.
+"""
+import logging
+import attr
+import pandas
+import pathlib
+import warnings
+import os
+import urllib.request
+from . import cache
+import zarr
+import tempfile
+
+logger = logging.getLogger(__name__)
+
+
+def zarr_to_dataframe(path):
+    """
+    converts zarr annotation file to
+    pandas dataframe for manipulations
+    """
+    z = zarr.open(path)
+    df = pandas.DataFrame(
+        {'seqid': z['seqid'],
+         'source': z['source'],
+         'type': z['type'],
+         'start': z['start'],
+         'end': z['end'],
+         'score': z['score'],
+         'strand': z['strand'],
+         'phase': z['phase']}
+    )
+    return df
+
+
+@attr.s
+class Annotation(object):
+    """
+    Class represnting a Annotation file
+    assume GFF3/GTF or similar
+
+    :ivar url: The URL where the packed and compressed GTF can be found
+    :vartype url: str
+    :ivar species_id: species id
+    :vartype id: str
+    :ivar species: a `stdpopsim.species` instance
+    :ivar annotation_description: description of annotation file
+    :vartype annotation_description: str
+    """
+    url = attr.ib(default=None)
+    zarr_url = attr.ib(default=None)
+    species = attr.ib(default=None)
+    id = attr.ib(default=None)
+    file_name = attr.ib(default=None)
+    description = attr.ib(default=None)
+    citations = attr.ib(factory=list)
+    long_description = attr.ib(default=None)
+
+    def __attrs_post_init__(self):
+        self.file_name = os.path.basename(self.zarr_url)
+
+    @property
+    def annot_cache_dir(self):
+        return pathlib.Path(cache.get_cache_dir()) / "annotations"
+
+    @property
+    def species_cache_dir(self):
+        return self.annot_cache_dir / self.species.id
+
+    def __str__(self):
+        s = "GTF Annotation:\n"
+        s += "\tspecies   = {}\n".format(self.species.name)
+        s += "\tid        = {}\n".format(self.id)
+        s += "\turl       = {}\n".format(self.url)
+        s += "\tzarr url       = {}\n".format(self.zarr_url)
+        s += "\tcached    = {}\n".format(self.is_cached())
+        s += "\tcache_dir = {}\n".format(self.species_cache_dir)
+        return s
+
+    def is_cached(self):
+        """
+        Returns True if this annotation is cached locally.
+        """
+        return os.path.exists(self.species_cache_dir)
+
+    def download(self):
+        """
+        Downloads the zarr from the source URL and stores it in the
+        cache directory. If the annotation directory already exists it is first
+        removed.
+        """
+        self.file_name = os.path.basename(self.zarr_url)
+        if self.is_cached():
+            logger.info(f"Clearing cache {self.species_cache_dir}")
+            with tempfile.TemporaryDirectory() as tempdir:
+                dest = pathlib.Path(tempdir) / "will_be_deleted"
+                os.rename(self.annot_cache_dir, dest)
+        logger.debug(f"Checking species cache directory {self.species_cache_dir}")
+        os.makedirs(self.species_cache_dir, exist_ok=True)
+        download_file = f'{self.species_cache_dir}/{self.file_name}'
+        logger.info(f"Downloading Zarr file '{self.id}' from {self.zarr_url}")
+        logger.info(f"download_file: {download_file}")
+        logger.info(f"species_cache_dir: {self.species_cache_dir}")
+        if os.path.exists(download_file):
+            warnings.warn("multiple downloads?")
+        try:
+            urllib.request.urlretrieve(self.zarr_url, filename=download_file)
+        except urllib.error.URLError:
+            print(f"could not connect to {self.zarr_url}")
+            raise
+        logger.debug("Download Zarr complete")
+        logger.info(f"Storing Zarr in {self.species_cache_dir}")
+
+    def get_chromosome_annotations(self, id):
+        """
+        Returns the pandas dataframe for
+        the chromosome with the specified id.
+        """
+        if not self.is_cached():
+            self.download()
+        annot_file = os.path.join(self.species_cache_dir, self.file_name)
+        if id is None:
+            raise ValueError("bad chrom id")
+        chr_prefix = "chr"  # building this in for future generalization
+        if id.startswith(chr_prefix):
+            id = id[len(chr_prefix):]
+        if os.path.exists(annot_file):
+            bed = zarr_to_dataframe(annot_file)
+            assert type(bed) == pandas.DataFrame
+            ret = bed[bed.seqid == id]
+            if len(ret) == 0:
+                raise ValueError
+        else:
+            ret = None
+            raise ValueError(
+                "Warning: annotation file not found for chromosome: '{}'"
+                " on annotation: '{}', no annotations will be used".format(
+                    id, self.id))
+        return ret
+
+    def get_annotation_type_from_chromomosome(self, a_type, chrom_id, full_table=False):
+        """
+        Returns all elements of
+        type a_type from chromosome
+        specified
+        """
+        annots = self.get_chromosome_annotations(chrom_id)
+        subset = annots[annots.type == a_type]
+        if subset.empty:
+            raise ValueError(f"annotation type '{a_type}' not found on chrom"
+                             f" {chrom_id}")
+        if full_table:
+            return subset
+        else:
+            return subset[['start', 'end']]
+
+    def get_genes_from_chromosome(self, chrom_id, full_table=False):
+        """
+        Returns all elements of
+        type gene from annotation
+        """
+        return self.get_annotation_type_from_chromomosome('gene', chrom_id,
+                                                          full_table)
--- a/stdpopsim/catalog/HomSap/__init__.py
+++ b/stdpopsim/catalog/HomSap/__init__.py
@@ -202,6 +202,36 @@ for pop in ["ACB", "ASW", "BEB", "CDX", "CEU", "CHB", "CHS", "CLM", "ESN",
     )
     _species.add_genetic_map(_gm)
 
+###########################################################
+#
+# Annotations
+#
+#
+#
+###########################################################
+
+_an = stdpopsim.Annotation(
+    species=_species,
+    id="Ensembl_GRCh38_gff3",
+    description="Ensembl GFF3 annotations on GRCh38",
+    long_description="""
+        These are the complete GFF3 annotations from
+        Ensembl. Please see
+        """,
+    url=(
+        "ftp://ftp.ensembl.org/pub/release-101/"
+        "gff3/homo_sapiens/Homo_sapiens.GRCh38.101.gff3.gz"),
+    zarr_url=(
+        "https://stdpopsim.s3-us-west-2.amazonaws.com/"
+        "annotations/HomSap.GRCh38.zip"),
+    citations=[
+        stdpopsim.Citation(
+            year=2018,
+            author="Hunt et al",
+            doi="https://doi.org/10.1093/database/bay119",
+            reasons={stdpopsim.CiteReason.ANNOTATION})]
+    )
+_species.add_annotations(_an)
 
 ###########################################################
 #
--- a/stdpopsim/species.py
+++ b/stdpopsim/species.py
@@ -68,6 +68,12 @@ def all_demographic_models():
             yield model
 
 
+def all_annotations():
+    for species in all_species():
+        for an in species.annotations:
+            yield an
+
+
 @attr.s()
 class Species:
     """
@@ -135,6 +141,7 @@ class Species:
     # intended to be used when the Species is initialsed.
     # Use add_genetic_map() instead.
     genetic_maps = attr.ib(factory=list, kw_only=True)
+    annotations = attr.ib(factory=list, kw_only=True)
 
     @ensembl_id.default
     def _default_ensembl_id(self):
@@ -236,3 +243,27 @@ class Species:
             if gm.id == id:
                 return gm
         raise ValueError(f"Genetic map '{self.id}/{id}' not in catalog")
+
+    def add_annotations(self, annotations):
+        if annotations.id in [an.id for an in self.annotations]:
+            raise ValueError(
+                    f"Annotations '{self.id}/{annotations.id}' "
+                    "already in catalog.")
+        annotations.species = self
+        self.annotations.append(annotations)
+
+    def get_annotations(self, id):
+        """
+        Returns a set of annotations with the specified ``id``.
+
+        :param str id: The string identifier for the set of annotations
+            A complete list of IDs for each species can be found in the
+            "Annotations" subsection for the species in the :ref:`sec_catalog`.
+        :rtype: :class:`Annotation`
+        :return: A :class:`Annotation` that holds genome annotation
+            information from Ensembl
+        """
+        for an in self.annotations:
+            if an.id == id:
+                return an
+        raise ValueError(f"Annotations '{self.id}/{id}' not in catalog")
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -1,0 +1,207 @@
+"""
+Tests for the annotations management.
+"""
+import unittest
+from unittest import mock
+import gzip
+import tempfile
+import os.path
+import shutil
+import urllib.request
+import pathlib
+
+import pandas
+import stdpopsim
+from stdpopsim import annotations
+import tests
+
+
+# Here we download all the annotations in one go and store
+# then in the local cache directory, _test_cache. Tests are then run
+# with the download URLs redirected to local files, which makes them
+# much faster and takes network errors out of the equation. The
+# tarball cache is also useful for developers as it means that these
+# files are only downloaded once.
+
+saved_urls = {}
+
+
+def setUpModule():
+    destination = pathlib.Path("_test_cache/zipfiles/")
+    for an in stdpopsim.all_annotations():
+        key = an.file_name
+        local_file = destination / key
+        if not local_file.exists():
+            cache_dir = local_file.parent
+            cache_dir.mkdir(exist_ok=True, parents=True)
+            print("Downloading", an.zarr_url)
+            urllib.request.urlretrieve(an.zarr_url, local_file)
+        saved_urls[key] = an.zarr_url
+        an.zarr_url = local_file.resolve().as_uri()
+
+
+def tearDownModule():
+    for an in stdpopsim.all_annotations():
+        an.zarr_url = saved_urls[an.file_name]
+
+
+class AnnotationTestClass(annotations.Annotation):
+    """
+    A Annotation that we can instantiate to get an annotation for testing.
+    """
+
+    def __init__(self):
+        genome = stdpopsim.Genome(chromosomes=[])
+        _species = stdpopsim.Species(
+            id="TesSpe", name="Test species", common_name="Testy McTestface",
+            genome=genome)
+        super().__init__(
+            species=_species,
+            id="test_annotation",
+            url="http://example.com/annotation.gff.gz",
+            zarr_url="http://example.com/annotation.zip",
+            file_name="annotation.gff.gz")
+
+
+# number of chromosomes etc.
+def get_annotation_file(custom_file_f=None, filter=None):
+    """
+    Returns a GFF file as a bytes object.
+
+    :param func custom_file_f: A function that accepts a single parameter
+        (a folder name), which may be used to create additional files under
+        the given folder. All files in the folder will be included in the
+        returned tarball.
+    :param func filter: A function which is passed as the ``filter`` argument
+        to ``TarFile.add()``. This function can be used to change the info
+        field for each file in the returned tarball. See tarfile documentation
+        for more details.
+    """
+    b = "21\t.\tbiological_region\t6111184\t6111187\t0.999\t+\t.\tlogic_name=eponine\n"
+    b += b
+    b += b
+
+    with tempfile.TemporaryDirectory() as map_dir:
+        with open(os.path.join(map_dir, "test.gff"), "w") as f:
+            print("##gff-version 3", file=f)
+            print("##sequence-region   21 1 46709983", file=f)
+            print(b, file=f)
+
+        if custom_file_f is not None:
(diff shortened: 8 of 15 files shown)


ndukler:
We should probably have some way of handling the sex chromosomes as part of this. Dunno if it's at the species level or the assembly level though.

andrewkern:
do you mean label sex chroms? they are in the GFFs

andrewkern:
ah got it. i don't *believe* that should hang out here though. i imagine we'd want those decorations in the Chromosome or Contig class

ndukler:
Sure, we're going to need to setup some logic to pass those along when we retrieve and parse the assembly.

andrewkern:
ugh i see... there is no `pybedtools` package on conda for windows... what do i do in this case?

mufernando:
Thanks for this, @andrewkern!

It looks good overall, but I have one suggestion:  I think we should make the class more general than `GTF` and instead call it `Annotation`. In principle, we should be able handle any `pybedtools` supported format, such as BED, GFF, GTF.

I also agree with @ndukler that this should ultimately not be a property of `species`, but of a assembly. For now, though, it's fine to stay where it's at.

Now, to start connecting this with the SLiM side we need to implement the following:

- `GenomicElement`
    - Intervals
    - Mutation types (and weights)

- Something more meta that houses all genomic elements which will handle the non-overlappingness of the elements. Further, deal with the neutral mutations and the fact that there has to be an element at the end of the chromosome (`chromosome_length-1`), so that SLiM actively records the right length.

- A mapping of Annotation objects to Elements. This gets tricky and might determine our support for GFF files vs. BED. GFF files have a feature column that could each be a GenomicElement. BED files on the other hand are usually made on a per-feature basis.

(btw. should I move this to a separate issue so we can discuss in more detail?)

andrewkern:
sure yeah happy to put this in a separate issue. The GTF() class is poorly named now, but this is the rough outline of how we will get annotations *in*

mufernando:
> ugh i see... there is no `pybedtools` package on conda for windows... what do i do in this case?

It seems for now you'll have to skip those tests in windows as they did in #397

grahamgower:
Hey @andrewkern, I don't about motivation to support cygwin-less windows. However, it seems that pybedtools states "windows is not supported" http://daler.github.io/pybedtools/main.html. Possibly this is because pybedtools depends upon bedtools and pysam. bedtools says nothing about windows on its installation page (https://bedtools.readthedocs.io/en/latest/content/installation.html), and there's this open ticket about pysam windows failures: https://github.com/pysam-developers/pysam/issues/575.

So what is the motivation for including pybedtools as a dependency? What parts of the GFF file are we expecting to use? Are we anticipating a need to do set operations on intervals? I vote we avoid that kind of complexity if possible. GFF is just tab separated data with a header, so it should be possible to extract the basics with `numpy.loadtxt()`, given an appropriate dtype.

mufernando:
We can swap pybedtools to [pyranges](https://github.com/biocore-ntnu/pyranges). It seems to support windows, [parses](https://pyranges.readthedocs.io/en/latest/autoapi/pyranges/readers/index.html) BED/GFF/GTF and does interval operations even faster than pybedtools.

andrewkern:
> So what is the motivation for including pybedtools as a dependency? What parts of the GFF file are we expecting to use? Are we anticipating a need to do set operations on intervals? I vote we avoid that kind of complexity if possible. GFF is just tab separated data with a header, so it should be possible to extract the basics with `numpy.loadtxt()`, given an appropriate dtype.

Yeah i can definitely avoid it, but was thinking we would want something to do interval operations with (e.g., to deal with overlapping annotations or quick filtering). I can easily just use numpy or pandas even. I'll look in to the pyranges thing.

@grahamgower how does the rest of this look to you?

stdpopsim/annotations.py
```diff
@@ -0,0 +1,133 @@
...
+import os
+import urllib.request
+from . import cache
+
+logger = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def cd(path):
+    """
```
grahamgower, line 20: I might have missed it, but I didn't see this `cd` function used anywhere.

stdpopsim/annotations.py
```diff
@@ -0,0 +1,133 @@
...
+        Returns the `pybedtools` `BedTool` object for
+        the chromosome with the specified id.
+        """
+        if not self.is_cached():
+            self.download()
+        annot_file = os.path.join(self.species_cache_dir, self.file_name)
+        if id is None:
+            raise ValueError("bad chrom id")
+        if "chr" in id:
+            id = id.replace("chr", "")
```
grahamgower, line 120: Probably want to just trim from the start of the `id` string. E.g.
```
if id.startswith("chr"):
    id = id[len("chr"):]
```

But shouldn't the chomosome names match those used along with the species/assembly definition? I guess the Assembly class needs to have a `chr_prefix` attribute, which says if the chromosome names have a prefix, so we know when to trim it and when not to...

stdpopsim/annotations.py
```diff
@@ -0,0 +1,133 @@
...
+    old_dir = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(old_dir)
+
+
+@attr.s
+class GTF(object):
```
grahamgower, line 33: I second @mufernando's motion that this class be named `Annotation`.

tests/test_annotations.py
```diff
@@ -0,0 +1,199 @@
...
+            species=_species,
+            id="test_annotation",
+            url="http://example.com/annotation.gff.gz",
+            file_name="annotation.gff.gz")
+
+
+# number of chromosomes etc.
+def get_annotation_gtf_file(custom_file_f=None, filter=None):
+    """
+    Returns a genetic map in hapmap format in a tarball as a bytes object.
```
grahamgower, line 71: The comment and docstring here seem misplaced or out of date?

stdpopsim/annotations.py
```diff
@@ -0,0 +1,133 @@
...
+        s += "\tcache_dir = {}\n".format(self.species_cache_dir)
+        return s
+
+    def is_cached(self):
+        """
+        Returns True if this map is cached locally.
+        """
+        return os.path.exists(self.species_cache_dir)
+
+    def download(self):
```
grahamgower, line 79: It might be nice to factor out the duplicated download/cache code between this class and the GeneticMap class.

tests/test_annotations.py
```diff
@@ -0,0 +1,199 @@
...
+            # Trick the download code into thinking there's several happening
+            # concurrently
+            gm.is_cached = lambda: False
+            with self.assertWarns(Warning):
+                gm.download()
+        finally:
+            gm.is_cached = saved
+
+
+class TestAllGeneticMaps(tests.CacheReadingTest):
```
grahamgower, line 164: 'TestAllGeneticMaps'... It looks like this is not related to annotations or the GTF class?

grahamgower:
> Yeah i can definitely avoid it, but was thinking we would want something to do interval operations with (e.g., to deal with overlapping annotations or quick filtering). I can easily just use numpy or pandas even. I'll look in to the pyranges thing.

I spent a few minutes playing with some annotations just now and can see the advantages. But I think we need to consider what annotations we're likely to use---if only to reduce the load times. I did some timing tests, just to load the (uncompressed) `Homo_sapiens.GRCh37.87.gff3` file (https://gist.github.com/grahamgower/edab09eb43fbfcf4a4c92c0606940039).

```
$ ./time_gff_loading.py Homo_sapiens.GRCh37.87.gff3
pybedtools_BedTool took 8 seconds
pyranges_read_gff3 took 30 seconds
load_gff_numpy took 18 seconds
load_gff_python took 3 seconds
```

andrewkern:
yeah we will want to stick with GFF files from Ensembl I think and we will want to keep them compressed. I wonder if we can use a pandas reader to speed this up further?

mufernando:
pyranges is [using pandas](https://github.com/biocore-ntnu/pyranges/blob/master/pyranges/readers.py) I think.

andrewkern:
well that's how i would have done it too. shit. okay i can implement a python implementation like @grahamgower has on that gist

mufernando:
Just tested [scikit-allel's GFF3 reader](https://scikit-allel.readthedocs.io/en/stable/io.html#gff3) and it's almost as fast as @grahamgower's pure Python version. With a bunch of added benefits (e.g., being able to extract chunks). See the time testing [here](https://gist.github.com/mufernando/750c45aec17d146036c7a27122595214).

```
pybedtools_BedTool took 21 seconds
pyranges_read_gff3 took 40 seconds
load_gff_numpy took 29 seconds
load_gff_python took 6 seconds
allel_gff3 took 11 seconds
```

andrewkern:
happy to try it out! will reimplement that bit

stdpopsim/annotations.py
andrewkern: the directory structure here is different enough that it would be a pain i think. little reward vs the bit of duplicated code IMO

grahamgower:
Hey @andrewkern, I'm still rather concerned about how slow it is to load up this 2.6 million line gff file. Presumably this is most problematic for HomSap, as lots of good annotations exist. Given that the download step is going to take a while and is a one-off, perhaps we could do some preprocessesing on the gff when the file gets downloaded into the cache? One obvious choice we might make is to split the gff by chromosome. But we could also split by the source or feature columns. (E.g. it would be possible to have it cached like `cache_dir/annotations/Homo_sapiens.GRCh37.87/1/ensembl/exon.gff.gz`)

Sources:
```
$ awk '!/^#/ {print $2}' Homo_sapiens.GRCh37.87.gff3 | sort  |
uniq -c
  22397 cpg
 360236 ensembl
 696421 ensembl_havana
  61761 Eponine
  87144 firstef
     84 GRCh37
1397884 havana
    124 insdc
    626 tRNAscan-SE
```
I don't know anything about gene annotations, so it's not clear to me whether we'd favour using `ensembl`, `ensembl_havana,` or `havana`. But I can't see an obvious use case for loading more than one at a time.

Feature/type
```
$ awk '!/^#/ {print $3}' Homo_sapiens.GRCh37.87.gff3 | sort  |
uniq -c
  25944 aberrant_processed_transcript
 171928 biological_region
 724218 CDS
     28 C_gene_segment
     25 chromosome
1195764 exon
 148015 five_prime_UTR
  26998 gene
    148 J_gene_segment
  11773 lincRNA
   7109 lincRNA_gene
   3174 miRNA
   3111 miRNA_gene
  81787 mRNA
     22 mt_gene
     25 nc_primary_transcript
  13046 NMD_transcript_variant
  11060 processed_pseudogene
  28551 processed_transcript
  17217 pseudogene
    858 pseudogenic_transcript
   2038 RNA
    537 rRNA
    533 rRNA_gene
   1531 snoRNA
   1459 snoRNA_gene
   1930 snRNA
   1923 snRNA_gene
     59 supercontig
 132262 three_prime_UTR
  13373 transcript
     37 VD_gene_segment
    194 V_gene_segment
```

grahamgower:
Oh, it seems there are chromosome-specific files already available. Can we start by using those?
http://ftp.ensembl.org/pub/grch37/release-87/gff3/homo_sapiens/

jeromekelleher:
I agree with @grahamgower here that we should use the cache to avoid some computation here. I have another suggestion to throw into the mix though: what if we (shock horror) didn't use text to store this stuff but used [zarr](https://zarr.readthedocs.io/en/stable/) instead? Then we're not quite so performance sensitive in the GFF parsing, and the client code is simplified by being able to work with numpy arrays rather than some upstream library's API.

stdpopsim/annotations.py
```diff
@@ -0,0 +1,120 @@
...
+        s += "\turl       = {}\n".format(self.url)
+        s += "\tcached    = {}\n".format(self.is_cached())
+        s += "\tcache_dir = {}\n".format(self.species_cache_dir)
+        return s
+
+    def is_cached(self):
+        """
+        Returns True if this map is cached locally.
+        """
+        return os.path.exists(self.species_cache_dir)
```
jeromekelleher, line 63: Not sure this logic holds up for ``is_cached``

stdpopsim/annotations.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def get_chromosome_annotations(self, id):
+        """
+        Returns the pandas dataframe for
+        the chromosome with the specified id.
+        """
+        if not self.is_cached():
+            self.download()
+        annot_file = os.path.join(self.species_cache_dir, self.file_name)
+        if id is None:
```
jeromekelleher, line 103: Why the specific check for None here? Seems like lots of other ways to have a bad id. I'd just remove the check.

grahamgower:
> think we should look into (a) retrieving stuff programatically from Ensembl first and then (b) storing the data in the cache as zarr or something. I'd like to avoid "classical" bioinfo data formats and the messy dependencies if at all possible.

Regards (a), it would seem that enseml don't provide access to resources for non-current assemblies. E.g. human reference GRCh37.p13 should be here but is not:
https://rest.ensembl.org/info/genomes/assembly/GCA_000001405.14
Am I missing something?

Caching using zarr sounds nice though!

andrewkern:
hey all-- zarr and preprocessing seems good to me [i quite like the zarr format]. let's talk about this on the call

andrewkern:
I will heavily edit this PR after we merge #563, although one question I have is can we easily pull genome annotation via the REST API? 

I see a lot of REST endpoints that are locus specific, rather than genomewide, but presumably this can be done?

jeromekelleher:
> I see a lot of REST endpoints that are locus specific, rather than genomewide, but presumably this can be done?

Good question. It might be a struggle to pipe that much data through JSON though - I think we want to convert to zarr now? So, we might add the script for converting GFF into zarr to our maintenance directory, and just make sure we're using the same version of Ensembl?

Although, if we *could* do it via the REST API that would be ideal.

andrewkern:
yeah after looking further I'm pretty sure that we will need to download the GFF and convert to zarr rather than do this through the REST API. I could add the GFF to zarr script to this PR once we merge #563. what do you think Jerome?

andrewkern:
hey @jeromekelleher check out this draft of the maintenance script for doing the download and conversion to zarr https://github.com/popsim-consortium/stdpopsim/pull/560/commits/55718c03710887ccbe0f2873464e6d58c70078c3

andrewkern:
it's still rough, but wanted to check if this is what you have in mind. this will go through all the species, download the indicated annotation, and then shove them in a zarr group.

I still need to figure out how then to upload these automatically to AWS. do we have a way to do that?

jeromekelleher:
Yep, looks ideal. We can tidy up the rough edges afterwards.

I'm sure there must be some way to automatically upload to AWS - you and @ndukler were the people dealing with this stuff though. How about you loop me in to whatever details you have and I'll see what we need to do?

andrewkern:
well circleci is failing to build the scikit-allel wheel for some reason (anyone know why?) but i think the code is in good enough shape for review. 

this code deals with the maintenance and distribution of annotations. GFF files are downloaded by a maintenance script and then coverted to zarr files and pushed to AWS. the stdpopsim client then automatically downloads / caches the zarr locally, and the user can interface with the GFF annotations via an pandas dataframe that is produced upon request.

grahamgower:
Cool! Thanks @andrewkern. It looks like you solved the CI issues here. I'll take a close look in the next few days.

andrewkern:
do to from here includes:
- writing a function to filter for DNA element type (e.g., 'gene'). this is trivial given that the output is a pandas dataframe
- writing a function that goes from the filtered df to the mutation type inputs through the SLiM interface

mufernando:
I assigned myself to review this so I can get started on linking this to SLiM. Hope to get to it by early next week at the most!

andrewkern:
> do to from here includes:
> 
> * writing a function to filter for DNA element type (e.g., 'gene'). this is trivial given that the output is a pandas dataframe

i'm going to go ahead and write this and commit more edits to this PR

stdpopsim/annotations.py
```diff
@@ -0,0 +1,165 @@
...
+            assert type(bed) == pandas.DataFrame
+            ret = bed[bed.seqid == id]
+            if len(ret) == 0:
+                raise ValueError
+        else:
+            warnings.warn(
+                "Warning: annotation file not found for chromosome: '{}'"
+                " on annotation: '{}', no annotations will be used".format(
+                    id, self.id))
+            ret = None
```
grahamgower, line 140: It isn't obvious to me how this case can happen. But I think it would be better if this raised a ValueError, instead of having a warning and returning None.

stdpopsim/annotations.py
```diff
@@ -0,0 +1,165 @@
...
+        except FileNotFoundError:
+            warnings.warn("can't connnect to url")
+
+        logger.debug("Download Zarr complete")
+        logger.info(f"Storing Zarr in {self.species_cache_dir}")
+
+    def get_chromosome_annotations(self, id):
+        """
+        Returns the pandas dataframe for
+        the chromosome with the specified id.
```
grahamgower, line 119: I suppose this will be the main entry point for consumers of annotations. So we'll need to have complete docstrings for this (and the other user-facing methods in this class). Happy for this to be punted down the road though. Maybe open a ticket tracking this once it's merged?

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -202,6 +202,37 @@
...
+        "http://ftp.ensembl.org/pub/grch37/current/gff3/"
+        "homo_sapiens/Homo_sapiens.GRCh37.87.gff3.gz"),
+    zarr_url=(
+        "https://stdpopsim.s3-us-west-2.amazonaws.com/"
+        "annotations/HomSap.zip"),
+    citations=[
+        stdpopsim.Citation(
+            year=2017,
+            author="Aken et al",
+            doi="http://dx.doi.org/10.1093/nar/gkw1104",
```
grahamgower, line 232: Ensembl seem to have a publication every year. How will we choose which one to cite? (I guess you chose this one because it corresponds with release 87?)

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -202,6 +202,37 @@
...
+    id="Ensembl_GRCh37_gff3",
+    description="Ensembl GFF3 annotations on GRCh37",
+    long_description="""
+        These are the complete GFF3 annotations from
+        Ensembl. Please see
+        ftp://ftp.ensembl.org/pub/grch37/current/gff3/homo_sapiens/README
+        """,
+    url=(
+        "http://ftp.ensembl.org/pub/grch37/current/gff3/"
+        "homo_sapiens/Homo_sapiens.GRCh37.87.gff3.gz"),
```
grahamgower, line 224: HomSap is now on GRCh38, so these annotations won't match the genome assembly in the catalog! Is there a way to make the version automagically match up?

stdpopsim/species.py
```diff
@@ -236,3 +243,17 @@ def get_genetic_map(self, id):
...
+
+    def add_annotations(self, annotations):
+        if annotations.id in [an.id for an in self.annotations]:
+            raise ValueError(
+                    f"Annotations '{self.id}/{annotations.id}' "
+                    "already in catalog.")
+        annotations.species = self
+        self.annotations.append(annotations)
+
+    def get_annotations(self, id):
```
grahamgower, line 255: docstring needed

tests/test_annotations.py
```diff
@@ -0,0 +1,198 @@
...
+            with gzip.open("test.gff.gz", mode="wb") as gz_file:
+                shutil.copyfileobj(f_in, gz_file)
+        # Read back the tarball
+        an = gzip.open("test.gff.gz", mode='rb')
+    return an
+
+
+class TestAnnotation(tests.CacheWritingTest):
+    """
+    Tests for the basic functionality of the genetic map class.
```
grahamgower, line 106: s/genetic map/annotation/

tests/test_annotations.py
```diff
@@ -0,0 +1,198 @@
...
+        gm = AnnotationTestClass()
+        os.makedirs(gm.species_cache_dir, exist_ok=True)
+        self.assertTrue(gm.is_cached())
+        shutil.rmtree(gm.cache_dir)
+        self.assertFalse(gm.is_cached())
+
+
+class TestAnnotationDownload(tests.CacheWritingTest):
+    """
+    Tests downloading code for the genetic maps.
```
grahamgower, line 129: s/genetic maps/annotations/

grahamgower (changes requested):
Hey @andrewkern, I've gone through this fairly thoroughly now. It looks great, and I think it's functionally complete now! There are a few rough edges that can be easily fixed up, and others that can be punted down the road. The only blocker I see here is that you've included HomSap annotations for GRCh37, but we now have chromosome lengths/units for HomSap GRCh38.

Also, it would be good if you could run `flake8 --max-line-length 89 maintenance/anno*.py` --- this is currently missing from the circle CI checks, which is just an oversight.

maintenance/annotation_maint.py
```diff
@@ -0,0 +1,38 @@
+import allel
```
mufernando, line 1: I tried to run this locally but got an error

```
~/projects/stdpopsim/maintenance annotation ?4 ❯ python3 annotation_maint.py                                                                                           stdpopsim-dev 13:25:30
  File "annotation_maint.py", line 21, in <module>
    x, y = urllib.request.urlretrieve(address, tmp_path)
  File "/opt/anaconda3/envs/stdpopsim-dev/lib/python3.7/urllib/request.py", line 288, in urlretrieve
    % (read, size), result)
urllib.error.ContentTooShortError: <urlopen error retrieval incomplete: got only 29639876 out of 37903989 bytes>

~/projects/stdpopsim/maintenance annotation ?5 ❯ python3 annotation_maint.py                                                                                   1m 48s  stdpopsim-dev 13:27:23
  File "annotation_maint.py", line 13, in <module>
    os.mkdir(annot_path)
FileExistsError: [Errno 17] File exists: 'annotations'
```

andrewkern: looks like there was a downloading error?

maintenance/__init__.py
```diff
@@ -1,4 +1,5 @@
 """
 Code used for internal maintenance tasks.
 """
+import zarr
```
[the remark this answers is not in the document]
andrewkern: yeah, i'm not sure. i was trying to defeat circleci and ended up adding imports. will try to remove it.

requirements/development.txt
```diff
@@ -14,3 +14,9 @@ attrs
 appdirs
 humanize
 pyslim>=0.401
+pandas
+numpy
+cython
+scikit-allel[full]
+zarr>=2.4
+boto3
```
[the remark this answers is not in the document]
andrewkern: yeah i can probably drop this. i was writing code to handle the AWS push for us, but because of user id issues this might not be feasible

stdpopsim/catalog/HomSap/__init__.py
andrewkern: looks like ensembl wants us to cite their most recent paper, if we are using their most recent releases: 

https://uswest.ensembl.org/info/about/publications.html

stdpopsim/catalog/HomSap/__init__.py
andrewkern: yup, this will be super easy. all one does is 

1. update the url string, 
2. rerun the maintenance code
3. upload to AWS

however the user will not be aware of an update if they have cached the annotations...

stdpopsim/species.py
```diff
@@ -236,3 +243,17 @@ def get_genetic_map(self, id):
             if gm.id == id:
                 return gm
         raise ValueError(f"Genetic map '{self.id}/{id}' not in catalog")
+
+    def add_annotations(self, annotations):
```
mufernando, line 247: i don't understand what this is doing

mufernando (commented):
great work, Andy! just a couple of qs/comments

requirements/conda-minimal.txt
```diff
@@ -5,3 +5,6 @@ tskit
 humanize
 attrs
 appdirs
+pandas
+scikit-allel
```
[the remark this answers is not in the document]
andrewkern: so i think circleci needs this here to run `test_maintence.py`

setup.py
```diff
@@ -33,7 +33,8 @@
     },
     # NOTE: make sure this is the 'attrs' package, not 'attr'!
     install_requires=["msprime>=0.7.1", "attrs>=19.1.0", "appdirs", "humanize",
-                      "pyslim>=0.401"],
+                      "pyslim>=0.401", "pandas", "scikit-allel", "zarr",
```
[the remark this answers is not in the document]
andrewkern: yeah again i added this for circleci... will clear these out

stdpopsim/annotations.py
```diff
@@ -0,0 +1,165 @@
...
+import warnings
+import os
+import urllib.request
+from . import cache
+import zarr
+
+logger = logging.getLogger(__name__)
+
+
+def zarr_to_dataframe(path):
```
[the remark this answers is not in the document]
andrewkern: not sure what you mean here? do you mean filter on the maintenance side and then dump those into memory? these are btw quite small files-- human genome is 40Mb, so i'm not worried about memory.

stdpopsim/annotations.py
```diff
@@ -0,0 +1,165 @@
...
+        annot_file = os.path.join(self.species_cache_dir, self.file_name)
+        if id is None:
+            raise ValueError("bad chrom id")
+        chr_prefix = "chr"  # building this in for future generalization
+        if id.startswith(chr_prefix):
+            id = id[len(chr_prefix):]
+        if os.path.exists(annot_file):
+            bed = zarr_to_dataframe(annot_file)
+            assert type(bed) == pandas.DataFrame
+            ret = bed[bed.seqid == id]
```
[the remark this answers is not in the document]
andrewkern: not sure what you mean here? do you mean filter on the maintenance side and then dump those into memory? these are btw quite small files-- human genome is 40Mb, so i'm not worried about memory.

stdpopsim/species.py
andrewkern: this is adding to the catalog of possible annotations in a Species

andrewkern:
pushed the requested changes from @grahamgower and @mufernando. this _may_ break circleci ...

andrewkern:
also worth mentioning that this now uses GRCh38 annotation from release 101 of Ensembl.

maintenance/annotation_maint.py
```diff
@@ -0,0 +1,38 @@
...
+import stdpopsim as stp
+import logging
+import urllib.request
+import os
+from zipfile import ZipFile
+
+logger = logging.getLogger(__name__)
+# make root directory for zarr annotations
+annot_path = "annotations"
+os.mkdir(annot_path)
```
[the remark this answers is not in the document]
andrewkern: right so i like this because it fails if it's already there

stdpopsim/annotations.py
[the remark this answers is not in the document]
mufernando: you could within `zarr_to_dataframe` do the filtering of rows on `seqid`.

Like

```
def zarr_to_dataframe(path, id):
    """
    converts zarr annotation file to
    pandas dataframe for manipulations
    """
    z = zarr.open(path)
   mask = z['seqid'][:] == id
    df = pandas.DataFrame(
        {'seqid': z['seqid'].get_mask_selection(mask),
         'source': z['source'].get_mask_selection(mask),
         'type': z['type'].get_mask_selection(mask),
         'start': z['start'].get_mask_selection(mask),
         'end': z['end'].get_mask_selection(mask),
         'score': z['score'].get_mask_selection(mask),
         'strand': z['strand'].get_mask_selection(mask),
         'phase': z['phase'].get_mask_selection(mask)}
    )
    return df


```

jradrion (approved):
LGTM, @andrewkern! The comments/requests from @grahamgower and @mufernando seem to have covered the relatively minor issues.

jradrion:
just saw that there are new changes since I last checked out this PR, I'll take another look

jradrion (approved):
`test_get_genes_full (tests.test_annotations.TestGetChromosomeAnnotations) ... ok`
LGTM

stdpopsim/species.py
```diff
@@ -236,3 +243,27 @@ def get_genetic_map(self, id):
...
+        annotations.species = self
+        self.annotations.append(annotations)
+
+    def get_annotations(self, id):
+        """
+        Returns a set of annotations with the specified ``id``.
+
+        :param str id: The string identifier for the set of annotations
+            A complete list of IDs for each species can be found in the
+            "Annotations" subsection for the species in the :ref:`sec_catalog`.
```
grahamgower, line 261: So we'll need some cleverness added to `docs/_ext/speciescatalog.py` in order to get the annotation id's listed in the catalog docs. (just mentioning this so we remember to open a ticket when this is merged)

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -202,6 +202,37 @@
...
+        These are the complete GFF3 annotations from
+        Ensembl. Please see
+        ftp://ftp.ensembl.org/pub/grch37/current/gff3/homo_sapiens/README
+        """,
+    url=(
+        "ftp://ftp.ensembl.org/pub/release-101/"
+        "gff3/homo_sapiens/Homo_sapiens.GRCh38.101.gff3.gz"),
+    zarr_url=(
+        "https://stdpopsim.s3-us-west-2.amazonaws.com/"
+        "annotations/HomSap.zip"),
```
grahamgower, line 227: How about including the ensembl version in the zarr zip filename? Consider what happens if a user has an old version in their cache, and then the genome assembly is updated -> no new annotation file is downloaded.

stdpopsim/annotations.py
```diff
@@ -0,0 +1,165 @@
...
+                    os.unlink(file_path)
+                except OSError:
+                    warnings.warn(f'could not unlink {file_path}')
+        logger.debug(f"Checking species cache directory {self.species_cache_dir}")
+        os.makedirs(self.species_cache_dir, exist_ok=True)
+        download_file = f'{self.species_cache_dir}/{self.file_name}'
+        print(download_file)
+        logger.info(f"Downloading Zarr file '{self.id}' from {self.zarr_url}")
+        if os.path.exists(download_file):
+            warnings.warn("multiple downloads?")
```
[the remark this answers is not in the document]
andrewkern: yes this is a test for multiple threads downloading simultaneously

andrewkern:
okay @grahamgower i think this is all now ready for merging. I've made all the requested changes-- thanks for the input @mufernando @jradrion @grahamgower

grahamgower (approved):
Great! Thanks @andrewkern! I'll leave it to you to merge, and open tickets for the remaining comments.

Issue #405, opened 2020-02-19
DroMel chrM is simulated using recombination rate > 0


grahamgower:
```
$ stdpopsim DroMel -c chrM -l 0.0001 -o /dev/null  10
Simulation information:
    Engine: msprime (0.7.5.dev48+ngef1bad9.d20200217)
    Model id: PiecewiseConstant
    Model desciption: Piecewise constant size population model over multiple epochs.
    Population: number_samples (sampling_time_generations):
        pop0: 10 (0)
Contig Description:
    Contig length: 1.9524000000000001
    Mean recombination rate: 8.4e-09
    Mean mutation rate: 5.49e-09
    Genetic map: None

If you use this simulation in published work, please cite:
[simulation engine]
  Kelleher et al., 2016: https://doi.org/10.1371/journal.pcbi.1004842
[genome assembly]
  dos Santos et al., 2015: https://doi.org/10.1093/nar/gku1099
[mutation rate]
  Schrider et al., 2013: https://doi.org/10.1534/genetics.113.151670
[population size, generation time]
  Li et al., 2006: https://doi.org/10.1371/journal.pgen.0020166

```

jeromekelleher:
Good catch @grahamgower, this is clearly a bug.

grahamgower:
Do we want to just yank out the mito, or add a recombination rate field to the `_chromosome_data` "table" like other species have?

jeromekelleher:
I guess we should be consistent with other species - what have we done for humans at the moment?

grahamgower:
I think DroMel is the only species with mito represented.

grahamgower:
Even if we fix the recombination rate, the mutation rate will still be the autosomal mutation rate. So I'm in favour of removing the mutation rate until we have mutation maps or at least per-chromosome mutation rates.

jeromekelleher:
Let's just drop the MT chromosome entirely then.

PR #563, opened 2020-06-15
Basic infrastructure for pulling genome data from Ensembl

jeromekelleher:
This seems a like a reasonable place to start with integrating with upstream genome data providers.


jeromekelleher:
Not sure how to test this, btw.

grahamgower:
I guess you could test the `taxon_id` via the REST API? E.g.: query `ext=f"/info/assembly/{taxon_id}"`, get the `assembly_accession` field, then query `ext=f"/info/genomes/accession/{assembly_accession}"` and finally translate the `name` field into a 3+3 to compare with the species id.

jeromekelleher:
There's some basic infrastructure here for pulling the genome/assembly data for a species from Ensembl here. The idea is that we use the REST API to get information on the assembly, and then write this info into a Python file. We can then easily read this in, and use it to populate the objects.

I decided that writing a simple Python file was better than trying to write the actual object definitions here because (a) it's a bit faffy writing out Python source code; and (b) we're still combining information from multiple sources, as the mutation/recombination rates aren't going to be available from Ensembl. We could equally use a JSON file, it probably doesn't matter. I just thought that a Black formatting python file like this would be an unequivocal and diff-friendly way to stating the information.

If we like this approach, then I think what we should do is to make each species module in the catalog a package directory, where the ``__init__.py`` brings together all the information. For now, the only extra information in the directory will be the ``genome_data.py`` file, but we might want others later.

There's a little bit more infrastructure needed to go over all the species automatically, but it's straightforward. Do we like the general layout, before I go any further?

Pinging @grahamgower @andrewkern @ndukler

jeromekelleher:
Oh yeah, Ensembl doesn't use the "chr" prefix and we currently do. Hooray.

ndukler:
Heh. I use a package in R called `GenomeInfoDb` that can detect chromosome name formats and convert between them to avoid this sort of thing when I work in R. Either we can switch everything to ensembl nomenclature or we can implement a similar functionality for our package. They have the mapping files [here](https://github.com/Bioconductor/GenomeInfoDb/tree/a0723e023623ff75eb569dd86db70494c1421cd3/inst/extdata/dataFiles) and we could probably re-implement their functionality in stdpopsim. That would also let people specify chromsomes from the command line or API however they want and we could convert them all to a single format internally.

maintenance/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Code used for internal maintenance tasks.
+"""
+from . net import *
+from . species_data import *
```
grahamgower, line 5: Did `species_data.py` get left off the commit? I didn't see the `maintenance.get_genome_data()` function.

grahamgower:
Very nice @jeromekelleher, I like it a lot!

maintenance/__init__.py
ndukler: I don't see it either

ndukler:
> Very nice @jeromekelleher, I like it a lot!

Seconded!

ndukler:
Is it worth me re-implementing the chromosome renaming functionality @jeromekelleher ? Or should we just agree that everyone always has to use ensembl nomenclature? Or just set up a specific ensembl to ucsc conversion when the genome data is imported and force everywhere else to be UCSC?

stdpopsim/catalog/HomSap_genome_data.py
```diff
@@ -0,0 +1,31 @@
+{
```
andrewkern, line 1: I'm wondering if we can pull this info too? why hard code it here if we are pulling all this assembly info anyway?

ndukler: Safer this way. Plus this is being generated programatically already.

maintenance/__init__.py
jeromekelleher: Yep, it was missing, thanks.

stdpopsim/catalog/HomSap_genome_data.py
jeromekelleher: Yeah, this file is autogenerated @andrewkern. I've added a comment header to say this, for clarity.

jeromekelleher:
I've pushed through the changes and it's mostly working pretty well. There's some details with some assemblies like PonAbe that we need to make decisions on, but there's no big issues.

The only real roadblock is the chromosome naming conventions. We either need to

- (a) find some mapping from the chromosome names to their synonyms (there's *surely* a web service to do this, right?) and build synonyms into the Chromosome object
- (b) Do it ourselves based on the data @ndukler pointed out above; or 
- (c) Standardise on just the Ensembl IDs, and fix the breakage in our tests (and break existing code).

I think (a) would be the ideal solution. Anyone like to do a bit of digging to see if we can find the required data? I'd be surprised if it wasn't in Ensembl somewhere.

stdpopsim/catalog/HomSap_genome_data.py
andrewkern: sweeeet

maintenance/ensembl.py
```diff
@@ -0,0 +1,99 @@
...
+            "assembly_accession": output["assembly_accession"],
+            "assembly_name": output["assembly_name"],
+        }
+        chromosomes = {}
+        for region in output["top_level_region"]:
+            if region["coord_system"] == "chromosome":
+                chromosomes[region["name"]] = {"length": region["length"]}
+        # Make sure the chromosomes are sorted by name
+        data["chromosomes"] = {
+            name: chromosomes[name] for name in sorted(chromosomes.keys())}
```
grahamgower, line 98: I think the "karyotype" output field gives a more natural ordering than sorting.
```
>>> sorted(["1", "2", "10"])
['1', '10', '2']
```

stdpopsim/catalog/PonAbe/__init__.py
```diff
@@ -17,40 +18,39 @@
...
+    "18": 4.57e-9,
+    "19": 7.56e-9,
+    "20": 5.83e-9,
+    "21": 4.98e-9,
+    "22": 6.03e-9,
+    "X": 9.50e-9,
+    "MT": 0,
+    # JK: I've no idea what this chr is. Is this the correct thing
+    # to do?
+    "Un": 0,
```
grahamgower, line 52: Maybe "Un" means unplaced?

stdpopsim/species.py
```diff
@@ -114,6 +117,14 @@ class Species(object):
...
     genetic_maps = attr.ib(factory=list, kw_only=True)
 
+    @property
+    def ensembl_id(self):
+        """
+        Returns the ID of this species for the Ensembl REST API.
+        This is the species name, underscore delimited and in lowercase.
+        """
+        return self.name.lower().replace(" ", "_")
+
```
grahamgower, line 127: Is this really needed? The REST API seems to accept space-delimited names, mixed case, taxon id, etc.

maintenance/ensembl.py
jeromekelleher: Yes, excellent idea. I'll change that.

stdpopsim/species.py
jeromekelleher: I dunno, I spent a long time fiddling with it, trying all sorts of things and this was the only thing I could get to work reliably. They do seem to use this form in the docs a lot, so maybe that's their basic internal representation and so less likely to break.

ndukler:
> * (a) find some mapping from the chromosome names to their synonyms (there's _surely_ a web service to do this, right?) and build synonyms into the Chromosome object

Do you want me to prototype this? I can just grab the tables from the place I mentioned then set up   a system for automatically attaching the synonyms to each chromosome when the genome is built. Then `get_chromosome()` can search the synonyms as well.

jeromekelleher:
That sounds good @ndukler, but I worry a bit about how extensive the species data will be and potential conflicts with the Ensembl data model. I'll ping a few more people to see if anyone knows, so maybe don't put too much effort into it.

jeromekelleher:
Some helpful comments on twitter [here](https://twitter.com/jeromekelleher/status/1273887681949425664)

jeromekelleher:
I've update this with your ideas @ndukler - it's still pretty messy tbh. I don't really know where we should get the data from - perhaps the best thing is to git submodule the R code? Or, maybe we should pull in the data using the R package by running R from Python? Then we can checkout a particular release tag from GenomeInfoDb.

I had a go at getting this information from UCSC via the database API too, but everything so bloody inconsistent :(

jeromekelleher:
This is nearly done, I think, so could use from review from those in the know. We can get some synonym information from Ensembl. It's not that comprehensive, but it's good enough - we can just say we only allow Ensembl IDs for the things we don't have (like DroMel, weirdly). I fixed the the test cases that broke because of this.

The only thing that seriously broken now is the recombination maps. These have UCSC IDs in them, which is fun. Hmm, I should be able to fix this.. wait on.

jeromekelleher:
OK! Recombination maps are fixed, so I think this is ready for review and final cleanups now.

Pinging @andrewkern, @grahamgower, @ndukler, @apragsdale

stdpopsim/catalog/DroMel/__init__.py
```diff
@@ -44,20 +32,18 @@
...
 _chromosomes = []
-for line in _chromosome_data.splitlines():
-    name, length = line.split()[:2]
+for name, data in genome_data.data["chromosomes"].items():
     _chromosomes.append(stdpopsim.Chromosome(
-        id=name, length=int(length),
+        id=name, length=int(data["length"]),
+        synonyms=data["synonyms"],
         mutation_rate=5.49e-9,  # citation: _SchriderEtAl
         recombination_rate=8.4e-9))  # WRONG, underestimate used in S&S!
```
grahamgower, line 40: I think this introduces a regression---the mito and Y will have a non-zero recombination rate. See #405.

stdpopsim/catalog/DroMel/genome_data.py
```diff
@@ -0,0 +1,15 @@
...
+    "assembly_name": "BDGP6.28",
+    "chromosomes": {
+        "2L": {"length": 23513712, "synonyms": []},
+        "2R": {"length": 25286936, "synonyms": []},
+        "3L": {"length": 28110227, "synonyms": []},
+        "3R": {"length": 32079331, "synonyms": []},
+        "4": {"length": 1348131, "synonyms": []},
+        "X": {"length": 23542271, "synonyms": []},
+        "Y": {"length": 3667352, "synonyms": []},
+        "mitochondrion_genome": {"length": 19524, "synonyms": []},
```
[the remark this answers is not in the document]
jeromekelleher: I don't see an easy way to do it, but I'm open to suggestions. Maybe something to raise an issue on for follow up? I think the primary ID must be the Ensembl ID, regardless of any inconsistencies we see.

stdpopsim/catalog/DroMel/__init__.py
jeromekelleher: Well spotted, ta.

stdpopsim/genomes.py
```diff
@@ -50,7 +52,7 @@ def get_chromosome(self, id):
         Returns the chromosome with the specified id.
         """
         for chrom in self.chromosomes:
-            if chrom.id == id:
+            if chrom.id == id or id in chrom.synonyms:
```
ndukler, line 55: Should we make this case insensitive?

stdpopsim/catalog/AraTha/genome_data.py
```diff
@@ -0,0 +1,14 @@
+# File autogenerated from Ensembl REST API. Do not edit.
+data = {
+    "assembly_accession": "GCA_000001735.1",
+    "assembly_name": "TAIR10",
+    "chromosomes": {
+        "1": {"length": 30427671, "synonyms": []},
```
ndukler, line 6: I don't see any synonyms here or in any of the other files? That means something is missing right?

ndukler: Actually I do see them in human. Do none of the other species have synonyms?

jeromekelleher: No, looks like Ensembl's list of synonyms is pretty limited. We could bug them about this, I guess.

stdpopsim/genomes.py
jeromekelleher: "No" would be my initial reaction (why accept CHR1 as well?), but I could be convinced.

grahamgower: `Mt`, `mt`, and even `MT` all seem reasonable to me. And while I don't see an instances here, I'm sure I've seen some UCSC references with captal C for `Chr1` etc., so accepting `chr1` as well would make sense in those cases.

stdpopsim/catalog/AraTha/genome_data.py
ndukler: I think we kinda have to, otherwise the synonyms aren't actually that useful. If they could at least add the ones that we've already found that would be great. Also I don't see us having a  `list_chromosomes()` function implemented. That would allow us to get the user to get things in the correct format for us as they could easily see the formatting themselves.

jeromekelleher:
Notes from the call 07/07

- Let's not worry about getting all the extra synoyms (unless Ensembl adds them)
- We can information on sex chrs as a follow up, once this is completed.

reedacartwright:
What information do we actually need for a sex chromosome? In my mutation calling software, what is important is to know the expected ploidy of an individual at a locus 0, 1, and 2 and how it is inherited. Chromosomes X, Y, etc. are used as shortcuts for determining the polidy of males and females and how it can be transmitted to offspring. This is then used to prune a family-history graph into the proper transmission structure.

In that sense, stdpopsim would need a low level API that specifies the polidy and transmission rules for chromosomes/fragments in species that have defined sexes. On top of that would be user-friendly labels like Autosomal, X, Y, Z, etc. And on top of that would be annotating a genome based on these labels.

jeromekelleher:
Good questions @reedacartwright! I think how we actually simulate sex chrs is a big question. All we're interested in for now, I think, is to basically record the information that something is not an autosome, so that we can emit a warning.

stdpopsim/genomes.py
andrewkern: yeah i agree with this. we should make this case-insensitive just to add some robustness.

stdpopsim/catalog/AraTha/genome_data.py
andrewkern: strikes me that we could easily add some default synonyms to each species e.g., 1, chr1, Ch1

andrewkern:
guys i've spent a bit of time playing with this code this AM. first off @jeromekelleher this is awesome. 

second, i think we should merge this as is and then fine tune things as we try to maintain it. undoubtedly we will come on myriad issues trying to align ourselves with ensembl (e.g. the synonym stuff), so i vote we merge this and give it a go while our user base is still small.

stdpopsim/catalog/AraTha/__init__.py
```diff
@@ -5,39 +5,39 @@
...
 # recombination value from Huber et al 2014 MBE
 # rho=200/Mb, assume Ne=124,000, rho=2*Ne*r
 _chromosomes = []
-for line in _chromosome_data.splitlines():
-    name, length = line.split()[:2]
+for name, data in genome_data.data["chromosomes"].items():
     _chromosomes.append(stdpopsim.Chromosome(
-        id=name, length=int(length),
+        id=name,
+        length=int(data["length"]),
```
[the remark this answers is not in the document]
jeromekelleher: Fixed, thanks.

stdpopsim/catalog/DroMel/genome_data.py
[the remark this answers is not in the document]
jeromekelleher: For now we have to live with whatever Ensembl gives us. We can follow up with more, if we like.

stdpopsim/genomes.py
jeromekelleher: In the interest of getting this much merged, moving the discussion to #569

stdpopsim/catalog/AraTha/genome_data.py
jeromekelleher: Moving discussion to #570

stdpopsim/catalog/DroMel/__init__.py
jeromekelleher: Fixed.

jeromekelleher:
I've updated this with the bug-fixes, and opened issues to track the outstanding stuff raised above. In the interest of making progress I suggest we merge this much.

@grahamgower, would you mind taking a last look and merging if you're happy please?

grahamgower:
Looks good @jeromekelleher! Merging.

Issue #565, opened 2020-06-18
simulate() allows (but ignores) random_seed as well as seed parameter


hyanwong:
I've just got bitten by doing `engine.simulate(..., random_seed=123)`, using the same terminology as `msprime.simulate`, which names the parameter `random_seed`. However, I now realise that in stdpopsim, the parameter is called `seed` instead, and calling `engine.simulate(..., random_seed=123)` return different results each time. Would it be possible either to make `random_seed` an alias for `seed`, or emit a warning, or something?

Code to reproduce is, e.g. 

```
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("chr20")
r_map = contig.recombination_map
model = species.get_demographic_model('OutOfAfrica_3G09')
assert len(r_map.get_rates()) == 2  # Ensure a single rate over chr
samples = model.get_samples(each_pop_n, each_pop_n, each_pop_n)
engine = stdpopsim.get_engine('msprime')
ts1 = engine.simulate(model, contig, samples, random_seed=random_seed)
ts2 = engine.simulate(model, contig, samples, random_seed=random_seed)
t1 = ts1.dump_tables()
t2 = ts2.dump_tables()
t2.provenances.clear()
t1.provenances.clear()
t1 == t2 # expected "True" but isn't, as the param name is "seed" not "random_seed"
```

grahamgower:
Thanks for the report! Accepting arbitrary keyword args is definitely problematic, so #566 will fix that. Probably an alias is a good idea too, as you're not the first person to have been bitten by the seed/random_seed difference here (e.g. #381).

grahamgower:
I'm going to close this without adding the alias. An alias would be nice, but it would also add a (minor) maintenance burden. I'll err on the side of less code in this instance, as the user will now at least get an error and look up the correct param name.

PR #566, opened 2020-06-19
Remove **kwargs from engine.simulate().

grahamgower:
This avoids silently accepting parameters that are misspelt.
See #565.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -14,6 +14,7 @@ import pathlib
 import shutil
 import os
 import re
+import inspect
 
 import msprime
 import tskit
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Nice, very clever,

Can we just get a quick unit test so we can be sure we're closing the bug and so we don't revert in future?

grahamgower:
I'm not sure what codecov is complaining about, but I added a test to check that we can't pass arbitrary params to `engine.simulate()`.

jeromekelleher:
Perfect, thanks. Codecov is quirky, you can't be too literal with it.

PR #572, opened 2020-07-31
Add pyrho genetic maps

jeffspence:
I added genetic maps for _Homo sapiens_ for the 26 populations in Phase 3 of the 1000 Genomes Project from my paper here: https://advances.sciencemag.org/content/5/10/eaaw9206

I also alphabetized the references at the top of `stdpopsim/catalog/HomSap/__init__.py`

I'd be happy to make any additional changes (e.g., are genetic maps unit tested?).

### DIFF
--- a/stdpopsim/catalog/HomSap/__init__.py
+++ b/stdpopsim/catalog/HomSap/__init__.py
@@ -61,6 +61,13 @@ _hapmap2007 = stdpopsim.Citation(
     author="The International HapMap Consortium",
 )
 
+_takahata1993 = stdpopsim.Citation(
+    doi="https://doi.org/10.1093/oxfordjournals.molbev.a039995",
+    year="1993",
+    author="Takahata",
+    reasons={stdpopsim.CiteReason.POP_SIZE}
+)
+
 _tian2019 = stdpopsim.Citation(
     doi="https://doi.org/10.1016/j.ajhg.2019.09.012",
     year="2019",
@@ -73,13 +80,6 @@ _tremblay2000 = stdpopsim.Citation(
     year="2000",
     author="Tremblay and Vezina",
     reasons={stdpopsim.CiteReason.GEN_TIME}
-)
-
-_takahata1993 = stdpopsim.Citation(
-    doi="https://doi.org/10.1093/oxfordjournals.molbev.a039995",
-    year="1993",
-    author="Takahata",
-    reasons={stdpopsim.CiteReason.POP_SIZE}
 )
 
 _chromosomes = []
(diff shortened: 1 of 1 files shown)


andrewkern:
thanks for the PR @jeffspence. it looks like one thing that is breaking the tests currently is the URL-- would you mind putting this on a publicly accessible address for a minute so that the PR can pass tests?

@ndukler do we have a SOP on how to deal with map downloading / caching before we merge?

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -174,6 +174,36 @@
...
+        Genetic maps are only available for the 22 autosomes.
+        See https://doi.org/10.1126/sciadv.aaw9206 for more
+        details.""".format(pop),
+        url=(
+            "https://stdpopsim.s3-us-west-2.amazonaws.com/genetic_maps/"
+            "HomSap/Pyrho{}_GRCh37.tar.gz".format(pop)),
+        file_pattern=(
+            "Pyrho{}_GRCh37_chr{{id}}.txt".format(pop)),
+        citations=[
+            stdpopsim.Citation(
```
andrewkern, line 199: maybe pull this citation out of here and set it to `_spence_song_1993` as with the other citations?

ndukler:
> thanks for the PR @jeffspence. it looks like one thing that is breaking the tests currently is the URL-- would you mind putting this on a publicly accessible address for a minute so that the PR can pass tests?
> 
> @ndukler do we have a SOP on how to deal with map downloading / caching before we merge?

I haven't actually uploaded the maps yet so I expect the tests to fail. I just asked @jeffspence to throw up the PR so I could take a look through and make sure everything was in order. It looks pretty good so I'll upload the maps now.

andrewkern:
it would be great to have a sandbox to stage these things in during development. at the start we were just hosting them locally on various servers, but since we moved to AWS it's a bit harder

ndukler:
> it would be great to have a sandbox to stage these things in during development. at the start we were just hosting them locally on various servers, but since we moved to AWS it's a bit harder

I agree, but it's fairly easy to just throw things up on AWS, even this only took a half hour or so and this was absolutely massive by our standards. In fact, we can only do something of this size twice more before we run out of our 2GB allowance :). Which is really my way of saying nice work @jeffspence , this is an awesome resource for us to have. Anyway, it's all up now so I just re triggered the travis build so lets see if it works.

ndukler:
Just as a record, I think I was wrong, the files do have to be flat in the `tar.gz` archive. I fixed it and will test. If that't the case I'll update the devdocs to make sure this bit of information doesn't have to be rediscovered

ndukler:
Ok, so things look good! I cannot trigger a manual rebuild for CircleCI or appveyor so add a comment somewhere, rebase and force push to get it going again @jeffspence .

jeffspence:
Thanks @ndukler and @andrewkern ! It was a really pleasure!

PR #574, opened 2020-08-05
added code for lifting over assemblies

jradrion:
Here is my **very unpolished** code for doing liftOver between assemblies. I've tested this on humans and flies so far, and it is running without issue for me. The code requires that you have the UCSC liftOver tool installed and that you point to the appropriate chain file using the argument `--chainFile`.  The `--winLen` option controls the interval length over which a weighted average recombination rate is calculated and `--gapThresh` control the length of gaps where recombination rates are set to zero rather than using the chromosome average rate. 

An example command for lifting over the human map would be as follows:
```
python liftOver_assemblies.py 
--species HomSap 
--map HapMapII_GRCh37
--chainFile ./liftOver_chains/hg19ToHg38.over.chain.gz
--winLen 1000
--gapThresh 1000000
```

### DIFF
--- a/maintenance/liftOver.py
+++ b/maintenance/liftOver.py
@@ -1,0 +1,695 @@
+"""
+Code for liftOver between assemblies
+"""
+
+import sys
+import os
+import multiprocessing as mp
+import subprocess as sp
+import numpy as np
+import matplotlib.pyplot as plt
+from datetime import date
+from matplotlib import gridspec
+
+
+def assign_task(tmpFile, task_q, nProcs):
+    """
+    puts tasks into the queue
+    """
+    siteID = tmpFile
+    c, i, nth_job = 0, 0, 1
+    while (i + 1) * nProcs <= len(siteID):
+        i += 1
+    nP1 = nProcs - (len(siteID) % nProcs)
+    for j in range(nP1):
+        task_q.put((siteID[c : c + i], nth_job))
+        nth_job += 1
+        c = c + i
+    for j in range(nProcs - nP1):
+        task_q.put((siteID[c : c + i + 1], nth_job))
+        nth_job += 1
+        c = c + i + 1
+
+
+def create_proc(nProcs, task_q, result_q, params):
+    """
+    starts a new process
+    """
+    for _ in range(nProcs):
+        p = mp.Process(target=worker, args=(task_q, result_q, params))
+        p.daemon = True
+        p.start()
+
+
+def hapmap_to_bed(inFile, outFile):
+    """
+    convert the hapmap-formatted file into a bed formatted file
+    """
+    chromName = None
+    outFile += ".tmp1"
+    oldRates = []
+    weights = []
+    oldRateDict = {}
+    ct = 0
+    try:
+        with open(inFile, "r") as fIN, open(outFile, "w") as fOUT:
+            fIN.readline()
+            outline = []
+            for line in fIN:
+                ar = line.split()
+                if len(outline) == 4:
+                    outline.insert(2, ar[1])
+                    oldRates.append([float(outline[1]), float(outline[3])])
+                    weights.append(int(outline[2]) - int(outline[1]))
+                    oldRateDict[outline[4]] = [
+                        int(outline[1]),
+                        int(outline[2]),
+                        float(outline[3]),
+                    ]
+                    fOUT.write("\t".join([str(x) for x in outline]) + "\n")
+                    outline = [x for x in ar[:3]]
+                    outline.append("int%s" % (ct))
+                    ct += 1
+                else:
+                    outline = [x for x in ar[:3]]
+                    outline.append("int%s" % (ct))
+                    ct += 1
+        chromName = outline[0]
+        old_average = np.average(
+            np.take(np.array(oldRates), 1, 1), weights=np.array(weights)
+        )
+        return (
+            outFile,
+            chromName,
+            np.array(oldRates),
+            np.array(weights),
+            oldRateDict,
+            old_average,
+        )
+    except FileNotFoundError:
+        return (
+            outFile,
+            chromName,
+            np.array(oldRates),
+            np.array(weights),
+            oldRateDict,
+            None,
+        )
+
+
+def lift_over(inFile, outFile, chainFile):
+    """
+    run the liftOver tool on the infile
+    """
+    newFile = outFile + ".tmp2"
+    unmappedFile = outFile + ".tmp2.unmapped"
+    cmd = [str(x) for x in ["liftOver", inFile, chainFile, newFile, unmappedFile]]
+    sp.run(cmd)
+    return newFile, unmappedFile
+
+
+def base_overlap(target, query):
+    """
+    Return the number of bases that overlap between the target and query
+    """
+    return len(range(max(target[0], query[0]), min(target[1], query[1])))
+
+
+def rescale_rate(oldRateDict, ar):
+    """
+    Return the rescaled rate from the new line
+    """
+    old_interval = oldRateDict[ar[4]]
+    return (
+        float(ar[3]) * (old_interval[1] - old_interval[0]) / (int(ar[2]) - int(ar[1]))
+    )
+
+
+def bed_to_hapmap(
+    inFile, outFile, chromName, winLen, gapThresh, jobN, oldRateDict, adjAvg
+):
+    """
+    Convert the lifted-over bed file to a hapmap-formatted
+    file using weighted average
+    """
+    max_pos = 0
+    rates = []
+    intervalLens = []
+    intervals = []
+    with open(inFile, "r") as fIN:
+        for line in fIN:
+            ar = line.split()
+            if ar[0] == chromName:
+                rescaled_rate = rescale_rate(oldRateDict, ar)
+                intervals.append([int(ar[1]), int(ar[2]), rescaled_rate, ar[0], ar[4]])
+                rates.append(rescaled_rate)
+                intervalLens.append(int(ar[2]) - int(ar[1]))
+                max_pos = max(max_pos, int(ar[2]))
+    rates = np.array(rates)
+    intervalLens = np.array(intervalLens)
+    sorted_intervals = sorted(intervals)
+    new_chrom_average = np.average(rates, weights=intervalLens)
+
+    # fill in gaps between intervals with the chromosome mean rate
+    gap_ct, over_ct = 0, 0
+    outLines = []
+    skip_idx = []
+    for i in range(len(sorted_intervals) - 1):
+        if i in skip_idx:
+            continue
+        gap = sorted_intervals[i + 1][0] - sorted_intervals[i][1]
+        if gap == 0:
+            outLines.append(
+                [
+                    sorted_intervals[i][3],
+                    sorted_intervals[i][0],
+                    sorted_intervals[i][1],
+                    sorted_intervals[i][2],
+                ]
+            )
+        elif gapThresh >= gap > 0:
+            gap_ct += 1
+            if adjAvg:
+                new_average = (
+                    sorted_intervals[i][2] + sorted_intervals[i + 1][2]
+                ) / 2.0
+            else:
+                new_average = new_chrom_average
+            outLines.append(
+                [
+                    sorted_intervals[i][3],
+                    sorted_intervals[i][0],
+                    sorted_intervals[i][1],
+                    sorted_intervals[i][2],
+                ]
+            )
+            outLines.append(
+                [
+                    sorted_intervals[i][3],
+                    sorted_intervals[i][1],
+                    sorted_intervals[i + 1][0],
+                    new_average,
+                ]
+            )
+        elif gap > gapThresh:
+            gap_ct += 1
+            outLines.append(
+                [
+                    sorted_intervals[i][3],
+                    sorted_intervals[i][0],
+                    sorted_intervals[i][1],
+                    sorted_intervals[i][2],
+                ]
+            )
+            outLines.append(
+                [
+                    sorted_intervals[i][3],
+                    sorted_intervals[i][1],
+                    sorted_intervals[i + 1][0],
+                    0.000000,
+                ]
+            )
+        else:
+            tmp_idx = [i]
+            tmp_rate = [sorted_intervals[i][2]]
+            tmp_weight = [sorted_intervals[i][1] - sorted_intervals[i][0]]
+            tmp_max = sorted_intervals[i][1]
+            j = 1
+            while True:
+                if sorted_intervals[i + j][0] < tmp_max:
+                    tmp_max = max(tmp_max, sorted_intervals[i + j][1])
+                    tmp_idx.append(i + j)
+                    tmp_rate.append(sorted_intervals[i + j][2])
+                    tmp_weight.append(
+                        sorted_intervals[i + j][1] - sorted_intervals[i + j][0]
+                    )
+                    j += 1
+                else:
+                    break
+            overlap_avg = np.average(np.array(tmp_rate), weights=np.array(tmp_weight))
+            skip_idx.extend(tmp_idx)
+            outLines.append(
+                [sorted_intervals[i][3], sorted_intervals[i][0], tmp_max, overlap_avg]
+            )
+
+            # check for a gap after the overlapping windows
+            gap = sorted_intervals[i + j][0] - tmp_max
+            if gapThresh >= gap > 0:
+                if adjAvg:
+                    new_average = (sorted_intervals[i + j][2] + overlap_avg) / 2.0
+                else:
+                    new_average = new_chrom_average
+                outLines.append(
+                    [
+                        sorted_intervals[i][3],
+                        tmp_max,
+                        sorted_intervals[i + j][0],
+                        new_average,
+                    ]
+                )
+            elif gap > gapThresh:
+                gap_ct += 1
+                outLines.append(
+                    [
+                        sorted_intervals[i][3],
+                        tmp_max,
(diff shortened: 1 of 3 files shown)


grahamgower:
Nice one, thanks Jeff!

I'm a noob when it comes to lifting over, so I have some questions about the nature of this liftOver tool that maybe are a bit offtopic, but here goes anyway...
* UCSC seem to only offer precompiled binaries (https://genome-store.ucsc.edu/). There is mention of source code availablity, but its not obvious where this can be obtained. The linked license terms suggest that one is not allowed to make derivatives (3.2a in https://genome-store.ucsc.edu/media/eula/2019/10/14/UCSC_GB_EULA.pdf). Jeff: does your script then constitute a derivative work?
* How does liftOver work? Is there a publication describing it? Or is the only description to be found in the sources (which I couldn't easily find).
* What is a chain file anyway, and how can one be created?

jeromekelleher:
Thanks @jradrion, this is great!

@jeffspence - it would be super helpful if we could get your thoughts on what we're doing here. What we need to do is lift over older genetic maps to GRCh38 (for humans), and to whatever genome build is used in Ensembl for other species. As such, it's something we do rarely so the code doesn't need to be especially polished or automated, but we'd like to have some assurance that we've not introduced artifacts into the maps. Any thoughts would be much appreciated.

jradrion:
@grahamgower thanks!

> does your script then constitute a derivative work?

I'm not a lawyer, but I would argue that this does not constitute a derivative work. All I'm doing is running the UCSC liftOver tool "as is", with some pre- and post-processing going on to convert the file formats for our use. To me this is no different than running liftOver using a shell script, which I'd have to imagine would not break the EULA. If this is a concern for people it would be pretty easy to strip out the call to liftOver and split this code into two scripts, where we run the first half, then we run liftover, then we run the second half.

> How does liftOver work?

I'm also not a liftOver expert, and this is a really good question. I guess I've always assumed it uses some kind of hash function to just grab the coordinates from the chain file. To be honest, I have no idea how it works.

> What is a chain file anyway, and how can one be created?

Chain files are better described [here](http://genomewiki.ucsc.edu/index.php/Chains_Nets), but it's basically just a file of coordinates from a whole-genome alignment mapping between the two alignments. The algorithm for building the chains is published (https://www.pnas.org/content/100/20/11484.full). [This page](http://genomewiki.ucsc.edu/index.php/LiftOver_Howto) describes how to make your own.

jeromekelleher:
> I'm not a lawyer, but I would argue that this does not constitute a derivative work. All I'm doing is running the UCSC liftOver tool "as is", with some pre- and post-processing going on to convert the file formats for our use. To me this is no different than running liftOver using a shell script, which I'd have to imagine would not break the EULA. If this is a concern for people it would be pretty easy to strip out the call to liftOver and split this code into two scripts, where we run the first half, then we run liftover, then we run the second half.

Let's not worry about this - we're not distributing the maintenance code to end users, so I don't think we need to worry about license issues.

jeffspence:
This looks great!  If I understand correctly -- you map the hapmap format file to a BED, use liftOver to remap the BED and then do a bit of post-processing (filling in gaps, and averaging within windows) to clean up the bed and map back to hapmap.  My only minor quibble is on line 140 you use the mean of the rates to fill in short gaps.  I think it would make a bit more sense to use the mean rate instead of the mean of the rates (the mean of the rates up-weights short intervals and down-weights long intervals; in species with hotspots, high rates will tend to be in short intervals and low rates will tend to be in long intervals, so my guess is that the mean of the rates would be a bit higher than the mean rate).

This overall approach is a little bit different than what Adam Auton did to lift the hapmap maps over.  This is from the README from the GRCh37 LiftOver of the HapMap recombination maps(ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/working/20110106_recombination_hotspots/README_hapmapII_GRCh37_map):

> The map was generated by lifting the HapMap Phase II genetic map from build 35 to GRCh37. The original map was generated using LDhat
> as described in the 2007 HapMap paper (Nature, 18th Sept 2007). The conversion from b35 to GRCh37 was achieved using the UCSC liftOver 
> tool.
> 
> Once the liftOver was completed, the map was inspected for regions in which the genome assembly had be rearranged. Such rearrangements
> result in a dip in the lifted genetic map (i.e. a negative recombination rate). These regions were removed by setting the 
> recombination to zero within, and for 50 SNPs either side of, such regions. These regions are fairly rare, and this method removed a 
> total of 2013 SNPs from the following chromosomes.
> 
> chr1: 224 SNPs
> chr2: 209 SNPs
> chr3: 104 SNPs
> chr4: 103 SNPs
> chr7: 547 SNPs
> chr9: 292 SNPs
> chr10: 101 SNPs
> chr13: 208 SNPs
> chr15: 101 SNPs
> chrX: 124 SNPs.
> 
> All other chromosomes did not have any SNPs removed by this method.
> 
> Adam Auton

I'm not 100% sure what this actually means.  It sounds like he might have looked at the actual position in genetic distance for each line in the file, and then back-calculated the rate (from the differences in physical distances and genetic distances between adjacent lines), setting negative rates to zero.  But I think that what you did makes more sense.

I did something similar to what you're doing in my paper (https://doi.org/10.1126/sciadv.aaw9206 ; Table S2) where I inferred maps on both GRCh37 and GRCh38, and then lifted over the maps from GRCh37 to GRCh38 and computed correlations between the map inferred on GRCh38 and the map inferred on GRCh37 but lifted over to GRCh38.  The way I lifted over the map was a bit more naive than what you're doing -- I just converted the recombination rates to bed files and then ran LiftOver on the bed files without post-processing.  I only computed correlations on positions that were included in an interval in both maps, and if positions appeared in more than one interval, I used the recombination rate from the last interval.  In any case, the correlations were actually quite good.  It will obviously depend on the method to infer the recombination maps, but I got correlations (spearman and pearson on the log-recombination rates) of ~0.98 and higher even at the single base pair resolution.  Pearson on the natural scale varied much more depending on the population. I'm guessing this is because hotspots were inferred to have somewhat different heats depending on the build,  driving the differences in correlation.  But overall, my take was that running LiftOver had less of an effect than I might have thought (although of course YMMV especially for lifting over between genome builds that are less close than GRCh37 and GRCh38; also my method also uses unphased data, and I suspect that methods that use phased data might change more between different builds).

Hope this helps, and overall it looks really great!
Jeff

jradrion:
@jeffspence thank you for taking a look at this!

> I think it would make a bit more sense to use the mean rate instead of the mean of the rates

Great catch, I hadn't considered the issue of biased means. I also just realized that I `rm` the unmapped positions files automatically, which is dumb because we may want to inspect those after running this code. I'll work on modifying this commit to reflect these changes.

> I'm not 100% sure what this actually means.

I'm not either. By rearrangements I was thinking he might have meant only where contiguous intervals are reordered in the lifted-over assembly, but I'm really not sure. This also doesn't address how they dealt with gaps in the new assembly when they presumably arose.

>  In any case, the correlations were actually quite good.

This is comforting to know!

jradrion:
I edited the code to use a properly weighted mean rate for the chromosome, per @jeffspence's suggestion. I also added the flag `--retainIntermediates` to retain all of the liftOver intermediate files for inspection. Finally, I added the option `--joinThresh` which joins adjacent intervals with same rate when rounded to `joinThresh` digits precision after the decimal point.

ndukler:
As per our conference call here's some code for seeing which target regions are covered in a chain file
```
#!/usr/bin/env bash

## A script designed to take a chain file and create a bed file showing its coverage on the target genome
## where chain has query --> target directionality. Places output file in same directory as source file

bname=`basename $1`
target=$2
query=$3
oname=`readlink -e $1`
oname=`dirname ${oname}`/${target}_covered_by_${query}.bed.gz

if [[ ${bname} == *.gz ]]; then
    zcat $1 | awk 'length {if($1 ~ "chain"){chrom=$8;qChrLen=$9;qStrand=$10; if(qStrand=="+"){start=$11} else{start=qChrLen-$11}} else{if(qStrand=="+"){print chrom,start,start+$1;start=start+$1+$3}; if(qStrand=="-"){print chrom,start-$1,start;start=start-$1-$3}}}' | sort-bed - | bedops -m - | awk 'BEGIN{OFS="\t"}{print $1,$2,$3}' | gzip -c > ${oname}
else    
    awk 'length {if($1 ~ "chain"){chrom=$8;qChrLen=$9;qStrand=$10; if(qStrand=="+"){start=$11} else{start=qChrLen-$11}} else{if(qStrand=="+"){print chrom,start,start+$1;start=start+$1+$3}; if(qStrand=="-"){print chrom,start-$1,start;start=start-$1-$3}}}' $1 | sort-bed - | bedops -m - | awk 'BEGIN{OFS="\t"}{print $1,$2,$3}' | gzip -c > ${oname}
fi
```

jradrion:
Thanks, @ndukler. I'll have a look.

jradrion:
I found some issues with the code that I will be fixing over the coming days so we should hold off on merging for the time being

jeromekelleher:
It's annoying we're failing on circle because we can't import the new script - I think this is because we're asking nose to track coverage on the maintenance package. Might be simplest to just put ``matplotlib`` into the list of dev dependencies, although this might slow down our install time a fair bit.

@grahamgower, any thoughts on the simplest way out here? I don't think we want to try to write tests to cover this code, so I'm not sure what's the best thing to do here.

grahamgower:
Hmm... interesting. Does this go away if we stop asking for coverage in the maintenance folder? If its as simple as that, then I'd say lets just not bother tracking coverage there.

jeromekelleher:
> Hmm... interesting. Does this go away if we stop asking for coverage in the maintenance folder? If its as simple as that, then I'd say lets just not bother tracking coverage there.

Yes, that's probably the best thing all right.

jradrion:
I've again updated this script. The newest version avoids doing a weighted average for all non-overlapping windows across the chromosome, and instead only performs weighted averaging for overlapping intervals in the lifted map (a relatively rare occurrence). This dramatically reduces the error when mapping back to the original assembly, and I think this is really the way to go. Additionally, I added a new option to fill gaps with either the chromosome mean, or with the mean of the flanking intervals which seems to be a better choice in most instances.

I've also added maintenance code for the liftover of the comeron map from dm3 to dm6, as discussed in #592.

grahamgower:
Thanks @jradrion, this is now at the top of my queue of stdpopsim things. It's quite long though, so I probably wont get to it for a few more days. (And Jerome is on parental leave). Perhaps you could look into avoiding the circleci failure in the mean time? It might be sufficient to remove the `maintenance` folder from the `nostests` command in `.circleci/config.yml` (and if that doesn't do the trick, let me know and we'll devise a new strategy).

jradrion:
@grahamgower OK, looks like removing `maintenance` from `nostests` did the trick.

andrewkern:
@jradrion do we really want to get rid of those tests?

grahamgower:
This change made here doesn't change which tests are run. What it does do, is change which folders code coverage is reported for when running the tests (it was looking under both `stdpopsim` and `maintenance`, now it will just look under `stdpopsim`). Annoyingly, the continuous integration was failing because it wanted to `import` everything under the maintenance folder to check its coverage. This failed because @jradrion's script(s) import stuff that isn't installed on the circleci machine (and we preferred not to depend on matplotlib et al.)

grahamgower:
Hey @jradrion, I've looked over this code a couple of times now, but haven't tried running it. I think there are a few code formatting issues that should be fixed. Can you try running `flake8 --max-line-length 89 maintenance/lift*.py`? We should be running that in the continuous integration as well, but I guess it hasn't been added yet (you're welcome to add it, if you like).

It would be nice to include a unittest or two for this in the test suite as well. How about including a small chain file for a small section of one chromosome, and store the pre-lifted and post-lifted rate maps for that? Then the unit test would check that the lift over script produces same output as the pre-calculated post-lifted maps.

jradrion:
Thanks for looking it over, @grahamgower. I'll fix the formatting, add some tests, and update.

maintenance/liftOver_assemblies.py
```diff
@@ -0,0 +1,540 @@
...
+import tarfile
+import urllib
+import shutil
+from datetime import date
+import stdpopsim
+import matplotlib.pyplot as plt
+from matplotlib import gridspec
+
+
+def progress_bar(percent, barLen = 50):
```
jeromekelleher, line 20: [tqdm](https://github.com/tqdm/tqdm) is a handy way of doing this.

maintenance/liftOver_assemblies.py
```diff
@@ -0,0 +1,540 @@
...
+    for i in range(barLen):
+        if i < int(barLen * percent):
+            progress += "="
+        else:
+            progress += " "
+    sys.stdout.write("[ %s ] %.2f%%" % (progress, percent * 100))
+    sys.stdout.flush()
+
+
+def assign_task(tmpFile, task_q, nProcs):
```
jeromekelleher, line 35: Seems pretty heavyweight to need our own task queue here - have you tried using concurrent.futures?

maintenance/liftOver_assemblies.py
```diff
@@ -0,0 +1,540 @@
...
+        finally:
+            task_q.task_done()
+
+
+def remap(intermediatesDir, validationDir, inFiles, outFiles, chainFile, winLen, gapThresh, adjAvg, nProc, validation=False, genome=False):
+    """
+    run all functions to translate and liftOver file types
+    """
+    # multithread remapping
+    task_q = mp.JoinableQueue()
```
jeromekelleher, line 430: This does seem unneccesarily low-level here - there's a good example in the [msprime code](https://github.com/tskit-dev/msprime/blob/main/verification.py#L4755) of using concurrent.futures along with a tqdm progress bar.

maintenance/liftOver_assemblies_dm3TOdm6.py
```diff
@@ -0,0 +1,538 @@
...
+import pandas as pd
+import tarfile
+import urllib.request
+import shutil
+from datetime import date
+import matplotlib.pyplot as plt
+from matplotlib import gridspec
+
+
+def progress_bar(percent, barLen = 50):
```
jeromekelleher, line 21: This looks like a near-identical copy of the liftover_assemblies.py file. Having multiple copies of the same code is almost never a good idea, and will be a significant maintenance burden over time (especially if we have lots of species - consider what  happens if we find a small bug in the liftover code, which has been copied n times).

How about we have a file ``liftover.py`` which provides the basic utilities that you're using here like ``hapmap_to_bed`` etc, and then the code specific to each liftover that we do is in its own python file that imports ``liftover``?

jeromekelleher: That is, basically just the ``main`` functions which you currently have?

jeromekelleher (changes requested):
Looks great, thanks @jradrion!

There's a few high-level structural issues here that are resulting in us having a lot more code than we really need, but I think they can be resolved pretty easily. I'm happy to provide more input if the comments aren't clear.

jradrion:
@jeromekelleher Thank you for your comments! I'll look into implementing both tqdm and concurrent.futures, and I'll fix the redundancies.

jradrion:
@jeromekelleher @grahamgower 
I made the requested flake8 formatting changes and moved all redundant functions to `liftOver.py`, which is then imported by both `liftOver_catalog.py` and `liftOver_comeron2012.py`. As a reminder, `liftOver_comeron2012.py` will be the archived code for how we generated our D. melanogaster dm6 map from the Comeron et al 2012 map, where `liftOver_catalog.py` is the code used when you want to liftOver from any of the current maps to a new map in the future (e.g. Hg37 -> Hg38).

I had quite a difficult time trying to wrap my head around how exactly I would modify this code to use `concurrent.futures`. I think this would take a sizeable amount of restructuring. I'm wondering if we could potentially hold off until after this code if merged and then open a new issue if we still want to restructure with `concurrent.futures`. Is the main motivation for using `concurrent.futures` just to have simpler maintenance code going forward? Let me know if you would be agreeable to this or this needs to be changed now.

jradrion:
Also, is there a consensus for whether we want to have tests for all maintenance code? This code is still failing on CircleCi. Should I be writing tests for these scripts next?

jradrion:
thank you @mufernando for helping me out of git hell.

jeromekelleher (approved):

jeromekelleher:
This is a big improvement, thanks @jradrion! I'm happy for this to be merged as-is.

There's still a bit of cleanup that could be done  (variable naming conventions are a bit inconsistent with a mixture of camelCase, CamelCase, ALLCAPS, and snake_case) but it's not that important for this maintenance code like this. The motivation for using concurrent.futures would be to remove redundant code, making it easier to understand what the file is actually for. Readers looking at the code might currently think there's some deep reason for needing to do tricky low-level concurrency stuff, and it takes a few minutes to figure out that there isn't. 

Anyway, this is nit-picking;  awesome work, thanks!

grahamgower:
Awesome! Thanks @jradrion! I've added a small commit to fix the circleci failure. I couldn't seem to push to your PR branch, so I had to open a new PR (#691). Will merge that and close this as soon as the CI lights are green.

grahamgower:
Thanks again @jradrion. And sorry for mess of CI changes you had to deal with during the life of this PR. Would you be able to open a couple of follow-up issues about the stuff raised by @jeromekelleher?

jradrion:
Awesome, thanks @grahamgower. I will open issues about the comments made by @jeromekelleher.

jradrion:
Thank you to @jeromekelleher, @grahamgower, @jeffspence, and @mufernando for all of your help with this.

PR #578, opened 2020-08-18
Adds CLI to make downloading species data easier

ndukler:
Added CLI to maintenance script for downloading species data

### DIFF
--- a/update_ensembl_data.py
+++ b/update_ensembl_data.py
@@ -11,7 +11,8 @@ import contextlib
 
 import stdpopsim
 import maintenance
-
+import argparse
+import os
 
 class DataWriter:
     """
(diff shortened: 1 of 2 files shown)


update_ensembl_data.py
```diff
@@ -29,7 +41,7 @@ def write(self, path):
             yield f
         # TODO we can remove this once we've moved to using pre-commit and
         # Black, as these will be autoformatted.
-        subprocess.check_call(["black", "-q", path])
+        # subprocess.check_call(["black", "-q", path])
```
jeromekelleher, line 44: I think we should keep the ``black`` call here - otherwise the code we're auto inserting will be horrible.

jeromekelleher: Maybe add a comment here for developers for now saying "use ``pip install black`` if this command fails".

update_ensembl_data.py
```diff
@@ -11,7 +11,19 @@
 
 import stdpopsim
 import maintenance
+import argparse
+import os
 
+# Parser that controls which species should be updated
+parser = argparse.ArgumentParser(description='Query ensembl API for chromosome lengths')
+parser.add_argument('--list-species', dest='list_species', action='store_true',
```
jeromekelleher, line 19: I think we can simplify this a bit - the ``--list`` isn't really needed since we can get the list of species from a bunch of other places. What about if we had the default of doing what it's currently doing (all species), so if we call
```
python update_ensembl_data.py
```
it'll run through all species and if we call 
```
python update_ensembl_data.py HomSap BosTau
```
it'll just do those two. Basically, it's just a case of putting one positional argument, rather than two mutually exclusive mandatory "options".

jeromekelleher (commented):
Looks good, thanks @ndukler. I think a small change to the UI would simplify it a lot, but it's definitely a really useful addition to the utility.

update_ensembl_data.py
ndukler: Ok, I was having trouble with it but once I'm done with local testing I'll un-comment it.

update_ensembl_data.py
ndukler: And the `--list` bit is just so that someone doesn't have to go hunting for a way to list all the registered species. I'm particularly thinking of my current case where a new developer is submitting a species but is not familiar with `stdpopsim` generally.

jeromekelleher: Well, the update_all is harmless as it'll just over-write identical data:
```
$ python3 update_ensembl_data.py 
Using Ensembl release 101
Writing genome data for AraTha arabidopsis_thaliana
Writing genome data for CanFam canis_familiaris
Writing genome data for DroMel drosophila_melanogaster
... [Ctrl-c'd out of boredom]
$ git diff
diff --git a/stdpopsim/catalog/ensembl_info.py b/stdpopsim/catalog/ensembl_info.py
index 4800dda..1d96ff5 100644
--- a/stdpopsim/catalog/ensembl_info.py
+++ b/stdpopsim/catalog/ensembl_info.py
@@ -1,2 +1,2 @@
-release = 100
+release = 101
```

So, we updated 3 species, but there's no diff in git (other than we seem to have updated to Ensembl release 101).

Having mandatory options (you must specify either --species or --update-all) gives me hives.

Fair enough having --list is handy, so let's keep that.

I have strong opinions about CLI structure I'm afraid @ndukler :wink:

update_ensembl_data.py
```diff
@@ -41,17 +53,31 @@ def write_genome_data(self, species):
...
             print("release = ", release, file=f)
 
 
-def main():
+def main(args):
+    # Lists species and exits if user requested list of species
+    if args.list_species:
+        for species in stdpopsim.all_species():
+            print(species.id)
+        exit(0)
```
jeromekelleher, line 66: Does this work? I think you need to call ``sys.exit(0)``, right?

Rather than yanking out of the interpreter directly here, it's a bit more pleasant to return here, and let control flow continue normally.

update_ensembl_data.py
ndukler: I'll change it. We can always catch a surprise unintentional update at the review stage.

ndukler:
@jeromekelleher Think I fixed the hairball by using ensembl_ids instead and generating the six character ids directly from those. Also added some error checking to make it less likely an invalid species name slips through to the step where ensembl is queried.

ndukler:
Just going to update the documentation and then we should be good to go

ndukler:
Ok, just squashed stuff. Pending the checks I think that this is good to go @jeromekelleher

jeromekelleher:
Looks good, thanks @ndukler

PR #590, opened 2020-09-30
Convert QC code to use functions instead of classes.

### LINKED ISSUE #304: Refactor QC module

grahamgower:
Also reformatted the code with Black.
Closes #304.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -686,7 +686,7 @@ Write the model function in the catalog source code
 In the ``stdpopsim`` catalog source code (found in ``stdpopsim/catalog/``),
 each species has a module that defines all of the necessary functions to run
 simulations for that species, including the demographic model. In each species module,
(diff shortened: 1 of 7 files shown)


jeromekelleher (approved):
LGTM! 

(Should we Blacken all the code? Seems like a good time to do it?

jeromekelleher:
Nice tidy up, thanks @grahamgower. Merging.

grahamgower:
> LGTM!
> 
> (Should we Blacken all the code? Seems like a good time to do it?

Yep. I'd like to go through a few issues this week, and that is definitely low hanging fruit!

Issue #542, opened 2020-05-26
QC for EscCol


grahamgower:
The **EscCol** species definition has not yet been QCed!

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [x] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [x] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [x] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [x] Population size.
- [x] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [x] Genome reference assembly.
- [x] Mutation rate.
- [x] Recombination rate.
- [x] Recombination map(s) (if relevant).
- [x] Population size.
- [x] Generation time.

grahamgower:
I've recently reviewed some items on this list, and thus they're already checked.

grahamgower:
We're using `recombination_rate=0` here, which I guess is ok. But the population size estimate seems ludicrous, in part because it was calculated using a mutation rate that is five orders of magnitude different to the mutation rate we're using to simulate. This seems wildly inconsistent.

Should we remove EscCol? I honestly don't see how this could be useful to anyone in its current form. Any E.coli experts in the audience interested in making this work (#593 might clean some of this up)? Ping @popsim-consortium/all

jeromekelleher:
I think @fbaumdicker brought this in initially - any thoughts?

fbaumdicker:
Yes, I brought this in! I am really convinced we should keep this in. If i remember correctly, @ndukler did the quality control back at the time when this was added. Not sure why this is not recorded in the repo??

[...]

PR #593, opened 2020-10-01
EscCol: use the K-12 MG1655 strain, rather HUSEC2011 (ensembl default).

grahamgower:
This better matches the parameter values we're using, but at the cost
of increased complexity. Note that we're still using parameter values
that do not necessarily apply to K-12.

Part of the species QC process in #542.

But is using this particular set of parameters for EscCol even useful to anyone? Fun fact: attemptying to simulate this (`stdpopsim EscCol 10 -o /dev/null`) eats all your memory until the process gets killed. It would be nice to have an E. coli expert weigh in here.


stdpopsim/catalog/ensembl_info.py
```diff
@@ -1,2 +1,2 @@
 # File autogenerated from Ensembl REST API. Do not edit.
-release = 100
+release = 101
```
jeromekelleher, line 2: Can you confirm that the script ran for all genomes? We don't want to have some at release 100 and others at 101 (although I'm sure there's probably no changes).

stdpopsim/catalog/EscCol/__init__.py
```diff
@@ -40,39 +37,52 @@
...
+            _wielgoss_et_al.because(stdpopsim.CiteReason.MUT_RATE),
             ],
         assembly_citations=[
             _blattner_et_al.because(stdpopsim.CiteReason.ASSEMBLY)])
 
-_species = stdpopsim.Species(
+
+# Nasty hack to overwrite ensembl_id property. Monkey patching won't work in
+# this case, as ensembl_id is a property without a setter.
+class Ecoli(stdpopsim.Species):
```
jeromekelleher, line 60: Should we have a different way of specifying the ensembl_id? It seems like this stuff might be common,  and it's only used for doing the maintenance stuff, so feel free to change it if it's getting in the way.

jeromekelleher (approved):
LGTM. Might be worth taking a look at how ensembl_id is defined/used - could be easier to change it than the current approach. It's not that big a deal to just drop a literal definition into the species files, e.g.

stdpopsim/catalog/ensembl_info.py
grahamgower: > Can you confirm that the script ran for all genomes?

Yep.

stdpopsim/catalog/EscCol/__init__.py
grahamgower: I changed ensembl_id to be a class attribute rather than a property.

PR #600, opened 2020-10-02
Add Bos taurus

grahamgower:
Hey @iechevarriaz, thanks for your hard work here! I've made some changes to the BosTau species, which are mostly cosmetic. The main non-cosmetic change I made was to use the ancestral Ne (62000 from MacLeod et al.) as the species default, rather than the estimate for recent years of Ne=90.


grahamgower:
I'm going to merge this now, and open a couple of follow-up issues.

PR #604, opened 2020-10-02
Rounded mutation & recomb rate to the 4th dec

ragreenburg:


grahamgower:
Hi @ragreenburg. I think it would be better to keep the precision of the parameters themselves, but just change how many digits are being printed in the `get_species_help()` function. https://github.com/popsim-consortium/stdpopsim/blob/403e350ec71e57e8c13355ea516f25cc2a4684bb/stdpopsim/cli.py#L182-L194

ragreenburg:
> 
> 
> Hi @ragreenburg. I think it would be better to keep the precision of the parameters themselves, but just change how many digits are being printed in the `get_species_help()` function.
> 
> https://github.com/popsim-consortium/stdpopsim/blob/403e350ec71e57e8c13355ea516f25cc2a4684bb/stdpopsim/cli.py#L182-L194

Okay, I will fix that!

grahamgower:
It might be simpler to use python's [format specifier](https://docs.python.org/3.4/library/string.html#format-string-syntax) here. I didn't see a simple example in the python docs, but I was thinking it would look something like `f"Mutation rate: {species.genome.mean_mutation_rate:.4g}\n"` (use scientific notation (`g`) with 4 digits after the decimal place `.4`).

grahamgower:
Perfect! Thanks @ragreenburg! Sorry for the run-around :-) One last thing, and I think it'll be ready to merge. Could you "squash" all your changes down to one commit? (See https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing)

I'm off to bed now, but I'll take another look first thing in the morning.

ragreenburg:
@grahamgower Sure, I can do that. Absolutely no problem haha. Have a good night!

grahamgower:
Looks good @ragreenburg. Merging now.

PR #613, opened 2020-10-09
Scale the selection coefficient when using a MutationType.

### LINKED ISSUE #612: slim engine doesn't apply scaling factor in `initializeMutationType()`
**Short description of the problem:**
When asking for a `MutationType` with non-zero selection coefficient (e.g. gamma distributed), the selection coefficient is used directly, without scaling by the scaling factor `Q`. Found in discussions with @RasaMukti.

**How to reproduce the problem:**
The DFE example in `docs/selection_example.py` is affected by this bug (but the adaptive introgression example is not).

**Suggested fixes:**
Generate SLiM code that multiplies the selection coefficient in the call to `initializeMutationType()`. This would need to do different things for, e.g. fixed selection-ceof types (`distrubtionType='f'`) versus gamma distributed selection-ceof types (`distributionType='g'`; multiply the mean value by the scaling factor here?).

Ping @petrelharp. Thoughts? Does this fix sound sane in general? (I know you don't like scaling, particularly for non-neutral sims!)

grahamgower:
Fixes #612.

Not sure how to unittest this. We could just check for `Q` in the initializeMutationType lines of the generated slim script, but this seems fairly fragile.

### DIFF
--- a/stdpopsim/ext/selection.py
+++ b/stdpopsim/ext/selection.py
@@ -49,6 +49,16 @@ class MutationType(object):
             raise ValueError(
                     f"{self.distribution_type} is not a supported distribution type")
 
+        # The index of the param in the distribution_args list that should be
+        # multiplied by Q when using --slim-scaling-factor Q.
+        self.Q_scaled_index = {
+            "e": 0,  # mean
+            "f": 0,  # fixed value
+            "g": 0,  # mean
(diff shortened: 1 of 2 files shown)


petrelharp (approved):
Looks good to me!

andrewkern (approved):
nice. gonna merge these changes.

Issue #622, opened 2020-10-15
conda issue with the documentation's builds


David-Peede:
On Tuesday I attempted to update the citation in the docs, but if you check the documentation you will notice that it still has the old citation https://stdpopsim.readthedocs.io/en/latest/introduction.html

Here are the steps used to diagnose this problem (S/o to @agladstein for the help!)

1. The PR was indeed pulled https://github.com/popsim-consortium/stdpopsim/pull/616
2. Double checked to make sure I actually updated the citation https://github.com/popsim-consortium/stdpopsim/pull/616/files
3. Make sure that my commit was the latest commit in the introduction section of the docs https://github.com/popsim-consortium/stdpopsim/blob/master/docs/introduction.rst
4. Make sure I was looking at the latest version of the docs (which I was)
5. Check the documentation's builds 
-the past three attempted builds have failed https://readthedocs.org/projects/stdpopsim/builds/
-there seems to be an issue with the conda packages https://readthedocs.org/projects/stdpopsim/builds/12108862/
6. I don't have any intuition about what the next steps are so I made an issue for it!

Thanks again to @agladstein for her help on what to do in these situations!

andrewkern:
weird. could it be that readthedocs hasn't updated yet? when i build the docs locally i get the version with this merged PR.

andrewkern:
this makes it look like py3.9 isn't compatible with the RTD conda env 
![image](https://user-images.githubusercontent.com/7674197/96145641-489eba80-0eba-11eb-8472-fd9c9e42db60.png)

grahamgower:
You beat me to it @andrewkern :)
Looks like we'll need to pin python 3.8 for now, at least until the rtd/conda/python stars align.

@David-Peede, would you like to have a go at adding a line to `requirements/rtd-conda-environment.yml` to pin python to version `3.8`?

David-Peede:
Sure! I can try to take a crack at it today @grahamgower ! But just to make sure I understand correctly when you say "pin python version to version 3.8" does that mean the upper bound will be python version 3.8.6 instead of 3.9.0, or the does that mean the upper bound will be version 3.8.0?

grahamgower:
I meant any 3.8.x version. I think, but am not certain, that if you say `python=3.8` it will install the latest version that starts with 3.8. If that doesn't do the trick though, you can say `python<3.9`.

grahamgower:
Fixed by #623

PR #623, opened 2020-10-15
pin python 3.8

David-Peede:
pinned python to version 3.8 in rtd-conda-environment.yml


petrelharp:
Why do we need to do this?

petrelharp:
Oh, never mind: #622

grahamgower:
Looks good! Annoyingly, the rtd build doesn't get kicked off during continuous integration. So let's merge and see what happens...

David-Peede:
Looks like it worked!

PR #624, opened 2020-10-16
lint fixes for maintenance folder

grahamgower:
See #615

PR #625, opened 2020-10-16
Test lint fix

platipo:
Just a test for the pinned requirements from #624


grahamgower:
Hmm... CircleCI says that succeeded, but I see errors in the logs. Maybe something is failing earlier, but the error doesn't get propagated through? Also, this message looks concerning:

> DEPRECATION: pomegranate was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. pip 21.0 will remove support for this functionality. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368.
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

platipo:
The deprecation warning have to be handled by the pomegranate team and the error is actually a warning about the new version of pip.

platipo:
I won't be too worried about it but this is a think to take care

platipo:
I'm closing this because I just wanted to test the errors

PR #629, opened 2020-10-16
dependabot following tskit's lead

### LINKED ISSUE #627: use dependabot to pin and auto update dependency versions
https://dependabot.com/
https://github.com/tskit-dev/tskit/pull/517

andrewkern:
just following tskit's lead here it looks like we only need this one file to set up dependabot.

@grahamgower i added you as default reviewer! feel free to not be

closes #627

### DIFF
--- a/.dependabot/config.yaml
+++ b/.dependabot/config.yaml
@@ -1,0 +1,9 @@
+version: 1
+update_configs:
+  - package_manager: "python"
+    directory: "/requirements/"
+    update_schedule: "live"
+    default_reviewers:
+      - "grahamgower"
+    default_labels:
+      - "dependancy-upgrade"


grahamgower:
Thanks @andrewkern, I ok with being the default reviewer for dependabot pull requests. I haven't looked at how this works at all, but don't we need to do something like https://github.com/tskit-dev/tskit/pull/516/files and also create a dependabot account?

andrewkern:
i just signed up the `stdpopsim` repo. i _think_ this is all we need. will merge this in and then see if it works..

andrewkern:
i think this is working @grahamgower? https://app.dependabot.com/accounts/popsim-consortium

grahamgower:
Yeah, it looks like its enabled and looking at the repository.

What Ben did with tskit is pin dependency versions for continuous integration, which gives greater reliability when dependent software releases new versions (e.g. CI wont automatically use a new version of numpy when it's released). The idea here is to avoid the maintenance burden associated with debugging broken CI if new dependency versions get released and break our software. Instead, when a new version is released, dependabot will submit a PR to update the pinned version, and this will show if CI will pass/fail when the version is bumped (and we can merge, or alternately ignore or delay the update). 

So we need to create a specific requirements file for CI (separate to development.txt---tskit has a `requirements/CI/requirements.txt` and points dependabot at that), in which all dependencies are pinned to a recent version. I think at the moment, dependabot won't submit PRs because we're not pinning versions (well, except for zarr).

andrewkern:
okay i can get a second PR done to pin things. will try to get to it later this AM.

PR #631, opened 2020-10-18
Document that mutation rate is set by the contig during simulation

hyanwong:


grahamgower:
LGTM! Thanks @hyanwong!

PR #633, opened 2020-10-18
pinning dependabot reqs into file; adjusted config

andrewkern:
will this do it @grahamgower?


grahamgower:
Yeah, I think that should do the trick, thanks @andrewkern! Let's merge it in and find out!

grahamgower:
Oh, no. I spoke too soon. The circleci config needs to be updated to use this too.

andrewkern:
@grahamgower is this where we need to change the circleci config? 
https://github.com/popsim-consortium/stdpopsim/blob/67e26c574ac6ae1a1d2b5e1882b500a3976e7171/.circleci/config.yml#L77

will all of the circleci stuff now point to the new requirements.txt I created? should I delete the old one? never have done this before...

grahamgower:
I think all of the circleci lines that reference `requirements/development.txt` need to be changed to `requirements/CI/development.txt`. The old requirements/development.txt should probably be left alone (at least for now?), because this is what we recommend to use in the dev-docs.

To be honest, I'm way out of depth will all this stuff, and just guessing. I'm about to go offline for today, so feel free to just merge it in when you're happy, and we'll see what happens.

andrewkern:
sounds good. will switch the refs in in the circleci config file, merge, and we can see if it all works!

Issue #638, opened 2020-10-19
streamline CI for PRs


petrelharp:
In #632, @hyanwong ran across a number of inscrutible problems when simply submitting a PR to fix up some documentation. We should make this easier, and document what needs to happen. Maybe @platipo has some insight (in particular, why @hyanwong needed to enable orbs on *his* circleCI account)?

hyanwong:
Or, indeed, why I would want my own circleCI account in the first place.

petrelharp:
Yeah - I think that we'd like it if you *don't* need a circleCI account.

hyanwong:
I have a dim recollection that one was semi-automatically set one for me when I clicked on the CI failure in the GitHub issue the first time round.

andrewkern:
I'm naive-- what does this orbs thing do for us?

grahamgower:
An orb is a third-party library for circleci (like a wheel for python, but this is for circleci).

I'm fairly certain there's a button for configuring whether actions get triggered in forks, we just need to figure out where the setting is hiding. I looked through the circleci config file and didn't see it, so maybe its in the popsim-consortium github settings, or the popsim-consortium circleci settings?

grahamgower:
I think what happened is that when @hyanwong pushed code to the https://github.com/hyanwong/stdpopsim/tree/patch-2 branch,  the circleci config triggered a circleci build using @hyanwong's circleci account (the default behaviour according to circleci https://circleci.com/docs/2.0/oss/#only-build-pull-requests). Then, because the commit hash for the PR commits are the same as for that fork/branch, circleci didn't rerun the build for the popsim-consortium circleci account (instead, just linking to the existing circleci build).

My bet is that this was happening before the config refactoring in #620, but no one ever noticed because (1) not many PRs come from folks with a circleci account configured, and (2) the build status would have been the same regardless of whos account it the build ran under. Maybe there's also a race going on here, where the fork triggered the circleci build before the PR did? I guess that one "fix" here, is for @hyanwong to toggle the "only build pull requests" setting in his circleci account?

grahamgower:
I've toggled the "Only build pull requests" setting for the stdpopsim project settings in circleci. I don't know if this is sufficient.

platipo:
Hi everyone, I'm joining late the conversation and speaking from my limited CircleCI experience of this month.

[...]

PR #640, opened 2020-10-21
test edit - do not merge

petrelharp:
Testing for #638.  Note this is a PR from the popsim-consortium repo (obtained by clicking 'edit' on a file).

PR #642, opened 2020-10-21
test PR

petrelharp:
Testing for #638, this time from my own repository.


petrelharp:
Well, this also successfully ran, but seems to be using the stdpopsim CircleCI account to do so, just like #640
Screenshot from 2020-10-21 07-40-44
Based on @hyanwong's experience in #638 I expected this to try to run on my account, where I hadn't enabled orbs.

Issue #587, opened 2020-09-29
support genomicElements and genomicElementTypes in the slim engine


grahamgower:
This will be needed to support, e.g. exon/intron differentiation, for background selection models. Probably we need a python class to match the SLiM `GenomicElementType` class, which will also contain a list of genomic intervals that use that type.

grahamgower:
Closed by #644.

PR #644, opened 2020-10-22
Genomic element

mufernando:
OK, so the plan is as follows:

One of the arguments to `simulate` is a `Contig` object. This is the class that represents the (contiguous) region to be simulated. It contains information about the mutation rate and the recombination map.

The goal of this PR is to implement a notion of genomic elements (sensu SLiM). These genomic elements are non-overlapping regions of the contig (as determined by start and end positions) in which certain mutation types can occur.

Mutation types can occur at different rates within each genomic element types, and so an important parameter is the proportion of *all* mutations that are going to be each mutation type. By default, a simulation will have one genomic element spanning the whole contig, in which only neutral mutations can occur (and are added via msprime after the SLiM run). 

You may create different mutation types (with different selection parameters and proportions) and assign them to genomic element types. In this case, SLiM is going to simulate the non-neutral mutations in the determined genomic elements, and at the end we will only add neutral mutations so that the total mutation rate (across all types, including neutral) is equal to the contig's mutation rate (see more detailed explanation here: #646).

WIP

Solves #587

### DIFF
--- a/stdpopsim/ext/selection.py
+++ b/stdpopsim/ext/selection.py
@@ -13,15 +13,8 @@ class MutationType(object):
     distribution_type = attr.ib(default="f", type=str)
     distribution_args = attr.ib(factory=lambda: [0], type=list)
     convert_to_substitution = attr.ib(default=True, type=bool)
-    # MutationTypes with non-zero weight will be simulated by SLiM,
-    # using rates obtained from the relative weights of the types.
-    # I.e. the weights will be used in the ``proportion`` parameter
-    # to SLiM's :func:`initializeGenomicElementType()`.
-    weight = attr.ib(default=0, type=float)
 
     def __attrs_post_init__(self):
-        if self.weight < 0:
-            raise ValueError("weight must be >= 0.")
         if self.dominance_coeff < 0:
             raise ValueError(f"Invalid dominance coefficient {self.dominance_coeff}.")
 
@@ -61,19 +54,6 @@ class MutationType(object):
             "n": 1,  # standard deviation
             "w": 0,  # scale
         }[self.distribution_type]
-
-
-def slim_mutation_frac(mutation_types):
-    """
-    The fraction of mutations that should be added by SLiM.
-    The remainder are added by msprime.mutate() once the SLiM part
-    of the simulation is complete.
-    """
-    if mutation_types is None:
-        weighted = False
-    else:
-        weighted = any(mut_type.weight > 0 for mut_type in mutation_types)
-    return 1 if weighted else 0
 
 
 @attr.s
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -1,7 +1,13 @@
 """
-Infrastructure for defining information about species' genomes.
+Infrastructure for defining information about species' genomes and genomic
+regions to be simulated.
 """
+import logging
 import attr
+import numpy as np
+import stdpopsim
+
+logger = logging.getLogger(__name__)
 
 
 @attr.s
@@ -107,12 +113,40 @@ class Chromosome:
     synonyms = attr.ib(factory=list, kw_only=True)
 
 
+@attr.s(kw_only=True)
+class GenomicElementType(object):
+    """
+    Class that represents a map between genomic intervals and a group of
+    mutation types and the rate in which they occur within those intervals.
+
+    :ivar mutation_type_ids: Indices for the mutation types that can occur in
+        genomic elements of this type.
+    :vartype mutation_type_ids: list
+    :ivar proportions: Proportions (out of the mutation rate) in which each
+        mutation type can happen in genomic intervals of this type.
+    :vartype: proportions: numpy.array, dtype=np.float32
+    :ivar intervals: (n, 2) numpy array with [left, right) genomic intervals
+        that will be of this GenomicElementType. The intervals are used to
+        specify SLiM GenomicElements. Intervals must be non-overlapping.
+    :vartype intervals: numpy.array, dtype=np.int32
+    """
+
+    mutation_type_ids = attr.ib(factory=lambda: [], type=list)
+    proportions = attr.ib(factory=lambda: np.array([]), type=np.ndarray)
+    intervals = attr.ib(default=None, type=np.array)
+
+    def __attrs_post_init__(self):
+        _check_mut_proportions(self.proportions, self.mutation_type_ids)
+        stdpopsim.utils.check_intervals_validity(self.intervals)
+
+
 @attr.s
 class Contig:
     """
     Class representing a contiguous region of genome that is to be
     simulated. This contains the information about mutation rates
     and recombination rates that are needed to simulate this region.
+    Genomic element types can be provided
 
     :ivar mutation_rate: The rate of mutation per base per generation.
     :vartype mutation_rate: float
@@ -129,6 +163,13 @@ class Contig:
     :ivar exclude: If True, ``mask_intervals`` specify regions to exclude. If False,
         ``mask_intervals`` specify regions in keep.
     :vartype exclude: bool
+    :ivar genomic_element_types: a list of stdpopsim.GenomicElementType
+        objects. By default, the only element is a single genomic element type
+        that spans the entire contiguous region.
+    :vartype genomic_element_types: list
+    :ivar mutation_types: a list of stdpopsim.ext.MutationType object. By
+        default, the only mutation type is neutral. See
+        stdpopsim.MutationType for more details.
 
     .. note::
         To run stdpopsim simulations with alternative, user-specified mutation
@@ -150,16 +191,154 @@ class Contig:
     genetic_map = attr.ib(default=None, kw_only=True)
     inclusion_mask = attr.ib(default=None, kw_only=True)
     exclusion_mask = attr.ib(default=None, kw_only=True)
+    genomic_element_types = attr.ib(factory=lambda: [], type=list, kw_only=True)
+    mutation_types = attr.ib(factory=lambda: [], type=list, kw_only=True)
+
+    @property
+    def length(self):
+        if self.recombination_map is None:
+            return None
+        return self.recombination_map.get_length()
+
+    @property
+    def slim_fractions(self):
+        map(lambda g: _check_mut_proportions(g.proportions), self.genomic_element_types)
+        props = np.array([sum(g.proportions) for g in self.genomic_element_types])
+        return props
+
+    @property
+    def genomic_intervals(self):
+        """
+        Returns an (n, 3) numpy array with n intervals (left, right, type)
+        """
+        all_intervals = np.concatenate(
+            [
+                np.column_stack(
+                    (g.intervals[:, :2], np.full((g.intervals.shape[0]), i))
+                )
+                for i, g in enumerate(self.genomic_element_types)
+            ],
+            axis=0,
+        )
+        all_intervals = stdpopsim.utils.build_intervals_array(all_intervals)
+        return all_intervals
+
+    @property
+    def msp_mutation_rate_map(self):
+        """
+        breaks = [0,10,20] means [0,10) and [10,20)
+        """
+        breaks = [0]
+        rates = []
+        for (start, end, t) in self.genomic_intervals:
+            if start not in breaks:
+                breaks.append(start)
+                rates.append(self.mutation_rate)
+            breaks.append(end)
+            rates.append((1 - self.slim_fractions[t]) * self.mutation_rate)
+        if not np.isclose(breaks[-1], int(self.length)):
+            breaks.append(int(self.length))
+            rates.append(self.mutation_rate)
+        assert len(breaks) == len(rates) + 1
+        return (breaks, rates)
+
+    @property
+    def slim_mutation_rate_map(self):
+        """
+        breaks = [10,20] means [0,10] and (10,20]
+        """
+        # making sure the mut types are still concordant with the ge types
+        breaks, rates = self.msp_mutation_rate_map
+        assert breaks[0] == 0
+        slim_breaks = [b - 1 for b in breaks[1:]]
+        slim_rates = [self.mutation_rate - r for r in rates]
+        return (slim_breaks, slim_rates)
+
+    def add_mutation_types(self, mutation_types, proportions, genomic_element_type_id):
+        """
+        Adds mutation types with their respective proportions to the genomic
+        element type provided.
+        """
+        _check_mut_proportions(proportions, mutation_types)
+        if genomic_element_type_id >= len(self.genomic_element_types):
+            raise ValueError(
+                f"Genomic element type of id {genomic_element_type_id} does not exist."
+            )
+
+        def _is_index(y, X):
+            for i, e in enumerate(X):
+                if e is y:
+                    return i
+            return -1
+
+        for mut_type, prop in zip(mutation_types, proportions):
+            mid = _is_index(mut_type, self.mutation_types)
+            if mid == -1:
+                self.mutation_types.append(mut_type)
+                mid = len(self.mutation_types) - 1
+            self.genomic_element_types[
+                genomic_element_type_id
+            ].mutation_type_ids.append(mid)
+            self.genomic_element_types[genomic_element_type_id].proportions = np.append(
+                self.genomic_element_types[genomic_element_type_id].proportions, [prop]
+            )
+
+    def add_genomic_element_type(self, intervals, mutation_types, proportions):
+        if self.recombination_map is None:
+            raise ValueError(
+                "You cannot add genomic element types to a "
+                "contig that has no recombination map."
+            )
+        stdpopsim.utils.check_intervals_validity(intervals)
+        ge_type = stdpopsim.GenomicElementType(intervals=intervals)
+        self.genomic_element_types.append(ge_type)
+        getid = len(self.genomic_element_types) - 1
+        self.add_mutation_types(
+            mutation_types, proportions, genomic_element_type_id=getid
+        )
+
+    def clear_genomic_mutation_types(self):
+        self.genomic_element_types = []
+        self.mutation_types = []
+
+    def fully_neutral(self, slim_mutations=False, convert_to_substitution=True):
+        self.clear_genomic_mutation_types()
+        self.add_genomic_element_type(
+            intervals=np.array([[0, int(self.length)]]),
+            mutation_types=[
+                stdpopsim.ext.MutationType(
+                    convert_to_substitution=convert_to_substitution
+                )
+            ],
+            proportions=[1 if slim_mutations else 0],
+        )
 
     def __str__(self):
         gmap = "None" if self.genetic_map is None else self.genetic_map.id
         s = (
             "Contig(length={:.2G}, recombination_rate={:.2G}, "
-            "mutation_rate={:.2G}, genetic_map={})"
+            "mutation_rate={:.2G}, genetic_map={}, genomic_element_types={}, "
+            "mutation_types={})"
         ).format(
-            self.recombination_map.get_length(),
+            self.length,
             self.recombination_map.mean_recombination_rate,
             self.mutation_rate,
             gmap,
+            self.genomic_element_types,
+            self.mutation_types,
         )
         return s
+
+
+def _check_mut_proportions(proportions, mutation_type=None, atol=1e-15):
+    if mutation_type is not None and len(proportions) != len(mutation_type):
+        raise ValueError(
+            "The list of proportions and mutation types should be of the same length."
+        )
+    if any([not (0 <= prop <= 1) for prop in proportions]):
+        raise ValueError("Proportions must lie within the [0,1] interval.")
+    if sum(proportions) > 1 + atol:
+        raise ValueError(
+            "The sum of the proportions within genomic element types cannot"
+            " exceed 1."
+        )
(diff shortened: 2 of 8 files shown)


stdpopsim/genomes.py
```diff
@@ -138,6 +139,19 @@ class Contig:
     recombination_map = attr.ib(default=None, kw_only=True)
     mutation_rate = attr.ib(default=None, type=float, kw_only=True)
     genetic_map = attr.ib(default=None, kw_only=True)
+    # TODO: figure out if ends are inclusive or not
```
andrewkern, line 142: we should probably do half open intervals to match bed specification

petrelharp: Left-inclusive, right-exclusive, hopefully? Matching what everything is in tskit/msprime?

stdpopsim/ext/genomic_elements.py
```diff
@@ -0,0 +1,22 @@
...
+    mutation_types = attr.ib(default=[stdpopsim.ext.MutationType()], type=list)
+    start_positions = attr.ib(default=None, type=np.array)
+    end_positions = attr.ib(default=None, type=np.array)
+    slim_mutations_fraction = attr.ib(frozen=True)
+
+    @slim_mutations_fraction.default
+    def _sum_proportions(self):
+        return sum([mut_type.proportion for mut_type in self.mutation_types])
+
+    def __attrs_post_init__(self):
```
mufernando, line 19: check that all `end_pos > start_pos`

stdpopsim/ext/selection.py
```diff
@@ -16,11 +16,11 @@ class MutationType(object):
...
     # I.e. the weights will be used in the ``proportion`` parameter
     # to SLiM's :func:`initializeGenomicElementType()`.
-    weight = attr.ib(default=0, type=float)
+    proportion = attr.ib(default=0, type=float)
 
     def __attrs_post_init__(self):
-        if self.weight < 0:
-            raise ValueError("weight must be >= 0.")
+        if self.proportion < 0 or self.proportion > 1:
+            raise ValueError("proportion must lie within the [0,1] interval.")
```
grahamgower, line 23: I agree that proportion is a better name for this, to match the SLiM name. SLiM allows proprtions to be greater than 1 though. Do we want to allow that too?

stdpopsim/ext/genomic_elements.py
petrelharp: `>=` since a segment of 1bp will have start==end

petrelharp: oh wait never mind, sorry! yes, you're right, a segment of 1bp will be `[a, a+1)`.

stdpopsim/genomes.py
```diff
@@ -138,6 +139,19 @@ class Contig:
...
+    genomic_element_types = attr.ib(default=
+                                    [stdpopsim.ext.GenomicElementType(
```
mufernando, line 144: same thing about the default (use a factory func)

stdpopsim/ext/genomic_elements.py
```diff
@@ -0,0 +1,22 @@
+import attr
+
+@attr.s(kw_only=True)
+class GenomicElementType(object):
+    """
+    TODO: docstring.
+
+    See SLiM documentation for initializeGenomicElementType().
+    """
+    mutation_types = attr.ib(default=[stdpopsim.ext.MutationType()], type=list)
```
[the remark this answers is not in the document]
mufernando: ah, thanks for this! I have never ussed `attrs`. 

I think we need a list of mutation types so we can check for consistency (e.g. sum of proportions must be at most 1). Also, from here on mutation types should always be tied to a genomic element type, that is `simulate` will take a list of genomic element types instead of a list of mutation types). But let's see how this plays out later.

stdpopsim/ext/selection.py
mufernando: I think these should be strict proportions, because then we can use them to deal with the overlaying of neutral mutations. Alternatively, we could parametrize as mutation rates, but I like the idea of having one knob for overall mutation rate and knobs for proportions (of the total mutation rate) for each mutation type.

stdpopsim/genomes.py
mufernando: yeah, I need to be careful when gluing this to the annotations because genomic elements in SLiM are 0-indexed left and right inclusive.

stdpopsim/ext/genomic_elements.py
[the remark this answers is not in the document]
mufernando: I agree with you re list of indexes now! We can do the consistency checking within the contig.

andrewkern:
@PopSim-bot ILU! ❤️

stdpopsim/genomes.py
```diff
@@ -119,6 +172,13 @@ class Contig:
         <https://msprime.readthedocs.io/en/stable/api.html#msprime.RecombinationMap>`_
         for more details.
     :vartype recombination_map: msprime.simulations.RecombinationMap
+    :ivar genomic_element_types: a list of stdpopsim.GenomicElementType
+        objects. By default, the only element is a single genomic element type
+        that spans the entire contiguous region.
+    :vartype genomic_element_types: list
+    :ivar mutation_types: a list of stdpopsim.ext.MutationType object. By
+        default, the only mutation type is neutral. See
+        stdpopsim.MutationType for more details.
```
grahamgower, line 181: Probably good to note here that these are only relevant when using the SLiM engine.

stdpopsim/genomes.py
```diff
@@ -2,6 +2,12 @@
 Infrastructure for defining information about species' genomes.
 """
 import attr
+import stdpopsim
+import tskit
+import numpy as np
+import logging
```
grahamgower, line 8: Imports should usually be grouped following [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) guidelines. (standard library imports first, then third-party imports, and finally local imports). Unfortunately neither flake8 nor black warn about this.

stdpopsim/genomes.py
```diff
@@ -138,14 +198,96 @@ class Contig:
...
+        return breaks
+
+    @property
+    def slim_mutation_rate_map(self):
+        # 2D array with end positions
+        ends = np.array([g.end_positions for g in self.genomic_element_types])
+        # bool 2D (# genome element types x # breaks)
+        break_in_ends = (self.genomic_element_breaks==ends[...,None]).any(axis=1)
+        # array with getype index for each break
+        which_getype = np.apply_along_axis(_where_is_true, 0, break_in_ends)
```
grahamgower, line 265: Maybe the `_where_is_true()` function could be moved up to here as a nested function, so that it's obvious that this is the only place the function is used. Unless there's a reason not to?

grahamgower (commented):
This is looking really good @mufernando, I just left a few small comments.

Regarding the API: it probably does make sense to include the `genomic_element_types`  and `mutation_types` as properties of the `Contig` class, as you've done here. Yes, they're slim-engine specific, but conceptually I think this is a better place for them than as parameters to the simulate function.

stdpopsim/genomes.py
mufernando: is this the right way to import stdpopsim? from within the module?

grahamgower: Yep, that should be fine. (And It's the only local import here).

tests/test_slim_engine.py
```diff
@@ -433,18 +433,20 @@ class PiecewiseConstantSizeMixin(object):
     """
```
mufernando, line 433: here @andrewkern

andrewkern: got it

.travis.yml
```diff
@@ -30,6 +30,7 @@ install:
   # include it in the 'minimal' requirements.
   - conda install --yes slim
   # pyslim is not yet on conda-forge.
+  - pip install git://github.com/tskit-dev/msprime.git@1cd73ef3f498903e022ebb0b2f6cfb9fa9564ab0#egg=msprime
```
grahamgower, line 33: I think you'll need a https url here.

requirements/CI/requirements.txt
```diff
@@ -8,8 +8,8 @@ sphinx-argparse==0.2.5
 sphinx-issues==1.2.0
 sphinx_rtd_theme==0.5.0
 sphinxcontrib-programoutput==0.16
-msprime==0.7.4
 attrs==20.2.0
+-e git://github.com/tskit-dev/msprime.git@1cd73ef3f498903e022ebb0b2f6cfb9fa9564ab0#egg=msprime
```
grahamgower, line 12: Similarly, I think this needs to be `git+https://`. And is `-e` really needed here?

requirements/development.txt
```diff
@@ -10,7 +10,7 @@ sphinx-argparse
 sphinx-issues
 sphinx_rtd_theme
 sphinxcontrib-programoutput
-msprime
+-e git://github.com/tskit-dev/msprime.git@1cd73ef3f498903e022ebb0b2f6cfb9fa9564ab0#egg=msprime
```
grahamgower, line 13: `git+https://`

.travis.yml
mufernando: thanks!!

mufernando:
can someone help with these travis/appveyor errors that I'm getting with the dev msprime? @grahamgower @jeromekelleher

stdpopsim/genomes.py
```diff
@@ -150,16 +204,162 @@ class Contig:
...
+        if self.is_slim_incompatible:
+            return None
+        return [stdpopsim.ext.MutationType()]
+
+    @property
+    def slim_fractions(self):
+        if self.is_slim_incompatible:
+            return None
+        props = [sum(g.proportions) for g in self.genomic_element_types]
+        if any([p > (1.0 + 1e-15) for p in props]):
```
mufernando, line 241: factor out consistency checks throughout

stdpopsim/genomes.py
```diff
@@ -150,16 +204,162 @@ class Contig:
...
+    mutation_types = attr.ib(type=list, kw_only=True)
+
+    @property
+    def length(self):
+        if self.recombination_map is not None:
+            return self.recombination_map.get_length()
+        return None
+
+    @property
+    def is_slim_incompatible(self):
```
mufernando, line 217: stdpopsim was not built to simulate continuous stuff anyway, so just one check `has_recombination_map`.

stdpopsim/genomes.py
```diff
@@ -150,16 +204,162 @@ class Contig:
...
             self.recombination_map.mean_recombination_rate,
             self.mutation_rate,
             gmap,
         )
         return s
+
+
+def _check_mutation_types(mutation_types, genomic_element_types):
+    mut_ids = [mtid for g in genomic_element_types for mtid in g.mutation_type_ids]
+    if set(range(len(mutation_types))) != set(mut_ids):
```
mufernando, line 338: they don't have to be the same, but the mut_ids in genomic element types must be a subset of all mutation types.

stdpopsim/slim_engine.py
```diff
@@ -800,29 +782,38 @@ def matrix2str(
...
-    printsc(
-        '    initializeGenomicElementType("g1", '
-        + f"seq(1, {len(mutation_types)}), c({mut_weights}));"
-    )
-    printsc()
+            printsc(f'    m{i}.convertToSubstitution = F;')
+    # Genomic element types.
+    genomic_element_types = contig.genomic_element_types
+    for j, g in enumerate(genomic_element_types):
+        mut_props = ", ".join([str(prop) for i, prop in enumerate(g.proportions)])
```
mufernando, line 800: remove enumerate

stdpopsim/genomes.py
```diff
@@ -150,16 +204,162 @@ class Contig:
...
+            return None
+        breaks = [0]
+        rates = []
+        for (start, end, t) in self.genomic_intervals:
+            if start not in breaks:
+                breaks.append(start)
+                rates.append(self.mutation_rate)
+            breaks.append(end)
+            rates.append((1 - self.slim_fractions[t]) * self.mutation_rate)
+        if not np.isclose(breaks[-1], int(self.length)):
```
mufernando, line 288: breaks are integers so no need for that.

stdpopsim/genomes.py
```diff
@@ -150,16 +204,162 @@ class Contig:
...
+        # One neutral mutation type, simulated with msprime
+        if self.is_slim_incompatible:
+            return None
+        return [stdpopsim.ext.MutationType()]
+
+    @property
+    def slim_fractions(self):
+        if self.is_slim_incompatible:
+            return None
+        props = [sum(g.proportions) for g in self.genomic_element_types]
```
mufernando, line 240: make this a numpy array

stdpopsim/slim_engine.py
```diff
@@ -800,29 +782,38 @@ def matrix2str(
...
+        printsc(
+            f"    initializeGenomicElementType({j}, c({mut_types}), c({mut_props}));"
+        )
+        printsc(
+            f"    initializeGenomicElement({j}, c({element_starts}), c({element_ends}));"
+        )
+    # Mutation rate map.
+    breaks, rates = contig.slim_mutation_rate_map
+    mut_rates = ", ".join([str(r) for r in rates])
+    mut_breaks = ", ".join([str(int(b)) for b in breaks])
```
mufernando, line 815: remove the int

stdpopsim/genomes.py
```diff
@@ -107,12 +113,53 @@ class Chromosome:
...
+        mutation type can happen in genomic elements of this type.
+    :vartype: proportions: list
+    :ivar intervals: (n, 2) numpy array with [left, right) genomic intervals
+        that will be of this GenomicElementType. The intervals are used to
+        specify SLiM GenomicElements. Intervals should be non-overlapping and
+        in asceding order.
+    :vartype intervals: numpy.array, dtype=np.int32
+    """
+
+    mutation_type_ids = attr.ib(factory=lambda: [0], type=list)
```
mufernando, line 133: make them mutation_types not ids

stdpopsim/genomes.py
```diff
@@ -129,6 +176,13 @@ class Contig:
     :ivar exclude: If True, ``mask_intervals`` specify regions to exclude. If False,
         ``mask_intervals`` specify regions in keep.
     :vartype exclude: bool
+    :ivar genomic_element_types: a list of stdpopsim.GenomicElementType
+        objects. By default, the only element is a single genomic element type
+        that spans the entire contiguous region.
+    :vartype genomic_element_types: list
+    :ivar mutation_types: a list of stdpopsim.ext.MutationType object. By
```
mufernando, line 183: mutation_type should be a property that pulls the types out of the ge types.

stdpopsim/genomes.py
```diff
@@ -107,12 +113,53 @@ class Chromosome:
     synonyms = attr.ib(factory=list, kw_only=True)
 
 
+@attr.s(kw_only=True)
+class GenomicElementType(object):
```
mufernando, line 117: MutationTypeMap

mufernando: it's not genomic element type because it also includes slim genomic elements

stdpopsim/slim_engine.py
```diff
@@ -1229,25 +1207,25 @@ def _recap_and_rescale(
...
-                    ts,
-                    rate=rate,
-                    keep=True,
-                    random_seed=s2,
-                    end_time=ts.slim_generation,
-                )
+        # Adding neutral mutations to the SLiM part
+        breaks, rates = contig.msp_mutation_rate_map
+        rate_map = msprime.RateMap(position=breaks, rate=rates)
+        ts = pyslim.SlimTreeSequence(
```
mufernando, line 1213: add slim mutations here and below

mufernando: allow_ancestral=True

mufernando: look at the highest mutation id and then use that for new overlaid muts

grahamgower:
> can someone help with these travis/appveyor errors that I'm getting with the dev msprime?

The errors are the same on both travis and appveyor: gsl headers are not being found when trying to build msprime. I guess that for the 0.7.4 release, msprime was installed from binary wheel, but for the dev version it needs to build from sources. Both travis and appveyor are using conda environments, so you might try adding a `conda install gsl` line for both.

stdpopsim/genomes.py
```diff
@@ -150,16 +191,172 @@ class Contig:
...
+            )
+        _check_mutation_types(self.mutation_types, self.genomic_element_types)
+
+    def add_genomic_element_type(self, intervals, mutation_types, proportions):
+        if self.recombination_map is None:
+            raise ValueError(
+                "You cannot add genomic element types to a "
+                "contig that has no recombination map."
+            )
+        intervals = stdpopsim.utils.build_intervals_array(intervals)
```
andrewkern, line 301: strikes me that we should assume the intervals coming in to `add_genomic_element_type` are already proper and have been built using this utility.

mufernando: fixed! thanks.

mufernando:
I think I have this in good shape for more pair of eyes.

The idea main idea is that `GenomicElementType` objects map `MutationType`s (and their proportions of the total mutation rate) to genomic regions. Within `Contig`, you can add as many `GenomicElementType`s you want (as long as all the intervals are non-overlapping). You do not need to define "neutral" regions if you don't want, and we will just overlay those with msprime neutral mutations. Similarly, for genomic intervals in which the sum of the proportions of all the different mutation types that can happen there do not sum to 1, neutral mutations will be added with the remaining rate. You may choose to simulate neutral mutations within SLiM, for that you'd only have to define the genomic element type and a neutral mutation type with non-zero proportion.

Things I could use input:
- User API (methods within `Contig` that are used to add genomic element and mutation types) and naming conventions -- see `docs/selection_example.py` for an example.
- Tests?
- As of msprime 1.0, you can use a SLiM mutation model to add the neutral mutations. There is one kink to figure out: we need to specify a mutation type,  but I'm not sure if we should always add a new type on top of all the existing. This could be confusing because the returned tree sequence would have more types than were specified by the user. We could always require that a neutral type be defined (the first in the list) and that will be used when overlaying mutations with msprime.
- The DrawMutation and ConditionOnAlleleFrequency API does not mesh super well with what I have so far. What is most weird about these is that they can use a MutationType that is not associated to any GenomicElementTypes. Also, we should provide better user interface in `Contig` to deal with these. Should they also live within `Contig`?

@grahamgower @andrewkern @petrelharp

tests/test_slim_engine.py
```diff
@@ -1105,15 +1138,14 @@ def test_no_drawn_mutation(self):
...
-                mutation_types=self.mutation_types,
                 extended_events=extended_events,
                 dry_run=True,
             )
 
 
 @unittest.skipIf(IS_WINDOWS, "SLiM not available on windows")
 class TestChangeMutationFitness(unittest.TestCase, PiecewiseConstantSizeMixin):
-    # Testing stdpopsim.ext.ChangeMutationFitness is challenging, because
+
```
grahamgower, line 1148: Was this removed deliberately?

grahamgower:
> Things I could use input:
> 
> * User API (methods within `Contig` that are used to add genomic element and mutation types) and naming conventions -- see `docs/selection_example.py` for an example.

The examples look quite neat IMO.

> * Tests?
> * As of msprime 1.0, you can use a SLiM mutation model to add the neutral mutations. There is one kink to figure out: we need to specify a mutation type,  but I'm not sure if we should always add a new type on top of all the existing. This could be confusing because the returned tree sequence would have more types than were specified by the user. We could always require that a neutral type be defined (the first in the list) and that will be used when overlaying mutations with msprime.

Yeah, this is awkward all right. I'm not sure what to do here either.

> * The DrawMutation and ConditionOnAlleleFrequency API does not mesh super well with what I have so far. What is most weird about these is that they can use a MutationType that is not associated to any GenomicElementTypes. Also, we should provide better user interface in `Contig` to deal with these. Should they also live within `Contig`?
> 

I think you're right. We should push the drawn mutation stuff into the Contig. But probably not in the present PR.

> What is missing to work with Annotations?
> 
> * Non-overlapping annotations
> * Infrastructure within `species.py`'s `get_contig` that creates an annotated contig already?

I think we should kick that down the road. There's already a lot of really good stuf in this PR, and it would be nice to merge it in and open issues for the remaining stuff.

grahamgower:
Hey @mufernando, do you think you could rebase this against the `msprime-1.0` branch and squash? You can also "retarget" the PR against that branch using the edit button near the PR's title. I'm happy to merge it in to that branch and deal with the various things you've outlined in follow-up issues.

tests/test_slim_engine.py
mufernando: no! I'll fix!

mufernando:
thanks @grahamgower! I'll add one or two more tests to get the right coverage, and then rebase against the msprime-1.0 branch so we can get this merged.

mufernando:
hey @grahamgower, I rebased against msprime-1.0 and expanded the tests a bit. I'll open follow up issues

grahamgower:
Thanks @mufernando. I just changed the "base" to target the `msprime-1.0` branch. But it looks like there's a couple of changes to the CI files in the diff now. I'm not sure what's going on. Can you double check that you're not modifying any CI files in your rebase against the new branch?

grahamgower:
All green! Merging.

Issue #663, opened 2020-10-28
Human genetic map not working as intended


Rosemeis:
**Using HapMap genetic map is not giving correct results:**
I simulated 500 individuals of each of the four populations of the demography model, AmericanAdmixture_4B11, with and without using the provided genetic map (HapMapII_GRCh37) on chromosome 22. Checked the results by performing PCA and saw that something was completely off for the scenario of using the provided genetic map. Similar results when using different demography model.

**How to reproduce the problem:**
Used the following commands using stdpopsim (v.0.1.2):
*With genetic map*
stdpopsim HomSap 500 500 500 500 -d AmericanAdmixture_4B11 -g HapMapII_GRCh37 -c chr22 -o amr.adm.hapmap.trees

*With uniform map*
stdpopsim HomSap 500 500 500 500 -d AmericanAdmixture_4B11 -c chr22 -o amr.adm.uniform.trees

I obtained the haplotype matrices of the chromosomes using tskit in Python (ts.genotype_matrix()). I performed standard PCA directly on standardized haplotype matrices and have attached the plot for each of the scenarios.

Cheers,
Jonas

![amr adm hapmap](https://user-images.githubusercontent.com/18143109/97487286-f8653500-195c-11eb-9cf6-807a10718e1b.png)
![amr adm uniform](https://user-images.githubusercontent.com/18143109/97487306-fef3ac80-195c-11eb-85fc-7ca460133223.png)

andrewkern:
hi there-- thanks for checking out the project.

i see that you are simulating chromosome 22. on that chromosome the first ~16Mb are nonrecombining-- did you remove variation from that portion of the chromosome before performing PCA? 

the uniform map will not have such large spans of the chromosome represented by one marginal tree, so i would expect to see differences here.

Rosemeis:
No, I was simply using all the common variants for PCA (maf > 0.05). But I see the same pattern for other chromosomes, so I guess it is due to the tails of the genetic map.

andrewkern:
for the analysis portion of the stdpopsim paper we use [these mask files](https://github.com/popsim-consortium/analysis/tree/master/n_t/masks) to remove regions which we considered to be low recombination, by some heuristic rule. This could be a good way to get started.

apragsdale:
[...]

Issue #104, opened 2019-07-16
Output masking feature


jeromekelleher:
On the conference call today, we decided to implement a feature to mask the tree sequences that we simulate from genetic maps. 

1. Provide an option for the user to specify a set of genomic intervals (as a bed file, I guess?), which we "delete" (see below) from the simulated tree sequence.
2. By default, when we simulate from a genetic map we compute a simple mask based on regions of zero recombination within this map. We will derive some heuristics for this which may or may not be organism/chromosome specific.

A planned feature for the next version of stdpopsim would be to curate a set of masks and provide methods to download and cache these transparently, like we do for genetic maps.

The 'delete' (or maybe 'dice' - we're still discussing what it should be called) operation will take the tree sequence simulated under the raw recombination map, and remove all topology and site data from the masked regions. Therefore, any statistics computed on this will not be affected by the zero recombination rate sections.  Anything plotted against genome coordinates will show big gaps (which is what we want, right?). This is being developed at https://github.com/tskit-dev/tskit/pull/261

grahamgower:
Should we close this since #664 has been merged, or should we consider adding a curated set of mask files? @jeromekelleher

jeromekelleher:
Let's close

PR #664, opened 2020-10-29
Masking simulated data

apragsdale:
Inspired by #663, and remembering that we had discussed adding masking features a while back (#104), I thought I'd sketch out a possible way of including a mask for simulating data. It's by no means complete, since I wanted to get feedback about it before putting much more time into it.

Essentially, at the time of defining a `contig`, we can specify the path to a mask file in bed format, where the mask specifies intervals to keep (following the format in the 1000G downloadable strict mask). Then the mask intervals are an attribute of the contig, and after simulating, we use `ts.keep_intervals(contig.mask_intervals)` to keep only the intervals in the tree sequence given by the mask, if given.

E.g.
```
import stdpopsim
contig = species.get_contig("chr22", length_multiplier=0.1, mask="path/to/masks/GRCh38/pilot_strict_combined.allChr.mask.bed")
engine = stdpopsim.get_engine('msprime')
ts = engine.simulate(model, contig, samples)
```

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -371,22 +371,33 @@ def add_simulate_species_parser(parser, species):
                 "Available maps: "
                 f"{', '.join(choices)}. "))
 
-    if len(species.genome.chromosomes) == 1:
-        species_parser.set_defaults(chromosome=species.genome.chromosomes[0].id)
-    else:
-        # To avoid listing too much stuff out in the help, we only list
-        # the actual IDs. We make all synonyms available as choices though.
-        choices = []
-        all_choices = []
-        for chrom in species.genome.chromosomes:
-            choices.append(chrom.id)
-            all_choices.extend([chrom.id] + chrom.synonyms)
-        species_parser.add_argument(
-            "-c", "--chromosome", choices=all_choices, metavar="", default=choices[0],
-            help=(
-                f"Simulate a specific chromosome. "
-                f"Options: {', '.join(choices)}. "
-                f"Default={choices[0]}."))
+    # To avoid listing too much stuff out in the help, we only list
+    # the actual IDs. We make all synonyms available as choices though.
+    choices = []
+    all_choices = []
+    for chrom in species.genome.chromosomes:
+        choices.append(chrom.id)
+        all_choices.extend([chrom.id] + chrom.synonyms)
+    species_parser.add_argument(
+        "-c", "--chromosome", choices=all_choices, metavar="", default=None,
+        help=(
+            f"Simulate a specific chromosome. If no chromosome is given, "
+            f"simulate a generic contig of given length, specified using --length. "
+            f"Options: {', '.join(choices)}. "
+            f"Default=None."))
+
+    species_parser.add_argument(
+        "-L", "--length", default=None, type=float,
+        help="Simulate a default contig of given length."
+    )
+    species_parser.add_argument(
+        "-i", "--inclusion-mask", default=None, type=str,
+        help="Path to inclusion mask specified in bed format."
+    )
+    species_parser.add_argument(
+        "-e", "--exclusion-mask", default=None, type=str,
+        help="Path to exclusion mask specified in bed format."
+    )
     species_parser.add_argument(
         "-l", "--length-multiplier", default=1, type=float,
         help="Simulate a sequence of length l times the named chromosome's length, "
@@ -440,8 +451,13 @@ def add_simulate_species_parser(parser, species):
                 "populations")
         samples = model.get_samples(*args.samples)
         contig = species.get_contig(
-            args.chromosome, genetic_map=args.genetic_map,
-            length_multiplier=args.length_multiplier)
+            args.chromosome,
+            genetic_map=args.genetic_map,
+            length_multiplier=args.length_multiplier,
+            length=args.length,
+            inclusion_mask=args.inclusion_mask,
+            exclusion_mask=args.exclusion_mask,
+        )
         engine = stdpopsim.get_engine(args.engine)
         logger.info(
             f"Running simulation model {model.id} for {species.id} on "
--- a/stdpopsim/engines.py
+++ b/stdpopsim/engines.py
@@ -64,8 +64,13 @@ class Engine:
     """
 
     def simulate(
-            self, demographic_model=None, contig=None, samples=None, seed=None,
-            dry_run=False):
+        self,
+        demographic_model=None,
+        contig=None,
+        samples=None,
+        seed=None,
+        dry_run=False,
+    ):
         """
         Simulates the model for the specified contig and samples.
 
@@ -99,25 +104,36 @@ class _MsprimeEngine(Engine):
     id = "msprime"  #:
     description = "Msprime coalescent simulator"  #:
     citations = [
-            stdpopsim.Citation(
-                doi="https://doi.org/10.1371/journal.pcbi.1004842",
-                year="2016",
-                author="Kelleher et al.",
-                reasons={stdpopsim.CiteReason.ENGINE}),
-            ]
+        stdpopsim.Citation(
+            doi="https://doi.org/10.1371/journal.pcbi.1004842",
+            year="2016",
+            author="Kelleher et al.",
+            reasons={stdpopsim.CiteReason.ENGINE},
+        )
+    ]
     # We default to the first model in the list.
     supported_models = ["hudson", "dtwf", "smc", "smc_prime"]
-    model_citations = {"dtwf": [
-             stdpopsim.Citation(
-                 doi="https://doi.org/10.1371/journal.pgen.1008619",
-                 year="2020",
-                 author="Nelson et al.",
-                 reasons={stdpopsim.CiteReason.ENGINE}),
-             ]}
+    model_citations = {
+        "dtwf": [
+            stdpopsim.Citation(
+                doi="https://doi.org/10.1371/journal.pgen.1008619",
+                year="2020",
+                author="Nelson et al.",
+                reasons={stdpopsim.CiteReason.ENGINE},
+            )
+        ]
+    }
 
     def simulate(
-            self, demographic_model=None, contig=None, samples=None, seed=None,
-            msprime_model=None, msprime_change_model=None, dry_run=False):
+        self,
+        demographic_model=None,
+        contig=None,
+        samples=None,
+        seed=None,
+        msprime_model=None,
+        msprime_change_model=None,
+        dry_run=False,
+    ):
         """
         Simulate the demographic model using msprime.
         See :meth:`.Engine.simulate()` for definitions of parameters defined
@@ -154,15 +170,22 @@ class _MsprimeEngine(Engine):
             demographic_events.sort(key=lambda x: x.time)
 
         ts = msprime.simulate(
-                samples=samples,
-                recombination_map=contig.recombination_map,
-                mutation_rate=contig.mutation_rate,
-                population_configurations=demographic_model.population_configurations,
-                migration_matrix=demographic_model.migration_matrix,
-                demographic_events=demographic_events,
-                random_seed=seed,
-                model=msprime_model,
-                end_time=0 if dry_run else None)
+            samples=samples,
+            recombination_map=contig.recombination_map,
+            mutation_rate=contig.mutation_rate,
+            population_configurations=demographic_model.population_configurations,
+            migration_matrix=demographic_model.migration_matrix,
+            demographic_events=demographic_events,
+            random_seed=seed,
+            model=msprime_model,
+            end_time=0 if dry_run else None,
+        )
+
+        if contig.inclusion_mask is not None:
+            ts = stdpopsim.utils.mask_tree_sequence(ts, contig.inclusion_mask, False)
+        if contig.exclusion_mask is not None:
+            ts = stdpopsim.utils.mask_tree_sequence(ts, contig.exclusion_mask, True)
+
         if dry_run:
             ts = None
         return ts
--- a/stdpopsim/species.py
+++ b/stdpopsim/species.py
@@ -9,6 +9,7 @@ import attr
 import msprime
 
 import stdpopsim
+import stdpopsim.utils
 
 logger = logging.getLogger(__name__)
 
@@ -33,8 +34,8 @@ def get_species(id):
     Returns a :class:`Species` object for the specified ``id``.
 
     :param str id: The string identifier for the requested species. E.g. "HomSap".
-         A complete list of species, and their IDs, can be found in the
-         :ref:`sec_catalog`.
+        A complete list of species, and their IDs, can be found in the
+        :ref:`sec_catalog`.
     :return: An object containing the species definition.
     :rtype: :class:`Species`
     """
@@ -47,6 +48,7 @@ def get_species(id):
 
 # Convenience methods for getting all the species/genetic maps/models
 # we have defined in the catalog.
+
 
 def all_species():
     """
@@ -151,15 +153,25 @@ class Species:
         """
         return self.name.lower().replace(" ", "_")
 
-    def get_contig(self, chromosome, genetic_map=None, length_multiplier=1):
+    def get_contig(
+        self,
+        chromosome=None,
+        genetic_map=None,
+        length_multiplier=1,
+        length=None,
+        inclusion_mask=None,
+        exclusion_mask=None,
+    ):
         """
         Returns a :class:`.Contig` instance describing a section of genome that
         is to be simulated based on empirical information for a given species
         and chromosome.
 
         :param str chromosome: The ID of the chromosome to simulate.
-             A complete list of chromosome IDs for each species can be found in the
-             "Genome" subsection for the species in the :ref:`sec_catalog`.
+            A complete list of chromosome IDs for each species can be found in the
+            "Genome" subsection for the species in the :ref:`sec_catalog`.
+            If the chromosome is not given, we specify a "generic" contig with given
+            ``length``.
         :param str genetic_map: If specified, obtain recombination rate information
             from the genetic map with the specified ID. If None, simulate
             using a default uniform recombination rate on a region with the length of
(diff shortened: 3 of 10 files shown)


grahamgower:
Nice idea, thanks @apragsdale.

```
contig = species.get_contig("chr22", length_multiplier=0.1, mask="path/to/masks/GRCh38/pilot_strict_combined.allChr.mask.bed")
```

This doesn't seem quite right though. What does it mean to use a mask and a length multiplier?

jeromekelleher:
Yes, this is a great idea, thanks @apragsdale. I'm also a bit uneasy about the ``length_multiplier`` here - what's the idea?

Also might be nice to use [numpy.loadtxt](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html) rather than parsing the file by hand - it'll be quite a lot faster and it'll save us worrying about parsing corner-cases.

jeromekelleher:
@grahamgower, how do you feel about the list-of-tuples representation of intervals? Seems good to me, since we can also interpret it as an ``(n, 2)`` numpy array which will be efficient to work with. Maybe we should keep it in numpy form from the start though?

apragsdale:
The `length_multiplier` is already an option in a contig. My understanding is that it compresses the contig length and the recombination map to match the new contig length. Since I expect a mask to also be in units of base pairs for a given chromosome, we'll also need to adjust the interval positions to match the new contig length if `length_multiplier` is anything other than 1.

Using `np.loadtxt` is surely better, I can fix that. Also storing as an Nx2 numpy array instead of a list of tuples, since `tskit` can handle either of those representations, I think. I'll fix both.

apragsdale:
Ah, I see now that the `length_multiplier` cannot currently be used with a genetic map, so it only works with a constant recombination rate. However, I imagine that that is a feature we would like to have in the future, to be able to also rescale the genetic map? Otherwise, I'm not sure what the purpose of the length multiplier is at all, really.

grahamgower:
> @grahamgower, how do you feel about the list-of-tuples representation of intervals? Seems good to me, since we can also interpret it as an `(n, 2)` numpy array which will be efficient to work with. Maybe we should keep it in numpy form from the start though?

I think numy recarrays are a rather nice thing for this. The look like a list of tuples during iteration, but you get named attributes for vectors of each column.

grahamgower:
> However, I imagine that that is a feature we would like to have in the future, to be able to also rescale the genetic map? Otherwise, I'm not sure what the purpose of the length multiplier is at all, really.

Does it really make sense to rescale the genetic map? I never really understood the idea of the length multiplier either. Personally, I always want to simulate specific fixed length. What's more, I'd rather like to sample this region uniformly from the recombination map (e.g. see #401/#402). It would be super cool to only sample from regions in a given mask too!

apragsdale:
I thought the main purpose of the length multiplier was to speed up simulation in slim. But maybe it's just to get sequence of a given length. I think we could improve API a lot here: Maybe have a generic contig that each species inherits, where you specify the exact length you want, and it uses the default mutation and average recombination rates for the species. That makes more sense to me than having to calculate what the length multiplier should be for a given chromosome to get the length you want..

grahamgower:
> I thought the main purpose of the length multiplier was to speed up simulation in slim.

Well, maybe. But the length multiplier was here before the slim engine.

> I think we could improve API a lot here:

Yeah, agreed! Right now, it's possible to get a specific desired length using the multiplier, but its quite awkward (find out the chr length, divide by that, then multiply by the length you actually want). And then you always get the recombination rate that applies to that one chromosome, not the genome-wide averge. Do you want to open a new issue with some ideas?

jeromekelleher:
> I thought the main purpose of the length multiplier was to speed up simulation in slim.

The ``length_multiplier`` is there so we can do things like "simulate 10% of chromosome 1". A big motivation for it was to allow us to write unit tests, so I could say "simulate 0.1% of chr20", which will run in a fraction of a second.

Fractions are useful for me as I can never follow the actual magnitudes of genome coordinates, there's just too many digits.

apragsdale:
I see - maybe there is a good argument to both keep the `length_multiplier` and also set up a generic contig for any given species that pulls from default rates, as suggested. And I also agree that the length multiplier only makes sense with a flat map, and since the masking only really makes sense (or is only really needed) when using a genetic map, we can also set it so that the mask cannot be used with a length multiplier that doesn't equal 1.

I'll open an issue this evening to start that discussion in it's own thread.

mufernando:
This looks really good @apragsdale! I definitely think we should support masking given the weirdness with no recombination regions.

I do think, however, we should invert the way we do things: `mask_intervals` should be the intervals to mask or remove, instead of the intervals to keep. In practice, people will have in their hands BED files with the inaccessible regions of the chromosome (in stdpopsim [analysis](https://github.com/popsim-consortium/analysis/tree/master/n_t/masks), we had BED files with the regions to remove e.g.).

apragsdale:
> I do think, however, we should invert the way we do things: `mask_intervals` should be the intervals to mask or remove, instead of the intervals to keep. In practice, people will have in their hands BED files with the inaccessible regions of the chromosome (in stdpopsim [analysis](https://github.com/popsim-consortium/analysis/tree/master/n_t/masks), we had BED files with the regions to remove e.g.).

Yes, I was thinking about this when writing it up yesterday. I made this choice initially because the human 1000G masks that you can directly download and use specify regions that are _kept_ instead of discarded. If (for humans at least) those are the masks that most users will be using, and I think they are, then to me it makes more sense to stick with that format. Otherwise, users will have to go through extra steps of reformatting masks, which can be a pain or unintuitive if the user isn't expecting to have to do that.

mufernando:
We could have both `keep_intervals` and `mask_intervals` and just use the appropriate tskit function (`tskit.keep_intervals()` vs `tskit.delete_intervals`)!

stdpopsim/utils.py
```diff
@@ -53,3 +54,24 @@ def is_valid_species_common_name(common_name):
...
+    Returns intervals to keep based on a bed file specified by the mask_fpath.
+    The mask must be in bed format (columns specify chrom, left, right) and
+    additional columns are ignored. The bed file specifies regions that are
+    to be kept, instead of deleted.
+    Intervals must be non-overlapping. The chromosome name must match
+    exactly ("22" is not equivalent to "chr22").
+    """
+    lines = np.loadtxt(
+        mask_fpath,
+        dtype={"names": ("chrom", "left", "right"), "formats": ("S15", "i4", "f4")},
```
grahamgower, line 70: How about: `"formats": (object, int, int)`? If you use `object` here as the type for the chrom column, then you won't need to specify the string length, nor encode() the chrom parameter below. Also, it looks like you have different types for `left` and `right`.

grahamgower:
> We could have both `keep_intervals` and `mask_intervals` and just use the appropriate tskit function (`tskit.keep_intervals()` vs `tskit.delete_intervals`)!

I wasn't aware of the tskit functions. But this seems very sensible (and better than my other suggestions). But how about `keep_intervals` and/or `delete_intervals`, to exactly match the tskit function names? I don't have a strong opinion of wether we support just one or both, but I'd lean slightly towards only including one (and picking the one for which bed files are more commonly found).

stdpopsim/utils.py
apragsdale: Heh, you're right, that's a bit of a mess. I think I was wrestling a lot with the `loadtxt` function, finally got something to work and didn't clean it up. I'll fix.

apragsdale:
To recap: The current iteration here lets you specify `mask`, which could be either a bed file or the intervals directly, as well as `exclude` which defaults to False, which means we keep those intervals, or optionally `True` which means we remove those intervals. `exclude` can only be used if the mask is specified, and we can only use a mask when we specify a chromosome without a length multiplier. If we do not specify a chromosome, we must specify the `sequence_length`, and the chromosome has default average recombination and mutation rates.

Does this look a bit cleaner/more intuitive to folks?

stdpopsim/engines.py
```diff
@@ -12,7 +12,6 @@
 def register_engine(engine):
     """
     Registers the specified simulation engine.
-
```
grahamgower, line 15: A whole bunch of line deletions seem to have snuck into this diff. I think this is why the docs are failing to build. It's fairly normal to have a single line docs summary at the start of docstring, and then the rest of the docs are separated by a blank line (https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

grahamgower:
I'm not entirely sold on having a separate parameter here (`exclude`) that controls the mask inclusion/exclusion. Given that we'll have lots of different parameters for this function, I think it would be clearer to have a single parameter for "inclusion mask" and a single parameter for "exclusion mask" (regardless of what they get called).

This PR also introduces intervals and bed parsing, which is already something that we sorta support with the maintenance/annotations code. My preference would be for us to have a think about how we can properly unify this, to avoid fragmentation and more work later.

stdpopsim/engines.py
apragsdale: Yeah, that's strange.. `black` seemed to be unhappy with those blank lines and removed them. Might have to manually tinker with things to solve that issue.

apragsdale:
That all sounds reasonable @grahamgower. And to me, `inclusion_mask` and `exclusion_mask` are nice descriptive keywords, for reducing possibilities of user confusion. I'll try some of these changes.

mufernando:
Agree with @grahamgower's point about unifying intervals/bed parsing with the work done on annotations. @andrewkern ideas on how to do that?

apragsdale:
I don't see where exactly bed files are read and parsed there, but I think something along the lines of the `get_intervals_to_keep` function in `stdpopsim/utils.py` in this PR could be generally used as a bed-reading function (with a name change and minor tweaks).

grahamgower:
> I don't see where exactly bed files are read and parsed there

Sorry, its not explicit. `scikit-allel` is being used to load GFF3 files in the maintenance code, and those are essentially a generalisation of BED files. We will need to do interval operations on the `start` and `end` columns of the annotations, and this ought to work identically for intervals read from bed files (with possible gotchas if we don't translate 1-based coords). At the very least, we'll want to check disjointness of intervals (for both annotations and masks) in the maintenance code and/or test suite.

We likely need to reorganise the zarr cache file(s) based on the operations we choose to allow users to do on the arrays versus what we do for them in the maintenance code. If we're going to cache mask files, and maybe change recombination map formats (!), then we should design something to handle all our needs ASAP so that we don't just have a mess to undo in 12 months time.

apragsdale:
> At the very least, we'll want to check disjointness of intervals (for both annotations and masks) in the maintenance code and/or test suite.

Do you think we need to catch disjointedness of masks in stdpopsim? tskit requires intervals to be disjoint and in increasing order, so if tskit already raises an informative error when that isn't met, we'd just be repeating it here.

> We likely need to reorganise the zarr cache file(s) based on the operations we choose to allow users to do on the arrays versus what we do for them in the maintenance code. If we're going to cache mask files, and maybe change recombination map formats (!), then we should design something to handle all our needs ASAP so that we don't just have a mess to undo in 12 months time.

While the GFF3 and bed files are similar, the columns don't quite align and reading a bed file into a 2D numpy array of intervals is really simple, so I don't know how much of a mess it would end up being. This is the first I've heard mention of changing recombination map formats - what's the reasoning there?

apragsdale:
Oh crap, I keep running `black` on files I edit out of habit and it adds a huge amount of changes...

grahamgower:
> Do you think we need to catch disjointedness of masks in stdpopsim? tskit requires intervals to be disjoint and in increasing order, so if tskit already raises an informative error when that isn't met, we'd just be repeating it here.

No maybe not.

> While the GFF3 and bed files are similar, the columns don't quite align and reading a bed file into a 2D numpy array of intervals is really simple, so I don't know how much of a mess it would end up being.

Yeah, ok. Let's get it in then, and then worry later about the zarr/cache side of things.

>This is the first I've heard mention of changing recombination map formats - what's the reasoning there?

https://github.com/popsim-consortium/stdpopsim/pull/676#discussion_r515931333

mufernando:
I think we should catch non-disjointness of intervals when masking. In my PR I'm just asserting that the starts are always greater than or equal to the end of the previous interval.

```
# intervals is (n, 2) array of start and ends 0-indexed, left inclusive and right exclusive
sorted_intervals = intervals[np.argsort(intervals[:,0])]
assert sorted_intervals[1:,0] >= sorted_intervals[:-1,1]
```

mufernando:
but maybe you should leave as is and we should think about refactoring all the code that uses intervals to use a class that checks this and other things upon initialization?

grahamgower:
The tskit code is acutally checking for disjointness [here](https://github.com/tskit-dev/tskit/blob/main/python/tskit/util.py#L222). But I agree, we should have some explicit interval datatype that works with data coming from GFF3, BED, or constructed programmatically.

apragsdale:
Really don't know what codecov is complaining about here. If it's testing the bit within the slim simulate function, I'm not really sure how to reach that.

apragsdale:
Also realizing we probably want CLI support here, probably just allowing a filepath to the mask.

Otherwise, are people happy with API, keywords, behavior? Would be nice to wrap this up soon.

grahamgower:
> Really don't know what codecov is complaining about here. If it's testing the bit within the slim simulate function, I'm not really sure how to reach that.

I guess you could construct a short contig with a custom mask (rather than test a whole chromosome). That way it's possible to complete a simulation reasonably quickly.

> Also realizing we probably want CLI support here, probably just allowing a filepath to the mask.

Yeah, that would be nice. The long options should use the same names as the contig arguments, with hyphens replacing the underscores, and then we might want to come up with short options too.
Maybe `--exclusion-mask`/`-e` and `--inclusion-mask`/`-i`?

> Otherwise, are people happy with API, keywords, behavior? Would be nice to wrap this up soon.

You've used `sequence_length` as the param to `get_contig()` here. Any major objection to `length`? I think this is unambiguous (what else would length mean for a contig?). But maybe with a CLI param `--length`/`-L` this would be less clear?

grahamgower:
Hmm.. Has something gone wrong with a rebase here at some point? This code now looks like an older version of the PR (ie,, still with the exclude param, and some other things that I think you fixed after we commented).

apragsdale:
> Hmm.. Has something gone wrong with a rebase here at some point? This code now looks like an older version of the PR (ie,, still with the exclude param, and some other things that I think you fixed after we commented).

Hmm, maybe - there was a bit of a rebase mess I had yesterday, and maybe I ended up patching together an older and newer version... The API still has you specify `inclusion_mask` or `exclusion_mask`, but the contig itself stores mask intervals and whether to exclude them as a bool. That can be cleaned up, and I'll have a go at the CLI as well.

> You've used `sequence_length` as the param to `get_contig()` here. Any major objection to `length`? I think this is unambiguous (what else would length mean for a contig?). But maybe with a CLI param `--length`/`-L` this would be less clear?

I don't mind shortening it. I chose `sequence_length` to mimic `msprime` keywords, but "length" is clear enough to me.

apragsdale:
I had a go at updating the cli.. I'm clearly missing something about what the cli tests are testing, because they are failing with errors that I don't really understand. Errors along the lines of `len(stdout) != 0` in the EndToEnd tests, for example.
Or another example, 

```
FAILED tests/test_cli.py::TestCacheDir::test_download_genetic_maps - AssertionError: "Namespace(cache_dir='/some/other/cache/d[297 chars]1)\n" != ' '

Don't really know what the expected behavior should be in the `verify` function there.

File: stdpopsim/cli.py [hunk 3]
```diff
@@ -371,9 +371,7 @@ def add_simulate_species_parser(parser, species):
                 "Available maps: "
                 f"{', '.join(choices)}. "))
 
-    if len(species.genome.chromosomes) == 1:
-        species_parser.set_defaults(chromosome=species.genome.chromosomes[0].id)
-    else:
+    if len(species.genome.chromosomes) > 0:
```
grahamgower, line 374: I think this statement will always be true. Codecov shows a yellow box for this line, which I think indicates that only one condition is tested.

stdpopsim/utils.py
```diff
@@ -116,3 +117,36 @@ def untar(filename, path):
...
+
+def mask_tree_sequence(ts, mask_intervals, exclude):
+    """
+    Return a masked tree sequence, based on the mask intervals and whether
+    they are kept or excluded.
+    """
+    if exclude is False:
+        ts = ts.keep_intervals(mask_intervals)
+    elif exclude is True:
+        ts = ts.delete_intervals(mask_intervals)
```
grahamgower, line 151: Codecov has a yellow box here too. It's not clear why, but maybe this will go away if you change `elif exclude is True` to `else:`?

stdpopsim/species.py
```diff
@@ -170,42 +182,112 @@ def get_contig(self, chromosome, genetic_map=None, length_multiplier=1):
             same chromosome-specific mutation and recombination rates.
             This option cannot currently be used in conjunction with the
             ``genetic_map`` argument.
+        :param inlusion_mask: If specified, ...
+        :param exclusion_mask: If specified, ...
```
grahamgower, line 186: s/inlusion/inclusion/ and docs missing

stdpopsim/species.py
```diff
@@ -170,42 +182,112 @@ def get_contig(self, chromosome, genetic_map=None, length_multiplier=1):
...
-
-        ret = stdpopsim.Contig(
-            recombination_map=recomb_map, mutation_rate=chrom.mutation_rate,
-            genetic_map=gm)
+                raise ValueError("Cannot use length multiplier for generic contig")
+            if inclusion_mask is not None or exclusion_mask is not None:
+                raise ValueError("Cannot use mask with generic contig")
+            if length is None:
+                raise ValueError("Must specify sequence_length of generic contig")
+            non_autosomal_lower = ["x", "y", "m", "mt", "chrx", "chry", "chrm"]
```
grahamgower, line 220: Could you move this `non_autosomal_lower` up and use in the predicate for the `NonAutosomalWarning`? Not sure if you aleady did this before, but the change has been lost.

stdpopsim/species.py
```diff
@@ -170,42 +182,112 @@ def get_contig(self, chromosome, genetic_map=None, length_multiplier=1):
...
+                    L_tot += chrom_data.length
+                    r_tot += chrom_data.length * chrom_data.recombination_rate
+                    u_tot += chrom_data.length * chrom_data.mutation_rate
+            u = u_tot / L_tot
+            r = r_tot / L_tot
+            recomb_map = msprime.RecombinationMap.uniform_map(length, r)
+            ret = stdpopsim.Contig(recombination_map=recomb_map, mutation_rate=u)
+        else:
+            if length is not None:
+                raise ValueError("Cannot use sequence_length with named contig")
```
grahamgower, line 235: `sequence_length` written here, rather than `length`. Maybe do a search for this and see if it's hiding anywhere else?

grahamgower:
Looks great, @apragsdale, thanks! One last thing: can you change the CLI params to use hyphens instead of underscores (`--inclusion-mask` rather than `--inclusion_mask`), so we're consistent with the other long options? Then squash/rebase and I think this is ready for merging.

apragsdale:
> Looks great, @apragsdale, thanks! One last thing: can you change the CLI params to use hyphens instead of underscores (`--inclusion-mask` rather than `--inclusion_mask`), so we're consistent with the other long options? Then squash/rebase and I think this is ready for merging.

Sounds great! I'll get to that now and squash/rebase.

grahamgower (approved):

Issue #668, opened 2020-10-29
Provide a generic contig for each species


apragsdale:
Currently, for each species, the only contigs available are the named chromosomes, which come with their set length, average recombination rate, etc. One usage of stdpopsim that I think will be common is to simulate under a demographic model with constant recombination rate for a given length (say 1 Mb). If I want to do this currently, I'd need to do something like
```
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("chr22")
L = contig.recombination_map.get_sequence_length()
frac = 1e6 / L
contig = species.get_contig("chr22", length_multiplier=frac)
```
This contig then is the desired length, with the mutation and recombination rate for human chromosome 22. To me, this is a bit cumbersome just to get a contig of a given length.

A simple solution could be to allow a "generic" contig, where the length is directly specified:
```
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("generic", length=1e6)
```
and this generic contig has the genome-wide average mutation and recombination rate for HomSap by default.

I think this would be a simple and useful addition to the API. Thoughts?

(This came out of discussions on masking simulated data that started here: #664)

jeromekelleher:
I like it.  What if we did
```python
species = stdpopsim.get_species("HomSap")
contig = species.get_contig(length=1e6)
```

That is, we make the first arg an optional named arg.

grahamgower:
Closed by #669.

PR #669, opened 2020-10-30
Specify generic contig

apragsdale:
This is to put up a quick illustration of what I mean in #668, which allows you to specify a "generic" contig (open to suggestions for naming - "generic" is maybe too close to "genetic"). The contig is called just as you would any chromosome, but with "generic" as its name, and you must specify `sequence_length`. The mutation and recombination rates are genome wide averages across autosomal chromosomes for that species.

### DIFF
--- a/stdpopsim/species.py
+++ b/stdpopsim/species.py
@@ -48,6 +48,7 @@ def get_species(id):
 # Convenience methods for getting all the species/genetic maps/models
 # we have defined in the catalog.
 
+
 def all_species():
     """
     Returns an iterator over all species in the catalog.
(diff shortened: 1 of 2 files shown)


grahamgower (commented):
Nice, thanks @apragsdale. I'm not sure about the API though---maybe we could just make `chromosome` a keyword with `chromosome=None` as the default?

jeromekelleher (commented):
I agree with Graham - let's make the first arg keyword and use ``chromosome=None`` to denote a species generic contig.

apragsdale:
Great suggestion to use `chromosome` as a keyword arg. And I'll fix those issues that `black` introduced. Didn't see that before.

PR #676, opened 2020-11-02
Factor out download/cache code and add checksums.

### LINKED ISSUE #561: compare checksums for downloaded files (genetic maps, annotations)
The GeneticMap class should have a checksum attribute, against which downloaded files are checked. The Annotations class should have this too. Proabably sha256 is appropriate here.
https://docs.python.org/3/library/hashlib.html

### LINKED ISSUE #583: downloading genetic maps is very slow
There's something funny going on with downloading genetic maps off amazon AWS. I tried running the test suite offline, and it looked like it had hung, so I killed it. Then I manually ran `stdpopsim download-genetic-maps`. This has been going more than 30 minutes now and looks like its only half way through the Pyrho maps. Manually downloading one of the Pyrho map files with wget seems fast enough though (~20 seconds).

Anyone else seeing slow speeds (maybe intermittently)? Maybe this is a regional thing, as the AWS servers we're using are in the US? It also looks like AWS are doing adaptive rate limiting, but I'm not sure how this would affect things.

### LINKED ISSUE #599: TestGeneticMapDownload.test_download_over_cache() is our slowest test
On my system this test takes about 50 seconds, out of 70 seconds for all tests. The test currently downloads and extracts all genetic map files (twice, usually from local cache). Probably we only need to do this for a couple of genetic maps.

grahamgower:
* Introduces a CachedData class, which deduplicates functionality
   from the GeneticMap and Annotation classes.
 * Compares SHA256 checksums for downloaded files.
 * Creates utils.untar() and utils.download(), for simpler reuse and testing.
 * Replaces urlretrieve so we can set a socket timeout.

Closes #561.
Closes #599.
Closes #583.

~WIP: Still a few things left to do.~

### DIFF
--- a/stdpopsim/cache.py
+++ b/stdpopsim/cache.py
@@ -4,8 +4,14 @@ Cache handling for downloaded data.
 import pathlib
 import logging
 import os
+import urllib.parse
+import tempfile
+import warnings
 
 import appdirs
+import attr
+
+from . import utils
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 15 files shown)


stdpopsim/cache.py
```diff
@@ -43,3 +49,128 @@ def get_cache_dir():
...
+        E.g. if we're on a unix system with cache under ``/path/to/cache``, and
+        ``namespace=pathlib.PurePath("foo", "bar")``, the cached data will
+        live under ``/path/to/cache/foo/bar``.
+    :vartype namespace: :class:`pathlib.PurePath`
+    :ivar str url: The URL of the data to be cached.
+    :ivar str sha256: The SHA256 checksum of the downloaded file.
+    :ivar bool extract: True if the downloaded file is a tarball that should be
+        extracted into the cached namespace, False otherwise.
+    """
+    namespace = attr.ib(type=pathlib.PurePath)
```
jeromekelleher, line 77: I wonder if we're abusing PurePath a little bit here, since the namespace will never correspond to either a relative or absolute path in the filesystem. It's sort of a "path segment".

What are we gaining here from making the type formally a PurePath vs a string?

stdpopsim/cache.py
```diff
@@ -43,3 +49,128 @@ def get_cache_dir():
...
+    :ivar namespace: The namespace under which the cache will be stored.
+        This will be converted into a folder, by constructing folders in the
+        cache corresponding to each component of the namespace.
+        E.g. if we're on a unix system with cache under ``/path/to/cache``, and
+        ``namespace=pathlib.PurePath("foo", "bar")``, the cached data will
+        live under ``/path/to/cache/foo/bar``.
+    :vartype namespace: :class:`pathlib.PurePath`
+    :ivar str url: The URL of the data to be cached.
+    :ivar str sha256: The SHA256 checksum of the downloaded file.
+    :ivar bool extract: True if the downloaded file is a tarball that should be
```
jeromekelleher, line 74: This is very useful right now, but I wonder if we should aim to deprecate the hapmap formatted files and package the maps up as zarrs also. Hapmap is a pretty stupid format, as it turns out. Just something to keep in mind: do we want tarballs to be something we support long-term, or do we want to phase this functionality out?

tests/test_utils.py
```diff
@@ -171,3 +174,153 @@ def test_good_common_names(self):
...
+                input_filename.resolve().as_uri(),  # local file
+                "http://example.com/nonexistant",  # remote file
+            ):
+                with self.assertRaises(OSError):
+                    utils.download(url, output_filename)
+                self.assertFalse(output_filename.exists())
+
+
+class TestSha256(unittest.TestCase):
+    def test_sha256(self):
```
jeromekelleher, line 209: Be nice to test the chunking functionality here somewhere too. Could we calculate the sha256 on files of varying length using a straight ``f.read()`` and comparing to the lib version?

jeromekelleher (approved):
This is really excellent @grahamgower, some minor comments/suggestions.

stdpopsim/cache.py
grahamgower: I think this will be a relative path in the filesystem though. I wanted to recreate the use case of `namespace="genetic_maps/HomSap/HapMapII_GRCh37"`, which will be a path relative to the cache dir. But maybe we can do `pathlib.Path("something") / "foo/bar/baz"` on Windows anyway? Or maybe its better to make the ``namespace`` param a tuple?

stdpopsim/cache.py
grahamgower: Sure. I'd rather not go down that road as part the present PR though. I don't anticipate exposing this new class as a user API though, so we should be able to simplify it to only do the `extract=False` case.

stdpopsim/cache.py
grahamgower: Indeed we can use forward slashes with pathlib on windows without problem. And pathlib is so awesome we can actually check that without needing a windows system.
```python
>>> from pathlib import PureWindowsPath
>>> PureWindowsPath("C:\\")
PureWindowsPath('C:/')
>>> PureWindowsPath("C:\\") / "foo/bar"
PureWindowsPath('C:/foo/bar')
>>> str(PureWindowsPath("C:\\") / "foo/bar")
'C:\\foo\\bar'
```

stdpopsim/cache.py
jeromekelleher: SGTM

grahamgower:
I'm going to merge this without travis. It passed before rebase, and travis is painfully slow right now.

PR #686, opened 2020-11-06
transition to github actions

grahamgower:


jeromekelleher (approved):

grahamgower:
This is still quite broken, unfortunately. I'm trying to figure it out in a test repository elsewhere so as to avoid creating more noise here.

grahamgower:
I'm fairly condfident this is doing the right thing now @jeromekelleher. It would be great if you could disable appveyor and travis now, as I've removed those config files in this PR. I've left circleci alone for now, because it's doing the codecov upload. We can move that to actions too, but I'd prefer to not look at this stuff again for a while. :)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,115 @@
...
+        id: cache-slim
+        uses: actions/cache@v2
+        env:
+          # Increase this to reset the cache if the key hasn't changed.
+          SLIM_CACHE_NUM: 0
+        with:
+          path: ~/slim
+          key: ${{ runner.os }}-slim-${{ env.SLIM_TAG }}-${{ env.SLIM_CACHE_NUM }}
+
+      - name: build SLiM
```
jeromekelleher, line 94: Why are we building SLiM? Can we not use the [conda package](https://anaconda.org/conda-forge/slim)?

jeromekelleher (commented):
LGTM - one question about SLiM though

jeromekelleher:
I've disabled builds on Travis and Appveyor. I haven't deleted the configs yet, though, we'll want to loop back around to do that at some point.

jeromekelleher:
Also: hooray! Thanks for getting this sorted @grahamgower - I know how painful getting this stuff up and running is.

.github/workflows/tests.yml
grahamgower: Umm... because for some reason I didn't even think of that. Will fix!

jeromekelleher:
Woohoo! I think there's a few stray `` env.SLIM_TAG ``s lying around that don't do anything now though?

grahamgower:
Because we can't install slim from conda on windows (it doesn't exist), we can't just include `slim==3.4` in the conda requirements file. So there's a conditional statement in the `install dependencies` step that installs slim, and I've used the `env.SLIM_TAG` to provide the version string there. This version tag is also used for the cache key in the `cache conda` step (so if the slim version gets bumped, then the cache will be invalidated).

PR #709, opened 2020-12-08
set slim generation on tables

petrelharp:
See discussion in #706 . I've also updated the pyslim version.


petrelharp:
Ah-ha: is this failing because mutation times are now in SLiM and aren't being scaled also?

petrelharp:
Yes, that was not happening; I've added that scaling also.  Something to consider is that we're not scaling the slim_times in mutation metadata, but I think that is OK.

grahamgower:
Thanks @petrelharp, this goods good to me!

PR #710, opened 2020-12-08
Document map cache dir

hyanwong:
Also move `file_pattern` into a property, which makes documentation nicer.

### DIFF
--- a/stdpopsim/genetic_maps.py
+++ b/stdpopsim/genetic_maps.py
@@ -18,9 +18,6 @@ class GeneticMap:
 
     :ivar url: The URL where the packed and compressed genetic map can be obtained.
     :vartype url: str
-    :ivar file_pattern: The pattern used to map name individual chromosome to
-        files, suitable for use with Python's :meth:`str.format` method.
-    :vartype file_pattern: str
     """
 
     def __init__(
@@ -40,7 +37,7 @@ class GeneticMap:
         self.long_description = long_description
         self.url = url
         self.sha256 = sha256
-        self.file_pattern = file_pattern
+        self._file_pattern = file_pattern
         self.description = description
         self.citations = citations
 
@@ -53,7 +50,25 @@ class GeneticMap:
 
     @property
     def map_cache_dir(self):
+        """
+        The path to the directory in which the files for this map are stored.
+        Names for individual chromosome files within this directory can be obtained
+        using :attr:`file_pattern`.
+
+        :type: pathlib.Path
+        """
         return self._cache.cache_path
+
+    @property
+    def file_pattern(self):
+        """
+        The pattern used to map individual chromosome id strings to files. To obtain
+        the file names, you can use Python's :meth:`str.format` method, for example
+        ``my_map.file_pattern.format(id=my_chromosome_id_string)``.
+
+        :type: str
+        """
+        return self._file_pattern
 
     def __str__(self):
         s = "GeneticMap:\n"


stdpopsim/genetic_maps.py
```diff
@@ -53,8 +50,25 @@ def __init__(
 
     @property
     def map_cache_dir(self):
+        """
+        The path to the directory in which the files for this map are stored.
+        Names for individual chromosome files within this directory can be obtained
+        using :attr:`file_pattern`.
+
+        :type: pathlib.PosixPath
```
grahamgower, line 58: The type here actually depends on the OS. On Unix it's `PosixPath`, but on Windows it will presumably be `WindowsPath`. So probably map_cache_dir's type should be listed as `pathlib.Path` ([the super type of both](https://docs.python.org/3/library/pathlib.html#concrete-paths)).

stdpopsim/genetic_maps.py
```diff
@@ -53,8 +50,25 @@ def __init__(
...
+        """
         return self._cache.cache_path
 
+    @property
+    def file_pattern(self):
+        """
+        The pattern used to map name individual chromosome to files, suitable for use
```
grahamgower, line 65: This sounds a bit clunky. Maybe just "the pattern used to map indiv..."? Also, as long as you're documenting this, would you also be able to mention the `id` variable? I'm not sure of the correct nomenclature here, but maybe @jeromekelleher knows?
```python
>>> gmap.file_pattern
'genetic_map_GRCh37_{id}.txt'
>>> gmap.file_pattern.format("40")
KeyError: 'id'
>>> gmap.file_pattern.format(id="40")
'genetic_map_GRCh37_40.txt'
```

grahamgower (changes requested):
Thanks @hyanwong! I've left a couple of comments below.

stdpopsim/genetic_maps.py
hyanwong: Sure. I just copied it from the original docs in the class header without thinking too hard!

hyanwong: And you are right - the grammar's all wrong. I hadn't even read it properly!

hyanwong:
Tidied up the docstring and gave an example: `my_map.file_pattern.format(id=my_chromosome_id_string)`. I also squashed all the commits together: hope that's OK.

grahamgower:
Thanks @hyanwong!

stdpopsim/genetic_maps.py
```diff
@@ -53,8 +50,26 @@ def __init__(
...
+        The path to the directory in which the files for this map are stored.
+        Names for individual chromosome files within this directory can be obtained
+        using :attr:`file_pattern`.
+
+        :type: pathlib.Path
+        """
         return self._cache.cache_path
 
+    @property
+    def file_pattern(self):
```
jeromekelleher, line 63: I don't really see the logic for adding a property here tbh - how is this different to putting the same docstring into the ivar section above? What's special about this property as compared to all the others?

hyanwong: I find the docs clearer if it's a property, and if it's meant to be read-only, then using a getter but not a setter seems a reasonable thing to do. If you want to mess by assigning to `self._file_pattern` then the underscore warns you that you are on your own.

hyanwong: I guess you could make the whole thing immutable instead, though.

jeromekelleher: Right, but all of the properties are immutable in the same way so now it's inconsistent and it looks like this property is special. This is Python: we should be concentrating on enabling people to write good code, not trying to prevent them writing bad code. I really don't see the point of adding a bunch of boilerplate for this, it just makes the code less readable. But it should certainly be consistent if we are going to do it.

hyanwong: True, although I didn't know if there were any mutable properties that a map should have? (I'm not familiar with the general usage). I *was* wondering whether to move the only other documented attribute (url) to a property too, for consistency at least within that class, but I thought it better to do one thing at a time. IMO it's more confusing to have some attributes listed using `:ivar:` at the top, and others (that have getters) further down the documentation. To the user, I don't think the distinction of whether they are "properties" or simple attributes should really matter. They are all obtained using `obj.attribute`. It seems clearer to *either* list them all in the definition of the class *or* in each getter. I went for the latter (but didn't finish it off by changing `url`).

hyanwong: There's also the fact that if you document it as a property, you can reference that specific property in the documentation, as I have done from within `map_data_dir`. That seems pretty useful, TBH.

jeromekelleher: That's all true, but if you're going to make these changes please be consistent and make the changes for *all* the attributes in the class. 

This isn't intended to be an externally documented interface anyway - the documentation was just put there for developers who are *adding* genetic maps, not for end users (which is why map_cache_dir wasn't documented).

hyanwong: Happy to change `url` too, which is the other documented property/attribute. Behind the scenes we could also shift the others to getters, although that seems heavyweight if they aren't even documented.

I chatted to @grahamgower on Slack about documenting `map_cache_dir`, and he thought that it was no worse than documenting `file_pattern`. Generally, I would like to know where a library is sticking files it's downloaded behind my back - I don't think we should hide this from the user in any way. I think the rationale behind documenting `file_pattern` is a bit less obvious, but (a) the documentation was there already anyway and (b) I did find it useful, as it happens, and it helped avoid a code bodge.

PR #713, opened 2020-12-09
Remove genetic_map from public API.

jeromekelleher:
The GeneticMap class is definitely not suitable for external code to use and we need to be able to change it going forward. It's not necessary for the documented workflow in the Python API (i.e., using ``species.get_contig()``).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -29,9 +29,6 @@ accessed through the main entry point, :func:`.get_species`.
 .. autoclass:: stdpopsim.Contig()
     :members:
 
-.. autoclass:: stdpopsim.GeneticMap()
-    :members:
-
 .. autoclass:: stdpopsim.Citation()
     :members:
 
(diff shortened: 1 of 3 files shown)


grahamgower:
Although I agree this should probably *not* be part of the API, the `GeneticMap` class is already part of the `0.1.2` documentation. Is there not already an implied API stability contract here? https://stdpopsim.readthedocs.io/en/stable/api.html#stdpopsim.GeneticMap

jeromekelleher:
I think it's early enough now that we can pull these out of the public API @grahamgower, particularly since it's fairly removed from the typical workflow. It's a fair point, but I think it's the lesser of two evils now to pull it out. Hopefully no-one else has been using this.

grahamgower (approved):
Okay.

grahamgower:
Should we go ahead and merge this @jeromekelleher? I think it probably is the right thing to do.

jeromekelleher:
Yep, let's do it. Just rebased, ready to go after CI finishes.

Issue #719, opened 2020-12-15
deprecation process for old genetic maps/ids


grahamgower:
We currently have a genetic map for HomSap with id `HapMapII_GRCh37`. But we've now moved to using `GRCh38` coordinates for chromosome lengths. And actually, these older HapMap genetic maps now extend beyond the new chromosome lengths. We don't want to use mismatched coordinate spaces anyway, so the HapMap maps have been lifted over to h38 (but still needs to be uploaded to AWS I think?). So, presumably we want to introduce a new genetic map with id `HapMapII_GRCh38`. We then either leave the old id present and doing the wrong thing (#701), or we deprecate/remove the old h37 ID (which is an API-breaking change). And then if folks really want to match h37-mapped data, they would need to use stdpopsim 0.1.2 (probably already necessary, without further changes in stdpopsim HEAD). This is getting a bit messy. Thoughs?

jeromekelleher:
I think the main principles should be to avoid breaking people's code, where possible, and to facilitate reproducibility. It should always be possible to ``pip install`` a given version of stdpopsim, run some code, and get the same answer as you would have at the time of release. This means, IMO, that we need to store these old maps forever. 

I agree we need a new genetic map with id HapMapII_GRCh38. I guess the right thing to do in the code is to put in a FutureWarning that HapMapII_GRCh37 will be removed from the API at some point (in, say, a year). Seem reasonable?

hyanwong:
Where are you getting the build 38 maps from @grahamgower ?

jeromekelleher:
We lifted them over: see code [here](https://github.com/popsim-consortium/stdpopsim/tree/master/maintenance) and #691

hyanwong:
Thanks @jeromekelleher . I guess some documentation describing this would be helpful (it took a while to find older lifted over files for elsewhere: it would be nice to make this easy for others to find and use). I'm happy to help here.

petrelharp:
Possibly @silastittes

Issue #701, opened 2020-11-25
Incorrect chromosome length


noscode:
Hi,

**Short description of the problem:**
I tried to simulate and get vcf file for Homo Sapiens chromosome 1. And even the length of chromosome 1 in documentation is [248956422](https://stdpopsim.readthedocs.io/en/latest/catalog.html#sec_catalog_HomSap_genome) there were SNP's in vcf file  with greater positions (e. g. the last SNP was on position 249250516).

I have checked and it looks like the length of chromosome is taken from recombination map and it is equal to 249250621.

I have not find information about it, have I missed it?

**How to reproduce the problem:**
```python
import stdpopsim
species = stdpopsim.get_species("HomSap")
contig = species.get_contig("chr1", length_multiplier=1)
print(contig)
print(contig.recombination_map.get_length())  # 249250621
model = stdpopsim.IsolationWithMigration(10000, 5000, 5000, 4000, 0, 0)
samples = model.get_samples(10, 10)
engine = stdpopsim.get_default_engine()
ts = engine.simulate(model, contig, samples)
with open("stdpopsim_out.vcf", "w") as vcf_file:
    ts.write_vcf(vcf_file, contig_id='chr1', ploidy=2)
```

**Suggested fixes:**
If it is correct behavior then it will be good to have some note. If not then I am not sure where is the problem: in recombination map or in simulation which takes length from map. I have not checked other chromosomes and species.

jeromekelleher:
Well spotted @noscode... I suspect I know what's wrong. Can you print out the last few site positions in the tree sequence before conversion to VCF please (.i.e., ``ts.tables.sites.position[-5:]``)?

noscode:
Yes, sure:

```
[2.49228583e+08 2.49228865e+08 2.49235466e+08 2.49245907e+08
 2.49250034e+08]
```

jeromekelleher:
Huh, that's weird, looks like the actual coordinates are spilling over. I thought this might be something to do with the VCF conversion. I'm going to mark this as a bug.

noscode:
I have looked to the code and found that length is taken from the recombination map, but I did not do further digging for the reason why recombination map has wrong length.

grahamgower:
Ah! Well the recombination maps usually stop before the end of the chromosome because recombination rates cannot be inferred there. So we should definitely be taking the contig length from the genome details for the species.

grahamgower:
[...]

PR #720, opened 2020-12-17
Check the genetic map length matches the chromsome length.

grahamgower:
* If the genetic map is shorter, we extend it to the end of the chromosome
  with a zero-rate region. This is the nominal case.
* If the genetic map is longer than the chromsome, we emit a warning and
  use the genetic map length. This can occur when there is a mismatch
  between the reference coordinates used for the genetic map and the
  chromosome. By using the genetic map length, we maintain some
  semblance of backwards compatibility.

See #701 and #719.

### DIFF
--- a/stdpopsim/genetic_maps.py
+++ b/stdpopsim/genetic_maps.py
@@ -100,14 +100,25 @@ class GeneticMap:
         # needs to be redownloaded.
         map_file = self.map_cache_dir / self.file_pattern.format(id=chrom.id)
         if map_file.exists():
-            ret = msprime.RecombinationMap.read_hapmap(str(map_file))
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM

hyanwong:
https://github.com/tskit-dev/msprime/pull/1346 implements a "sequence_length" parameter in the new `read_hapmap` code, so we could probably use that instead of the code here. Not that this also does the right thing of not reporting the inserted 0 length region in the `mean_rate` value, but it *doesn't* truncate the map if the map is longer than the sequence_length. We might want to do the latter, with a warning, to match the behaviour here.

PR #726, opened 2020-12-30
add 4pop HomSap ooa model from Jouganous et al 2017

rwaples:
I have attempted to implement the 4 population model presented in Jouganous et al 2017. 

This is my first time contributing to stdpopsim.

### DIFF
--- a/stdpopsim/catalog/HomSap/__init__.py
+++ b/stdpopsim/catalog/HomSap/__init__.py
@@ -1650,3 +1650,164 @@ def _AJ():
 
 
 _species.add_demographic_model(_AJ())
+
+
+def _ooa_4pop():
+    id = "OutOfAfrica_4J17"
+    description = "4 population out of Africa"
+    long_description = """
+        Demographic model for a four population out-of-Africa history,
+        taken from Jouganous et al. (2017). Parameter values were taken
+        from table 4 in the main text. This model was fit based on joint
+        allele frequecy spectrum (AFS) data from 1000 Genomes exomes
+        from the YRI, CEU, CHB, and JPT poulation samples. The demography
+        follows the previous three-populations out-of-Africa models
+        with an additional population split in Asia leading to the
+        Japanese (JPT) population. Parameter values were estimated with
+        the program Moments assuming a mutation rate of 1.44e-8 and a
+        generation time of 29 years.
+    """
+    populations = [
+        _yri_population,
+        _ceu_population,
+        _chb_population,
+        stdpopsim.Population(
+            id="JPT",
+            description="1000 Genomes JPT (Japanese in Tokyo, Japan)",
+            sampling_time=0,
+        ),
+    ]
+    citations = [
+        stdpopsim.Citation(
+            author="Jouganous et al.",
+            year="2017",
+            doi="https://doi.org/10.1534/genetics.117.200493",
+            reasons={stdpopsim.CiteReason.DEM_MODEL},
+        )
+    ]
+
+    generation_time = 29
+
+    # Parameter values from Table 4 (in bold)
+
+    # Population sizes
+    N_A = 11293  # ancestral YRI population size
+    N_AF = 23721  # YRI population size
+    N_B = 2831  # B population size (out-of-Africa)
+    N_EU0 = 2512  # CEU population size (intial)
+    N_AS0 = 1019  # Asian population size (inital)
+    N_JP0 = 4384  # Japanese population size (inital)
+
+    # growth rates
+    r_EU = 0.0016  # European growth rate
+    r_JP = 0.0129  # Japanese growth rate
+    r_AS = 0.0026  # Asian growth rate
+
+    # migration rates
+    m_AF_B = 16.8e-5  # YRI <--> B (out-of-Africa)
+    m_AF_EU = 1.14e-5  # YRI <--> CEU
+    m_AF_AS = 0.56e-5  # YRI <--> CHB
+    m_EU_AS = 4.75e-5  # European <--> CHB
+    m_CH_JP = 3.3e-5  # CHBHan <--> JPT
+
+    # epochs (specified in years, converted to generations)
+    T_AF = int(357000 / generation_time)  # time ancestral to African
+    T_B = int(119000 / generation_time)  # time of Out of Africa split
+    T_EU_AS = int(46000 / generation_time)  # time of Europe/Asia split
+    T_CH_JP = int(9000 / generation_time)  # time of Han/Japanese split
+
+    # also
+    r_AF = 0  # YRI and B growth rate
+
+    # A and YRI
+    pop0 = msprime.PopulationConfiguration(
+        initial_size=N_AF * math.exp(r_AF * T_AF),
+        growth_rate=r_AF,
+        metadata=populations[0].asdict(),
+    )
+
+    # B and CEU
+    pop1 = msprime.PopulationConfiguration(
+        initial_size=N_EU0 * math.exp(r_EU * T_EU_AS),
+        growth_rate=r_EU,
+        metadata=populations[1].asdict(),
+    )
+
+    # Asian and CHB
+    pop2 = msprime.PopulationConfiguration(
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Hi @rwaples, welcome to popsim! The model looks good to me, thanks very much for contributing. I think we should be able to merge this one pretty quickly, and then hopefully find someone to do the QC model. However, @grahamgower is more up-to-date on the protocols for adding new models and we should probably wait for his :+1: (I think he's back next week)

rwaples:
Thanks, sounds good! 

In regards to the model, I wasn't quite sure how to deal with population sizes prior to when the populations should exist, so I set them to zero, but looking at other model I see that this isn't always done.

jeromekelleher:
> In regards to the model, I wasn't quite sure how to deal with population sizes prior to when the populations should exist, so I set them to zero, but looking at other model I see that this isn't always done.

It's optional - once there's no lineages in a particular population, it doesn't really matter what its size is.

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -1650,3 +1650,175 @@ def _AJ():
...
+
+    # JPT splits from CHB
+    JPT_event = [
+        # seg migration rate JPT <--> CHB to zero
+        msprime.MigrationRateChange(time=T_CH_JP, rate=0.0, matrix_index=(2, 3)),
+        msprime.MigrationRateChange(time=T_CH_JP, rate=0.0, matrix_index=(3, 2)),
+        # JPT splits from CHB
+        msprime.MassMigration(time=T_CH_JP, source=3, destination=2, proportion=1.0),
+        # set growth rate in JPT to zero
+        msprime.PopulationParametersChange(
```
apragsdale, line 1769: This parameters change shouldn't be necessary, as no lineages remain in the JPT population.

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -1650,3 +1650,175 @@ def _AJ():
...
+
+    # CEU and CHB merge into B
+    AS_event = [
+        # set migration rates to zero
+        msprime.MigrationRateChange(time=T_EU_AS, rate=0.0),
+        # CHB splits from CEU
+        msprime.MassMigration(time=T_EU_AS, source=2, destination=1, proportion=1.0),
+        # change size of B to N_B
+        # set growth rates in B and CHB to zero,
+        msprime.PopulationParametersChange(
```
apragsdale, line 1782: Here, we probably don't want to be adding 0.0001 to the time. The demographic events in msprime are applied in the order that they are specified, so keeping the events in the current order and setting `time=T_EU_AS` should do it.

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -1650,3 +1650,175 @@ def _AJ():
...
+            time=T_EU_AS + 0.0001, initial_size=N_B, growth_rate=0.0, population_id=1
+        ),
+        msprime.PopulationParametersChange(
```
apragsdale, line 1785: Also not needed - doesn't change the behavior of the model, but might end up causing the strict equality tests in QC to fail.

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -1650,3 +1650,175 @@ def _AJ():
...
+            time=T_EU_AS + 0.0001, growth_rate=0.0, population_id=2
+        ),
+        # set up migration between YRI and B
+        msprime.MigrationRateChange(
```
apragsdale, line 1789: Here, I think we also want to avoid adding the 0.0001 buffer to the time of the migration rate changes.

stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -1650,3 +1650,175 @@ def _AJ():
...
+
+    # Out-of-Africa event
+    EU_event = [
+        # B splits from YRI
+        msprime.MigrationRateChange(time=T_B, rate=0.0),
+        msprime.MassMigration(time=T_B, source=1, destination=0, proportion=1.0),
+    ]
+
+    # Change size of YRI/African population
+    AF_event = [
```
apragsdale, line 1805: I think this should have `time=T_AF` instead of `T_B`.

apragsdale (changes requested):
Thanks for the contribution, @rwaples! This looks great. I left a few comments here. Like @jeromekelleher said, resetting parameters of populations with no lineages doesn't change the behavior, but might end up causing confusion during QC. Also, adding the timing buffer of 0.0001 to some events shouldn't be needed, since events are applied in the order (backwards in time) that they are specified (unless this causes some issue with the slim engine?).

Thanks again, and welcome to stdpopsim!!

stdpopsim/catalog/HomSap/__init__.py
rwaples: Good catch!

stdpopsim/catalog/HomSap/__init__.py
rwaples: I will remove this and the other 0.0001 buffers.  It is nice they are not necessary.

stdpopsim/catalog/HomSap/__init__.py
rwaples: Yes, that makes sense.

stdpopsim/catalog/HomSap/__init__.py
rwaples: Yes, that makes sense.  I was struggling a bit how to deal with pops before they should exist.

rwaples:
Thanks for the helpful comments, you folks have really made it easy to contribute here. As you both noticed I did have a bit of confusion in how to deal with populations before they really exist.  

I have now removed the 0.0001 time buffers and also removed the demographic events that set pop sizes to zero when no lineages should exist in the population.  I will resubmit the pull request soon. 

I do plan to use this model (and others from stdpopsim) with the SLiM engine, via slim_script=True and with a bit of  subsequent editing of the SLiM script file.  I would be happy to help/test with that process if I can.

grahamgower (approved):
Thanks @rwaples, this looks excellent! Would you be able to squash your commits into a single commit following the dev docs. I think this is ready for merging after that.

rwaples:
Great.  I tried to squash into a single commit

grahamgower:
Merged. Thanks again @rwaples! And please do let me know if you have any issues with modifying/using the script output by the SLiM engine (I trust you've seen the [`recap_and_rescale()`](https://stdpopsim.readthedocs.io/en/latest/api.html#stdpopsim.slim_engine._SLiMEngine.recap_and_rescale) function?).

rwaples:
Thanks!
I have seen recap_and_rescale(), but I didn't end up using that as I am pretty familiar with simulate() and didn't do any rescaling, nice to have though.  I'm just using msprime.simulate(from_ts=).  

The only real issue I have found with stdpopsim -> SLiM so far is that an msprime.CensusEvent() within the demographic events does not seem to have any impact when running with the SLiM engine.  I have ended up manually editing the SLiM code to add "treeSeqRememberIndividuals".  I can raise an issue about this if it would help., but I also understand if census events are not really meant to be supported. 

FYI, this same paper has a related 5-pop model as well, so I will implement that once I get a chance.

grahamgower:
> The only real issue I have found with stdpopsim -> SLiM so far is that an msprime.CensusEvent() within the demographic events does not seem to have any impact when running with the SLiM engine. I have ended up manually editing the SLiM code to add "treeSeqRememberIndividuals". I can raise an issue about this if it would help., but I also understand if census events are not really meant to be supported.

I don't see any reason why we couldn't support `msprime.CensusEvent()`, so feel free to open an issue for that.


> FYI, this same paper has a related 5-pop model as well, so I will implement that once I get a chance.

:+1:

rwaples:
Quick question - There is an error in the description of this model (docs/parameter_tables/HomSap/OutOfAfrica_4J17.csv) (as noted in #757).  I can fix this, but not sure how to proceed - should I do this with a new pull request?

petrelharp:
Yes - a PR is the way to go. Thanks!

PR #731, opened 2021-01-11
NewWorldSettlement_4G09

David-Peede:
New demographic model for settling the new world as outlined in Gutenkunst et al. 2009.

### DIFF
--- a/stdpopsim/catalog/HomSap/__init__.py
+++ b/stdpopsim/catalog/HomSap/__init__.py
@@ -1650,3 +1650,129 @@ def _AJ():
 
 
 _species.add_demographic_model(_AJ())
+
+
+def _nws_4():
+    id = "NewWorldSettlement_4G09"
+    description = "Three population out-of-Africa and the settlment of the New World"
+    long_description = """
+        The three population Out-of-Africa model followed by a fourth population
+        settling into the New World from Gutenkunst et al (2009).
+        The fourth population splits from the East Asian lineage with no
+        recurrent migration and experiences admixture with the European lineage
+        at time 0.
+        All model parameters are the maximum likelihood values of the various
+        parameters given in Table 2 of Gutenkunst et al except for all
+        Africa-related parameters which are given in Table 1 of
+        Gutenkunst et al.
+        A schematic of this model is outlined in Figure 3B of Gutenkunst et al.
+    """
+    populations = [
+        stdpopsim.Population(
+            id="YRI", description="Environmental Genomes YRI (West Africans)"),
+        stdpopsim.Population(
+            id="CEU", description="Environmental Genomes CEU (Europeans)"),
+        stdpopsim.Population(
+            id="CHB", description="Environmental Genomes CHB (East Asians)"),
+        stdpopsim.Population(
+            id="MXL", description="Environmental Genomes MXL (Mexican-Americans)"),
+    ]
+
+    citations = [stdpopsim.Citation(
+        author="Gutenkunst et al.",
+        year=2009,
+        doi="https://doi.org/10.1371/journal.pgen.1000695",
+        reasons={stdpopsim.CiteReason.DEM_MODEL})
+    ]
+
+    generation_time = 25
+
+    # First we set out the maximum likelihood values of the various parameters
+    # all Africa-related parameters are given in Table 1 and all other
+    # parameters are given in Table 2.
+    N_A = 7300
+    N_B = 2100
+    N_AF = 12300
+    N_EU0 = 1500
+    N_AS0 = 590
+    N_MX0 = 800
+    # Times are provided in years, so we convert into generations.
+    T_AF = 220e3 / generation_time
+    T_B = 140e3 / generation_time
+    T_EU_AS = 26.4e3 / generation_time
+    T_MX = 21.6e3 / generation_time
+    T_ADMIX = 0
+    # We need to work out the starting (diploid) population sizes based on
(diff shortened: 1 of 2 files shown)


stdpopsim/catalog/HomSap/__init__.py
```diff
@@ -1552,3 +1552,129 @@ def _AJ():
...
+    migration_matrix=[
+        [      0, m_AF_EU, m_AF_AS, 0],
+        [m_AF_EU,       0, m_EU_AS, 0],
+        [m_AF_AS, m_EU_AS,       0, 0],
+        [      0,       0,       0, 0]
+        ]
+    demographic_events=[
+        # CEU and MXL admix with proportion f_MX at T_ADMIX
+        msprime.MassMigration(
+            time=T_ADMIX, source=3, destination=1, proportion=f_MX),
```
grahamgower, line 1644: Having `T_ADMIX = 0` here is going to produce undesireable behaviour with the SLiM engine, because it will be impossible to sample the MX population after the admixture event. Maybe we should bodge it so that T_ADMIX is one generation in the past? Thoughts @apragsdale, @jeromekelleher?

grahamgower:
Thanks @David-Peede! Would you be able to run the `pre-commit` checks on your workstation to fix some minor formatting issues (https://stdpopsim.readthedocs.io/en/latest/development.html#pre-commit-checks)? In #732, you mentioned that the admixture time for `MXL` is at time=0... I'm not sure what to do about this, because it will certainly produce incorrect behaviour with the current SLiM engine.

stdpopsim/catalog/HomSap/__init__.py
David-Peede: In ms they noted that to model admixture in the MXL population they generated a 5th population that exists for 0 time.

David-Peede:
I am having some issues. I fixed the formatting issues, but am having some issues with the pre-commit hooks. I keep on getting the error .pre-commit-config.yaml does not exist despite running pre-commit install and pre-commit clean. Any advice?

grahamgower:
Is your working tree up to date with the stdpopsim repository? I.e., maybe you need to do a `git fetch` and rebase onto the `master` branch?

David-Peede:
I seem to have messed something up... I believe I am upstream, but am still running into the same issue. This is what it should look like correct?
Screen Shot 2021-01-11 at 9 17 34 AM

stdpopsim/catalog/HomSap/__init__.py
jeromekelleher: I agree having T_ADMIX exactly equal to zero is odd and will lead to weirdness. I wonder why have a MassMigration at time zero at all - wouldn't it be simpler to set up the initial conditions so that this MassMigration had already happened?

David-Peede: Right, I was initially trying to explicitly follow what the authors did, but I see the hiccup now. If I were to preset the admixture event from CEU to MXL wouldn't I still need to specify a time at which this occurred?

apragsdale: That is quite strange, and I didn't realize there was that quirk to this model. My first reaction is that we might not want to support this model in that case, because (1) it has this difficulty with parameterization that makes it difficult to faithfully implement, and (2) it's not very realistic, since admixture was more like ~10 generations ago. Probably doesn't matter much for the SFS, but it won't be correct for IBD or LD. People who run this model to test recent admixture methods could be really led astray by it. What do you all think?

grahamgower: Thanks @apragsdale, those are very good points. We certainly don't want to include a model and then recommend against using it (for whatever reason). Maybe @RyanGutenkunst has an opinion here too?

RyanGutenkunst: Aaron's correct: the model is this way because a difference of ~0.005 (in genetics) units of admixture timing doesn't matter to the SFS. And in dadi it was more computationally convenient to put it at 0. Since we have well-documented history here, we could easily push this event to 5-10 generations ago. If that's not kosher to modify the model, we can always leave it out. (It was inferred in a strange way anyways, integrating out the African pop.)

David-Peede: @grahamgower it seems like it might be best to ditch this model because of all the quirks? I am happy to adjust the admixture time if you think it is worth keeping.

grahamgower: How were you planning on using the model @David-Peede?

David-Peede: I was personally going to use it so simulate Chr6 under different evolutionary histories in SLiM and see if I could recapitulate the patterns of variation we observe in the empirical data from the 1000 Genomes Project. But it seems like this might not be the best model for that (or be useful to others), perhaps AmericanAdmixture_4B11 might better fit my needs.

RyanGutenkunst: Yes, that's likely a better bet.

David-Peede: I am going to go ahead and close this PR if that's ok with you @grahamgower

grahamgower: Yep, that's fine @David-Peede. It can always be reopened and picked up again in the future if we change our minds.

Issue #735, opened 2021-01-20
kwargs for `simulate` functions


awohns:
I recently noticed that there's no option to set `record_migrations=True` when using `simulate()`.

Since this might be an issue if people want to use other options such as `record_full_arg `, I propose we use **kwargs with engine.simulate(), which would then be passed to the slim and msprime engines.

I'm happy to implement this if others agree it's a good idea!

jeromekelleher:
Seems OK to me, once users are aware that extra kwargs might not work across different engines.

@grahamgower?

grahamgower:
Closed by #736.

PR #736, opened 2021-01-21
msprime kwargs

awohns:
As mentioned in #735, I noticed that there's no option to pass further arguments to `msprime.simulate`. I've added `kwargs` to the relevant `simulate` function and added a test that the information is passed through to msprime successfully.

I aslo thought it would be sensible to require the `demographic_model`, `contig`, and `samples` parameters in `simulate` and make all the other arguments keyword-only. When I first started using `stdpopsim` I found the error messages a bit opaque when any of the first three weren't specified and defaulted to `None`. Happy to take this part out of the PR though others don't agree!

### DIFF
--- a/tests/test_engines.py
+++ b/tests/test_engines.py
@@ -4,6 +4,7 @@ Tests for simulation engine infrastructure.
 import unittest
 
 import stdpopsim
+import msprime
 
 
 class TestEngineAPI(unittest.TestCase):
@@ -49,7 +50,8 @@ class TestEngineAPI(unittest.TestCase):
 
     def test_abstract_base_class(self):
         e = stdpopsim.Engine()
-        self.assertRaises(NotImplementedError, e.simulate)
+        with self.assertRaises(NotImplementedError):
+            e.simulate(None, None, None)
         self.assertRaises(NotImplementedError, e.get_version)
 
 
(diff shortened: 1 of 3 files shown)


awohns:
Unsure why the "Build and deploy docs" test failed, would appreciate guidance on that!

tests/test_engines.py
```diff
@@ -68,3 +69,20 @@ def test_simulate_nonexistent_param(self):
...
+            with self.assertRaises(TypeError):
+                engine.simulate(model, contig)
+
+    def test_msprime_kwargs(self):
+        species = stdpopsim.get_species("HomSap")
+        model = species.get_demographic_model("AshkSub_7G19")
+        contig = species.get_contig("chr22", length_multiplier=0.01)
+        samples = model.get_samples(10)
+        engine = stdpopsim.get_engine("msprime")
+        replicates = engine.simulate(model, contig, samples, num_replicates=2)
```
jeromekelleher, line 87: ``num_replicates`` might not be a great example here as this could get implemented in the main API some day. Something that's msprime specific like ``record_full_arg`` is probably better, and you can test for the existance of recomb nodes.

Good to set the seed as well to make this deterministing.

jeromekelleher (approved):
LGTM. Let's I think the mandatory args thing is a good idea.

Let's see what @grahamgower thinks.

grahamgower:
Sorry for not seeing this right away, I'm party AFK for a few days.

The mandatory args, and support for `random_seed` as well as `seed` looks good (although the latter may need a docstring). But I'm luke warm about supporting arbitrary kwargs in this way. I like that python makes this easy, so the code is simple and maintainable. But on the other hand, I'm wary about kwargs passing in this way because it can lead to labyrintine documentation like in matplotlib (kwargs get passed through multiple functions/classes and it's difficult for users to quickly identify the documentation for some arg seen in some user-level code---it's not in the function docstring). Probably this isn't likely to be an issue here, so I'd be willing to concede the point. But how do you feel about explicitly adding support for the `msprime.simulate()` args you want to use?

jeromekelleher:
I think the kwargs thing is probably OK as an "experts" interface @grahamgower - this is just for people who know they are using msprime under the hood and want to do something funky with it (which we should enable). Otherwise we'll have to duplicate and document the entire msprime API here, which would be tedious and error prone, and the documentation would inevitably drift. So, I think we basically document that kwargs sends any extra args to the underlying engine, and that's the end of it.

grahamgower:
Yeah, fair points @jeromekelleher. The PR is fine with me then, it just needs some small docs addition to explain the behaviour ("additional kwargs are passed to msprime.simulate()"). @awohns the docs build is currently broken and can be ignored.

grahamgower (approved):
LGTM! Thanks @awohns!

PR #750, opened 2021-01-28
Fix Zigzag_1S14 population sizes.

### LINKED ISSUE #745: HomSap/Zigzag_1S14 population sizes don't match Schiffels & Durbin (2014)
**Short description of the problem:**
The population sizes in the `Zigzag_1S14` demographic model swing between 1431 and 14312. However, the values in [Schiffels & Durbin (2014)](https://doi.org/10.1038/ng.3015) are five times higher.

**How to reproduce the problem:**
Compare the zigzag parameters in [stdpopsim's catalog](https://stdpopsim.readthedocs.io/en/latest/catalog.html#sec_catalog_homsap_models_zigzag_1s14) with Figure 2 in Schiffels & Durbin (2014). Also see the `ms` command in the supplementary material, section 7, pg. 25.

This appears to have been deliberate, according to #105, but is not documented, and doesn't seem to have come up during the model QC process (#498).

**Suggested fixes:**
The original `ms` command given in the supplement includes a `-eN 0 5` command, which has not been reproduced in the stdpopsim model. From the ms manual:
`-eN t x` `Set all subpop’s to size x ∗ N0 and growth rates to zero.`

We should at the very least document the difference between the stdpopsim ZigZag model, and that in the original publication.

CC: @Chris1221, @apragsdale.

[EDIT: [translation of the ms command, as stolen from the QC implementation](https://gist.github.com/grahamgower/7a2c0115493547e5ddc02887030abee9)]

grahamgower:
This makes the population sizes match those used in Schiffels & Durbin (2014),
and outputs a warning when simulating the Zigzag_1S14 model.

Closes #745.


grahamgower:
Odd that the circleci failure doesn't also occur on the github-workflow tests. I can't produce locally either, but code inspection suggests this test should fail (simulating the model outputs a warning, so we *should* have `len(stderr) > 0`).

jeromekelleher:
Huh, weird.

jeromekelleher (approved):

grahamgower:
I'm just going to remove the failing test. It doesn't add anything that's not covered elsewhere, and I identifying why its failing only on circleci is not worth the time (I already spent half an hour and didn't come up with anything).

PR #754, opened 2021-02-01
Add a "did you mean" to dem model matching error

### LINKED ISSUE #753: List available DemographicModels when non-existent one accessed
If I do: `model = stdpopsim.get_species("HomSap").get_demographic_model("OOA")`, I currently get:

```
  File "/Users/yan/opt/miniconda3/lib/python3.8/site-packages/stdpopsim/species.py", line 295, in get_demographic_model
    raise ValueError(f"DemographicModel '{self.id}/{id}' not in catalog")
ValueError: DemographicModel 'HomSap/OOA' not in catalog
```

It would be nice to (a) list the available models for this species in the raised error, or even, in addition (b) suggest possible "did you mean" options, perhaps simply on the basis of `[models.startswith(id) for models in catalog]` or `['id' in models for models in catalog]`

hyanwong:
Fixes #753 by adding a "Did you mean model XXX" to the error output, but (a) I haven't added any tests (not sure how we do that unless model names are hardcoded) and (b) we might want to abstract the function into `utils.py` so that we recommend possibilities for e.g. `stdpopsim.get_species("HomMisspelt")` or `get_genetic_map()` too?

I set a lower than default cutoff so that `stdpopsim.get_species("HomSap").get_demographic_model("Out")` suggested some "Out of Africa" models, but tweak it to whatever seems best. I can't see any harm in liberally recommending options though.

### DIFF
--- a/stdpopsim/species.py
+++ b/stdpopsim/species.py
@@ -2,6 +2,7 @@
 Infrastructure for defining basic information about species and
 organising the species catalog.
 """
+import difflib
 import logging
 import warnings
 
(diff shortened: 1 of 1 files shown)


hyanwong:
We could also put a link in the error message such as:

```
f"See https://stdpopsim.readthedocs.io/en/latest/catalog.html#sec_catalog_{self.id}_models" for more options.
```

I don't know how stable this URL is, through?

stdpopsim/species.py
```diff
@@ -292,7 +293,21 @@ def get_demographic_model(self, id):
         for model in self.demographic_models:
             if model.id == id:
                 return model
-        raise ValueError(f"DemographicModel '{self.id}/{id}' not in catalog")
+        suggestions = difflib.get_close_matches(  # Can't harm to use a liberal cutoff
```
jeromekelleher, line 296: Let's make a function for this in ``util`` as you suggest: ``get_suggestions(query, possibilities)`` which returns the list of suggestions, and write some test cases for that. We can then make another function which takes the results of this and formats a nice error string from this, which we can then reuse to make error messages anywhere we're doing ID lookup like this.

We should do some tests on some hard-coded examples from the catalog to make sure it does what we expect, too.

jeromekelleher (commented):
Looks like a good start. It's worth doing it properly though and writing the test cases systematically.

jeromekelleher:
I'm going to close this due to inactivity. Please feel free to reopen if you'd like to finish this up @hyanwong.

hyanwong:
Sure. Perhaps someone at the stdpopsim hackathon today or tomorrow might like to pick it up too?

PR #759, opened 2021-02-10
Update docs github action.

grahamgower:
This removes the docs preview for pull requests in favour
of just building the docs. When a pull request is merged,
a "push" is triggered and so the docs will still be committed
to the stdpopsim-docs repository.

The docs preview relied on a feature which turned out to be
a security risk (which is why it was previously disabled).
No simple alternative is available, so this won't be reenabled.


jeromekelleher (approved):
LGTM

PR #764, opened 2021-02-19
Transition to msprime 1.0 demography model

jeromekelleher:
This is an initial pass at converting the codebase to use the msprime 1.0 Demography model, rather than the msprime 0.x model as we currently do.

This will be a messy change, but I think it's something we can do incrementally, which will also move us closer to demes (which will be the future basis for demography).

The first pass here is to just remove the model comparison code we have here, which is using internal msprime APIs, and use msprime's APIs instead. 

My proposal for the transition would be the following:

1. Remove the ``population_configurations``, ``demographic_events`` and ``migration_matrix`` replace with a ``demography`` object. (May get rid of the ``populations`` list as well, since this is redundant, but not in the first pass)
2. Go through all the catalog files and call ``Demography.from_old_style`` on all the places we create a model object.

This much should work, and we should be able to merge.

Once that's working, we can then start tweaking the model definitions to make them more idiomatic/more demes-like. Ultimately, I'd like to make some demes-like Demographys for the stuff in the catalog, where we use population splits etc rather than mass migrations. We then compare them with the qc models using the ``is_equivalent()`` method, which is able to take into account populations being reused, via a mapping of equivalent populations epoch-by-epoch. That way, we should be able to keep the QC models as they are, but incrementally modernise the catalog models.

Upgrading from the msprime demes-like Demography model should be straightforward enough then at a later date.

What do you think @grahamgower?

### DIFF
--- a/tests/test_HomSap.py
+++ b/tests/test_HomSap.py
@@ -32,12 +32,13 @@ class TestGenome(unittest.TestCase, test_species.GenomeTestMixin):
         genetic_map = "HapMapII_GRCh37"
         species = stdpopsim.get_species("HomSap")
         for chrom in self.genome.chromosomes:
-            if chrom.id == "chrY":
+            if chrom.id == "Y":
                 with self.assertWarns(Warning):
                     contig = species.get_contig(chrom.id, genetic_map=genetic_map)
+                print("HERE")
             else:
                 contig = species.get_contig(chrom.id, genetic_map=genetic_map)
             self.assertAlmostEqual(
                 chrom.recombination_rate,
-                contig.recombination_map.mean_recombination_rate,
+                contig.recombination_map.mean_rate,
             )
--- a/stdpopsim/catalog/BosTau/__init__.py
+++ b/stdpopsim/catalog/BosTau/__init__.py
@@ -115,7 +115,7 @@ def _HolsteinFriesan_1M13():
     inspection of the plots.
     """
     populations = [
-        stdpopsim.Population(id="Holstein-Friesian", description="Holstein-Friesian"),
+        stdpopsim.Population(id="Holstein_Friesian", description="Holstein Friesian"),
     ]
     citations = [_MacLeodEtAl.because(stdpopsim.CiteReason.DEM_MODEL)]
 
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -577,7 +577,7 @@ def write_simulation_summary(engine, model, contig, samples, seed=None):
     dry_run_text += f"{indent}Population: number_samples (sampling_time_generations):\n"
     sample_counts = [0] * model.num_sampling_populations
     for s in samples:
-        sample_counts[s.population] += 1
+        sample_counts[s.population] += s.num_samples
     for p in range(0, model.num_sampling_populations):
         pop_name = model.populations[p].id
         sample_time = model.populations[p].sampling_time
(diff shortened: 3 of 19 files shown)


grahamgower:
Yeah, that sounds like a good transition plan. Of course, the devil will be in the details. 😈

jeromekelleher:
@grahamgower, it'd be good to get your eyes on this at this point.

I've got stuff mostly working - the CLI is working with the msprime engine, and I think we can work through all the rest of the test failures easily enough.

The basic idea is to basically get rid of all the code we have for modelling populations/demography etc and to push this (first) out to msprime and then later to Demes. If people have strayed off the simple "run a simulation" example code and written stuff that depends on the internal structure of the stdpopsim objects, then we'll break their code. But, mostly I think the update shouldn't affect people.

The next step here would be to make the minimal changes to get the QC and production models agreeing, via ``is_equivalent()``, but I didn't want to get into the weeds here in case I'm on  the wrong track entirely.

What do you think?

grahamgower:
There's quite a lot of breakage here in the `DemographicModel` class compared to the documented API (https://stdpopsim.readthedocs.io/en/stable/api.html#stdpopsim.DemographicModel). I think we can probably minimise this though, and include deprecation warnings? In particular for `get_demography_debugger()`. And we could provide `@property`s for the `demographic_events`, `population_configurations`, and `migration_matrix` attributes.

jeromekelleher:
> There's quite a lot of breakage here in the DemographicModel class compared to the documented API

I guess we can try to keep compatibility with this, but I wonder if we should. Most of the API documentation there was "premature documentation", it should never have been exposed as public in the first place. I think we only put the API documentation in here for models etc for stdpopsim *developers* - I certainly didn't have a long-term stable API in mind there. In general, we're going to be pretty hidebound if we try to keep full compatibility with all these details, and end up doing an awful lot of extra work, probably for nothing.

Do you think anyone is actually using the DemographicModel API?

grahamgower:
> Do you think anyone is actually using the DemographicModel API?

That's hard to guess at. Maybe it's worth asking on slack?

jeromekelleher:
@petrelharp @andrewkern This is worth taking a look at now. There's still some details to be worked out, but the basic idea is that we get rid of as much of our local code for defining demographic models as we can, and lean on the msprime tools. I've also updated to use msprime 1.0 APIs for the msprime engine, and updated from RecombinationMap to RateMap everywhere.

What do we think of the basic direction? I think we just have to view the old implementation of DemographicModel as an internal implementation detail that wasn't documented. We'll never get through the transition to demes otherwise.

petrelharp:
This looks good! I'm not sure I have enough of the big picture to be more specific, though.

jeromekelleher:
This is ready for review @grahamgower. I think most things went fairly smoothly - the plan for incrementally modernising the catalog should work, I think. The only thing I'm not sure about is non-sampling pops, which I've commented on elsewhere.

There'll be a batch of stuff to do after this, which I guess is best handled by opening issues. If you see things that are outstanding (like, say, updating the recapitation bit of the SLiM engine) which aren't quick fixes, can you open some issues to track please?

tests/test_HomSap.py
```diff
@@ -31,13 +31,13 @@ def test_recombination_rates(self):
         # compare the results to the current recombination rates for each chromosome
         genetic_map = "HapMapII_GRCh37"
         species = stdpopsim.get_species("HomSap")
-        for chrom in self.genome.chromosomes:
+        for chrom in self.genome.chromosomes[2:]:
```
grahamgower, line 34: What's going on here with the first two chromosomes?

tests/test_HomSap.py
```diff
@@ -31,13 +31,13 @@ def test_recombination_rates(self):
...
             if chrom.id == "chrY":
```
grahamgower, line 35: Hmm... I don't think we have a `chrY` anymore, only a `Y`. So this probably doesn't get executed anymore.

grahamgower (approved):
Looks good!

tests/test_HomSap.py
jeromekelleher: Good spot! Left over from earlier debugging.

tests/test_HomSap.py
jeromekelleher: Changed to "Y", and confirmed it's running now.

jeromekelleher:
OK, I think this is ready to go. @grahamgower, I'll let you do the merging honours!

PR #767, opened 2021-02-25
Remove time deltas from AmericanAdmixture_4B11

jeromekelleher:
These are not necessary and make comparing the model with other implementations impossible.

I guess we should consider changelog updates for things like this, but I think it's such an inconsequential change for users we probably don't need to bother.


grahamgower (approved):
LGTM. This has been a long time coming, I think.

apragsdale (approved):
Looks good to me, too

PR #741, opened 2021-01-25
Require msprime >= 1.0.0a4.

grahamgower:

PR #770, opened 2021-03-02
Minimal changes for msprime 1.0 alpha.

jeromekelleher:
Basically the same as #741 

Hopefully will all work with updated msprime and pyslim.


jeromekelleher:
Looks like we need an msprime package for windows. Working on a building a conda-prerelease for msprime [here](https://github.com/conda-forge/msprime-feedstock/pull/44)

PR #772, opened 2021-03-02
template for requesting addition of a new species

petrelharp:
This would be a way for someone to compile the relevant information for adding a new species without doing the coding.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-species-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/new-species-issue-template.md
@@ -1,0 +1,69 @@
+---
+name: New species
+about: Adding a new species to the stdpopsim catalogue.
+title:
+labels: enhancement
+assignees: ''
+
+---
+
+- **Name:** scientific name of the species
+- **Common name:**
+
+**Description:** any other discussion or useful background? (optional)
+
+*Note: Maybe you don't know a lot of the things below - but, that's OK!
+The criteria are that (a) the choice is reasonable, and (b) the rationale is clear.
+What we want is a "reasonable estimate" (or, "best guess") - if you were to use a number in a paper,
+what would you use and how would you justify it to your collaborators?
+For instance, for "generation time", best would be to cite a paper that actually measures
+or estimates generation time. But most species don't have this;
+next best would be to give a number used in the literature, and provide a citation that used it
+(and hopefully the publication gives a justification also).
+Lacking this, you might provide a number from a related species (and a citation for this).*
+
+*Here is the checklist of things that we need to add a species to the catalogue.
+Each thing should be provided, with a justification (maybe short) and a citation.*
+
+**Demographic information:**
+
+- [] generation time in years
+- [] "default" population size
+
+
+**Chromosome structure:**
+
+- [] ensembl assembly ID *or*, if the assembly is not in Ensembl, a list of chromosomes with *name* and *length* (in bp)
+
+*Note: species in Ensembl can be found in one of these lists:
+[vertebrates](https://uswest.ensembl.org/info/about/species.html),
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/new-species-issue-template.md
```diff
@@ -0,0 +1,61 @@
...
+
+**Demographic information:**
+
+- [] generation time in years
+- [] "default" population size
+
+
+**Chromosome structure:**
+
+- [] list of chromosomes with *name* and *length* (in bp)
```
grahamgower, line 36: These are coming from ensembl now, so this isn't strictly necessary, right?

grahamgower (approved):
LGTM. Maybe we don't need the chromosome list though?

.github/ISSUE_TEMPLATE/new-species-issue-template.md
jeromekelleher: Speaking of which, the Ensembl ID is a necessary bit of info.

.github/ISSUE_TEMPLATE/new-species-issue-template.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+**Recombination rates:**
+
+- [] genetic map (as a .csv) of recombination rates **(optional)**
```
jeromekelleher, line 41: Or Hapmap? Don't want to people to convert to CSV if it's already in the right format.

jeromekelleher (approved):
LGTM, but we should give more info on how to get the Ensembl ID, etc.

petrelharp:
> Speaking of which, the Ensembl ID is a necessary bit of info.

Hm. So, here's the species that are apparently available on ensembl:

- [mostly mammals](https://uswest.ensembl.org/info/about/species.html)
- ["metazoa"](https://metazoa.ensembl.org/species.html)
- [plants](https://plants.ensembl.org/species.html)
- [fungi](https://fungi.ensembl.org/species.html)
- [protists](http://protists.ensembl.org/species.html)
- [bacteria](http://bacteria.ensembl.org/species.html)

I think that's all? I can't find a list of ensembl sites.

And, what do we do about species not in Ensembl (e.g., there are no Mimulus, looks like)? Do we say that we're only for species with annotations uploaded to Ensembl? Any idea how difficult getting a new species on there is?

jeromekelleher:
I guess we could put the data files in by hand for species that aren't in Ensembl. We'd have to introduce some level of QC then, though.

> Any idea how difficult getting a new species on there is?

No - can't imagine it's a quick process though, by the time the data gets into a release.

petrelharp:
> I guess we could put the data files in by hand for species that aren't in Ensembl. We'd have to introduce some level of QC then, though.

Can we do this currently or would this require a lot of changes to the infrastructure? In other words, is this something we want to allow in the "adding to the zoo" workshop?

petrelharp:
Ok, I've updated this - I think it's OK now, unless we think we will never want to support non-Ensembl species.

petrelharp:
More Ensembl info: from [this page](https://uswest.ensembl.org/info/genome/genebuild/assembly.html)
> Ensembl does not produce genome assemblies, instead we provide annotation on genome assemblies that have been deposited into the [INSDC](http://www.ebi.ac.uk/ena) (GenBank, ENA, DDBJ) and are publicly available. We select species to annotate on a case-by-case basis ...

jeromekelleher:
> Can we do this currently or would this require a lot of changes to the infrastructure? In other words, is this something we want to allow in the "adding to the zoo" workshop?

No, probably not. I think we should focus on Ensembl species for the workshop.

jeromekelleher (approved):
LGTM!

PR #773, opened 2021-03-02
Check for minimum SLiM version at runtime.

### LINKED ISSUE #771: Check minimum supported version of SLiM at run time
We need to check that SLiM is at least version X when we try to use the SLiM Engine, or the user will get cryptic errors.

I think the minimum is 3.5 now, after the transition to msprime 1.

grahamgower:
Closes #771.


jeromekelleher (approved):
LGTM

PR #775, opened 2021-03-03
Minimal changes for msprime 1.0 alpha.

jeromekelleher:
Bump slim to version 3.5

Another version of #770, trying to rule out weird cache issues. Conda can't seem to find the msprime packages on Windows (but can for other OSs)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,8 +77,9 @@ jobs:
       - name: install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
-          # Workaround weird issues on windows.
-          conda install --yes -c conda-forge msprime==1.0.0a6
+          # Install pre-release from non standard label until msprime
+          # 1.0 is tagged.
+          conda install -c conda-forge/label/mprime_rc -c conda-forge msprime
           conda install --yes --file=requirements/CI/conda.txt
           if [ "$RUNNER_OS" != "Windows" ]; then
             conda install --yes slim==${{ env.SLIM_TAG }}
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Hooray, looks like this is working! But boo, I have some seriously weird behaviour around the alpha package on conda-forge. I had tried to put this in a non-standard label, but it didn't seem to work, so conda is not randomly installing 0.7.4 or 1.0.0a6, randomly. I'll have to sort this out because we don't want users installing the pre release without knowing about it, so this CI config may well break soon.

Should we merge this now @grahamgower, or wait until the packaging stuff for msprime has been sorted out?

grahamgower:
Well, the alternative is to do something [like this](https://github.com/popsim-consortium/stdpopsim/blob/msprime-1.0/.github/workflows/tests.yml#L85-L108), so the current pr looks just fine to me.

I'm pretty sure we'll need to change a bunch of these files when msprime 1.0 comes along anyway, so that there's no alpha version in any of the requirements. Pip has some non-obvious behaviour when you "require" by comparing with an alpha version of something (the setup.py here asks for `msprime>=1.0.0a6`). Suppose some time after 1.0.0 is released, a 1.1.0a1 is released. Then "pip install stdpopsim" will happily install the newest 1.1.0a1 alpha version to fullfill the dependency requirement rather than the non-alpha 1.0.0. So definitely we'll poke these requirements again soon.

grahamgower (approved):

jeromekelleher:
Let's not merge just yet, the package has been moved on conda-forge. I'll update.

jeromekelleher:
> Well, the alternative is to do something like this, so the current pr looks just fine to me.

We'll never get that working on Windows...

grahamgower:
> > Well, the alternative is to do something like this, so the current pr looks just fine to me.
> 
> We'd never get that working on Windows...

I seem to recall it working.

jeromekelleher:
It works! Let's merge.

Issue #708, opened 2020-12-08
CicleCI should use a pinned version of SLiM


grahamgower:
SLiM 3.5 just got released, and this broke CircleCI tests for an unrelated PR (presumably the issue here is that pyslim must be updated *before* updating SLiM).
https://app.circleci.com/pipelines/github/popsim-consortium/stdpopsim/1611/workflows/cf56e409-1157-4ca1-91dc-1d8283c760c7/jobs/2681

We currently have CI using pinned versions of almost everything. This avoids problems whereby an upstream dependency releases a new version which breaks our builds. And for pip-installable dependencies, @dependabot will open a PR to bump the version shortly after the new release occurs. So, great! The potential build failures are isolated and we also get notified about them.

However... We also have SLiM as a dependency, and it's not pip-installable. We should pin the version (it currently is in github actions, but not in CircleCI). Unfotunately we then won't get notified about potential problems with newer versions. (Note that SLiM is conda-installable, but dependabot doesn't know about conda requirements files, which have a subtly different format to pip requirements files).

jeromekelleher:
Closed in #777 (pinned to 3.5)

PR #777, opened 2021-03-04
Basic changes for pytest

jeromekelleher:
Ping @grahamgower. The plan is to make the smallest update we can here, and then transition the code over to use pytest features over time.

One thing I was confused about was all the stuff around running coverage. Is that needed now?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,10 @@ jobs:
             git clone --bare https://github.com/MesserLab/SLiM /tmp/SLiM_top/SLiM
             cd /tmp/SLiM_top/SLiM
             git describe --tags $(git rev-list --tags --max-count=1) > ${TAG_FILE}
+            # FIXME Pinning to SLiM 3.5 for now; looks like there's an issue with 3.6
+            echo v3.5 > ${TAG_FILE}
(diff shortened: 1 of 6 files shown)


grahamgower:
> One thing I was confused about was all the stuff around running coverage. Is that needed now?

No, those lines can be removed. To get the equivalent output from `pytest`, you can use the `term-missing` coverage report. E.g. see the pytest line in the demes-python repo: https://github.com/popsim-consortium/demes-python/blob/main/.github/workflows/tests.yml#L55-L56

grahamgower:
We could upload the codecov report from the `tests.yml` action too. Then we could remove Circle CI (it's equally difficult to manage as github actions, so having both is just maintenance burden).

jeromekelleher:
> We could upload the codecov report from the tests.yml action too. Then we could remove Circle CI (it's equally difficult to manage as github actions, so having both is just maintenance burden).

Sure - feel free to file an issue to track. I don't want to get too deep down CI rabbit holes here though.

jeromekelleher:
Finally! Jeez that was painful. @grahamgower, can you take a quick look please?

I don't know about you, but I'm not particularly worried about the coverage drop.

grahamgower (approved):
LGTM! You're also closing #708 here.

PR #789, opened 2021-03-07
Refactor catalog

### LINKED ISSUE #785: Refactor catalog code into submodules
The ``__init__.py`` file in the catalog subpackages currently contains a lot of code, which will confuse contributors I think. It would be better if this was split into separate files, .e.g., ``__init__.py`` would become

```
from . import genome
from . import genetic_maps
from . import annotations
from . import demographic_models
```
This would probably help contributors see where their code is supposed to go and make things less confusing.

Thoughts @grahamgower? It's not hard to do, just a bit of file rejiggering.

jeromekelleher:
Closes #785

I haven't changed anything here, just rearranged the code. Also refactored the species testing code to use pytest.


grahamgower (approved):
Thanks @jeromekelleher, this looks good.

I only have a minor concern---some of the citations are now declared in multiple files (e.g. `_locke2011` citation created in both `PonAbe/species.py` and `PonAbe/demographic_models.py`).

jeromekelleher:
> I only have a minor concern---some of the citations are now declared in multiple files (e.g. _locke2011 citation created in both PonAbe/species.py and PonAbe/demographic_models.py).

Yes, this was a compromise. I figured we could add a ``citations.py`` file in the species directory which is cross imported if this gets out of hand, but it was simpler for now to just repeat the citations (which it doesn't matter all that much if there's slight differences in the different copies). What do you think?

grahamgower:
It's probably not a big deal having this small duplication, so I'm happy to leave it as you have it for now.

jeromekelleher:
OK, merging.

PR #796, opened 2021-03-10
Improve tooling around adding species.

jeromekelleher:
A bunch of changes around the process of adding a new species. Hopefully this will make clearer what needs to be done by the original implementers and the follow-up QC.

I'll open a PR for mus_musculus so we can see roughly what it looks like. (now open in #797)

### DIFF
--- a/maintenance/ensembl.py
+++ b/maintenance/ensembl.py
@@ -9,6 +9,9 @@ import urllib.parse
 import urllib.request
 
 
+logger = logging.getLogger("ensembl")
+
+
 class EnsemblRestClient:
     """
     A client for the Ensembl REST API. Based on the example code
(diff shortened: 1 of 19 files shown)


andrewkern (commented):
this looks good @jeromekelleher. one question i had when reviewing the earlier version of this code last night is is there a way to autopopulate some of the citations? e.g., does ensembl give references for their assemblies?

if not would it be worth `add-species` creating a blank citation template?

jeromekelleher:
I don't think we'll be able to get citation data from Ensembl, but yes, adding a citation template for the essential stuff would be useful.

grahamgower:
> @grahamgower, any objections to me changing the "x_citations" stuff to a single citations list with `reasons`? It really doesn't make sense having the separate lists, and might as well get it right while we're setting the template for new additions here.

No, that seems reasonable. If there was a reason these were in separate lists, I don't recall what it was.

jeromekelleher:
OK, done. I think this is ready to go if you want to merge @andrewkern. Hopefully you have what you need for writing the basic directions.

andrewkern:
LGTM @jeromekelleher. do you want to squash?

andrewkern (approved):

jeromekelleher:
Na, these are logically separate commits, merge away (I'm on baby duty)

PR #797, opened 2021-03-10
An example of adding a new species.

jeromekelleher:
Not to be merged, just to show what #796 produces.


andrewkern:
gonna close this cause we now have full example over [here](https://github.com/popsim-consortium/workshops/blob/main/adding_species/contributing.ipynb)

PR #801, opened 2021-03-11
added mosquito genome woot

andrewkern:
THIS IS A TEST FROM THE WORKSHOP

IGNORE THIS PR

but

- added Anopheles gambiae to the catalog
- added recombination rates from each chromosome
- added mutation rate from literature


andrewkern:
wow this passed all tests!?! should we merge? it needs docs still i think

PR #809, opened 2021-03-18
Bos tau edits

gregorgorjanc:


gregorgorjanc:
I am getting this error, but I don't know what to do about it. Can anyone guide me? Thanks! ```=================================== FAILURES ===================================
________________ TestBosTauHolsteinFriesian_IM13.test_id_valid _________________

self = <tests.test_models.TestBosTauHolsteinFriesian_IM13 testMethod=test_id_valid>

    def test_id_valid(self):
>       self.assertTrue(utils.is_valid_demographic_model_id(self.model.id))
E       AssertionError: False is not true```

petrelharp:
From the code: https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/utils.py#L19
it looks like HolsteinFriesian_IM13 should have a 1 instead of an I?

stdpopsim/catalog/BosTau/demographic_models.py
```diff
@@ -1,31 +1,34 @@
...
-    doi="https://doi.org/10.1093/molbev/mst125",
-    year="2013",
     author="MacLeod et al.",
+    year=2013,
+    doi="https://doi.org/10.1093/molbev/mst125",
+    reasons={
+        stdpopsim.CiteReason.GEN_TIME,
+        stdpopsim.CiteReason.POP_SIZE,
+        stdpopsim.CiteReason.DEM_MODEL,
+    },
```
grahamgower, line 16: I think for this citation, it's better to have the reasons given in the different locations where it's used, with the `.because(...)` form. It is possible someone will use the species definition with a different demographic model (or a constant size model), in which case the citation reasons shouldn't include DEM_MODEL.

I.e. The "please cite" output for running:
```
python -m stdpopsim BosTau -D -c 1 10
```
Shouldn't mention "demographic model" for the McLeod citation.

stdpopsim/catalog/BosTau/demographic_models.py
```diff
@@ -1,31 +1,34 @@
...
 )
 
 
-def _HolsteinFriesan_1M13():
-    id = "HolsteinFriesian_1M13"
+def _HolsteinFriesian_IM13():
+    id = "HolsteinFriesian_IM13"
```
grahamgower, line 21: As Peter suggested, this id needs to have `_1M13` (one-em-thirteen) rather than `_IM13` (eye-em-thirteen). The `1` refers to the number of populations in the model.

grahamgower (changes requested):
Looks good @gregorgorjanc; just a couple of minor tweaks and this should be good to merge.

stdpopsim/catalog/BosTau/demographic_models.py
gregorgorjanc: Done!

stdpopsim/catalog/BosTau/demographic_models.py
gregorgorjanc: Done!

grahamgower (approved):
Looks great, thanks @gregorgorjanc! If you'd like to squash down the commits, this can be merged.

jeromekelleher (approved):
LGTM

gregorgorjanc:
@grahamgower commits squashed

PR #821, opened 2021-03-26
ask for chromosome-level assembly

petrelharp:
A good number of genomes in ensembl are just scaffold-level, with thousands of scaffolds. I don't *think* we want those.


jeromekelleher (approved):
Agreed!

Issue #557, opened 2020-06-09
What to do when our sims don't line up with data?


andrewkern:
Okay so just ran into an interesting case with @cjbattey.

CJ has been comparing stdpopsim to 1000 Genomes data as a baseline for a project we are working on. Simulating under `OutOfAfrica_3G09` yielded too few mutations in comparison to equivalent data from 1kg. e.g., for chr22 and considering only sites in the 'P' state of the 1kg masking `OutOfAfrica_3G09` gives roughly 2x fewer SNPs than are seen in phase 3 samples.

One reason for this difference is that stdpopsim default mutation rate (1.29e-08) is almost half of the mutation rate estimate used by @RyanGutenkunst et al in the original paper (2.38e-08). 

So this is an issue that we might deal with in a number of ways (see #530).

1. have paper specific mutation rates override defaults (as others have suggested)
2. trigger warnings if defaults are used (our rationale for having defaults is still good-- we want up to date, best info in those slots, but users need to beware of caveats)
3. allow users to simply define there own custom parameters. perhaps we would want this to extend to the CLI. i just submitted #556 as part of this.

thoughts? it's kinda thorny

jeromekelleher:
I think users should probably use the Python API and override the values at that point. I don't think we can accommodate that level of detail in the CLI.

petrelharp:
The other option is

4. Have "mutation rate used for inference" be part of the model, and then rescale times accordingly.

This is thorny because in some methods the mutation rate isn't just a simple time scaling, though.

PR #839, opened 2021-04-07
Allow demographic models to store mutation rates

apragsdale:
See #557.

The issue: implemented demographic models were originally inferred with certain assumptions about mutation rates, generation times, etc, which then are used to calibrate population size and demographic event timings. Since #556, #631, users can specify and simulate with contigs with user-defined mutation rates, so that mutation rates could be set to match what was assumed when a model was first inferred. However, unless a user is very familiar with the literature, they might not be aware that the species' default mutation rate does not match the model they want to use, so that outputs will have measures of diversity that are far too large or too small for that species without realizing it.

One solution, as proposed in #557, is to allow models to have their own mutation rates that can be stored and used instead of the default rate. The default behavior of this PR is to use the species' default rate if a mutation rate is not specified, but now mutation rates would be attributes of the demographic model that can be easily passed to `species.get_contig()`.

For example:
```python
import stdpopsim
model = species.get_demographic_model("OutOfAfrica_3G09")
contig = species.get_contig("chr1", mutation_rate=model.mutation_rate)
```

This contig now has a mutation rate of 2.35e-8 instead of 1.29e-8. There are some models that don't have associated mutation rates (for example, they were calibrated using recombination rates), and those models can be left without mutation rates so that they use the species' default rate (`None` would default to species' default). This would require going back through and adding mutation rates to each species' demographic model and then have them be QC'd and verified, but I think it would be worth the effort.

I didn't make a ton of changes to each demographic model here or implement tests, since feedback would be really helpful here before spending the time to do that.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -75,7 +75,9 @@ chromosome, recombination map, and number of samples.
 The most basic simulation we can run is to simulate two (haploid) genomes
 - i.e., two samples -
 using the species' defaults as seen in the species help (``stdpopsim HomSap --help``).
-These defaults include constant size population and a uniform recombination map.
+These defaults include constant size population, a uniform recombination map based
+on the average recombination rate (either genome-wide or within a chromosome, if
+specified), and the mutation rate shown above.
 To save time we will specify that the simulation use
 chromosome 22, using the ``-c`` option. We also specify that the resulting
 tree-sequence formatted output should be written to the file ``foo.ts`` with the
@@ -123,6 +125,15 @@ model has two populations that we can sample from
 The order of those numbers is the same as the order
 specified in the model documentation. In this case, ``2 3`` means
 we are simulating two African American samples and three European American samples.
+
+.. note::
+    Many demographic models were inferred or calibrated using a mutation rate that
+    differs from the cataloged species' mutation rate. Simulations using the CLI now
+    automatically use the *model's* specified mutation rate instead of the species
+    rate, so that expected levels of diversity more closely match those observed in
+    the data that were used to infer the demographic model. For generic demographic
+    models or those without associated mutation rates, the species mutation rate is
+    used.
 
 Now we want to add an empirical recombination map to make the simulation more
 realistic. We can look up the available recombination maps using the
@@ -282,7 +293,7 @@ using SLiM with a (very large!) scaling factor of 1000, we could run
 
 .. code-block:: console
 
-   $ stdpopsim -e slim --slim-scaling-factor 1000 DroMel \
+   $ stdpopsim -e slim --slim-scaling-factor 1000 DroMel\
    $     -c chr2L -l 0.05 -o foo.ts -d African3Epoch_1S16 100
 
 The scaling factor of 1000 makes this model run very quickly,
@@ -350,31 +361,21 @@ Running a published model
 The first example uses a mostly default genome
 with a published demographic model.
 
-Pick a species and a contig
----------------------------
+Pick a species and demographic model
+------------------------------------
 
-First, we will pick a species (humans), contig (chromosome 22),
-and genetic map (a flat map, the default):
+First, we will pick a species (here, humans) and the published demographic
+model to simulated under. In ``stdpopsim`` there are two types of model: ones
+taken to match the :ref:`demographic history reported in published papers
+<sec_catalog>`, and :ref:`"generic" models <sec_api_generic_models>`. We'll
+first simulate using a published model from the catalog. Let's see what
+demographic models are available for humans:
 
 .. code-block:: python
 
    import stdpopsim
 
    species = stdpopsim.get_species("HomSap")
-   contig = species.get_contig("chr22")  # default is a flat genetic map
-
-Choose a demographic model
---------------------------
-
-Now, we need a demographic model.
-In ``stdpopsim`` there are two types of model:
-ones taken to match the :ref:`demographic history reported in published papers <sec_catalog>`,
-and :ref:`"generic" models <sec_api_generic_models>`.
-First, we'll simulate samples from a published model of human prehistory.
-Let's see what demographic models are available
-for humans:
-
-.. code-block:: python
 
    for x in species.demographic_models:
        print(x.id)
@@ -387,11 +388,13 @@ for humans:
    # Zigzag_1S14
    # AncientEurasia_9K19
    # PapuansOutOfAfrica_10J19
+   # AshkSub_7G19
+   # OutOfAfrica_4J17
 
 These models are described in detail in the :ref:`Catalog <sec_catalog>`.
 We'll look at the first model, "OutOfAfrica_3G09", from
 `Gutenkunst et al (2009) <https://doi.org/10.1371/journal.pgen.1000695>`__.
-First we need to know how many populations it has, and what they are:
+We can check how many populations exist in this model, and what they are:
 
 .. code-block:: python
 
@@ -400,14 +403,50 @@ First we need to know how many populations it has, and what they are:
    # 3
    print(model.num_sampling_populations)
    # 3
-   print([pop.id for pop in model.populations])
+   print([pop.name for pop in model.populations])
    # ['YRI', 'CEU', 'CHB']
 
-This model has 3 populations, named YRI, CEU, CHB,
-and all three can be sampled from.
+This model has 3 populations, named YRI, CEU, CHB, and all three can be sampled from.
 The number of "sampling" populations could be smaller than the number of populations,
 since some models have ancient populations which are currently not allowed to be
 sampled from - but that is not the case in this model.
+
+Set up the contig
+-----------------
+
+We'll next define the contig, which contains information about the genome length we
+want to simulate and recombination and mutation rates. Here, we use the human
+chromosome 22. If no recombination map is specified, we assume a uniform genetic map
+based on the average recombination rate for that chromosome.
+
+.. code-block:: python
+
+   contig = species.get_contig("chr22")
+
+   # default is a flat genetic map
+   print("mean recombination rate:", f"{contig.recombination_map.mean_rate:.3}")
+   # mean recombination rate: 1.44e-08
+
+   # and the default mutation rate is based on the species default
+   print("mean mutation rate:", contig.mutation_rate)
+   # mean mutation rate: 1.29e-08
+
+   # but note that the mutation rate differs from the model's assumed rate
+   print("model mutation rate:", model.mutation_rate)
+   # model mutation rate: 2.35e-08
+
+The Gutenkunst OOA model was inferred using a mutation rate much larger than the
+default mutation rate in the `stdpopsim` catalog. As such, simulating using this
+model and default rate will result in levels of diversity substantially lower than
+expected for the human population data that this model was inferred from. To match
+observed diversity in humans, we should instead use the mutation rate associated
+with the demographic model:
+
+.. code-block:: python
+
+   contig = species.get_contig("chr22", mutation_rate=model.mutation_rate)
+   print(contig.mutation_rate == model.mutation_rate)
+   # True
 
 Choose a sampling scheme and simulate
 -------------------------------------
@@ -459,19 +498,16 @@ so we use the json module to easily parse it:
 Running a generic model
 =======================
 
-Next, we will simulate using a "generic" model,
-with piecewise constant population size.
-This time,
-we will simulate a small portion of human chromosome 2,
-again with a flat recombination map,
-using the current best estimate of the human
-effective population size from the :ref:`sec_catalog`.
+Next, we will simulate using a "generic" model, with piecewise constant
+population size. This time, we will simulate a given genome length under
+a flat recombination map, using an estimate of the
+human effective population size from the :ref:`sec_catalog`.
 
 Choose a species
 ----------------
 
-Although the model is generic, we still need a species,
-to get contig information.
+Although the model is generic, we still need a species in order
+to get the contig information.
 Again, we'll use `Homo sapiens`, which has the id "HomSap".
 (But, you could use any species from the :ref:`sec_catalog`!)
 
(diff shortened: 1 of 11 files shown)


grahamgower:
Looks like a sensible approach. Thanks @apragsdale!

jeromekelleher:
Yep, LGTM also. Just needs a few tests.

petrelharp:
Hm - it'd be nice if a mutation rate in the model would be used directly instead of having to remember to pass it in. (I agree, it needs to be an attribute of the model, though, so this is good.) Maybe simulation should throw a warning if the model's mutation rate and the contig's mutation rate don't match?

grahamgower:
> Hm - it'd be nice if a mutation rate in the model would be used directly instead of having to remember to pass it in.

That's certainly an option. We'd want to make sure the citation for the species mutation rate isn't printed too.

apragsdale:
> Hm - it'd be nice if a mutation rate in the model would be used directly instead of having to remember to pass it in. (I agree, it needs to be an attribute of the model, though, so this is good.) Maybe simulation should throw a warning if the model's mutation rate and the contig's mutation rate don't match?

> That's certainly an option. We'd want to make sure the citation for the species mutation rate isn't printed too.

Those are good points. Agree that using the model parameter as default would be good, but I'm not sure how to use it directly without completely pulling the mutation rate out of the contig class and moving it to the model entirely. We still want to preserve users' abilities to set custom mutation rates as well. A warning makes sense and would avoid a giant refactor

apragsdale:
I'm going to go through and add mutation rates to the catalog models here. They will need to be QC'd by someone else most likely, and mutation rate equality between the models and their QC counterparts should also be checked in the tests.

stdpopsim/engines.py
```diff
@@ -112,6 +112,15 @@ def _warn_zigzag(self, demographic_model):
                 "https://github.com/popsim-consortium/stdpopsim/issues/745"
             )
 
+    def _warn_mutation_rate_mismatch(self, contig, demographic_model):
+        if demographic_model.mutation_rate is not None and demographic_model.mutation_rate != contig.mutation_rate:
+            warnings.warn(
+                "The demographic model mutation rate differs from the contig mutation "
+                "rate (simulation uses the contig mutation rate). Diversity levels "
+                "may be different than expected for this species. For details see: "
+                "(url of some documentation for this)"
```
petrelharp, line 124: maybe stick in the mutation rates? Like `f"The demographic model has mutation rate {demographic_model.mutation_rate}, but this simulation used mutation rate {contig.mutation_rate}. ..."

apragsdale: Good idea. Updated now to include those rates

apragsdale:
One obvious thing still missing are tests to ensure that the model mutation rate matches the QC model mutation rate (which have not yet been included, and should be compiled and added by someone other than myself). Am I missing anything else that is obvious here (or non-obvious)?

apragsdale:
Added a test for mutation rate equality between model and its QC, and also added a test for generation time equality (I also slightly refactored the qc register so that it registers the stdpopsim model object instead of just the msprime Demography object, so that it's easier to check other attributes of the model).

The tests that fail now are because the QC models need mutation rates to be added.

stdpopsim/catalog/AraTha/demographic_models.py
```diff
@@ -102,6 +102,8 @@ def _sma_1pop():
         )
     ]
 
+    mutation_rate = 7.1e-9
```
jeromekelleher, line 105: Maybe we should separate out these catalog changes from the code updates? These should be QC'd more closely and would require looking up the original papers etc, which is a lot to do at once.

Specifying a mutation rate shouldn't be necessary for the models, it can default to None so the basic change PR can go through.

stdpopsim/cli.py
```diff
@@ -293,6 +293,8 @@ def write_citations(engine, model, contig, species):
     """
     cite_str = ["If you use this simulation in published work, please cite:"]
     for citation in get_citations(engine, model, contig, species):
+        if "mutation rate" in citation.reasons and model.mutation_rate is not None:
```
jeromekelleher, line 296: We have constants for this I think, don't need to use a hard coded string. citations.CiteReason.MUT_RATE I think?

stdpopsim/cli.py
```diff
@@ -587,7 +590,11 @@ def write_simulation_summary(engine, model, contig, samples, seed=None):
     # Get information about relevant contig
     gmap = "None" if contig.genetic_map is None else contig.genetic_map.id
     mean_recomb_rate = contig.recombination_map.mean_rate
-    mut_rate = contig.mutation_rate
+    # use the model mutation rate, if provided with the demographic model
+    if model.mutation_rate is not None:
```
jeromekelleher, line 594: probably good to add a logging.info() saying where the mutation rate is coming from

stdpopsim/engines.py
```diff
@@ -112,6 +112,20 @@ def _warn_zigzag(self, demographic_model):
                 "https://github.com/popsim-consortium/stdpopsim/issues/745"
             )
 
+    def _warn_mutation_rate_mismatch(self, contig, demographic_model):
+        if (
+            demographic_model.mutation_rate is not None
+            and demographic_model.mutation_rate != contig.mutation_rate
```
jeromekelleher, line 118: math.isclose() would be better here, exact equality will lead to noise I think.

stdpopsim/models.py
```diff
@@ -79,20 +82,22 @@ def __init__(
...
         model=None,
+        mutation_rate=None,
     ):
         self.id = id
         self.description = description
         self.long_description = long_description
         self.generation_time = 1 if generation_time is None else generation_time
+        self.mutation_rate = mutation_rate
         self.citations = [] if citations is None else citations
-        self.qc_model = qc_model
```
jeromekelleher, line 95: What's the reasoning for this change?

stdpopsim/species.py
```diff
@@ -169,6 +170,8 @@ def get_contig(
             same chromosome-specific mutation and recombination rates.
             This option cannot currently be used in conjunction with the
             ``genetic_map`` argument.
+        :param float mutation_rate: The per-base mutation rate. If none is given,
```
jeromekelleher, line 173: Why do we need a species mutation rate as well?

jeromekelleher (commented):
Basically looks good @apragsdale, but there's an awful lot changing here at once. I think we should at least separate this into the basic changes to the object model and the changes to the catalog data. I think we're also increasingly muddling up simulation logic with the job of declaratively encoding information about species in the catalog. To me, there should be no concept of the semantics of running simulations in the catalog code, it should just be about declaring information that we know about species.

I'm also increasingly negative about the idea of a "generic" contig - it really seems like more trouble than it's worth.

stdpopsim/models.py
apragsdale: Previously, the QC model (just the `Demography` object) was attached to a model to check 1) that a model has been QC'd (is not None), and 2) to check for equivalency in tests. I think it would also be good to check equivalency of other attributes of the model to its QC counterpart, such as generation time and mutation rate, so that those are more easily available to compare to each other if the full QC stdpopsim DemographicModel object is attached instead.

stdpopsim/species.py
apragsdale: Previously, the QC model (just the Demography object) was attached to a model to check 1) that a model has been QC'd (is not None), and 2) to check for equivalency in tests. I think it would also be good to check equivalency of other attributes of the model to its QC counterpart, such as generation time and mutation rate, so that those are more easily available to compare to each other if the full QC stdpopsim DemographicModel object is attached instead.

apragsdale:
> I think we should at least separate this into the basic changes to the object model and the changes to the catalog data.

That makes sense - I had gone through and looked up all mutation rates from implemented demographic models already, so I'll move them to their own PR that can be QC'd separately.

> I think we're also increasingly muddling up simulation logic with the job of declaratively encoding information about species in the catalog. To me, there should be no concept of the semantics of running simulations in the catalog code, it should just be about declaring information that we know about species.

I'm not sure I follow, what is being added to the catalog that is being muddled here, aside from mutation rates to the demographic models? (in my view, the mutation rates are just as model-specific as the generation times, population sizes, and demographic event times)

> I'm also increasingly negative about the idea of a "generic" contig - it really seems like more trouble than it's worth.

I think it's an important basic feature. I can only speak for my own usage, but I far more often simulate a fixed length (say 1Mb with a rec rate that I specify) than simulate under a listed chromosome.

stdpopsim/species.py
jeromekelleher: I think this comment is about the ``qc_model`` -> ``qc`` change above? (I don't see the connection with Species.mutation_rate?)

stdpopsim/models.py
apragsdale: Previously, the QC model (just the Demography object) was attached to a model to check 1) that a model has been QC'd (is not None), and 2) to check for equivalency in tests. I think it would also be good to check equivalency of other attributes of the model to its QC counterpart, such as generation time and mutation rate, so that those are more easily available to compare to each other if the full QC stdpopsim DemographicModel object is attached instead.

stdpopsim/species.py
apragsdale: Yeah sorry, replied to wrong review.. edited comment.

jeromekelleher: Hmm - Github seems to have messed up the comments here - ignore

jeromekelleher: Hmm, well, let's take the Species.mutation_rate out of this PR and get the other changes merged first. It's not clear why we'd have a Species.mutation_rate and not a Species.recombination_rate, etc, and why this wouldn't be derived from the per-chromosome values that we're trying to catalog.

apragsdale: This is in the `get_contig` function, so that you can specify a contig with the demographic model's mutation rate instead of the chromosome rate. I don't think there is a Species.mutation_rate still?

jeromekelleher: Ah, sorry, my bad.

stdpopsim/cli.py
apragsdale: I'm not sure if what I have now here is exactly what you mean... let me know if you had something else in mind.

stdpopsim/catalog/AraTha/demographic_models.py
apragsdale: Done - I'll open a separate PR with these rates and their tests for equality.

apragsdale:
I think this fix is just about ready to merge after a rebase/squash. Once it is, I can open a PR with the catalog mutation rate changes and fix up the tests using those specified mutation rates.

stdpopsim/engines.py
```diff
@@ -112,6 +113,19 @@ def _warn_zigzag(self, demographic_model):
...
+    def _warn_mutation_rate_mismatch(self, contig, demographic_model):
+        if demographic_model.mutation_rate is not None and not math.isclose(
+            demographic_model.mutation_rate, contig.mutation_rate
+        ):
+            warnings.warn(
+                "The demographic model has mutation rate "
+                f"{demographic_model.mutation_rate}, but this simulation used the "
+                f"contig's mutation rate {contig.mutation_rate}. Diversity levels "
+                "may be different than expected for this species. For details see "
+                "documenation at "
```
petrelharp, line 125: "documen**t**ation"

stdpopsim/models.py
petrelharp: hm, but this is changing the API for the Demographic Model - if we really want to change this, we should probably leave `qc_model` as a deprecated synonym... but, this *is* a "model" after all, so `qc_model` doesn't seem so bad?

apragsdale: That's true - I'm fine with either name, and that avoids a param name change. In the equivalency tests we would then check `DemographicModel.model == DemographicModel.qc_model.model`

stdpopsim/engines.py
apragsdale: Fixed! Good spot

stdpopsim/cli.py
andrewkern: could be good to print the actual rate in that logging message so the user can see it

stdpopsim/models.py
andrewkern: yeah i agree with Peter here. I think we should probably avoid changing the param name

apragsdale:
Just rebased. Would love to get this merged soon!

grahamgower (approved):
LGTM!

docs/tutorial.rst
```diff
@@ -282,7 +293,7 @@ using SLiM with a (very large!) scaling factor of 1000, we could run
 
 .. code-block:: console
 
-   $ stdpopsim -e slim --slim-scaling-factor 1000 DroMel \
```
jeromekelleher, line 285: Does this work? I thought ``--engine`` options were to the top level parser and not available to the subcommands.

tests/test_models.py
```diff
@@ -315,3 +327,34 @@ def test_zigzag_produces_warning(self):
         for engine in stdpopsim.all_engines():
             with self.assertWarnsRegex(UserWarning, "Zigzag_1S14"):
                 engine.simulate(model, contig, samples, dry_run=True)
+
+
+class TestMutationRates(unittest.TestCase):
+    def test_mutation_rate_warning(self):
+        species = stdpopsim.get_species("HomSap")
+        model = species.get_demographic_model("OutOfAfrica_3G09")
```
jeromekelleher, line 335: Better do a copy.deepcopy() on this object, as you're potentially modifying the global state and this might affect other test. Ditto below.

jeromekelleher (approved):
LGTM, minor comments.

docs/tutorial.rst
apragsdale: Ah, you're right. It guess it looked weird to have the species "positional" argument coming in the middle of some optional flags. In general, the CLI interface is confusing and doesn't follow standard guidelines, but that's another matter. I'll change it back so it works with what we've got.

apragsdale:
I have no clue how to test these logging outputs that codecov is complaining about...

agladstein:
Hey, I'm not sure if I'll be able to review this today. If you want to merge it go for it. I can always add an issue later if something comes up.

jeromekelleher:
I'm happy to merge this as it is, the coverage things look pretty minor.

apragsdale:
Is this ok to merge? I have a follow-up PR with updates to the catalog to include the mutation rate, but waiting for this to be pulled in.

Issue #851, opened 2021-04-08
Converting recombination maps to Zarr


jeromekelleher:
The current approach of storing recombination maps is very error prone and inefficient. There's a lot of manual effort involved in actually packaging the maps into the format that's required, and parsing the hapmap format is slow. The current approach evolved out of something that was quick to get working initially, while other functionality was being developed - it was never intended as a long-term thing.

Here's a proposal:

1. Store the maps in Zarr format, one Zarr store per map (i.e., ``HapMapII_GRCh38`` would be one file, which would contain the data for each chromosome).
2. Add a "add-recombination-map" command to the maintenance script which would take a directory of chromosome hapmap files as input, and create the corresponding Zarr file. This would do various sanity checks on the input data, to make sure that (e.g.) the chromosome lengths agree with the genome build. 
3. (Optional) Add command to the maintenance script to upload maps to AWS.  Once the user has the right credentials in their environment, this should be easy enough to do.

PR #852, opened 2021-04-08
Initial script to convert hapmap to zarr

jeromekelleher:
A first pass at #851.

Opening the zarr and getting the recombination map now takes a fraction of a second, massively faster than parsing hapmap.

### DIFF
--- a/maintenance/main.py
+++ b/maintenance/main.py
@@ -12,6 +12,8 @@ import logging
 import click
 import black
 import daiquiri
+import msprime
+import zarr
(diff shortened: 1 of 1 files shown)


andrewkern:
would we still store hapmap formatted files in AWS or would we push a new directory full of zarrs?

jeromekelleher:
Just push a new directory full of zarrs. We'd keep the old maps around forever, so that existing code won't break, but this would be a clean start.

PR #863, opened 2021-04-08
Added Chlamydomonas reinhardtii genome

### LINKED ISSUE #828: Add Chlamydomonas reinhardtii to catalogue
- **Name:** _Chlamydomonas reinhardtii_
- **Common name:** n/a

**Description:** unicellular isogamous alga with a haploid 110 Mb genome - used in studies of cell motility, but also more recently as a model for evolutionary genetics (especially in light of its facultatively sexual life cycle) 

---

*Note: Maybe you don't know a lot of the things below - but, that's OK!
The criteria are that (a) the choice is reasonable, and (b) the rationale is clear.
What we want is a "reasonable estimate" (or, "best guess") - if you were to use a number in a paper,
what would you use and how would you justify it to your collaborators?
For instance, for "generation time", best would be to cite a paper that actually measures
or estimates generation time. But most species don't have this;
next best would be to give a number used in the literature, and provide a citation that used it
(and hopefully the publication gives a justification also).
Lacking this, you might provide a number from a related species (and a citation for this).*

*Here is the checklist of things that we need to add a species to the catalogue.
Each thing should be provided, with a justification (maybe short) and a citation.*

**Demographic information:**

- [x] generation time in years
  - ~2.4 mitotic doublings per day -> 365 * 2.4 = 876 doublings per year (Vitova et al 2011, [link][vitova2011])
  - 1/876 = 0.0011 years/generation
  - I hope I did this right - please correct me if I'm wrong! 
- [x] "default" population size
  - 1.4 * 10^7 (Ness et al 2016, [link](https://academic.oup.com/mbe/article/33/3/800/2579627?login=true))

[vitova2011]: https://link.springer.com/article/10.1007/s00425-011-1427-7

**Chromosome structure:**

*Note:* The assembly should be *chromosome-level*, i.e., not composed of thousands of scaffolds.

- [x] ensembl assembly ID *or*, if the assembly is not in Ensembl, a list of chromosomes with *name* and *length* (in bp)
  - GCA_000002595.3 ([Ensembl (Plants) link][ensembl], [ENA link][ena])

[ensembl]: http://plants.ensembl.org/Chlamydomonas_reinhardtii/Info/Annotation/#assembly
[ena]: https://www.ebi.ac.uk/ena/browser/view/GCA_000002595.3

**Recombination rates:**

- [ ] genetic map of recombination rates (as a hapmap or csv file)
- [x] genome-wide mean recombination rate (which will be used as the default)
  - 0.015 cM/Mb (Hasan and Ness 2020, [link][ld-paper])
  - This is = rho / 2Ne, and is two orders of magnitude lower than [lab estimates][kathir2003] (9.15 cM/Mb)! But if simulating wild populations of C. reinhardtii, in the absence of a means of modifying the rate of sex in natural populations, this value already takes the infrequency of sex into account
  - If a means of modifying the rate of sex is implemented into `stdpopsim`, then the 9.15 cM/Mb value is appropriate to use when simulating natural populations, provided the rate of sex is set to 1/840 (see [Hasan and Ness 2020][ld-paper])
  - But it is important to note that this is NOT the expected recombination rate in a single cross of C. reinhardtii, and is in fact substantially lower! 

[kathir2003]: https://ec.asm.org/content/2/2/362.short
[ld-paper]: https://academic.oup.com/gbe/article/12/4/370/5809146?login=true

**Mutation rate:**

- [x] genome-wide mean mutation rate
  - μ = 9.63 × 10^−10 (Ness et al 2015, [link][ness-2015])

[ness-2015]: https://genome.cshlp.org/content/25/11/1739.short

**Demographic model:**

- [ ] as a list of population sizes, growth rates, migration rates, etcetera. **(optional)**
- [ ] citation

**Other information:**

These are things we don't currently use, but will want to use in the future:

- [x] sex determination system, and which chromosomes are the sex chromosome(s)
  - C. reinhardtii is an isogamous, heterothallic species with mating types determined by a ~500 kb locus on chromosome 6
  - individuals are either MT+ or MT- based on which allele is present at said locus
  - see [Umen 2011][umen-2011] for more details
- [x] ploidy
  - haploid

[umen-2011]: www.sciencedirect.com/science/article/pii/S1369527411001755

aays:
No demographic models for this one yet - I intend to do some digging after this, though, even if I'm not super confident there's many out there given how much Chlamy pop gen is only just taking off. But in either case, here's the reinhardtii genome! 

An important note re: the recombination rate - sex takes place once every ~840 generations in the species (it's facultatively sexual), and the recombination rate estimates in here reflect that (e.g. they are r = rho / 2Ne, where rho was estimated directly from natural populations). To get 'true' physical recombination rate estimates, each of the values would be divided by the rate of sex, 0.001194 (explained further in the results/discussion [here](https://academic.oup.com/gbe/article/12/4/370/5809146?login=true#supplementary-data)) which I've left as a note in the script, but I figure that if simulating wild populations it makes more sense to have 'effective' recombination rates in there as opposed to the physical rates (since that would otherwise inflate the amount of recombination happening in the simulation).

If there's functionality to modify the rate of sex in a given simulation I think it'd make sense to encode the physical rates instead and save 0.001194 as a separate parameter. Happy to hear thoughts though! 

Closes #828.


jeromekelleher:
LGTM - can you squash the commits please @aays? https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing

aays:
Sounds good @jeromekelleher - on it! I've just heard back from Rob about per-chromosome mutation rates though, so I'm going to nab those as well before squashing the commits. 

Re: selfing rate, that sounds good as well - it's less so selfing and more infrequent sex altogether (Chlamy is happy to have many mitotic generations before periodically outcrossing) but I figure the same idea applies in that that's already baked into the estimate of Ne.

jeromekelleher:
Looks good! Merging.

Issue #884, opened 2021-04-09
Set tolerances explicitly in pytest.appox for species QC stubs


jeromekelleher:
It turns out that he default tolerances aren't appropriate for comparing recombination rate values, and we need to set them explicitly. 

@percyfal, can you outline your experience with the drosophila QC here please?

jeromekelleher:
#885

percyfal:
Sure. So I calculated recombination values from [Table 2](https://journals.plos.org/plosgenetics/article/figure?id=10.1371/journal.pgen.1002905.t002) in Comeron et al. with the following R code:

```
gmap <- data.frame(
    #col.names=c("X", "2L", "2R", "3L", "3R"),
    "X"=c(71.8, 66.2, 56.4, 63.4, 67.5, 66.6, 73.0, 57.7),
    "2L"=c(49.8, 55.8, 64.9, 62.5, 55.6, 52.4, 55.9, 57.5),
    "2R"=c(45.8, 53.8, 60.0, 51.4, 49.0, 55.0, 58.6, 59.3),
    "3L"=c(54.9, 52.7, 49.9, 51.8, 55.9, 50.3, 53.4, 53.9),
    "3R"=c(47.6, 60.8, 56.7, 61.3, 56.2, 66.5, NA, NA)
)
gmap.mean <- apply(gmap, 2, mean, na.rm=TRUE)
gmap.mean["4"] = mean(gmap.mean)
chrlen = c(22909512, 24956976, 21536224, 28131630, 30464902, 1277805)
gmap.mean / chrlen / 100
```
The entries were manually entered from the table which is a potential source of error (Friday evening; I'm tired...). I had squinted at the genome_data file enough to see that the values for chromosome 4 were calculated as an average. I hadn't noticed that the average was weighted though, so my value is one order of magnitude off (4.49e-7). I did think this was wrong, but decided to run the test to verify that the tests fail, which to my surprise they didn't. The weighted average for chromosome 4  in species.py is 2.00e-08. pytest approx gives two ways to specify tolerance: relative tolerance (default 1e-6) and absolute tolerance (default 1e-12). I had to change the relative tolerance for the test to fail, but shouldn't the default values suffice?

petrelharp:
See #914.

petrelharp:
I think there was something else going on originally (or pytest has changed?), since
```
>>> 4.49e-7 == pytest.approx(2e-8)
False
```
[...]

Issue #799, opened 2021-03-11
Genome-wide recombination rate lowered by rate 0 chromosomes


jeromekelleher:
When we create a "generic" config of a given length for a species, we compute the recombination and mutation rates as the length weighted recombination rate over the whole genome, [here](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/species.py#L206). We do exclude non-autosomes, but the detection is weak and won't work for most species.

Perhaps it would be simpler to just exclude chrs with zero recombination or mutation rate from the computation?

grahamgower:
Good idea. What about also excluding regions of rate 0 (e.g. centromeres)?

andrewkern:
but what would we do with a chromosome, like chr4 is drosophila that never recombines? we usually include that "0" when averaging

grahamgower:
> but what would we do with a chromosome, like chr4 is drosophila that never recombines? we usually include that "0" when averaging

Interesting! I feel it would be a mistake to simulate using a value averaged from both recombining and non-recombining chromosomes though. Surely the user is better served by simulating something that either recombines (using a rate averaged over recombining regions), or simulating something that doesn't recombine?

jeromekelleher:
Yes, that was my take as well @grahamgower.

Issue #630, opened 2020-10-18
averaged recombination rates for some species are wrong


grahamgower:
**Short description of the problem:**
Some species definitions (such as `HomSap`) use incorrect per-chromosome recombination rates. I suppose these were calculated using `msprime.RecombinationMap.mean_recombination_rate`. Unfortunately, [msprime versions <= 0.7.4 give an incorrect value](https://github.com/tskit-dev/msprime/issues/844) because they assume the genetic map starts at position 0.

**How to reproduce the problem:**
```Python
$ python
Python 3.8.6 (default, Sep 30 2020, 04:00:38)
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import msprime, stdpopsim
>>> msprime.__version__
'0.7.4'

>>> human = stdpopsim.get_species("HomSap")
>>> chr1 = human.get_contig("1", genetic_map="HapMapII_GRCh37")
>>> chr1.recombination_map.mean_recombination_rate
1.1487055306951004e-08

>>> chr22 = human.get_contig("22", genetic_map="HapMapII_GRCh37")
>>> chr22.recombination_map.mean_recombination_rate
1.4466102768915285e-08
```

This is not so much of a problem with chr1.
```
$ head -n 2 ~/.cache/stdpopsim/genetic_maps/HomSap/HapMapII_GRCh37/genetic_map_GRCh37_chr1.txt
Chromosome      Position(bp)    Rate(cM/Mb)     Map(cM)
chr1    55550   2.981822        0.000000

$ tail -n 1 ~/.cache/stdpopsim/genetic_maps/HomSap/HapMapII_GRCh37/genetic_map_GRCh37_chr1.txt
chr1    249218992       0.000000        286.279234

$ python -c "print(286.279234 / (249218992 - 55550) / 100)"
1.1489616281669442e-08
```

But some chromosomes (e.g. human chr22) have a large telomere at the start, and hence the map starts well into the chromosome.
```
$ head -n 2 ~/.cache/stdpopsim/genetic_maps/HomSap/HapMapII_GRCh37/genetic_map_GRCh37_chr22.txt
Chromosome      Position(bp)    Rate(cM/Mb)     Map(cM)
chr22   16051347        8.096992        0.000000

$ tail -n 1 ~/.cache/stdpopsim/genetic_maps/HomSap/HapMapII_GRCh37/genetic_map_GRCh37_chr22.txt
[...]

Issue #540, opened 2020-05-26
QC for CanFam


grahamgower:
The **CanFam** species definition has not yet been QCed!

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [ ] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [ ] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [ ] Population size.
- [ ] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [ ] Genome reference assembly.
- [ ] Mutation rate.
- [ ] Recombination rate.
- [ ] Recombination map(s) (if relevant).
- [ ] Population size.
- [ ] Generation time.

grahamgower:
Is there a vollunteer to do some species QC on CanFam? I brought this species in, so I feel it would be inappropriate to do the QC. Just copy/paste the QC template from https://raw.githubusercontent.com/popsim-consortium/stdpopsim/master/.github/ISSUE_TEMPLATE/species-qc-issue-template.md into a new comment here, so that you have permission to push the check boxes.

David-Peede:
---
name: Species QC issue template
about: Quality control process for species addition
title: QC for CanFam
labels: Species QC
assignees: 'David Peede'

---

**PR for new species:** {link to Pull Request}

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
[...]

PR #885, opened 2021-04-09
Add values for recombination test for DroSec

percyfal:
WIP: I have added recombination values for DroSel based on Table 2 in [Comeron et al](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1002905). I don't get the same values as @jradrion which was probably a good thing because I realized the tests, despite a large difference for chromosome 4, *didn't* fail. It seems the tolerance (rel I think) needs to be set on pytest.approx. ping @jeromekelleher

### DIFF
--- a/tests/test_HelAnn.py
+++ b/tests/test_HelAnn.py
@@ -21,40 +21,37 @@ class TestSpeciesData(test_species.SpeciesTestBase):
     # independently referring to the citations provided in the
     # species definition, filling in the appropriate values
     # and deleting the pytest "skip" annotations.
-    @pytest.mark.skip("Population size QC not done yet")
     def test_qc_population_size(self):
-        assert self.species.population_size == -1
+        assert self.species.population_size == 673968
 
-    @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 1
 
 
 class TestGenomeData(test_species.GenomeTestBase):
 
     genome = stdpopsim.get_species("HelAnn").genome
 
-    @pytest.mark.skip("Recombination rate QC not done yet")
     @pytest.mark.parametrize(
         ["name", "rate"],
         {
-            "1": -1,
-            "2": -1,
-            "3": -1,
-            "4": -1,
-            "5": -1,
-            "6": -1,
-            "7": -1,
-            "8": -1,
-            "9": -1,
-            "10": -1,
-            "11": -1,
-            "12": -1,
-            "13": -1,
-            "14": -1,
-            "15": -1,
-            "16": -1,
-            "17": -1,
+            "1": 0.4e-8,
+            "2": 0.4e-8,
+            "3": 0.4e-8,
+            "4": 0.4e-8,
+            "5": 0.4e-8,
+            "6": 0.4e-8,
+            "7": 0.4e-8,
+            "8": 0.4e-8,
+            "9": 0.4e-8,
+            "10": 0.4e-8,
+            "11": 0.4e-8,
+            "12": 0.4e-8,
+            "13": 0.4e-8,
+            "14": 0.4e-8,
+            "15": 0.4e-8,
+            "16": 0.4e-8,
+            "17": 0.4e-8,
         }.items(),
     )
     def test_recombination_rate(self, name, rate):
@@ -62,27 +59,26 @@ class TestGenomeData(test_species.GenomeTestBase):
             self.genome.get_chromosome(name).recombination_rate
         )
 
-    @pytest.mark.skip("Mutation rate QC not done yet")
     @pytest.mark.parametrize(
         ["name", "rate"],
         {
-            "1": -1,
-            "2": -1,
-            "3": -1,
-            "4": -1,
-            "5": -1,
-            "6": -1,
-            "7": -1,
-            "8": -1,
-            "9": -1,
-            "10": -1,
-            "11": -1,
-            "12": -1,
-            "13": -1,
-            "14": -1,
-            "15": -1,
-            "16": -1,
-            "17": -1,
+            "1": 6.1e-9,
+            "2": 6.1e-9,
+            "3": 6.1e-9,
+            "4": 6.1e-9,
+            "5": 6.1e-9,
+            "6": 6.1e-9,
+            "7": 6.1e-9,
+            "8": 6.1e-9,
+            "9": 6.1e-9,
+            "10": 6.1e-9,
+            "11": 6.1e-9,
+            "12": 6.1e-9,
+            "13": 6.1e-9,
+            "14": 6.1e-9,
+            "15": 6.1e-9,
+            "16": 6.1e-9,
+            "17": 6.1e-9,
         }.items(),
     )
     def test_mutation_rate(self, name, rate):


percyfal:
I should add that I set rel to 1e-9 manually to make sure the tests fail in this case.

jeromekelleher:
Fantastic, QC process is working!

@jradrion, any thoughts here?

jradrion:
I think there are a few things going on here and I am going to look into them in ASAP. 

1) IIRC the mean chromosomal rates currently implemented for DroMel in `/DroMel/species.py` were taken directly from the map itself (ftp://ftp.flybase.org/flybase/associated_files/Comeron.2012.10.15.xlsx) and not as the means between crosses from Table 2 in Comeron et al 2012. I directly copied the rates from `/DroMel/species.py` over to `/DroSec/species.py`. 

2) I am not sure the mean chromosomal rates for DroMel found in `/DroMel/species.py` were ever corrected since the most recent map liftover to dm6 (the Comeron map itself is in dm3 and was lifted over to dm6 using the maintenance script `liftover_comeron2012.py`). I will revisit this.

3) @jeromekelleher suggested that perhaps we should just use the numbers from Table 2 for the chromosomal averages instead of taking them from the liftover map. This is fine with me, it just means that we will need to change the mean chromosomal rates in `/DroMel/species.py` and re-QC to make sure that they are correct. I can then correct them for DroSec.

jeromekelleher:
@andrewkern any thoughts here on point (3) - given we have summary estimates in the paper, maybe we should use these directly rather than recomputing from the map?

andrewkern:
interesting. are the current chromosome average rates calculated from the liftedOverMap @jradrion? that would be my preference I think

jradrion:
@andrewkern I had assumed so, but the current chromosome averages aren't an exact match (although they are close) with what I'm getting from either the older or new liftover map, so I don't know where those numbers actually came from. I tried looking through the commit history of that file, but I can't seen anything prior to the 7th of March. Personally, I like the idea of sticking to the numbers from Table 2 for the chromosome averages, since we currently have an older (elife ms version) and newer map with very slightly different numbers. Seems less complicated that way.

jradrion:
It looks like @percyfal divided the average map lengths from Table 2 by the DroSec chromosome lengths. If we use the numbers from Table 2 I think we need to scale by DroMel dm6 only. That will give us the per-base rates and we just use those same rates for DroSec. Does that seem right?

jradrion:
I'll be in the new remo link this afternoon if anyone wants to talk it over there.

jradrion:
After talking with @andrewkern it sounds like we want to stick to using the averages from the actual map file and not the map lengths from Table 2. #897 now has the average rates recalculated from "ComeronCrossoverV2_dm6". I am currently using these same rates for DroSec as well.

petrelharp:
Do you agree with this solution, @percyfal? If so, want to update the PR?

And, good catch on the `approx`. I'm opening an issue for that.

percyfal:
@petrelharp Yes that sounds good to me. I'm planning on getting back to the PR later this week (probably won't have time today). I think @jeromekelleher has opened an issue already? cf #884

percyfal:
Having little experience of liftovers, I naively started out by using
Bioconductor's liftover function to independently gauge the rates. It
turns out for some chromosomes (in particular X) the number of
returned intervals differs substantially from the UCSC implementation.
Consequently, the rates were also off.

However, even when using UCSC liftover results, and the maintenance
script, I cannot reproduce the values, or at least I'm not sure what
values were used. I feel I'm missing something obvious here. For instance, printing the chromosome name and `new_chrom_average` value at the end of bed_to_hapmap gives
2.996015709213288 for chrX. @jradrion could you verify/refute that this is indeed the intended rate? 

The values that I currently use in the test are based on reading the
generated comeron_map tables and calculating weighted means (by
interval length) of the rates.

grahamgower:
Hi @percyfal. There has been some discussion of how to deal with telomeres or zero-rate regions in other species (see #630 and #799). Perhaps the differences are (partially) accounted for by whether zero-rate regions are given any weight in the weighted mean calculation?

percyfal:
@grahamgower thanks for the feedback, that could very well be the case. I'll try it out on some of the chromosomes that show larger discrepancies (notably 2R and 3R), based on the information in #540 and the issues you referred to.

jradrion:
Hi @percyfal, thanks for looking into this. I don't know anything about how bioconductor implements liftover. Is it using the UCSC chain files to do the remapping? I think it's worth considering that there are different ways to do a liftover, but I'm partial to sticking with the UCSC tool and the USCS chain files. I'm open to other opinions, and it might be a good idea to compare different liftovers to see how close to one another they are.

With respect to the maintenance script, I think I see what's going on here. Similar to what @grahamgower mentioned, the issue is that `new_chrom_avg` from `liftover.py` does not account for newly introduced gap intervals in the final lifted-over coordinates. `new_chrom_avg` is the weighted average of the raw intervals that were able to be lifted over between assemblies. The maintenance script takes any new gaps that were introduced during liftover and fills them by one of two ways, either with `new_chrom_avg` directly or by taking the average of the intervals adjacent to the gap. The latter is what we ended up doing for the liftover maps that were uploaded to AWS. Additionally, gaps larger than the specified value for `--gapThresh` are filled with rate zero, and an interval of rate zero is assumed for the start of the chromosome if the file doesn't already start at position 0. 

Ultimately, we need to decide what intervals should be used to calculate the chromosome averages found is `species.py`.

If we just take chrX as an example, the result is that `new_chrom_avg = 2.999964904556056`, with 221 intervals lifted over. However 3 gaps were introduced during liftover (and the start interval is assumed to have rate zero) and the filling of those gaps results in 224 intervals included in the final hapmap file. Those gaps were filled by taking the average of the adjacent windows, so the resulting weighted average for the 224 intervals (plus the interval from position 0, so 225 intervals in total) was 2.8899257342235853. This number is different from my recent upload of 2.89650687913, but that is because I mixed up whether the position column in the hapmap formatted file was the start or end of the interval during the hackathon, so my interval weighting was off. Those numbers will need to be corrected in a new PR. Thanks for helping me catch this.

At the end of the day, the chromosome averages I'm getting for are as follows (all in cM/Mb):
chrX = 2.8899257342235853
chr2L = 2.4048709303512266
chr2R = 2.234585953584805
chr3L = 1.795563724129453
chr3R = 1.7164204577707685

Can you check to see if we are getting similar results using the resulting lifted-over map from the maintenance script?

Perhaps we should also round these numbers when we use them in the `species.py` file?

jradrion:
This still doesn't match your value of 2.996015709213288 for `new_chrom_avg` for chrX. We might be using different commands? I am using the following: `python liftOver_comeron2012.py --winLen 1000 --gapThresh 1000000 --useAdjacentAvg --retainIntermediates`.

jradrion:
Also I'm curious about your Bioconductor liftover. Were you lifting over the Comeron map between dm3 and dm6?

izabelcavassim:
Trying to get this finished but it seems like a lot of discussion and thoughts have been given into this. 
My parsimonious suggestion is similar to @jeromekelleher opinion back in 2021 that we just use the recombination averages from Cameron 2010 and explicitly describe what table we have used to extract that information. 
If we decided to do the whole leftover approach then we need to be even more descriptive about how we reached out our values. This will be applied to Melanogaster and DroSec since we are assuming that these species have the same recombination rate. 
@petrelharp @percyfal @andrewkern

percyfal:
@izabelcavassim that sounds fine to me. I would need to put in some effort to revisit the liftover results.

PR #592, opened 2020-10-01
DroMel: add assembly ref, and use recomb rates from Comeron map.

grahamgower:
Part of the long-overdue species QC (#541).

PR #897, opened 2021-04-09
Recalculate DroMel average chromosome rate

jradrion:
- Recalculated the average chromosome recombination rate for DroMel based off of the newer liftover map, "ComeronCrossoverV2_dm6". 
- Added a description to explain how this was done and which map was used. 
- Changed the rates in DroSec to reflect these changes.

### DIFF
--- a/stdpopsim/catalog/DroMel/species.py
+++ b/stdpopsim/catalog/DroMel/species.py
@@ -35,16 +35,20 @@ _ComeronEtAl = stdpopsim.Citation(
     year=2012,
 )
 
-# Mean chromosomal rates, calculated from the Comeron 2012 map.
-# Chromosome 4 isn't in this map, so the weighted mean of 2L, 2R, 3L and 3R
-# was used instead.
+# Mean chromosomal rates, calculated by taking the
+# average of all rates in the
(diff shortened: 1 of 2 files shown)


jradrion:
The changes are so minuscule that they might not be worth merging....just wanted to put them up so that anyone QCing DroSec would know where these rates came from.

andrewkern (approved):
LGTM Jeff

grahamgower:
> The changes are so minuscule that they might not be worth merging....just wanted to put them up so that anyone QCing DroSec would know where these rates came from.

Ah, sorry @jradrion, I should have been much more explicit when I updated the DroMel values a couple of days ago. Given the recent changes in both stdpopsim (the lift over maps) and msprime (RecombinationMap -> RateMap -> including NaNs now), there are a lot of different ways one might choose to calculate average rates. I did see there were very small differences between the calculated rates with the most recent everything, and as you say here the differences are rather small. Finally, I figured there will be further drift from these averages whenever there's a new liftover or further curation of the maps (when the conversion to zarr happens), and no egregious problems would result from using the values that were already in the months old #592 PR.

PR #898, opened 2021-04-09
Myotis lucifugus genome scaffolds

lauterbur:
Myotis lucifugus just has scaffolds, no chromosome-scale assembly yet
Change the maximum file size limit to 1000kb from 500kb because genome_data.py gets big with lots of scaffolds


stdpopsim/catalog/MyoLuc/species.py
```diff
@@ -0,0 +1,95 @@
...
+_genome = stdpopsim.Genome(
+    chromosomes=_chromosomes,
+    citations=[_LindbladTohEtAl, _DumontAndPayseur, _RayEtAl],
+)
+
+_species = stdpopsim.Species(
+    id="MyoLuc",
+    ensembl_id="myotis_lucifugus",
+    name="Myotis lucifugus",
+    common_name="Microbat",
```
petrelharp, line 87: I think it's "Little brown bat" (and is a type of microbat).

stdpopsim/catalog/MyoLuc/species.py
```diff
@@ -0,0 +1,95 @@
...
+
+# mutation rate:
+# Ray et al. 2008 estimate 2.366 × 10^−9 per site per year
+# Humphrey and Cope 1976, Frick et al. 2010 state generation time of 2 years
+_overall_mutation_rate = 0.000000004732
+
+# no recombination rate estimate available, use "standard" mammalian rate 10^-8
+# similar to ~1cM/Mb for horses and cows
+# (bats closest relatives) from Dumont and Payseur 2008
+_overall_recombination_rate = 0.00000001
```
petrelharp, line 64: `1e-8` is easier to read?

stdpopsim/catalog/MyoLuc/species.py
lauterbur: Definitely, thanks. Don't know why I let ensembl talk me into that one when I know better

petrelharp:
Looks like we should take a different approach to the `test_MyoLuc.py` file (that doesn't make a file that's so big).  Can you edit that to be smaller? It should be easy in this case because they all have the same rate. (except maybe the mt, is that in there?)

More generally: I was not at first convinced we wanted to add species with a ton of contigs. But - especially once you think about simulating with selection on an annotation - there's still lots of things you can do with an assembly like this. Simulation won't get the long-range LD... but on the other hand, it could be *quicker* than simulating whole chromosomes (and is like what people do already, breaking up a chromosome into chunks). So, after discussing with @lauterbur I was convinced it's a good idea.

But, do we need to not put all the contig info in this file itself, and either host it on AWS or pull it from ensembl? Since it's so big?

lauterbur:
I think limiting the scaffold size to some lower bound would help with file size a bit. There's a certain size point at which a scaffold is unlikely to have anything annotated there, for example.

It will probably be the case that most scaffold-level genomes will use a constant recombination rate across scaffolds (and maybe mutation rate too). Except mt, I don't see that in there. Will look at making test_MyoLuc.py smaller, good point.

jeromekelleher:
I'm not at all convinced we want to put this information into stdpopsim - my take is that we should probably limit ourselves to species that have pretty well characterised genomes. That's still extremely ambitions, IMO...

grahamgower:
> I'm not at all convinced we want to put this information into stdpopsim 

Agreed.

> my take is that we should probably limit ourselves to species that have pretty well characterised genomes.

I'm on the fence here. Not everyone wants or needs chromosome length simulations. In particular, it can be useful to simulate fixed contig lengths with a uniform recombination rate. One could argue that the species doesn't need to be in stdpopsim to do such a simulation. While I agree that the scaffold lengths are unlikely to be useful, there's potentially lots of other advantages to be gained from the species inclusion in stdpopsim, over and above the chromosome/scaffold lengths. E.g. for simulating selection. Or for a fixed length contig, have the mutation rate and/or recombination drawn from an empirically-derived distribution.

So maybe for species inclusion we should consider: what is to be gained (for the contributor) by including the species in stdpopsim?

lauterbur:
My thinking is that there are people already doing (or wanting to do) such simulations on non-model species that have decent scaffold-level genomes, but not chromosome-level genomes. 

@docmanny was the catalyst for adding a Myotis bat and can chime in here. We chose MyoLuc because it's one of the best studied bat species (because of white nose syndrome). Despite the current state of the genome, there have been a number of recent studies on white nose syndrome adaptation that would have benefited from a standardized way of simulating variation on the scaffolds, some of which have annotated putative adaptive loci.

I'm probably showing my inexperience with stdpopsim, but what is the downside to including species like this (not specifically MyoLuc), if the contributor would find it useful?

jeromekelleher:
> I'm probably showing my inexperience with stdpopsim, but what is the downside to including species like this (not specifically MyoLuc), if the contributor would find it useful?

There's a significant maintenance burden involved for every species we add, and this is significantly multiplied if it's "weird" in any way. In this case, I'm pretty sure that having thousands of scaffolds will break the CLI (we provide all the options for -c in the help text) and probably numerous other little things where we assumed the chromosomes would be, well, chromosomes. We'll also massively increase the size of the code, which will make the repo much larger and unwieldy, the code package much larger and also likely substantially increase load times.

So, I'm afraid there are quite a few downsides here... I guess we could hook up dynamic links to NCBI or whatever and fix the various problems with assuming chromosomes, but this will be another huge refactor, and we've just gone through one to enable all the stuff we're doing at the moment.

petrelharp:
I do think we'll want to allow such species eventually, since if stdpopsim is useful outside of humans & drosophila, then it'd be useful for the great many spcies without chromosome-length assemblies.  Note thaa MyoLuc does have an annotation, so the various applications of simulating with selection could apply here, when that works. Currently, though, maybe all this would be doing is making a somewhat roundabout way to simulate a whole bunch of genome sequences with the same lengths as the contigs - so I agree that putting it off may be the right thing to do.  But - maybe I'm missing something - what would stdpopsim help a lot with, currently?

docmanny:
I have two pressing use cases for MyoLuc in StdPopSim: 
- I would like to simulate genetic variability in different populations of lucifugus to justify sample sizes of individuals and populations, both for my sake as well as at reviewer's requests. 
- The different Myotis species are theorized to have migrated over the Bering Strait, and speciated as they spread across the Americas. There's several open questions about which demographic models fit best with the observed variation in Myotis lucifugus that could be tested with StdPopSim. 

While those are low-hanging fruits of things to want to do, without a MyoLuc implementation in StdPopSim, I can't really explore the possibilities of things that we can do with it. 

I agree that it becomes an issue if too many scaffold-level chromosomes are added to the code base; however, without any scaffold-level chromosomes, StdPopSim won't be of use to the vast majority of biologists. Since this is the first batch of new species being added, it might be worth adding them just to see what works and if anything breaks, but I think that ultimately, we'll need to seriously consider an even more modular approach for StdPopSim as more and more species are added to avoid a monolithic repo.

grahamgower:
> ... but I think that ultimately, we'll need to seriously consider an even more modular approach for StdPopSim as more and more species are added to avoid a monolithic repo

Yes, it seems we're bumping up against the design of "every species is a submodule" already. Just look at #912, which needs to duplicate a change in each species.py, for what is a rather simple feature addition.

petrelharp:
(update for others - we're going to chat next week about what exactly is the use case and how stdpopsim will help)

stdpopsim/catalog/MyoLuc/genome_data.py
```diff
@@ -0,0 +1,11661 @@
...
+        "AAPE02069950": {"length": 6637, "synonyms": []},
+        "AAPE02069951": {"length": 6636, "synonyms": []},
+        "AAPE02069952": {"length": 6635, "synonyms": []},
+        "AAPE02069953": {"length": 6634, "synonyms": []},
+        "AAPE02069954": {"length": 6634, "synonyms": []},
+        "AAPE02069955": {"length": 6633, "synonyms": []},
+        "AAPE02069956": {"length": 6633, "synonyms": []},
+        "AAPE02069957": {"length": 6632, "synonyms": []},
+        "AAPE02069958": {"length": 6631, "synonyms": []},
+        "AAPE02069959": {"length": 6631, "synonyms": []},
```
grahamgower, line 8941: Huh, weird. Github got to line 8938 and then said: no more syntax highlighting for you!

petrelharp:
I think we've decided that (a) this would be a great thing to do eventually, but (b) it at least depends on some adjustments to the code base. What do you think, @lauterbur - shall we close this for now (but not delete the branch)?

PR #906, opened 2021-04-10
QC for AnoGam - closes #859

### LINKED ISSUE #859: QC for AnoGam
**PR for new species:** #856 

refs used for the parameters/assembly follow:
https://doi.org/10.1186/gb-2007-8-1-r5
https://doi.org/10.4269/ajtmh.2006.75.901
https://doi.org/10.1038/nature24995

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [ ] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [ ] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [ ] Population size.
- [ ] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [ ] Genome reference assembly.
- [ ] Mutation rate.
- [ ] Recombination rate.
- [ ] Recombination map(s) (if relevant).
- [ ] Population size.
- [ ] Generation time.

petrelharp:
There's a number of discrepancies that we need to work out yet, @andrewkern...

### DIFF
--- a/stdpopsim/catalog/AnoGam/species.py
+++ b/stdpopsim/catalog/AnoGam/species.py
@@ -21,6 +21,14 @@ _PombiEtAl = stdpopsim.Citation(
     reasons={stdpopsim.CiteReason.REC_RATE},
 )
 
+_ZhengEtAl = stdpopsim.Citation(
+    # there's no DOI?!?!
+    doi="https://www.genetics.org/content/143/2/941",
+    year=1996,
+    author="Zheng et al.",
+    reasons={stdpopsim.CiteReason.REC_RATE},
+)
+
 _SharakhovaEtAl = stdpopsim.Citation(
     doi="https://doi.org/10.1186/gb-2007-8-1-r5",
     year=2006,
@@ -28,17 +36,30 @@ _SharakhovaEtAl = stdpopsim.Citation(
     reasons={stdpopsim.CiteReason.ASSEMBLY},
 )
 
+_KeightleyEtAl = stdpopsim.Citation(
+    doi="https://doi.org/10.1101/gr.091231.109",
+    year=2009,
+    author="Keightley et al.",
+    reasons={stdpopsim.CiteReason.MUT_RATE},
+)
+
+# values are in Pombi et al, but except for the X,
+# they refer to Zheng et al for the data
+# and have no data for 2L so we set equal to 2R
+# the value for X is the average from Table 1 in Zheng
 _recombination_rate = {
-    "2L": 0,  # setting to zero because of inversion
+    "2L": 1.3e-8,
     "2R": 1.3e-8,
-    "3L": 1.6e-8,
-    "3R": 1.3e-8,
-    "X": 1e-8,
+    "3L": 1.3e-8,
+    "3R": 1.6e-8,
+    "X": 2.04e-8,
     "Mt": 0,
 }
 
 
-_overall_rate = 5.49e-9
+# the rate for DroMel as inferred by Keightley et al 2009,
+# which was used for the stairwayplot demographic inference
+_overall_rate = 3.5e-9
 _mutation_rate = {
     "2L": _overall_rate,
     "2R": _overall_rate,
@@ -56,6 +77,8 @@ _genome = stdpopsim.Genome.from_data(
         _SharakhovaEtAl,
         _Ag1000G,
         _PombiEtAl,
+        _KeightleyEtAl,
+        _ZhengEtAl,
     ],
 )
 stdpopsim.utils.append_common_synonyms(_genome)
@@ -67,7 +90,8 @@ _species = stdpopsim.Species(
     common_name="Anopheles gambiae",
     genome=_genome,
     generation_time=1 / 11,
-    population_size=6e6,  # Ghana population
+    # based on theta = 4 Ne u in Gabon population, rounded
+    population_size=1e6,
     citations=[_Ag1000G],
 )
 
--- a/tests/test_AnoGam.py
+++ b/tests/test_AnoGam.py
@@ -17,37 +17,52 @@ class TestSpeciesData(test_species.SpeciesTestBase):
     def test_common_name(self):
         assert self.species.common_name == "Anopheles gambiae"
 
-    # QC Tests. These tests are performed by another contributor
-    # independently referring to the citations provided in the
-    # species definition, filling in the appropriate values
-    # and deleting the pytest "skip" annotations.
-    @pytest.mark.skip("Population size QC not done yet")
     def test_qc_population_size(self):
-        assert self.species.population_size == -1
+        # based on theta = 4 Ne u since diversity in most pops is 1.5%
+        assert self.species.population_size == pytest.approx(
+            0.015 / (4 * 3.5e-9), rel=0.1
+        )
 
-    @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 1 / 11
 
 
 class TestGenomeData(test_species.GenomeTestBase):
 
     genome = stdpopsim.get_species("AnoGam").genome
+    mut_rate = 3.5e-9
 
-    @pytest.mark.skip("Recombination rate QC not done yet")
+    # these values are given in the text of Pombi et al
+    # except no single value is given for the X; the number
+    # given is the average across chromosome from Pombi's Table 1
+    # No value is given for 2L, because of the inversion,
+    # so we set it equal to 2R.
     @pytest.mark.parametrize(
         ["name", "rate"],
-        {"2L": -1, "2R": -1, "3L": -1, "3R": -1, "X": -1, "Mt": -1}.items(),
+        {
+            "2L": 1.3e-8,
+            "2R": 1.3e-8,
+            "3L": 1.3e-8,
+            "3R": 1.6e-8,
+            "X": round(41.379 / (22.104923 - 1.825909) * 1e-8, 10),
+            "Mt": 0,
+        }.items(),
     )
     def test_recombination_rate(self, name, rate):
         assert rate == pytest.approx(
             self.genome.get_chromosome(name).recombination_rate
         )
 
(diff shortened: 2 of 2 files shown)


tests/test_AnoGam.py
```diff
@@ -17,35 +17,46 @@ def test_name(self):
...
 
-    # QC Tests. These tests are performed by another contributor
-    # independently referring to the citations provided in the
-    # species definition, filling in the appropriate values
-    # and deleting the pytest "skip" annotations.
-    @pytest.mark.skip("Population size QC not done yet")
     def test_qc_population_size(self):
-        assert self.species.population_size == -1
+        # based on theta = 4 Ne u since diversity in most pops is 1.5%
+        assert pytest.approx(self.species.population_size, 0.015 / 3.5e-9)
```
petrelharp, line 22: I'm very confused on this, because the code says the pop size of 6e6 is from the "Ghana population", but there's very few references to Ghana in the paper (one colony used for 'calibration and validation' is from Ghana, but there didn't seem to be demographic inference done on that one?). This number is what I would have put in, based on nucleotide diversity?

tests/test_AnoGam.py
```diff
@@ -17,35 +17,46 @@ def test_name(self):
...
-    @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 1 / 11
 
 
 class TestGenomeData(test_species.GenomeTestBase):
 
     genome = stdpopsim.get_species("AnoGam").genome
+    mut_rate = 3.5e-9
```
petrelharp, line 31: The paper provided lists three mutation rates from Drosophila: 2.5e-9, 3.5e-9, and 5.5e-9. The number in the code is 5.49e-9, which doesn't appear in the Ag1000g paper, but I assume is a more precise number from Schrider et al, the source of the third number. Based on only the Ag1000g paper I would have chosen 3.5e-9, since (a) it's the one in the middle, and (b) it's the number which is used to scale some of the plots. But, maybe the third one is a better estimate? If so, we should say something about why we're choosing that one. And, add the citation to the original Drosophila paper that we get the estimate from.

tests/test_AnoGam.py
```diff
@@ -17,35 +17,46 @@ def test_name(self):
...
+    recomb_rate = 1.4e-8
```
petrelharp, line 32: whoops ignore this

tests/test_AnoGam.py
```diff
@@ -17,35 +17,46 @@ def test_name(self):
...
 
-    @pytest.mark.skip("Recombination rate QC not done yet")
+    # obtained as total lengths of maps at
+    # ftp://ngs.sanger.ac.uk/production/ag1000g/phase1/AR3.1/recombination_maps/
     @pytest.mark.parametrize(
         ["name", "rate"],
-        {"2L": -1, "2R": -1, "3L": -1, "3R": -1, "X": -1, "Mt": -1}.items(),
+        {
```
petrelharp, line 38: I'm not sure where the recombination rates in the code come from either, other than the 0 rate on 2L? I couldn't find them in the paper. The numbers I got below from the average across the AR3.1-released maps, but they don't match.

andrewkern: got the rates from here:

image

tests/test_AnoGam.py
andrewkern: yes 5.49e-9 is what we used for the demographic analysis in the paper I believe. @dschride does that sound right to you?

tests/test_AnoGam.py
andrewkern: Ugh typo here-- this should be "Gabon" not "Ghana" 

As for population size- I have no clue where that number came from now. I must have gotten it from somewhere... that nucleotide diversity number above is approx the average across species- should we just pin everything on the Gabon sample / demographic history? That is the GAS model I implemented already.

tests/test_AnoGam.py
petrelharp: The main demographic analysis gives ranges of times, corresponding to the range of mutation rates. I didn't find any demograhpic inference that used only 5.5e-9. But, maybe I missed it? Is that the mutation rate for  the times for the AnoGam demographic model you've put in?

tests/test_AnoGam.py
petrelharp: hah, right, sorry. was looking at the wrong paper.

tests/test_AnoGam.py
petrelharp: Computing from theta = 4 Ne u and the species mutation rate for Gabon sounds good to me!

tests/test_AnoGam.py
andrewkern: yep. IIRC that's the rate for all of the Stairwayplot runs we did (1 for each population)

andrewkern: crap this is wrong! we used 3.5e-9 
image

so that should be edited on the QC PR

tests/test_AnoGam.py
andrewkern: actually 0.015 looks about correct for \pi. this is from the Phase 2 paper: 
image

petrelharp: That gets an answer of 1,000,000 if I did my factors of 4 right.

tests/test_AnoGam.py
petrelharp: Hm - but, those numbers are for 3L, 3R, and 2R, and are actually from Zheng et al. Pombi et al report a map for the X but don't give an overall rate. Computing an overall rate from the endpoints of their table 1 gets a rate of about 2cM/Mb, not 1cM/Mb like we've got here. But, they say that the X has possibly more weird stuff than 2L, which you've already set to 0 because of the inversion. (I'm not sure setting to 0 is the right thing, honestly...)  Wouldn't it be better to use endpoints of the genetic map that was released by Ag1000g?

andrewkern: i think these are the numbers from this study, not Zheng et al? We can use the ag1000g map though, yes

petrelharp: well, it says "based on the data of Zheng and others" - this study only collected data on the X.

Happy to go with whatever you want - maybe could you put in the numbers you like so I can QC them? Either push a commit to this PR or just post them in comments?

tests/test_AnoGam.py
petrelharp: perhaps you can double-check?

tests/test_AnoGam.py
andrewkern: I gave the Zheng et al. and the Pombi et al papers a second read. Zheng et al. is a linkage map from the pre-genome era, so there are no estimates of physical distance there. What _seems_ to be going on is that Pombi et al, with a genome in hand, are now able to combine the physical and genetic maps to get these numbers in units of cM/Mb. Indeed in doing so they are able to confirm some of the observations from Zheng et al about discordance between polytene band size and local recombination rate on the X. 

So- I think the original numbers in my PR were correct here and that adding the Zheng et al citation is important.

andrewkern (commented):
left a couple comments here. I was looking further at the Ag1000G recombination data and it's really coarse... ftp://ngs.sanger.ac.uk/production/ag1000g/phase2/AR1/recombination_maps/

not sure we want to use that versus what we have in the original PR

tests/test_AnoGam.py
petrelharp: Ok, sounds good. And, what to do about X (and, 2L)?

stdpopsim/catalog/AnoGam/species.py
```diff
@@ -38,7 +55,9 @@
 }
 
 
-_overall_rate = 5.49e-9
+# the rate for DroMel as inferred by Keightley et al 2009,
+# which was used for the stairwayplot demographic inference
+_overall_rate = 3.5e-9
```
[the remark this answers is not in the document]
petrelharp: What's the disagreement?

tests/test_AnoGam.py
petrelharp: (andy checked this)

tests/test_AnoGam.py
petrelharp: We've decided to take the overall rough rate given in Pombi for X (1e-8) and the 2R rate for 2L (1.3.e-8)

petrelharp:
Whoops, @andrewkern - there is information in Pombi about the X - their Table 1 (below) lets us compute a chromosome-wide average. The number from this table is what I'd put in QC; want to double-check it?
Screenshot from 2021-04-27 10-16-13

izabelcavassim:
Hey @andrewkern, some work here to finish the QC for AnoGam. Can you do this? Thanks!

stdpopsim/catalog/AnoGam/species.py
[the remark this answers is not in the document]
petrelharp: ping @andrewkern

petrelharp:
Could you have a look here, @andrewkern?

andrewkern:
just looked through this with @petrelharp and while we say cryptic things in the paper wrt mutation rate ranges being used, `3.5e-9` is the correct number in this case

petrelharp:
Discussed with @andrewkern just now - this is ready to go!

Issue #902, opened 2021-04-09
Not all genomes have chromosome synonyms


apragsdale:
While updating some of the docs, realized that the tutorial section "Calculating the allele frequency spectrum" would not run as written, as "chr4" does not exist as a synonym for chromosome 4. We should probably include these basic synonyms for all species genomes.

jeromekelleher:
This is a function of the upstream data repos @apragsdale - no would be the short answer. 

They should exist for humans though - do the synonyms exist for the other chrs?

apragsdale:
I see. That's too bad - as far as I can see, it's only humans that have synonyms. This means that while we previously could specify chr2L for DroMel and chr4 for AraTha (as the docs currently do), these are errors now. I think it's worth adding these obvious synonyms by hand for all species.

jeromekelleher:
Well, feel free (but in a way that doesn't get clobbered by automatic updates from the repository).

apragsdale:
Closed by #912

PR #912, opened 2021-04-13
Synonyms

apragsdale:
An idea (WIP) to tackle #902. Would fill chromosome synonyms if the chrom.id is a common name (e.g. "1" -> "chr1", "2L" -> "chr2L", "MT" -> "chrM"). It's fairly automated, should be fairly universal, and protects against common synonym errors when ensembl doesn't fill that field.

### DIFF
--- a/stdpopsim/utils.py
+++ b/stdpopsim/utils.py
@@ -148,3 +148,25 @@ def mask_tree_sequence(ts, mask_intervals, exclude):
     else:
         ts = ts.delete_intervals(mask_intervals)
     return ts
+
+
+def append_common_synonyms(genome):
+    """
+    For common chromosome IDs, add their obvious synonyms if they do not exist already.
+    The input is a stdpopsim.Genome object, and synonyms are appended in place.
+    """
+
+    def add_if_unique(chrom, syn):
+        if syn not in chrom.synonyms:
+            chrom.synonyms.append(syn)
+
+    for chrom in genome.chromosomes:
+        # "1" -> "chr1" and "2L" -> "chr2L"
+        if chrom.id[0].isdigit():
+            add_if_unique(chrom, "chr" + chrom.id)
+        # sex chroms
+        if chrom.id in ["X", "Y", "W", "Z"]:
+            add_if_unique(chrom, "chr" + chrom.id)
+        # Mt
+        if chrom.id == "MT":
+            add_if_unique(chrom, "chr" + "M")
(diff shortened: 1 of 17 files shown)


stdpopsim/genomes.py
```diff
@@ -49,6 +50,7 @@ def from_data(genome_data, *, recombination_rate, mutation_rate, citations):
                     recombination_rate=recombination_rate[name],
                 )
             )
+        append_common_synonyms(chromosomes)
```
jeromekelleher, line 53: I don't think we should do it here - best to make a case-by-base choice when adding the species.

jeromekelleher (approved):
Very nice, I like it a lot. Non-invasive and simple. :+1:

stdpopsim/utils.py
```diff
@@ -148,3 +148,24 @@ def mask_tree_sequence(ts, mask_intervals, exclude):
...
+    """
+
+    def add_if_unique(chrom, syn):
+        if syn not in chrom.synonyms:
+            chrom.synonyms.append(syn)
+
+    for chrom in chromosomes:
+        # "1" -> "chr1" and "2L" -> "chr2L"
+        if chrom.id[0].isdigit():
+            add_if_unique(chrom, "chr" + chrom.id)
```
grahamgower, line 165: Do we need the reverse as well? I don't know if we'll get "chr"-prefixed names by default --- maybe from the ncbi genomes?
```python
        if chrom.id.startswith("chr"):
            add_if_unique(chrom, chrom.id[len("chr"):])
```

grahamgower (approved):
Looks good to me too!

stdpopsim/genomes.py
apragsdale: Ok, makes sense. I'll reformat a bit.

stdpopsim/utils.py
apragsdale: Hmm, I'm not sure about ncbi genomes. It seems like all the ensembl genomes have chromosome names without the chr prefix, at least from what I've seen and in a few other test cases I've pulled down. If that does come up at some point, we can update this function then to handle those cases?

stdpopsim/genomes.py
apragsdale: I moved it around a bit to append synonyms to a genome object, so that it doesn't affect the current default setup of the species.py file. Can then choose to either add or not to that file after calling `_genome = species.Genome.from_data(...)`.

grahamgower (approved):
Still LGTM.

PR #914, opened 2021-04-13
corrected pytest.approx usage (closes #913)

### LINKED ISSUE #913: pytest.approx( ) misused
Turns out that
```
assert pytest.approx(a, b)
```
is *always* True - the way you're supposed to use this is
```
assert a == pytest.approx(b)
```
and in the first call `b` is taken as the value of `rel`.

I'm going to fix this now.

petrelharp:
Whoopsie! Fortunately we just started using `approx`.

Can't say I'm a fan of `pytest.approx( )`'s syntax; it's kinda counterintuitive.

### DIFF
--- a/maintenance/main.py
+++ b/maintenance/main.py
@@ -140,14 +140,14 @@ class TestGenomeData(test_species.GenomeTestBase):
         ["name", "rate"],
         $chromosome_rate_dict.items())
     def test_recombination_rate(self, name, rate):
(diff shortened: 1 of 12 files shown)


petrelharp:
Darn it, looks like we've got some tests to fix up now...

jeromekelleher (approved):
LGTM!

:facepalm: :facepalm: :facepalm:

petrelharp:
I'm going to merge with three commits, as they do different things.

PR #916, opened 2021-04-13
cM/Mb = 1e-8 (closes #915)

### LINKED ISSUE #915: fix AedAeg recombination rates
As discovered in #914, actually 0.306 is not approximately 0.306e-8 - whoops - and the [recombination rate values](https://github.com/popsim-consortium/stdpopsim/blob/5d4ba60449b220bfd7001ef5c79ff563ef2215c0/stdpopsim/catalog/AedAeg/species.py#L9) for AedAeg are in cM/Mb, so need a `e-8` appended.

Here's the relevant table, to confirm:
Screenshot from 2021-04-13 03-53-09

petrelharp:
See #915.

### DIFF
--- a/stdpopsim/catalog/AedAeg/species.py
+++ b/stdpopsim/catalog/AedAeg/species.py
@@ -3,7 +3,7 @@ import stdpopsim
 from . import genome_data
 
 # These are in Table 1 of Juneja et al:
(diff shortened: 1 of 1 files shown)


petrelharp:
Actually, never mind - this is so minor I'm rolling it into #914.

PR #919, opened 2021-04-14
Catalog mutation rates

apragsdale:
Add mutation rates to the cataloged demographic models for each species and the parameter tables in docs.

I'm not sure the best way to do this, but each QC model will also need to have mutation rates added. It took me maybe a half hour to follow the doi in each demographic model to the source publication to find the mutation rate used in inference, and *most* were quite obvious. Perhaps someone would be willing to do the same for the QC models and commit to this same PR, so that tests will make sure that mutation rates match between models and their QCs? Any takers? :slightly_smiling_face:

### DIFF
--- a/docs/parameter_tables/HomSap/PapuansOutOfAfrica_10J19.csv
+++ b/docs/parameter_tables/HomSap/PapuansOutOfAfrica_10J19.csv
@@ -1,4 +1,3 @@
-Generation time (yrs.),29,Generation time
 Population size,"48,433",African pop. size
 Population size,"6,962",European pop. size
 Population size,"9,025",East Asian pop. size
@@ -51,3 +50,5 @@ Migration rate (x10^-4),5.72,(European + East Asian)--Papuan migration rate
 Migration rate (x10^-4),4.42,(European + East Asian)--Ghost migration rate
 Time (yrs.),"37,497",(European + East Asian)--Papuan migration begins
 Time (yrs.),"37,497",(European + East Asian)--Ghost migration begins
+Generation time (yrs.),29,Generation time
+Mutation rate,1.4e-8,Per-base per-generation mutation rate
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -5,6 +5,9 @@ import unittest
 import sys
 import textwrap
 import copy
+import csv
+import pathlib
+import os
 
 import numpy as np
 import msprime
@@ -168,6 +171,7 @@ class TestModelOutput:
             description="abc",
             long_description="ABC " * 50,
             generation_time=1234,
+            mutation_rate=888,
             model=msprime.Demography.isolated_model([1]),
         )
         s = str(model)
@@ -179,6 +183,7 @@ class TestModelOutput:
         ║                     ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC
         ║                     ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC ABC
         ║  generation_time  = 1234
+        ║  mutation_rate    = 888
         ║  citations        = []
         """  # noqa 501
         assert textwrap.dedent(expected) in s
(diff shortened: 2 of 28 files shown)


petrelharp:
Sure, I can do this.

petrelharp:
Unsurprisingly, the hardest-to-find one so far is in a Science paper.

docs/parameter_tables/HomSap/PapuansOutOfAfrica_10J19.csv
```diff
@@ -51,3 +50,5 @@ Migration rate (x10^-4),5.72,(European + East Asian)--Papuan migration rate
 Migration rate (x10^-4),4.42,(European + East Asian)--Ghost migration rate
 Time (yrs.),"37,497",(European + East Asian)--Papuan migration begins
 Time (yrs.),"37,497",(European + East Asian)--Ghost migration begins
+Generation time (yrs.),29,Generation time
+Mutation rate,1.25e-8,Per-base per-generation mutation rate
```
petrelharp, line 54: where'd you get this, @apragsdale ? Here's what I find: 
Screenshot from 2021-04-15 16-44-55
Screenshot from 2021-04-15 16-46-02

petrelharp:
One other question @apragsdale - I see you've put mutation rates in for the ZigZag_1S14 and AmericanAdmixture_4B11 models. I don't think these should have mutation rates specified, because the first is not inferred from data, and the second does not depend on mutation rate at all (I don't think?).  I put `None`, but I could imagine other ways to deal with it?

petrelharp:
Ok, I think I'm done here other than the questions above.

apragsdale:
> One other question @apragsdale - I see you've put mutation rates in for the ZigZag_1S14 and AmericanAdmixture_4B11 models. I don't think these should have mutation rates specified, because the first is not inferred from data, and the second does not depend on mutation rate at all (I don't think?). I put `None`, but I could imagine other ways to deal with it?

That makes sense to me! I just added the mutation rate in the Schiffels paper used in simulation, but you're right that since it's just a toy model that wasn't fit to data, it shouldn't be needed. Happy to go with `None`.

docs/parameter_tables/HomSap/PapuansOutOfAfrica_10J19.csv
apragsdale: Hmm, I see. Within the implementation, there was already a comment `# mutation_rate = 1.25e-8  # per gen per site, Scally & Durbin 2012`, and I had assumed that it was citing the mutation rate used in the Jacobs et al paper, and I didn't check the paper myself in that case. Good catch, and sorry for the confusion. Looks like the 1.4e-8 value is the correct one here.

apragsdale:
For the Browning model, it's a bit curious. It's the paper that presents the IBDNe method, and uses it on a number of populations for inference. But the model we have specified is their _simulation_ model that they used to test data, and wasn't actually inferred. It's based on the Gravel model which had u=2.35e-8, but they simulated under 1.25e-8. But important to note that the admixture fractions and size they picked for the additional admixed population were not actually inferred. Should we go with the original Gravel value (2.35e-8), their simulated value (1.25e-8), or `None`? Maybe `None` is the correct more conservative option here.
Screenshot from 2021-04-16 16-55-12

petrelharp:
Ah, that's odd. I guess I would vote for the Gravel rates, as we'll presumably get data that's closer to real data, which is the goal.

petrelharp:
I've changed the Browning rate to that of Gravel (and fixed the citation year and added Gravel).

petrelharp:
Ok, looks good? Check what I've done.

apragsdale:
Looks good! Just updated the American Admixture param table for catalog docs.

tests/test_models.py
```diff
@@ -361,3 +364,26 @@ def test_mutation_rate_match(self):
         self.assertNotEqual(model.mutation_rate, contig.mutation_rate)
         contig = species.get_contig(length=100, mutation_rate=model.mutation_rate)
         self.assertEqual(model.mutation_rate, contig.mutation_rate)
+
+    def test_params_match_docs_tables(self):
```
petrelharp, line 368: oh good idea!

apragsdale:
Ok great! Should I squash/rebase, or anything else to take care of here?

grahamgower (approved):
This LGTM. Happy to merge after a squash/rebase.

PR #945, opened 2021-05-25
minor fixes to selection_example.py because of API changes

mufernando:


docs/selection_example.py
```diff
@@ -188,7 +188,8 @@ def OutOfAfrica_3G09_with_DFE(seed):
 
     # neutral and deleterious mutations occur across the whole contig
     contig.add_genomic_element_type(
-        intervals=np.array([[0, int(contig.length)]]), **KimDFE()
+        intervals=np.array([[0, int(contig.recombination_map.sequence_length)]]),
```
grahamgower, line 191: Did the contig `.length` property get removed?

mufernando: yeah, but I think it needs to come back... @jeromekelleher didn't like that I defined it based off of the recombination map, and so if a contig didn't have a recombination map then it would return `None`. I don't know what the msprime engine does with a contig that doesn't have a rec map, so not sure how to deal with this? but I think `length` is a pretty useful property to have in the contig object...

petrelharp: Seems like we need to think through this - not to be fixed here.

petrelharp:
LGTM!

docs/selection_example.py
grahamgower: The contig should always have a recombination map.

grahamgower: See `Species.get_contig()`.

PR #949, opened 2021-05-25
start at selection tutorial

petrelharp:
I've moved the code from `docs/selection_example.py` over into the Tutorials section. It needs a lot of improvements.  But since we want some people to start using this stuff, maybe we should merge it as-is and improve?

To see these docs, check out this PR; go to the `docs/` directory and run `make`; then point your web browser at `docs/_build/html`.


grahamgower (approved):
LGTM!

petrelharp:
@mufernando is going to add something to this, and I'm hoping to take a pass through with some more explanation, so let's hold off on merging it for a minute.

petrelharp:
The annotation part of this is not working: see #952.

mufernando:
- [ ] remove annotation example
- [ ] add sweep example, conditioning of freq also

mufernando:
I've been trying to add a simpler example with the DrawMutation/etc API and I can't figure this out. 

First, I noticed the example we had was not working. This was because we were forgetting to add the mutation type to the contig. So I added a new method to easily add mutation types not tied to a genomic element type.

But still, the example under "selective sweeps" in the tutorial is not working. No mutation is placed at that site.

@grahamgower

grahamgower:
That's very strange. I'm fairly sure we have tests that check the frequency of the drawn allele after conditioning. Are these tests broken in some way that silently prevents them from running?

grahamgower:
I think what's going on here is that the allele is present in the population (at low frequency), but not present in the 10 sampled individuals.

And assuming you sample an individual with the drawn mutation, the following code overrides the mutation types for your example code, so the mutation's `selection_coeff` metadata is set to zero.
https://github.com/popsim-consortium/stdpopsim/blob/81bc0144bb144c83cfcd2393bd235872dbc6a257/stdpopsim/slim_engine.py#L1143-L1145

mufernando:
@petrelharp I added an example of a selective sweep. this looks good to me, but you said you wanted to do another pass.

One last thing we should showcase here is the ChangeMutationFitness, which can be used to simulate soft sweeps.

docs/tutorial.rst
```diff
@@ -869,6 +869,302 @@ of realistic size for real data.
...
+            # Save state before the mutation is introduced.
+            # This allows the simulation to restart from here if the mutation is lost.
+            save=True,
+        ),
+        # Condition on not loosing the mutation from the generation the mutation
+        # is added until the end. Note that our conditioning must start one
+        # generation after the mutation is placed. You cannot simply specify
+        # T_mut+1 if rescaling is present, otherwise the conditioning would start
+        # at the same generation when the mutation is placed.
+        stdpopsim.ext.ConditionOnAlleleFrequency(
```
petrelharp, line 1101: This conditioning is redundant - it can't have been lost if it's at frequency above 80% at the end. Is the purpose to restart the simulation earlier, if it's lost?

mufernando: yeah, the purpose is so that we don't wait until the end to check, instead we will be checking for presence every generation.

docs/tutorial.rst
```diff
@@ -869,6 +869,302 @@ of realistic size for real data.
...
+    T_mut = 1000
+    extended_events = [
+        # Draw the mutation
+        stdpopsim.ext.DrawMutation(
+            time=T_mut,
+            mutation_type_id=mut_id,
+            population_id=0,
+            coordinate=coordinate,
+            # Save state before the mutation is introduced.
+            # This allows the simulation to restart from here if the mutation is lost.
```
petrelharp, line 1093: how does it know to restart from this point? Does it just go back to the last saved point?

petrelharp:
Hm: I see the extended events are not yet documented.

And - I've split up the last example by adding more words. Maybe this is helpful?

docs/tutorial.rst
```diff
@@ -1078,46 +1083,78 @@ of at least 0.8.
...
-    # Simulate.
+
+Next,we condition on the mutation not being lost.
+Since in the next step we condition on the mutation being at 80% frequency at
+the end, this is redundant, but it allows the simulation to immediately restart
+from any generation in which the mutation is lost, rather than waiting until the end.
+Note that this conditioning must start one
+generation after the mutation is placed, for which we use
+``stdpopsim.ext.GenerationAfter(T_mut)``.
+We cannot simply specify ``T_mut - 1`` if rescaling is present,
```
mufernando, line 1115: it would be + 1 because this API has forward times

petrelharp: good catch - I meant to ask. If it has forwards times, though, why are "start time" and "end time" equal to 0 for the final conditioning?

grahamgower: The API here is confusing, sorry. Times have units of "generations ago", just like msprime. The start_time/end_time parameters for the `ChangeMutationFitness` and `ConditionOnAlleleFrequency` need `start_time => end_time`, which is opposite to what msprime uses (e.g in the demography debugger). I'm not sure why I did it that way.

There are various comments and docstring stubs in https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/ext/selection.py which explain some things.

petrelharp: Ah-ha, thanks. ... so, should it have been `T-1` or `T+1` there?

docs/tutorial.rst
```diff
@@ -869,6 +869,337 @@ of realistic size for real data.
...
+otherwise the conditioning would start
+at the same generation when the mutation is placed.
+
+.. code-block:: python
+
+    extended_events.append(
+        [
+            stdpopsim.ext.ConditionOnAlleleFrequency(
```
grahamgower, line 1123: This doesn't look right. I think this should be appending an element, not a list containing one element.

docs/tutorial.rst
```diff
@@ -869,6 +869,337 @@ of realistic size for real data.
...
+    )
+
+Finally, we condition on the mutation being above 80% at the end of the simulation.
+(The "end" is at time 0, since "time" is in generations before the end of the simulation.)
+
+.. code-block:: python
+
+    extended_events.append(
+        [
+            stdpopsim.ext.ConditionOnAlleleFrequency(
```
grahamgower, line 1141: Likewise here, I think it shouldn't be a list.

docs/tutorial.rst
```diff
@@ -869,6 +869,337 @@ of realistic size for real data.
...
+    plt.plot(sweep_pi, "r", label="sweep")
+    plt.legend()
+    plt.xlabel("Genomic window")
+    plt.ylabel("Diversity")
+    plt.show()
+
+.. image:: _static/tute-sweep.png
+    :width: 500px
+    :align: center
+    :alt: Plot with nucleotide diversity along the chromosome for simulations with a without a selective sweep.
```
grahamgower, line 1200: Cool! It works!

docs/tutorial.rst
grahamgower: `T-1` looks right to me.

mufernando: gee, sorry for the confusion!

docs/tutorial.rst
mufernando: I wanted sphinx to render the plot using matplotlib, but it looks like the page that describes how to do this hasn't been updated in many years and the instructions don't work anymore.

docs/tutorial.rst
mufernando: I think it goes back to the last saved point. @grahamgower

docs/tutorial.rst
petrelharp: if/when we switch to jupyterbook then this does work...

petrelharp:
Thanks - fixed those things (and actually ran the code =). I'll let you all merge it if you think it's ready?

docs/tutorial.rst
grahamgower: Yep, exactly.

grahamgower:
> Thanks - fixed those things (and actually ran the code =). I'll let you all merge it if you think it's ready?

Looks good. Maybe it should be squashed though?

mufernando (approved):
LGTM!

PR #950, opened 2021-05-29
Add per-chromosome mutation rates to C. reinhardtii and update refs

aays:
Apologies for the significant delay in following up here (the Growing the Zoo hackathon was a fair bit ago now...) but I've finally been able to both obtain as well as upload the per-chromosome mutation rates for C. reinhardtii to figshare! I've also uploaded both the mutation and recombination rate data to figshare as recommended during the hackathon. 

Technically covers a bit more of #828, though that issue was obviously closed after the merging of #863 back then.

### DIFF
--- a/stdpopsim/catalog/ChlRei/species.py
+++ b/stdpopsim/catalog/ChlRei/species.py
@@ -2,13 +2,13 @@ import stdpopsim
 
 from . import genome_data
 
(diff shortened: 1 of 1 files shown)


grahamgower (approved):
LGTM! Thanks @aays!

PR #953, opened 2021-06-01
fix zigzag parameter table

awohns:
During the fix to the population sizes in the ZigZag model (detailed in #745), the parameter table in the docs weren't updated to reflect that the ancient and recent population sizes should be 5x higher

### DIFF
--- a/docs/parameter_tables/HomSap/Zigzag_1S14.csv
+++ b/docs/parameter_tables/HomSap/Zigzag_1S14.csv
@@ -1,5 +1,5 @@
-Population size,"1,431",Ancient pop. size
-Population size,"14,312",Recent pop. size
+Population size,"7,156",Ancient pop. size
+Population size,"71,560",Recent pop. size
 Growth rate (per gen.),"8.99448 x 10^(-5)",Growth rate for 1st growth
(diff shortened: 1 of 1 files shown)


docs/parameter_tables/HomSap/Zigzag_1S14.csv
```diff
@@ -1,5 +1,5 @@
-Population size,"1,431",Ancient pop. size
-Population size,"14,312",Recent pop. size
+Population size,"7,155",Ancient pop. size
```
petrelharp, line 1: I think this [should be 7,156](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/catalog/HomSap/demographic_models.py#L645)

petrelharp (changes requested):

docs/parameter_tables/HomSap/Zigzag_1S14.csv
awohns: Ah, thanks. Was just looking at the previous values.

petrelharp (approved):
LGTM, thanks! off by 1 isn't a big deal, except it might confuse people in the future...

PR #959, opened 2021-06-15
Make contigs fully_neutral() by default.

grahamgower:
We should do this in the Contig class's __attrs_post_init__, but this requires lots of tests to be updated so I've deferred that for now.

Issue #391, opened 2020-01-29
References for BGS models


ckyriazis:
Here are some DFE inference papers to get the BGS modeling conversation started. Please feel free to add any that I have missed!

Keightley & Eyre Walker 2007 (humans and drosophila) https://www.genetics.org/content/177/4/2251

Boyko 2008 (humans) https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1000083

Torgerson 2009 (non coding loci in humans) https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1000592

Tataru 2017 (chimps) https://www.genetics.org/content/207/3/1103.abstract

Kim 2017 (humans) https://www.genetics.org/content/206/1/345.abstract

Huber 2017 (humans, mice, flies) https://www.pnas.org/content/114/17/4465

Booker & Keightley 2018 (mice) https://academic.oup.com/mbe/article/35/12/2971/5123518

Zhen 2018 (humans, mice, flies) https://www.biorxiv.org/content/10.1101/427583v1

grahamgower:
So the main things this needs are an exon model and a distribution of fitness effects? Is there anything else this requires?

The DFE should be straightforward to add into the slim engine. Is it expected that we'll use exon positions from a given species annotation, or have a parameterized exon model (i.e. generate exon positions from a distribution)? The former will require a framework in stdpopsim for importing and using annotations.

dschride:
I think we would eventually want the flexibility to take an annotation that has more than one sequence type (e.g. CDS vs. UTR vs. intron, etc, rather than just exon vs non-exon). Then each sequence type (GenomicElementType in SLiM) could have its own DFE (or mixture of DFEs, or MutationTypes in SLiM).

petrelharp:
I'm leaving this open because it has references to a bunch more DFEs in non-human species that it'd be nice to implement!

petrelharp:
Closing!

PR #960, opened 2021-06-16
Annot fixup

andrewkern:
okay here is a PR that fixes up the Annotation class to return non-overlapping intervals using @grahamgower's approach from the Analysis2 repo. 

basics points to this PR:

1. refactored the data type stored to be a numpy array of intervals. the `Annotation.get_chromosome_annotations()` returns this array for use with SLiM
2. the annotation_maint.py script now does the download of the GFF and the interval merge, before tarring stuff up for aws

things to do:
1. i don't know how fragile the `gff.source == "ensembl_havana"` requirement is. should we worry?
2. the way things are written now, species can have more than one annotation. it would be cool to extend this to other choices
3. add a front end for doing the download of genome annotations in the same way that we have for genome assemblies

### DIFF
--- a/maintenance/annotation_maint.py
+++ b/maintenance/annotation_maint.py
@@ -1,38 +1,125 @@
 import allel
-import zarr
+import operator
+import pathlib
 import numpy as np
-import stdpopsim as stp
 import logging
-import warnings
-import urllib.request
 import os
+import glob
+import tarfile
+import stdpopsim
 
 logger = logging.getLogger(__name__)
 # make root directory for zarr annotations
 annot_path = "annotations"
-os.mkdir(annot_path)
-# loop through species and download
-for spc in stp.all_species():
-    if spc.annotations:
-        address = spc.annotations[0].url
-        genome_version = os.path.basename(address).split(".")[1]
-        logger.info(f"Downloading GFF file {spc.id}")
-        tmp_path = f"{spc.id}.tmp.gff.gz"
-        try:
-            x, y = urllib.request.urlretrieve(address, tmp_path)
-        except FileNotFoundError:
-            warnings.warn("can't connnect to url")
-        logger.info(f"creating zarr arrays {spc.id}")
-        # create zarr store and zarr root
-        spc_path = os.path.join(annot_path, spc.id + "." + genome_version + ".zip")
-        store = zarr.ZipStore(spc_path)
-        root = zarr.group(store=store, overwrite=True)
-        x = allel.gff3_to_dataframe(tmp_path)
-        for col_name in x.columns:
-            if x[col_name].dtype == "O":
-                tmp = root.array(col_name, np.array(x[col_name], dtype=str))
+os.makedirs(annot_path, exist_ok=True)
+
+
+def merged(intervals, *, closed: bool):
+    """
+    Merge overlapping and adjacent intervals.
+
+    :param intervals: An iterable of (start, end) coordinates.
+    :param bool closed: If True, [start, end] coordinates are closed,
+        so [1, 2] and [3, 4] are adjacent intervals and will be merged.
+        If False, [start, end) coordinates are half-open,
+        so [1, 2) and [3, 4) are not adjacent and will not be merged.
+    """
+
+    def iter_merged(intervals, *, closed: bool):
+        """
+        Generate tuples of (start, end) coordinates for merged intervals.
+        """
+        intervals = sorted(intervals, key=operator.itemgetter(0))
+        if len(intervals) == 0:
+            return
+        start, end = intervals[0]
+        for a, b in intervals[1:]:
+            assert a <= b
+            if a > end + closed:
+                # No intersection with the current interval.
+                yield start, end
+                start, end = a, b
             else:
-                tmp = root.array(col_name, np.array(x[col_name]))
-        # cleanup
-        os.unlink(tmp_path)
-        store.close()
+                # Intersects, or is contiguous with, the current interval.
+                end = max(end, b)
+        yield start, end
+
+    return list(iter_merged(intervals, closed=closed))
+
+
+def gff_recarray_to_stdpopsim_intervals(gff):
+    """
+    Merge overlapping intervals and convert coordinates. GFF intervals are
+    1-based [i,j], but stdpopsim intervals are 0-based [i-1,j).
+    """
+    intervals = np.array(merged(zip(gff.start, gff.end), closed=True))
+    intervals[:, 0] = intervals[:, 0] - 1
+    return intervals
+
+
+def get_gff_recarray(url, sha256):
+    local_path = pathlib.Path(url).name
+
+    if not pathlib.Path(local_path).exists():
+        print(f"downloading {url}")
+        stdpopsim.utils.download(url, local_path)
+
+    print("checking sha256")
+    local_sha256 = stdpopsim.utils.sha256(local_path)
+    if local_sha256 != sha256:
+        print(
+            f"{local_path}: sha256: expected {sha256}, but found {local_sha256}. "
+            "Delete the file to download it again."
+        )
+        exit(1)
+
+    print(f"loading {local_path} into numpy recarray")
+    gff = allel.gff3_to_recarray(local_path)
+    return gff
+
+
+def make_tarfile(output_filename, source_dir, dest):
+    if os.path.exists(output_filename):
+        os.remove(output_filename)
+    with tarfile.open(output_filename, "w:gz") as tar:
+        tar.add(source_dir, arcname=dest)
+    tar.close()
+
+
+# loop through species and download
+for spc in stdpopsim.all_species():
+    if spc.annotations:
+        for an in spc.annotations:
+            CHROM_IDS = [chrom.id for chrom in spc.genome.chromosomes]
+            genome_version = os.path.basename(an.url).split(".")[1]
+            logger.info(f"Downloading GFF file {spc.id}")
+            tmp_path = f"{spc.id}.tmp.gff.gz"
+            gff = get_gff_recarray(an.url, an.gff_sha256)
+            logger.info("extracting exons {spc.id}")
+            # this is fragile-- is ensembl_havana always a feature?
+            exons = gff[
+                np.where(
+                    np.logical_and(gff.source == "ensembl_havana", gff.type == "exon")
+                )
+            ]
+            logger.info(f"merging overlapping regions {spc.id}")
+            # create zarr store and zarr root
+            spc_name_path = os.path.join(annot_path, spc.id)
+            os.makedirs(spc_name_path, exist_ok=True)
+            for chrom_id in CHROM_IDS:
+                chrom_exons = exons[np.where(exons.seqid == chrom_id)]
+                if len(chrom_exons) == 0:
+                    continue
+                intervals = gff_recarray_to_stdpopsim_intervals(chrom_exons)
+                # double check that the intervals can be used in stdpopsim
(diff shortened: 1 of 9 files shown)


maintenance/annotation_maint.py
```diff
@@ -1,38 +1,147 @@
...
+            logger.info(f"Downloading GFF file {spc.id}")
+            tmp_path = f"{spc.id}.tmp.gff.gz"
+            gff = get_gff_recarray(GFF_URL, GFF_SHA256)
+            logger.info("extracting exons {spc.id}")
+            # this is fragile-- is ensembl_havana always a feature?
+            exons = gff[
+                np.where(
+                    np.logical_and(gff.source == "ensembl_havana", gff.type == "exon")
+                )
+            ]
```
grahamgower, line 126: The ensembl_havana `source` [is only available for human, mouse, zebrafish and rat](https://m.ensembl.org/info/genome/genebuild/annotation_merge.html). But there's also `ensembl` and `havana` sources in thus gff, so we certainly should be judicious in our choice. I chose the ensembl_havana source in the (analysis2 repo)[https://github.com/popsim-consortium/analysis2] by guessing --- I have no clue what we actually want. Is it even possible to know what is going to be in a gff without opening it and looking?

I hate to say it, but I think we need to keep all the annotation info, so that BGS model implementers can choose this stuff downstream. Probably what we want the maintenance code to do is to split the gff up into smaller pieces, by chromosome, and possibly other fields, so that loading/parsing is quick for the user.

maintenance/annotation_maint.py
```diff
@@ -1,38 +1,147 @@
...
+                start, end = a, b
+            else:
+                # Intersects, or is contiguous with, the current interval.
+                end = max(end, b)
+        yield start, end
+
+    return list(iter_merged(intervals, closed=closed))
+
+
+def test_merged():
```
grahamgower, line 50: The test should go in `tests/test_maintenance.py`.

requirements/development.txt
```diff
@@ -24,3 +24,4 @@ numpy
 scikit-allel
 zarr>=2.4
 biopython
+boto3
```
grahamgower, line 27: What's this for? I didn't see it in the imports.

grahamgower (commented):
I'm not sure this is the direction we should take @andrewkern. We still don't know how the annotations are going to be used by the BGS models, so I think it's premature to do something concrete here where we choose which annotations folks can use. I think we really need to make some BGS models (for 2+ species), apply them to the appropriate annotations, then step back and look at how we create an API (both internal and external).

requirements/development.txt
andrewkern: ah thats for the AWS push script I wrote. still not sure how to share that script cause of passwords.

maintenance/annotation_maint.py
andrewkern: so while choosing this one gene model track -- ensemble_havana -- is fragile, i don't think there is any reason we want all the annotations. at most i reckon we want genes and conserved noncoding regions. CNSs will be challenging to get for many non-human species, so that leaves us with genes as the major locus of selection and I think that's good.

with respect to splitting the gff-- it seems quiet fast now to load and the files are small. humans will have more annotation than anything else so this is the slowest example we'll work with.

grahamgower: The discussion in #391 suggests we'll quickly outgrow just exons. I chose the ensembl_havana source and exon type as an example, fully expecting someone to come back and say: "we don't want that, we want `xyz` instead", or "we need `abc` as well". I really do think we want to let the BGS models guide us here.

grahamgower: One reason folks might want access to other annotation types/sources is to do simulation-based inference. If we lock ourselves to one or just a few things, this use case becomes difficult or impossible.

andrewkern: just had a look back at #391, and yeah I agaree with what is being said there. we've already got this handled i reckon though in the currently implementation of `Annotation`,  because species can be associated with more than one set of annotations. so it's simple enough for us to create a list of features like UTR5, UTR3, CDS, etc. then on the maintenance side add those as annotations and push them to AWS. 

right now, i see this single case of the human genome with only exons as a proof of concept that we can tweak going forward, but i think all the major bits are here.

mufernando: if we decide we want users to access as many annotation types as they want, then we will have to make interval operations in the user side of things -- at least merging and subtracting. I think also we will ultimately want this, but don't see why we can't just merge this and open follow up issues to expand?

jeromekelleher:
I'm with @grahamgower here - I think we should get something simple working in terms of simulations first before making a more general annotations API.

requirements/development.txt
andrewkern: this is for the `aws_push_annotations.py` script that i've written to deal with getting these info to AWS. problem that we haven't yet figured out is how to share those passwords securely and have the code on github

andrewkern:
so i think we're chasing our own tail here. if we merge this code then we have a good place to move forward for building the simulation API. we need intervals do do selection sims, this provides them in a general way.

we can nitpick the maintenance side later as to what sorts of annotations we provide.

jeromekelleher:
We had decided a few weeks ago to avoid this complexity and just make some simple gene annotations based on a RateMap that we pass around between ourselves, so that we can get the simulations going though. I guess we can see how much work it is to go in either direction from here, though.

maintenance/annotation_maint.py
andrewkern: i think `tests/test_annotation.py` is actually the right place

tests/test_annotations.py
```diff
@@ -26,33 +26,34 @@
 saved_urls = {}
 
 
-def setup_module():
-    destination = pathlib.Path("_test_cache/zipfiles/")
+def setUpModule():
```
grahamgower, line 29: The conventions for this in pytest are `setup_module()`/`teardown_module()`. https://docs.pytest.org/en/latest/how-to/xunit_setup.html?highlight=setup_module

tests/test_annotations.py
```diff
@@ -7,14 +7,14 @@
...
-import pandas
 import pytest
-
+import numpy as np
+import logging
 import stdpopsim
 from stdpopsim import utils
 import tests
-
+import unittest
```
grahamgower, line 16: I've recently [removed `unittest` imports so that we use pytest more extensively](https://github.com/popsim-consortium/stdpopsim/pull/959/commits/80470ca2008b6caa6b32de56b7c48b9e907f8c86). Pytest uses raw asserts, test classes don't need to inherit from `unittest.TestCase` anymore, and warnings are filtered using a class/method decorator. This leads to clearer tests, usually.

tests/test_annotations.py
```diff
@@ -130,15 +133,15 @@ def test_correct_url(self):
             # The destination file will be missing.
             with pytest.raises(FileNotFoundError):
                 an.download()
-        mocked_get.assert_called_once_with(an.zarr_url, filename=mock.ANY)
+        mocked_get.assert_called_once_with(an.intervals_url, filename=unittest.mock.ANY)
```
grahamgower, line 136: s/unittest.mock/mock/

tests/test_annotations.py
```diff
@@ -150,8 +153,8 @@ def test_download_over_cache(self):
         assert an.is_cached()
 
 
-@pytest.mark.xfail  # HomSap annotation not currently available
-class TestGetChromosomeAnnotations(tests.CacheReadingTest):
+# @pytest.mark.xfail  # HomSap annotation not currently available
+class TestGetChromosomeAnnotations(unittest.TestCase):
```
grahamgower, line 157: I think this really should inherit from `tests.CacheReadingTest`. (and unittest.TestCase isn't needed anymore)

maintenance/annotation_maint.py
grahamgower: Ok, but the test function can be removed from here now.

maintenance/annotation_maint.py
```diff
@@ -1,38 +1,147 @@
...
-        os.unlink(tmp_path)
-        store.close()
+        for an in spc.annotations:
+            GFF_URL = an.url
+            GFF_SHA256 = an.gff_sha256
+            CHROM_IDS = [chrom.id for chrom in spc.genome.chromosomes]
+            genome_version = os.path.basename(GFF_URL).split(".")[1]
+            logger.info(f"Downloading GFF file {spc.id}")
+            tmp_path = f"{spc.id}.tmp.gff.gz"
+            gff = get_gff_recarray(GFF_URL, GFF_SHA256)
```
grahamgower, line 119: I think the UPPERCASE variables aren't really needed here. Just replace the use of these variables with `an.url` and `an.gff_sha256`, etc.

grahamgower:
I'm not sure what's going on with the failing test `tests/test_cli.py::TestDryRun::test_dry_run_quiet`. That test checks that there's no output when the `-q` flag is used, so probably there's some warning or print statment leaking into the output. If you modify the test to print the output, I'm sure you'll narrow it down.

stdpopsim/catalog/HomSap/annotations.py
```diff
@@ -7,13 +7,15 @@
...
-        "gff3/homo_sapiens/Homo_sapiens.GRCh38.101.gff3.gz"
+        "ftp://ftp.ensembl.org/pub/release-104/"
+        "gff3/homo_sapiens/Homo_sapiens.GRCh38.104.gff3.gz"
     ),
-    zarr_url=(
-        "https://stdpopsim.s3-us-west-2.amazonaws.com/annotations/HomSap.GRCh38.zip"
+    gff_sha256="313ad46bd4af78b45b9f5d8407bbcbd3f87f4be0747060e84b3b5eb931530ec1",
+    intervals_url=(
+        "https://stdpopsim.s3-us-west-2.amazonaws.com/"
+        "annotations/HomSap/HomSap.GRCh38.tar.gz"
```
grahamgower, line 16: Should we keep the ensembl release number in this?

tests/test_annotations.py
```diff
@@ -130,15 +132,15 @@ def test_correct_url(self):
...
 
     def test_incorrect_url(self):
         an = AnnotationTestClass()
-        an.zarr_url = "http://asdfwersdf.com/foozip"
+        an.intervals_url = "http://asdfwersdf.com/foozip"
         with pytest.raises(OSError):
             an.download()
 
-    @pytest.mark.xfail  # HomSap annotation not currently available
+    # @pytest.mark.xfail  # HomSap annotation not currently available
```
grahamgower, line 143: If the test now passes, the xfail decorator can be removed. Ditto for test below.

tests/test_cli.py
```diff
@@ -818,9 +818,10 @@ def test_dry_run_quiet(self):
                 filename = path / "output.trees"
                 cmd = (
                     f"{sys.executable} -m stdpopsim -q HomSap -D -L 1000 "
-                    "-o {filename} 2"
+                    f"-o {filename} 2"
```
grahamgower, line 821: Nice catch!

tests/test_cli.py
```diff
@@ -818,9 +818,10 @@ def test_dry_run_quiet(self):
...
                 )
                 subprocess.run(cmd, stderr=stderr, shell=True, check=True)
+                print(cmd)
```
grahamgower, line 824: Leftover from debugging? I guess this can be removed?

tests/test_annotations.py
andrewkern: done

andrewkern:
gonna go for it

PR #963, opened 2021-06-17
Ensure we're using discrete recombination maps.

### LINKED ISSUE #441: When msprime 1.0 is released, we should use discrete recombination maps
Also for the slim engine, see #438 and #439.

grahamgower:
Closes #441.

### DIFF
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -258,7 +258,7 @@ class Contig:
             )
         if chromosome is None:
             if genetic_map is not None:
-                raise ValueError("Cannot use genetic map with generic contic")
+                raise ValueError("Cannot use genetic map with generic contig")
             if length_multiplier != 1:
                 raise ValueError("Cannot use length multiplier for generic contig")
             if inclusion_mask is not None or exclusion_mask is not None:
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -1293,11 +1293,10 @@ class _SLiMEngine(stdpopsim.Engine):
             migration_matrix=recap_epoch.migration_matrix,
         )
 
-        # TODO: using recombination map in recapitate. maps should be discrete.
         ts = msprime.sim_ancestry(
             initial_state=ts,
             demography=demography,
-            recombination_rate=contig.recombination_map.mean_rate,
+            recombination_rate=contig.recombination_map,
             random_seed=s1,
         )
 
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -119,18 +119,19 @@ class TestAPI:
         )
         assert "sim.registerLateEvent" in out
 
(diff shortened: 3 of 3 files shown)


stdpopsim/genomes.py
```diff
@@ -223,6 +223,12 @@ class Contig:
     genomic_element_types = attr.ib(factory=list)
     mutation_types = attr.ib(factory=list)
 
+    @recombination_map.validator
+    def _check_recombination_map_is_discrete(self, _attribute, _value):
+        discrete_position = np.array(self.recombination_map.position, dtype=int)
+        if any(discrete_position != self.recombination_map.position):
+            raise ValueError("recombination map must be discrete")
```
petrelharp, line 230: I'm not sure if we should have this check, actually - regardless of whether the positions in the recombination map are integers, the resulting tree sequence will still have breakpoints at integer values. I guess that a recombination map with noniteger values might be in error, but the problem isn't because it's not discrete. (And, it'd be easy to do, e.g. if there was a breakpoint at `ts.sequence_length / 2`...)

tests/test_genomes.py
```diff
@@ -72,3 +72,7 @@ def test_add_genomic_element_type(self):
         contig.add_genomic_element_type(intervals2, [m1], [0.2])
         assert len(contig.genomic_element_types) == 2
         assert len(contig.mutation_types) == 2
+
+    def test_recomb_map_is_discrete(self):
+        with pytest.raises(ValueError):
+            stdpopsim.Contig.basic_contig(length=100.1)
```
petrelharp, line 78: ah-ha. Testing if the *endpoint* is an integer makes sense, though.

petrelharp (commented):
I'd have thought the way to check this was whether all the breakpoints coming out the end are at integers?

stdpopsim/genomes.py
jeromekelleher: I agree - the breakpoints in the recombination map will have no effect on the output coordinates.

jeromekelleher (commented):
I don't think there's much point in this tbh - if people want to use continuous coords, let em I say.

mufernando:
I think we wanted to only use discrete recombination maps in stdpopsim because SLiM is discrete?

Also, all the maps in the catalog are discrete.

jeromekelleher:
> Also, all the maps in the catalog are discrete

I think all maps are discrete, full stop, unless someone goes out of their way to construct one especially. We're not planning on providing an API for specifying your own map AFAIK, so I just don't see the value we gain from the complexity here.

grahamgower:
Thanks for the feedback. I guess I was missing the fundamental point that non-integer coords will work fine (and I forget the details of our former problems...). I've wound back the PR: use the recombination map during recapitation, and make sure there's a test that actually uses this functionality.

jeromekelleher (approved):
LGTM

mufernando (approved):
looks good!

PR #970, opened 2021-06-29
Remove numcodecs from conda deps.

grahamgower:
Just checking if this is still needed...


grahamgower:
I've removed the numcodecs dependency from the conda requirements file. I guess there were build issues previously that warranted adding this, but this problem appears to have gone away.

jeromekelleher (approved):
Not sure why we put this in, tbh

PR #979, opened 2021-07-13
Fix the dominance coefficient in Kim DFE

apragsdale:
The Kim et al paper simulated under multiple DFEs, some with fixed dominance and some where h is a function of s. Stdpopsim currently only supports fixed h for a given class of selected mutations, but the current parameterization copies the DFE with an h(s) function. Maybe h=0.5 is a better option here, or h could be a an argument of the DFE function. (See popsim-consortium/analysis2#5)


mufernando (approved):
looks good! thanks!

grahamgower:
Thanks @apragsdale. I think the same change should also be made in `docs/tutorial.rst`?

grahamgower (approved):

PR #993, opened 2021-08-11
Boyko 2008 African demographic model 

izabelcavassim:
Here's an African demographic model from Boyko et al 2008.


petrelharp:
LGTM, thanks!

PR #995, opened 2021-08-11
QC for Boyko et al

petrelharp:
Here's my implementation - looks like we have some differences, @izabelcavassim -

1. How'd you find the mutation rate? I could not find it.
2. Since they work in units of generations, I think it makes sense for us to implement the model in terms of years, translating quantities (times and rates) using the default generation time - this will match other things better, I think?


petrelharp:
Update:
- mutation rate was in the supplement
- they don't specify, or use, a generation time really

docs/parameter_tables/HomSap/Africa_1B08.csv
```diff
@@ -1,6 +1,4 @@
 Population size,"7,778",Ancestral pop. size
-Population size,"25,636",Current pop. size
-Growth rate (per gen.),0.000175165,AFR pop. growth rate expansion
-Epoch Time (gen.),"6,809",Expansion time of ancestral pop.
-Generation time (yrs.),1,Generation time
+Population size,"25,636",AFR current pop. size
```
petrelharp, line 2: I'd not say `AFR`, as that's the code from 1000 genomes; this is a different sample. Shouldn't this be "African" to match the populatiuon name?

stdpopsim/catalog/HomSap/demographic_models.py
```diff
@@ -1554,12 +1554,18 @@ def _africanBoyko():
...
+        the least European ancestry, using coalescent simulations with
+        recombination with the maximum likelihood method of Williamson et al 2005;
+        times were calibrated assuming 3e5 generations since human-chimp divergence
+        and fitting the number of synonymous human-chimp differences. Mutation and
+        recombination rates were assumed to be the same (1.8e-8).
     """
     populations = [
-        stdpopsim.Population(id="AFR_Boyko", description="African Boyko"),
+        stdpopsim.Population(
+            id="African", description="African-Americans from Boyko et al 2008"
```
petrelharp, line 1567: It feels weird to call this "african" when it's african-americans; what do you think?

petrelharp:
I've adjusted that test so that if the generation time is not given it checks the `.generation_time` attribute is 1 (the default).

petrelharp:
ok - tests are passing, and I've squashed the commits. if it looks good to you, @izabelcavassim, I'll merge this.

izabelcavassim:
It looks good to me. Thanks

PR #1002, opened 2021-08-24
Adding DFE class

izabelcavassim:
Adding a DFE class to stdpopsim. The structure is similar to how DemographicModel class was built. 
This is the first attempt, and still not working properly. I need some help to integrate the stuff that is in ext/ to this new class. So we won't necessarily need to do stdpopsim.ext.MutationType(), to specify a given mutation type, for instance.
Maybe @grahamgower can help me with that?

Thanks!

### DIFF
--- a/stdpopsim/catalog/HomSap/dfes.py
+++ b/stdpopsim/catalog/HomSap/dfes.py
@@ -1,0 +1,47 @@
+import stdpopsim
+
+_species = stdpopsim.get_species("HomSap")
+
+###########################################################
+#
+# DFEs
+#
+###########################################################
+
+
+def _KimDFE():
+    id = "Gamma_K17"
+    description = "Deleterious Gamma DFE"
+    long_description = """
+    Return neutral and negative MutationType()s representing a human DFE.
+    Kim et al. (2017), https://doi.org/10.1534/genetics.116.197145
+    """
+    citations = [
+        stdpopsim.Citation(
+            author="Kim et al.",
+            year=2017,
+            doi="https://doi.org/10.1534/genetics.116.197145",
+            reasons="to be defined",  # include the dfe_model reason
+        )
+    ]
+    neutral = stdpopsim.ext.MutationType()
+    gamma_shape = 0.186  # shape
+    gamma_mean = -0.01314833  # expected value
+    h = 0.5  # dominance coefficient
+    negative = stdpopsim.ext.MutationType(
+        dominance_coeff=h,
+        distribution_type="g",  # gamma distribution
+        distribution_args=[gamma_mean, gamma_shape],
+    )
+
+    return stdpopsim.DFE(
+        id=id,
+        description=description,
+        long_description=long_description,
+        mutation_types=[neutral, negative],
+        proportions=[0.3, 0.7],
+        citations=citations,
+    )
+
+
+_species.add_dfe(_KimDFE())
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -1,10 +1,14 @@
 """
 Common infrastructure for specifying demographic models.
 """
+import collections.abc
 import copy
+import numpy as np
 import textwrap
 
 import msprime
+
+from . import ext
 
 
 class Population:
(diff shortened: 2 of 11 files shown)


petrelharp:
I can help!

petrelharp:
>  So we won't necessarily need to do stdpopsim.ext.MutationType(), to specify a given mutation type, for instance.

I'm not sure what you mean - do you mean not having to call `stdpopsim.ext.MutationType( )` [here](https://github.com/popsim-consortium/stdpopsim/pull/1002/files#diff-969ce029f7c31b37201e94e249ff493e10fd349a7124aaa48ff5aadc49da531cR27), for instance? If not, how are you thinking it would work?

stdpopsim/species.py
```diff
@@ -219,6 +229,29 @@ def add_demographic_model(self, model):
             )
         self.demographic_models.append(model)
 
+    def get_dfe(self, id):
+        """
+        Returns a dfe with the specified ``id``.
```
petrelharp, line 234: I'd have thought we'd capitalize DFE, so `get_DFE` and `Returns a DFE`? It's an abbreviation?

stdpopsim/models.py
```diff
@@ -299,3 +299,67 @@ def __init__(self, NA, N1, N2, T, M12, M21):
...
+        *,
+        id,
+        description,
+        long_description,
+        mutation_types=None,
+        proportions=None,
+        citations=None,
+        # qc_model=None, # we may want to include qc of dfe in the future
+    ):
+        self.id = id
```
petrelharp, line 345: Should we check that `mutation_types` and `proportions` are of the same length, and raise an error if not?

petrelharp:
Looking good so far!

stdpopsim/species.py
```diff
@@ -117,6 +123,9 @@ class Species:
     :ivar demographic_models: This list of :class:`DemographicModel`
         instances in the catalog for this species.
     :vartype demographic_models: list
+    :ivar dfes: This list of :class:`DFE`
+        instances in the catalog for this species.
+    :vartype demographic_models: list
```
petrelharp, line 128:
```suggestion
    :vartype dfes: list
```

stdpopsim/catalog/HomSap/dfes.py
```diff
@@ -0,0 +1,49 @@
...
+    gamma_shape = 0.186  # shape
+    gamma_mean = -0.01314833  # expected value
+    h = 0.5  # dominance coefficient
+    negative = stdpopsim.ext.MutationType(
+        dominance_coeff=h,
+        distribution_type="g",  # gamma distribution
+        distribution_args=[gamma_mean, gamma_shape],
+    )
+
+    # neutral mutations have 0 proportion because they are not simulated by SLiM
```
petrelharp, line 37: hm, but it doesn't have 0 proportion below?

stdpopsim/models.py
petrelharp: Also check that proportions are sensible (eg not negative, don't sum to more than 1).

petrelharp: and check that all the things in `mutation_types` are of class MutationType

petrelharp:
TODO:
- fixup module import order
- add a flag to the CLI for choosing a DFE (maybe a different PR?)
- add test for DFEs like [this one](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_species.py#L119)
- add test that uses a DFE to [test_slim_engine.py](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_slim_engine.py)

and write new tests in [test_genomes.py](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_genomes.py) -- see [test_models.py](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_models.py) for inspiration. Like:
- make sure that the info that we pass in to a DFE object is there in a correct way, e.g., if we specify the `id` then get the `.id` property then its the same; same for mtuation types
- check that it's ok to have proportions that sum to less than one, but can't have negative numbers or sum to more than 1 (maybe we change this?)
- test that we can print it out

izabelcavassim:
TODO (updated my side):

- [ ]  fixup module import order
- [ ]  add a flag to the CLI for choosing a DFE (maybe a different PR?)
- [ ] add test for DFEs like this one
- [ ] add test that uses a DFE to test_slim_engine.py (OR NOT since this will be done in test_genomes.py)
- [ ] add a DFE argument to contig.add_genomic_element_type that pulls the proportions and mutation types

and write new tests in test_genomes.py -- see test_models.py for inspiration. Like:

- [ ] make sure that the info that we pass in to a DFE object is there in a correct way, e.g., if we specify the id then get the .id property then its the same; same for mtuation types
- [ ] check that proprtions must sum to 1 and not be negative
- [ ] test that we can print it out
- [ ] automatic test that the DFE runs in a simulation as in test_models.py

petrelharp:
>  add a DFE argument to contig.add_genomic_element_type that pulls the proportions and mutation types

Do we still want to do this? I think `contig.add_DFE` does everything we want here?

stdpopsim/slim_engine.py
```diff
@@ -1132,7 +1132,7 @@ def simulate(
         if slim_burn_in < 0:
             raise ValueError("slim_burn_in must be non-negative")
 
-        if slim_scaling_factor != 1:
+        if slim_scaling_factor > 1:
```
petrelharp, line 1135: Hm, I'm not sure about this change - what prompted it?

tests/test_genomes.py
```diff
@@ -72,3 +73,222 @@ def test_add_genomic_element_type(self):
...
+    """
+
+    def test_id_valid(self):
+        assert utils.is_valid_dfe_model_id(self.dfe.id)
+
+
+qc_test_classes = []
+for species in stdpopsim.all_species():
+    for model in species.demographic_models:
+        for dfe in species.dfes:
```
petrelharp, line 281: ah, I see! Nice work figuring this out. But, I don't think we need to test all combinations of DFE and model. I don't think we need to even test DFEs with non-generic models at all - they're pretty much completely separated?

petrelharp: Hm - and, this won't test DFEs at all in species without demographic models. I'm replacing this with a generic model.

petrelharp:
I may have gotten the tests working - there seem to be some other slim-related errors for certain demographic models, but that's separate from this PR.

izabelcavassim:
Fixed the coverage and rebased. 
it is all yours @petrelharp, you can read it through!

Thank you 🐢🐢

petrelharp:
Looks good (still)! I tidied up a few more loose ends, and will squash again.

PR #1020, opened 2021-10-03
Prototype for transforming the catalog to yaml

jeromekelleher:
This is a prototype in which we decouple the catalog from Python and instead encode our species descriptions declaratively using yaml. For example, this yaml:

```yaml
id: DroMel
name: Drosophila melanogaster
common_name: D. melanogaster
generation_time: 0.1
population_size: 1720600
ensembl_id: drosophila_melanogaster
genome:
  assembly_name: BDGP6.32
  assembly_accession: GCA_000001215.4
  chromosomes:
    2L:
      length: 23513712
      recombination_rate: 2.40462600791e-08
      mutation_rate: 5.49e-09
      synonyms:
      - chr2L
    2R:
      length: 25286936
      recombination_rate: 2.23458641776e-08
      mutation_rate: 5.49e-09
      synonyms:
      - chr2R
    3L:
      length: 28110227
      recombination_rate: 1.79660308862e-08
      mutation_rate: 5.49e-09
      synonyms:
      - chr3L
    3R:
      length: 32079331
      recombination_rate: 1.71642045777e-08
      mutation_rate: 5.49e-09
      synonyms:
      - chr3R
    '4':
      length: 1348131
      recombination_rate: 2.00579550709e-08
      mutation_rate: 5.49e-09
      synonyms:
      - chr4
    X:
      length: 23542271
      recombination_rate: 2.89650687913e-08
      mutation_rate: 5.49e-09
      synonyms:
      - chrX
    Y:
      length: 3667352
      recombination_rate: 0
      mutation_rate: 5.49e-09
      synonyms:
      - chrY
   mitochondrion_genome:
      length: 19524
      recombination_rate: 0
      mutation_rate: 5.49e-09
      synonyms: []
```

encodes the same information (excluding citations, which I've left out for clarity) as two Python files, with a lot less boiler plate. I think this is much more useful because it decouples the details of recording the information required to simulate a species from Python, and in particular from our frontend for msprime/slim.

We're missing (at least) two big things here: genetic maps and demographic models. The demographic models are already scheduled to be turned into static yaml in Demes format, so I think that works well. In practise, I think we'd structure the catalog like this:
```
catalog
    HomSap/
          species.yaml
          genetic_maps.yaml
          demography
               OutOfAfrica_3G09.yaml
               OutOfAfrica_2T12.yaml
               ....
```

The "genetic_maps.yaml" file could contain a list of the information about genetic maps, encoding the class structure like we're doing here - or we could simply merge this information into the species.yaml file, whatever is simpler.

The catalog then is *entirely* static and encoded in json/yaml, and we could therefore just think of it as a static website hosted at a URL somewhere. (Of course we could also thing of it as a JSON webservice, where this just gives the logical structure of the API - but let's not go there yet).

So, my proposal initially is that we take the species information out of Python, and we put it into a directory hierarchy like this. We can then instantiate the classes we need on demand, doing something like this:
```python
def get_species(name):
    species_yaml = config.catalog_path / name / species.yaml
    return Species.fromdict(yaml.load(species_yaml))
```
So, our startup times should be *much* faster, since we're not parsing all the species information on startup.

Most of the code in this PR is for transforming the current object model into JSON and back again. The following code generated the yaml in the included files (I've put then in here just so we can see what they'd look like) and checks that we can exactly recover the same information:

```python
from ruamel.yaml import YAML
import json
import stdpopsim

for species in stdpopsim.all_species():

    d = species.asdict()
    yaml = YAML() #typ="safe")
    with open(f"yamls/{species.id}.yaml", "w") as f:
        yaml.dump(d, f)

    species2 = stdpopsim.Species.fromdict(d)
    print(species.asdict() == species2.asdict())
```

What do we think?

### DIFF
--- a/catalog/AedAeg.yaml
+++ b/catalog/AedAeg.yaml
@@ -1,0 +1,65 @@
+id: AedAeg
+name: Aedes aegypti
+common_name: Yellow fever mosquito
+generation_time: 0.06666666666666667
+population_size: 1000000.0
+ensembl_id: aedes_aegypti_lvpagwg
+genome:
+  assembly_name: AaegL5
+  assembly_accession: GCA_002204515.1
+  citations:
+  - doi: https://doi.org/10.1126/science.1138878
+    author: Nene et al.
+    year: 2007
+    reasons:
+    - genome assembly
+  - doi: https://doi.org/10.1371/journal.pntd.0002652
+    author: Juneja et al.
+    year: 2014
+    reasons:
+    - recombination rate
+  - doi: https://doi.org/10.1186/s12915-017-0351-0
+    author: Crawford et al.
+    year: 2017
+    reasons:
+    - population size
+    - mutation rate
+    - generation time
+  - doi: https://doi.org/10.1101/gr.091231.109
+    author: Keightley et al.
+    year: 2009
+    reasons:
+    - mutation rate
+  chromosomes:
+    '1':
+      length: 310827022
+      recombination_rate: 3.06e-09
+      mutation_rate: 3.5e-09
+      synonyms:
+      - chr1
+    '2':
+      length: 474425716
+      recombination_rate: 2.49e-09
+      mutation_rate: 3.5e-09
+      synonyms:
+      - chr2
+    '3':
+      length: 409777670
+      recombination_rate: 2.91e-09
+      mutation_rate: 3.5e-09
+      synonyms:
+      - chr3
+    MT:
+      length: 16790
+      recombination_rate: 0
+      mutation_rate: 3.5e-09
+      synonyms:
+      - chrM
+citations:
+- doi: https://doi.org/10.1186/s12915-017-0351-0
+  author: Crawford et al.
+  year: 2017
+  reasons:
+  - population size
+  - mutation rate
+  - generation time
--- a/catalog/AnaPla.yaml
+++ b/catalog/AnaPla.yaml
@@ -1,0 +1,280 @@
+id: AnaPla
+name: Anas platyrhynchos
+common_name: Mallard
+generation_time: 4
+population_size: 156000
+ensembl_id: anas_platyrhynchos
+genome:
+  assembly_name: ASM874695v1
+  assembly_accession: GCA_008746955.1
+  citations:
+  - doi: https://doi.org/10.1111/mec.15343
+    author: Lavretsky et al.
+    year: 2019
+    reasons:
+    - mutation rate
+    - generation time
+  - doi: https://dx.doi.org/10.1534%2Fgenetics.105.053256
+    author: Huang et al.
+    year: 2006
+    reasons:
+    - recombination rate
+  chromosomes:
+    '1':
+      length: 208326429
+      recombination_rate: 1.52e-08
+      mutation_rate: 4.83e-09
+      synonyms:
+      - chr1
+    '2':
+      length: 162939446
+      recombination_rate: 1.39e-08
+      mutation_rate: 4.83e-09
+      synonyms:
+      - chr2
+    '3':
+      length: 119723720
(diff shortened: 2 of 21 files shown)


grahamgower:
I really like this idea. What we have now with every species, genetic map, etc. being imported is not sustainable long term, and this refactoring nicely opens the possibility of not loading something until it is queried. E.g. load the list of species, but only load info about some given species (e.g. demographic models available) when we do get_species("foo"). Hopefully the separation of species definitions from the Python code will make it easier for new contributors too!

apragsdale:
I think this is a great idea - and in addition to not having to load everything up front, it makes it much easier to understand and contribute from a user/developer's perspective.

I wonder if this data (species and demographic models) should be placed in a separate repository, which then `stdpopsim` imports as data files? We've gone back and forth over in Demes about whether example data files should be included within that package, and mostly decided against it since verified demographic models exist here. If the YAMLs were placed in their own repo, anyone else (such as dadi, fwdpy11, moments, etc) could also then import those QC'd models as package data, without needing the whole stdpopsim suite as a dependency. I think that would be widely useful.

jeromekelleher:
Good point @apragsdale, I completely agree. There's no need for the catalog to be a Python package at all. If it's just a directory of yaml files, then we might was well package it as a directly of yaml files, so that the data is available to any programming language.

One way we could go is to make a new repo ``stdpopsim-catalog`` which is just for maintaining the catalog. The main focus would be the actual ``catalog`` directory containing the yaml files, and we could move over some of the existing infrastructure we've built for maintaining it. Since it's a git repo, it's naturally versioned, and we could tag versions as new species etc are added to the catalog/have been QC'd etc. The current ``stdpopsim`` Python package could use ``git submodule`` to pull in a specific release of the catalog. 

That way we completely decouple the current Python-based SLiM/msprime frontend software from the actual species data. I think this is really important, as the current repo where we mush both things together will become unmanagable pretty quickly otherwise.

grahamgower:
Really good question @vsbuffalo! We've had lots of discussions of this exact thing for Demes, and yes, it's entirely possible to construct a "mult document" yaml file where each document has the same structure but distinct parameter values (e.g. for posterior samples from an ABC).

fbaumdicker:
I like this idea.
I recognized this issue as I just "rediscovered" my long-standing to-do from the hackathon to add gene conversion to stdpopsim. Now I am wondering whether gene conversion rates would need a special treatment before the yaml representation is fixed?
So would gene conversion be a change to the data model that is incompatible with a static YAML representation or could we simply write something like:
```yaml
genome:
  assembly_name: XY
  assembly_accession: ABC
  chromosomes:
    2L:
      length: 23513712
      recombination_rate: 0.0
      gene_conversion_rate: 3.14e-08
      gene_conversion_tract_length: 15.4
      mutation_rate: 5.49e-09
      synonyms:
      - chr2L
```
and that would work?

petrelharp:
> my long-standing to-do from the hackathon to add gene conversion to stdpopsim

Oh, that'd be lovely!

I'm not actually quite clear on why the YAML representation needs to be 'fixed' - can't we always add things (we just don't want to move around existing things)?

jeromekelleher:
> I'm not actually quite clear on why the YAML representation needs to be 'fixed' - can't we always add things (we just don't want to move around existing things)?

I guess the idea is that we document the structure of the yaml so that this becomes an artefact by itself, independent of the Python API. This means "fixing" in terms of deciding on a structure and sticking with it (within reason). So, it's not a "protocol" in that we encourage other people to use it for interchange, but it's a reasonably stable representation of the data in the catalog that people can write code against in a variety of languages.

But yeah, we can totally add things like gene conversion later on.

I'm going to close this PR for now, as realistically I don't have any time to work on this until at least February and I don't want to give the expectation that these changes are imminent.

Issue #921, opened 2021-04-15
QC for {MallardBlackDuck_2L19} ({AnaPla})


petrelharp:
**PR for new model:** #883
**Original paper:** [Lavretsky et al 2019](https://www.utep.edu/science/lavretskylab/Articles/2020.Lavretsky-et-al.MEC_ddRAD.Capture_ABMA.pdf)
**Parameter values:** in the paragraph below this figure: 
![params](https://user-images.githubusercontent.com/1046249/114813473-066e0d00-9d67-11eb-995c-51b5ba8b632d.png)
... and the ancestral population size was not reported; the authors say by email that:

>  Based on the contemporary dataset, the ancestral population for the Black duck/Mall dadi model was 819535.

**Population indices:** 0: "Mallard", contemporary; 1: "Black_duck", contemporary, 2: "Ancestral", before the split.

**Potential issues that might lead to differences between the model implementations:**
The migration rate is as in a "split_mig" model of dadi, which has symmetric migration rate and for which [the documentation](https://dadi.readthedocs.io/en/latest/user-guide/specifying-a-model/) says

> [the migration rate is] the fraction of individuals each generation in pop 1 that are new migrants from pop 2, times the 2Nref

... and the migration rate we want to input into msprime is precisely the proportion of each population that are replaced by migrants.

igronau:
I'm going to try to QC this model

PR #1021, opened 2021-10-03
Added QC for duck/mallard (AnaPla)

igronau:
The Demographic model implemented for Black duck and Mallard. See Original Model QC Issue [#921](https://github.com/popsim-consortium/stdpopsim/issues/921#issue-858468039)

I reviewed the demographic parameters reported by Lavretsky et al. (2019) (Figure 6), and added the ancestral population size reported to @petrelharp in personal communication. I used these values to implement a demographic model and ran the unit tests. All seems good.

Regarding @petrelharp 's comment about the migration rate, I'm pretty sure that the rate we use in the model is consistent with what Lavretsky and colleagues infer using dadi. The raw dadi estimates (after scaled by N_anc) should be equivalent to the per-generation rate used by the coalescent simulator of msprime. To confirm this with 100%, one may wish to generate data using this model and run dadi on the data to see if the expected values are indeed inferred.

### DIFF
--- a/stdpopsim/qc/AnaPla.py
+++ b/stdpopsim/qc/AnaPla.py
@@ -1,0 +1,54 @@
+import msprime
+
+import stdpopsim
+
+
+_species = stdpopsim.get_species("AnaPla")
+
+
+def LavertskyEtAl2019TwoPop():
+    id = "QC-MallardBlackDuck_2L19"
+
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM

grahamgower (approved):

petrelharp:
Late to the party here, but: awesome, thanks @igronau!

PR #1025, opened 2021-10-06
Added and edited the data for Apis mellifera catalog

janaobsteter:


grahamgower (approved):
LGTM. I just have one question... The chromosome names are not so user friendly---are these the canonical names or are can we do better than what enseml provides?

janaobsteter:
> LGTM. I just have one question... The chromosome names are not so user friendly---are these the canonical names or can we do better than what enseml provides?

@grahamgower yes, I agree. I put the commonly used chromosome names in the synonyms, but I can try to add them as main names. Andy suggested in Slack to use `add-species-ncbi`, I can do this and recommit?

janaobsteter:
> > > LGTM. I just have one question... The chromosome names are not so user friendly---are these the canonical names or can we do better than what enseml provides?
> > 
> > 
> > @grahamgower yes, I agree. I put the commonly used chromosome names in the synonyms, but I can try to add them as main names. Andy suggested in Slack to use `add-species-ncbi`, I can do this and recommit?
> 
> So the `NC_xxx` names are the commonly used names? Having them as synonyms should be fine, as long as the `update-genome-data` maintenance subcommand doesn't overwrite them.

Yes, but some people also use just 1, 2, 3, 4 ... Do you think it would be better to have them as chromosome names and then have both NC_xxx and CM_xxx as synonyms?

grahamgower:
It wont work for your specific case, but for some inspiration see the `append_common_synonyms()` utility function:
https://github.com/popsim-consortium/stdpopsim/blob/31e84af11860d743d407efe5e244b3d10f8e957b/stdpopsim/utils.py#L165-L185

As used for, e.g. AedAeg:
https://github.com/popsim-consortium/stdpopsim/blob/31e84af11860d743d407efe5e244b3d10f8e957b/stdpopsim/catalog/AedAeg/species.py#L55-L66

PR #1042, opened 2021-10-19
generalize annotations, add Drosoohila

andrewkern:
okay here is a PR that adds Drosophila melanogaster exons to the annotation. I'm using BDGP6 annotations that I've garnered via Ensembl. 

One note here-- take a look at the way I'm dealing with the "source" of the exon annotations in `maintenance/annotation_maint.py`. Basically I'm doing this is a species specific manner at this level, which I _think_ should be decently sustainable for the time being?


grahamgower:
Thanks @andrewkern, this looks like it be a manual process for any given species but at least it won't be a ton of work. :)

> actually i have a different idea about how to deal with varying annotation "sources" for different species -- i think i might include that as an attribute to be defined in the `Annotation` object, much in the same way that I currently define the `file_pattern` attribute. I wonder what folks would think of that?

Yeah, I like that idea. Maybe this should be slightly more general and include the 'type" field in addition to the "source"? So if we want to support multiple annotations from distinct sources/types, we just add another Annotation object with a different source/type even if it comes from the same GFF? That's perhaps not ideal, but the redundancy only exists in the maintenance side so probably that's ok. Certainly the "source" and "type" values will need to be included in the filename on AWS so there are no conflicts.

andrewkern:
Okay @grahamgower I altered this PR to include new attributes in the `Annotation` class so now we can specify the source and type at the level of individual species level annotations and things should be less fragile than before

andrewkern:
just discussed this PR with @mufernando and we agreed this is probably a good way to go. the tests will have to wait for #1042 to merge

grahamgower (approved):
LGTM!

PR #1045, opened 2021-10-20
draft of a new add_dfe function

andrewkern:
here is a mock up of a new `add_dfe` type function that assumes you want a neutral background behind the intervals for which you specify the DFE. starting this merely as a draft to get comments @mufernando @izabelcavassim @petrelharp


andrewkern:
i haven't written any tests yet so codecov is complaining, but wanted to look at this design together

andrewkern:
just discussed this PR with @mufernando. we think this is a good way to go for now, that allows users simple functionality.

Issue #1040, opened 2021-10-19
add gamma DFE for Drosophila from Huber et al


petrelharp:

izabelcavassim:
You can close this! @petrelharp

PR #1046, opened 2021-10-21
Drosophila dfe

izabelcavassim:
Adding a new gamma DFE for Drosophila based on [Huber et al., 2017](https://www.pnas.org/content/114/17/4465)
following the issue described in #1040 
Please have a look at it @petrelharp and let me know if I need to do any modifications. 
Thank you!


andrewkern:
this looks good to me @izabelcavassim. can you please rebase and squash these commits before we merge? i think i can use the github interface myself to do that in this case but don't want to screw it up...

andrewkern:
awesome @izabelcavassim!

Issue #1041, opened 2021-10-19
Add both annotation and DFE to the catalog documentation 


izabelcavassim:
As discussed today in the stdpopsim meeting, 
We need to describe how we invoke annotation, based on @andrewkern's implementation #960 
And how we invoke the DFE based on the implementation @petrelharp and I had worked on #999 

We should include those examples in this specific section of the doc: 
https://stdpopsim.readthedocs.io/en/latest/tutorial.html#incorporating-selection

And modifications to the doc can be done here: https://github.com/popsim-consortium/stdpopsim/blob/main/docs/tutorial.rst

PR #1048, opened 2021-10-21
added annotations to the catalog docs

### LINKED ISSUE #948: Add annotations to catalog
The annotations should be automatically parsed in the same way that genetic maps, demographic models, etcetera are as well. To do this, we need to edit [`docs/_ext/speciescatalog.py`](https://github.com/popsim-consortium/stdpopsim/blob/main/docs/_ext/speciescatalog.py) to pull in that information.

andrewkern:
closes #948 but note that #1041 is still open. need to add the DFEs to the catalog next.


izabelcavassim:
I can also just use your template above to do for DFEs. It is fine by me. Will do it today.

PR #1050, opened 2021-10-22
DFE class in CLI

izabelcavassim:
OK, I don't know why it is giving me the "Can't automatically merge", maybe some conflict, but I am PRequesting anyway.

So here is the first attempt to include selection (DFE) in the CLI and simulations. 
The way I do so is to allow the user to do the following:
if `-dfe (or --dfe-model)` is equal to **None**:
- then simulations will be done as previously (**without selection**). 
if `-dfe (or --dfe-model)` is equal to a specified model but demography is None:
- then selection is included and a single population constant size model is used for the demographic scenario.
if `-dfe (or --dfe-model)` and `--demographic-model` is defined by our smart user:
- then both parameters are taken into consideration.

There is a specific [place](https://github.com/popsim-consortium/stdpopsim/compare/main...izabelcavassim:DFE-CLI-new?diff=unified#diff-a2f706ab045cd6f2975f074024f6793138850eee3db24a9434a55f79c3c7230aR642), in which @andrewkern would work on getting the annotation specified.

This is at least what I wanted to do, glad to receive any comments from people @petrelharp @andrewkern @jeromekelleher 

Will dig into the PR conflict problem. 

Thanks


izabelcavassim:
Also, I always think it is a bit awkward how I call the DFE. Maybe if we could agree on terminology that would be great! And I would correct it throughout the files.

stdpopsim/cli.py
```diff
@@ -183,6 +190,42 @@ def __call__(self, parser, namespace, values, option_string=None):
         parser.exit()
 
 
+def get_dfes_help(species_id, dfe_id):
+    """
+    Generate help text for the given genetic map. If map_id is None, generate
```
andrewkern, line 195: "genetic map" -> "dfe"

andrewkern:
so a bigger design decision here is how should the DFEs interface with annotations for the CLI? 

One proposal would be to default to the behavior that @izabelcavassim has implemented here (where the entire contig is under the specified DFE) and then allow users to secondarily specify an annotation as the set of selected intervals. 

A second way to do this I suppose would be to force users to choose a set of intervals for each DFE. This might allow us in the future to extend things to layering more than 2 DFEs (neutral + 1 selected class).

PR #1052, opened 2021-10-22
DFE command line interface

izabelcavassim:
So here is the first attempt to include selection (DFE) in the CLI and simulations. 
The way I do so is to allow the user to do the following:
if `-dfe (or --dfe-model)` is equal to **None**:
- then simulations will be done as previously (**without selection**). 
if `-dfe (or --dfe-model)` is equal to a specified model but demography is None:
- then selection is included and a single population constant size model is used for the demographic scenario.
if `-dfe (or --dfe-model)` and `--demographic-model` is defined by our smart user:
- then both parameters are taken into consideration.

There is a specific [place](https://github.com/popsim-consortium/stdpopsim/compare/main...izabelcavassim:DFE-CLI-new?diff=unified#diff-a2f706ab045cd6f2975f074024f6793138850eee3db24a9434a55f79c3c7230aR642), in which @andrewkern would work on getting the annotation specified.

This is at least what I wanted to do, glad to receive any comments from people @petrelharp @andrewkern @jeromekelleher 

Thanks

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -21,6 +21,7 @@ import tskit
 import humanize
 
 import stdpopsim
+import numpy as np
 
 # resource is from the standard library, but it's not available on
 # Windows. We break from the usual import grouping conventions here
@@ -99,6 +100,13 @@ def get_model_wrapper(species, model_id):
 def get_genetic_map_wrapper(species, genetic_map_id):
     try:
         return species.get_genetic_map(genetic_map_id)
+    except ValueError as ve:
+        exit(str(ve))
+
+
+def get_dfe_wrapper(species, dfe_id):
+    try:
+        return species.get_dfe(dfe_id)
     except ValueError as ve:
         exit(str(ve))
 
@@ -183,6 +191,42 @@ class HelpGeneticMaps(argparse.Action):
         parser.exit()
 
 
+def get_dfes_help(species_id, dfe_id):
+    """
+    Generate help text for the given DFE. If dfe_id is None, generate
+    help for all DFEs. Otherwise, it must be a string with a valid DFE
+    ID.
+    """
+    species = stdpopsim.get_species(species_id)
+    if dfe_id is None:
+        dfes_text = f"\nAll DFEs for {species.name}\n\n"
+        dfes = [dfe.id for dfe in species.dfes]
+    else:
+        dfes = [dfe_id]
+        dfes_text = "\nDFE description\n\n"
+
+    indent = " " * 4
+    wrapper = textwrap.TextWrapper(initial_indent=indent, subsequent_indent=indent)
+    for dfe_id in dfes:
+        gdfe = get_dfe_wrapper(species, dfe_id)
+        dfes_text += f"{gdfe.id}\n"
+        dfes_text += wrapper.fill(textwrap.dedent(gdfe.long_description))
+        dfes_text += "\n\n"
+
+    return dfes_text
+
+
+class HelpDFEs(argparse.Action):
+    """
+    Action used to produce DFE help text.
+    """
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        help_text = get_dfes_help(namespace.species, values)
+        print(help_text, file=sys.stderr)
+        parser.exit()
+
+
 def get_species_help(species_id):
     """
     Generate help text for the given species with some of the species attributes
@@ -262,7 +306,7 @@ def write_output(ts, args):
         ts.dump(args.output)
 
 
-def get_citations(engine, model, contig, species):
+def get_citations(engine, model, contig, species, dfe):
     """
     Return a list of all the citations.
     """
@@ -273,26 +317,28 @@ def get_citations(engine, model, contig, species):
     if contig.genetic_map is not None:
         citations.extend(contig.genetic_map.citations)
     citations.extend(model.citations)
+    if dfe is not None:
+        citations.extend(dfe.citations)
     return stdpopsim.Citation.merge(citations)
 
 
-def write_bibtex(engine, model, contig, species, bibtex_file):
+def write_bibtex(engine, model, contig, species, bibtex_file, dfe):
     """
     Write bibtex for available citations to a file."""
-    citations = get_citations(engine, model, contig, species)
+    citations = get_citations(engine, model, contig, species, dfe)
     for citation in citations:
         bibtex_file.write(citation.fetch_bibtex())
     bibtex_file.close()
 
 
-def write_citations(engine, model, contig, species):
+def write_citations(engine, model, contig, species, dfe):
     """
     Write out citation information so that the user knows what papers to cite
     for the simulation engine, the model and the mutation/recombination rate
     information.
     """
     cite_str = ["If you use this simulation in published work, please cite:"]
-    for citation in get_citations(engine, model, contig, species):
+    for citation in get_citations(engine, model, contig, species, dfe):
         if (
             stdpopsim.citations.CiteReason.MUT_RATE in citation.reasons
             and model.mutation_rate is not None
@@ -400,6 +446,18 @@ def add_simulate_species_parser(parser, species):
             ),
         )
 
+    if len(species.dfes) > 0:
+        species_parser.add_argument(
+            "--help-dfes",
+            action=HelpDFEs,
+            nargs="?",
+            help=(
+                "Print list of DFEs and exit. If a DFE ID is "
+                "given as an argument, show help for this DFE. Otherwise show "
+                "help for all available DFEs"
+            ),
+        )
+
     # To avoid listing too much stuff out in the help, we only list
     # the actual IDs. We make all synonyms available as choices though.
     choices = []
@@ -463,8 +521,8 @@ def add_simulate_species_parser(parser, species):
     )
 
     model_help = (
-        "Specify a simulation model. If no model is specified, a single population"
-        "constant size model is used. Available models:"
+        "Specify a simulation model. If no model is specified, a single population "
+        "constant size model is used. Available models: "
         f"{', '.join(model.id for model in species.demographic_models)}"
         ". Please see --help-models for details of these models."
     )
@@ -476,6 +534,38 @@ def add_simulate_species_parser(parser, species):
         choices=[model.id for model in species.demographic_models],
         help=model_help,
     )
+    dfe_help = (
+        "Specify a Distribution of Fitness Effects (DFE) model. "
+        "If no DFE is specified, all mutations are neutral. "
+        "Available DFE models:"
+        f"{', '.join(dfe.id for dfe in species.dfes)}"
+        ". Please see --help-dfes for details of these DFE models."
+    )
+    species_parser.add_argument(
+        "--dfe",
+        default=None,
+        type=str,
+        metavar="",
+        choices=[dfe.id for dfe in species.dfes],
+        help=dfe_help,
+    )
+    interval_help = (
+        "Specify the interval where selection (given a DFE) is simulated. "
+        "Anything outside of the interval is simulated as neutral "
+        "If no interval is specified, "
+        "selection is simulated across the entire region (contig). "
+        "Interval (for now) is writen as 'left,right' (separated by a comma, "
+        "with no space, for instance: --dfe-interval 1000,2000"
+        # say something here about multiple dfes?
+    )
+    species_parser.add_argument(
+        "--dfe-interval",
+        default=None,
+        type=str,
+        metavar="",
+        help=interval_help,
+    )
+
     species_parser.add_argument(
         "-o",
         "--output",
@@ -505,7 +595,10 @@ def add_simulate_species_parser(parser, species):
             model = stdpopsim.PiecewiseConstantSize(species.population_size)
             model.generation_time = species.generation_time
             for citation in species.citations:
-                reasons = {stdpopsim.CiteReason.POP_SIZE, stdpopsim.CiteReason.GEN_TIME}
+                reasons = {
+                    stdpopsim.CiteReason.POP_SIZE,
+                    stdpopsim.CiteReason.GEN_TIME,
+                }
                 if len(citation.reasons & reasons) > 0:
                     model.citations.append(citation)
             qc_complete = True
@@ -532,9 +625,32 @@ def add_simulate_species_parser(parser, species):
             f"Running simulation model {model.id} for {species.id} on "
             f"{contig} with {len(samples)} samples using {engine.id}."
         )
+        dfe_interval = args.dfe_interval
+        dfe = args.dfe
+        if dfe_interval is None:
+            dfe_interval = ",".join(
+                map(str, [0, int(contig.recombination_map.sequence_length)])
+            )
+        if dfe is not None:
+            dfe = species.get_dfe(args.dfe)
+            left, right = dfe_interval.split(",")
+            contig.add_dfe(
+                intervals=np.array([[int(left), int(right)]]),
+                DFE=dfe,
+            )
+            logger.info(
+                f"Applying selection under the DFE model {dfe.id} "
+                f"in interval [{left}, {right})."
+            )
 
         write_simulation_summary(
-            engine=engine, model=model, contig=contig, samples=samples, seed=args.seed
+            engine=engine,
+            model=model,
+            contig=contig,
+            samples=samples,
+            dfe=args.dfe,
+            dfe_interval=dfe_interval,
+            seed=args.seed,
         )
         if not qc_complete:
             warnings.warn(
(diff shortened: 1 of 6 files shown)


izabelcavassim:
@grahamgower can you help me out here, are the errors related to the fact that some tests are processed with msprime and not slim?

grahamgower:
It looks like the cli tests need to also include `--engine slim` in the command. It might be good to catch the (probably common) case when users specify `--dfe` but not `--engine slim` to provide a user-friendly error message.

izabelcavassim:
> It looks like the cli tests need to also include `--engine slim` in the command. It might be good to catch the (probably common) case when users specify `--dfe` but not `--engine slim` to provide a user-friendly error message.

Will do, thank you!

izabelcavassim:
OK, I need to get a bit more of coverage. But this is what I have so far, happy to get comments, feedback, suggestions.

stdpopsim/cli.py
```diff
@@ -474,6 +532,36 @@ def add_simulate_species_parser(parser, species):
...
+        "selection is simulated across the entire region."
+        # say something here about multiple dfes?
+    )
+    species_parser.add_argument(
+        "--interval",
+        default=None,  # the contig length but should we add it here?
+        # type="numpy array", # can I say like that here?
+        metavar="",
+        help=interval_help,
+    )
```
grahamgower, line 563: It's not clear what the user should pass as the interval. Should this be the fiename of a file containing the intervals?

tests/test_cli.py
```diff
@@ -662,12 +677,28 @@ def test_genetic_map_citations(self, caplog):
...
         engine = stdpopsim.get_default_engine()
+        dfe = None
         stdout, stderr = capture_output(
-            cli.write_citations, engine, model, contig, species
+            cli.write_citations, engine, model, contig, species, dfe
         )
         assert len(stdout) == 0
         self.check_citations(engine, species, genetic_map, model, caplog.text)
 
+    def test_dfe_citations(self, caplog):
```
grahamgower, line 687: I think its hard to see where the `caplog` argument comes from (its a "magic" word that pytest recognises). To make this easier for future code readers, could we mark the test with the `@pytest.mark.usefixtures("caplog")` decorator?

stdpopsim/docs/_ext/speciescatalog.py
```diff
@@ -0,0 +1,622 @@
+"""
```
grahamgower, line 1: Not sure why this file is in the pr as a complete file, rather than a diff. Something funny must be going on, but probably need to do a rebase at least.

grahamgower (approved):
LGTM. I think you need to rebase though, because there's something funny going on with the speciescatalog docs script.

stdpopsim/cli.py
izabelcavassim: so, for now, we have a string, but later I will work with @andrewkern to get a bed file as an input and the idea is to parse the bed files and introduce selection to any interval from that file.

tests/test_cli.py
izabelcavassim: just putting this @pytest.mark.usefixtures("caplog") above the test?

stdpopsim/docs/_ext/speciescatalog.py
izabelcavassim: rebasing now

tests/test_cli.py
petrelharp: yep!

stdpopsim/cli.py
petrelharp: so, the help should explain what/how to put in there (and be accurate for now). Maybe an example, even?

petrelharp: For `default`, I'd say to leave it as None but say in the help text what the default is.

stdpopsim/cli.py
```diff
@@ -499,40 +587,112 @@ def add_simulate_species_parser(parser, species):
...
-            model = stdpopsim.PiecewiseConstantSize(species.population_size)
-            model.generation_time = species.generation_time
-            for citation in species.citations:
-                reasons = {stdpopsim.CiteReason.POP_SIZE, stdpopsim.CiteReason.GEN_TIME}
-                if len(citation.reasons & reasons) > 0:
-                    model.citations.append(citation)
-            qc_complete = True
+        if args.dfe is None:
+            dfe = None
+            if args.demographic_model is None:
```
petrelharp, line 592: It looks like this bit of code defining the demographic model and contig is duplicated - can't we put the `if args.dfe is None:` bit after the contig is defined?

stdpopsim/docs/_ext/speciescatalog.py
petrelharp: Wow, this was confusing! This took me a while to spot, but you've accidentally copied the file: now in your PR are both
```
docs/_ext/speciescatalog.py
```
and
```
stdpopsim/docs/_ext/speciescatalog.py
```
Only the first version should be there. (And, the two files are identical, I think, so you can safely delete the second.)

izabelcavassim: thanks! Deleted it!

stdpopsim/cli.py
izabelcavassim: you are right! :S thanks!

stdpopsim/cli.py
izabelcavassim: done

stdpopsim/cli.py
izabelcavassim: done

stdpopsim/cli.py
```diff
@@ -474,6 +532,38 @@ def add_simulate_species_parser(parser, species):
         choices=[model.id for model in species.demographic_models],
         help=model_help,
     )
+    dfe_help = (
+        "Specify a Distribution of Fitness Effects (DFE) model."
+        "If no demography is specified,"
+        "selection is simulated assuming a constant size demographic model."
```
petrelharp, line 538:
```suggestion
        "If no DFE is specified, all mutations will be neutral."
```

stdpopsim/cli.py
```diff
@@ -530,9 +623,50 @@ def run_simulation(args):
...
+                    ),
+                    DFE=dfe,
+                )
+                engine = stdpopsim.get_engine(args.engine)
+                logger.info(
+                    f"Running simulation model {model.id} for {species.id} on "
+                    f"{contig} with {len(samples)} samples using {engine.id}."
+                    f"With selection under the DFE model {dfe.id}"
+                    f"in interval {args.interval}."
+                )
```
petrelharp, line 660: We could simplify things a bit more here, I think:
```suggestion
            dfe = species.get_dfe(args.dfe)
            if args.interval is None:
                # if not specified, whole contig is under selection
                intervals=np.array(
                        [[0, int(contig.recombination_map.sequence_length)]]
                )
            contig.add_dfe(
                    intervals=intervals,
                    DFE=dfe,
            )
            engine = stdpopsim.get_engine(args.engine)
            logger.info(
                    f"Running simulation model {model.id} for {species.id} on "
                    f"{contig} with {len(samples)} samples using {engine.id}."
                    f"With selection under the DFE model {dfe.id}"
                    f"in interval {args.interval}."
            )
```

stdpopsim/cli.py
```diff
@@ -530,9 +623,50 @@ def run_simulation(args):
...
 
         write_simulation_summary(
-            engine=engine, model=model, contig=contig, samples=samples, seed=args.seed
+            engine=engine,
+            model=model,
+            contig=contig,
+            samples=samples,
+            seed=args.seed,
+            dfe=args.dfe,
+            interval=args.interval,
```
petrelharp, line 669: Whoops: shouldn't this be:
```suggestion
            dfe=dfe,
            interval=interval,
```

stdpopsim/cli.py
```diff
@@ -561,22 +695,24 @@ def run_simulation(args):
...
         if args.bibtex_file is not None:
-            write_bibtex(engine, model, contig, species, args.bibtex_file)
+            write_bibtex(engine, model, contig, species, args.bibtex_file, dfe)
 
     species_parser.set_defaults(runner=run_simulation)
 
 
-def write_simulation_summary(engine, model, contig, samples, seed=None):
+def write_simulation_summary(
+    engine, model, contig, samples, interval, seed=None, dfe=None
```
petrelharp, line 706:
```suggestion
    engine, model, contig, samples, seed=None, dfe=None,  intervals=None
```

stdpopsim/cli.py
```diff
@@ -610,6 +746,13 @@ def write_simulation_summary(engine, model, contig, samples, seed=None):
     dry_run_text += f"{indent}Mean recombination rate: {mean_recomb_rate}\n"
     dry_run_text += f"{indent}Mean mutation rate: {mut_rate}\n"
     dry_run_text += f"{indent}Genetic map: {gmap}\n"
+    # Get the information about dfe if any
+    if dfe is not None:
+        dry_run_text += f"{indent}DFE: {dfe}\n"
+        if interval is not None:
```
petrelharp, line 752: If we change the call to this function as I suggested above (passing in `intervals` not `args.intervals`) then we don't need this conditional here.

tests/test_cli.py
```diff
@@ -622,7 +635,8 @@ def test_number_of_calls(self):
         dois = set([ref.doi for ref in local_cites])
         ncite = len(dois)
         assert ncite == len(local_cites)
-        cli_cites = cli.get_citations(engine, model, contig, species)
+        dfe = None
```
petrelharp, line 638: probably we should test with a DFE here?

tests/test_slim_engine.py
```diff
@@ -315,6 +315,65 @@ def test_simulate(self):
...
+        n_mut_types = count_mut_types(ts)
+        assert n_mut_types[0] > 0
+        assert n_mut_types[1] > 0
+
+        # test a DFE without demography with interval
+        with tempfile.NamedTemporaryFile(mode="w") as f:
+            cmd = (
+                f"-q -e slim --slim-scaling-factor 20 --slim-path {slim_path} "
+                f"HomSap -c chr22 -l0.001 -o {f.name} --dfe Gamma_K17 "
+                "--interval [[1000,2000]] 10"
```
petrelharp, line 351: does this work? I'd thought you'd need quotes around the interval.

petrelharp (commented):
Looking good - a few things here and there.

stdpopsim/cli.py
```diff
@@ -474,6 +532,38 @@ def add_simulate_species_parser(parser, species):
...
+        choices=[dfe.id for dfe in species.dfes],
+        help=dfe_help,
+    )
+    interval_help = (
+        "Specify the interval where selection (given a DFE) is simulated."
+        "Anything outside of the interval is simulated as neutral"
+        "If no interval is specified,"
+        "selection is simulated across the entire region (contig)."
+        "Interval (for now) is writen between double brackets"
+        "such as: [[1000, 2000]]"
```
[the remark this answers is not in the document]
izabelcavassim: right now it only accepts one interval (I know, useless), but the suggestion you are giving is easy to implement (split the string above by commas and loop through that list).
Do I do the above, or do we assume that the person will always give a bed file?

stdpopsim/cli.py
```diff
@@ -474,6 +532,38 @@ def add_simulate_species_parser(parser, species):
...
+        # say something here about multiple dfes?
+    )
+    species_parser.add_argument(
+        "--interval",
```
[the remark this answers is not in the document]
izabelcavassim: agreed

tests/test_slim_engine.py
izabelcavassim: i mean the test pass and non-neutral sites are observed, so I guess it works?

stdpopsim/cli.py
izabelcavassim: I am following the rest of the way the code is written, should we change everything then?

izabelcavassim: nevermind

stdpopsim/cli.py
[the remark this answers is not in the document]
izabelcavassim: done

tests/test_slim_engine.py
petrelharp: oh! That's because there's no space in `[[1000,2000]]`! =)

stdpopsim/cli.py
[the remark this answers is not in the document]
petrelharp: I don't know? If we plan to remove the ability to pass in a single interval entirely then it doesn't matter; do whatever. If we want to leave it in (as well as passing in the name of an annotation) then I'd say do the "start,end" format. @andrewkern ?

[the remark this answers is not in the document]
petrelharp: But, I think that "name of annotation" or "name of bed file" is just fine.

izabelcavassim:
Can you have a look at this again @petrelharp?

petrelharp:
Looking!

Did you mean to resolve [this comment](https://github.com/popsim-consortium/stdpopsim/pull/1052/files#r747978430)? I think it's a copy-paste error in the docs, but let me know what's going on otherwise?

stdpopsim/cli.py
```diff
@@ -474,6 +532,39 @@ def add_simulate_species_parser(parser, species):
...
+        choices=[dfe.id for dfe in species.dfes],
+        help=dfe_help,
+    )
+    interval_help = (
+        "Specify the interval where selection (given a DFE) is simulated."
+        "Anything outside of the interval is simulated as neutral"
+        "If no interval is specified,"
+        "selection is simulated across the entire region (contig)."
+        "Interval (for now) is writen between double brackets"
+        "such as: 1000,2000"
```
petrelharp, line 557:
```suggestion
        "Interval (for now) is writen as 'left,right' (separated by a comma, "
        "with no space, for instance: --dfe-interval 1000,2000"
```

stdpopsim/cli.py
```diff
@@ -530,9 +624,43 @@ def run_simulation(args):
...
+                    ),
+                    DFE=dfe,
+                )
+            # Including selection in a specific interval
+            else:
+                inter = args.dfe_interval.split(",")
+                contig.add_dfe(
+                    intervals=np.array([[int(inter[0]), int(inter[1])]]),
+                    DFE=dfe,
+                )
```
petrelharp, line 647: Slightly more concise:
```suggestion
            if args.dfe_interval is None:
                left, right = 0, int(contig.recombination_map.sequence_length)
            else:
                left, right = args.dfe_interval.split(",")
            contig.add_dfe(
                intervals=np.array([[int(inter[0]), int(inter[1])]]),
                DFE=dfe,
            )
```

stdpopsim/cli.py
```diff
@@ -530,9 +624,43 @@ def run_simulation(args):
...
+            engine = stdpopsim.get_engine(args.engine)
+            logger.info(
+                f"Running simulation model {model.id} for {species.id} on "
+                f"{contig} with {len(samples)} samples using {engine.id}."
+                f"With selection under the DFE model {dfe.id}"
+                f"in interval {args.dfe_interval}."
```
petrelharp, line 653:
```suggestion
                f"in interval [{left}, {right})."
```

stdpopsim/cli.py
```diff
@@ -530,9 +624,43 @@ def run_simulation(args):
...
 
         write_simulation_summary(
-            engine=engine, model=model, contig=contig, samples=samples, seed=args.seed
+            engine=engine,
+            model=model,
+            contig=contig,
+            samples=samples,
+            seed=args.seed,
+            dfe=args.dfe,
+            dfe_interval=args.dfe_interval,
```
petrelharp, line 663:
```suggestion
            dfe_interval=[[left, right]],
```

stdpopsim/cli.py
```diff
@@ -610,6 +740,12 @@ def write_simulation_summary(engine, model, contig, samples, seed=None):
     dry_run_text += f"{indent}Mean recombination rate: {mean_recomb_rate}\n"
     dry_run_text += f"{indent}Mean mutation rate: {mut_rate}\n"
     dry_run_text += f"{indent}Genetic map: {gmap}\n"
+    # Get the information about dfe and dfe-interval if any
+    dry_run_text += f"{indent}DFE: {dfe}\n"
+    if dfe_interval is not None:
+        dry_run_text += f"{indent}DFE in interval: {dfe_interval}\n"
+    else:
+        dry_run_text += f"{indent}DFE in interval: {str([[0,contig_len]])}\n"
```
petrelharp, line 748: I think that the actual interval that's used should be passed in - i.e., we do the swapping out for None above, and None won't arrive here. Right now, so that this message is accurate, we need to replicate the behavior of `None` in two places: here, and above, which can be error-prone.

petrelharp:
LMK if it'd be easier to talk through these suggestions!

izabelcavassim:
Ok I am not doing the left right suggestions but the rest I took into account, and I also changed the default to dfe_interval.

petrelharp:
Uh-oh, how'd 0ba1dd5 get in here?

stdpopsim/cli.py
```diff
@@ -476,6 +534,39 @@ def add_simulate_species_parser(parser, species):
         choices=[model.id for model in species.demographic_models],
         help=model_help,
     )
+    dfe_help = (
+        "Specify a Distribution of Fitness Effects (DFE) model."
+        "If no demography is specified,"
+        "selection is simulated assuming a constant size demographic model."
```
petrelharp, line 540: And, whoops - I think this line is still in error? Or not?

PR #1066, opened 2021-10-27
Oo a nea ext puls demo

LeonardoIasi:
Additional HomSap demography of an simple OoA with Nenadertal introgression following Iasi et al. 2021 including the extended admixture pulse with a duration of 800 generations.


grahamgower:
Hi @LeonardoIasi! Thanks for the model addition! I haven't gone through the details, but it looks like there are a few code paths that aren't used in your nested `extended_pulse()` function. Given that you're calling the function only once with fixed values, maybe you could just `assert` that the conditions for these code paths dont happen and then remove that code?

LeonardoIasi:
Hello, 
yes the function has some if statements. But true since the parameters will not vary the function does not need to check them. I'll change the code and  commit again. Thanks

grahamgower (approved):
LGTM.

grahamgower:
@LeonardoIasi, would you please open an issue to track the QC of the model implementation as per the dev docs? https://stdpopsim.readthedocs.io/en/latest/development.html#demographic-model-review-process

PR #1073, opened 2021-10-29
added section to development on coverage output from pytest

andrewkern:
added a new section of the dev docs to show how to use pytest locally to get code coverage... this is something that i personally haven't known how to do until now and wow it helps!!

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -1142,6 +1142,23 @@ before::
 Finally, rename ``pre-commit.sample`` to simply ``pre-commit``
 
 *************
+Code Coverage
+*************
+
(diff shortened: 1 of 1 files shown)


izabelcavassim:
There is this very nice plug-in that I use to get some HTML reports of errors: 
https://pytest-html.readthedocs.io/en/latest/installing.html

If you do your pytest tests the following way: pytest --html=report.html --self-contained-html
you get a nice HTML report of errors, I prefer that than to stare at the terminal.

andrewkern:
yep basically the same thing

petrelharp (approved):
Works for me, awesome!

PR #1076, opened 2021-10-29
Create codecov.yml

andrewkern:


petrelharp:
What would this change - do you know? This is all a big black box to me.

andrewkern:
so i think adding this yaml is automatically looked at by the codecov orb thingy. i'm not sure exactly where it should live though. it might go at the repo root directory?

petrelharp:
Right - I'm wondering if you know what the effect on the behavior of codecov is going to be. How is it changing things?

andrewkern:
oh i'm trying to move the target here to 95%. that's all.

.circleci/codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+        enabled: yes           # must be yes|true to enable this status
+        target: 95             # specify the target coverage for each commit status
+                               #   option: "auto" (must increase from parent commit or pull request base)
+                               #   option: "X%" a static target percentage to hit
+        threshold: null        # allowed to drop X% and still result in a "success" commit status
+        if_not_found: success  # if parent is not found report status as success, error, or failure
+        if_ci_failed: error    # if ci fails report status as success, error, or failure
+    patch:
+      default:
+        enabled: no
```
petrelharp, line 18: any reason you're disabling `patch`?

petrelharp:
Currently, what's the policy? Is it
```
                                #   option: "auto" (must increase from parent commit or pull request base)
```

If so, I guess I'd vote to actually keep it at `auto` but then we know that sometimes we'll need to merge PRs that don't "pass" codecov just because they've decreased the denominator (say).

In any case - right now it looks like we're at 99.52% covered, so 95% is a big drop?

.circleci/codecov.yml
andrewkern: i think that's what it's supposed to be by default? no sure

andrewkern:
don't know?

andrewkern:
target: `auto` looks good actually: https://docs.codecov.com/docs/common-recipe-list

petrelharp:
So - maybe we leave things the way they are and not merge this PR?

jeromekelleher:
I had a look at the settings and it looks like we're on auto settings for codecov. The msprime config is [here](https://github.com/tskit-dev/msprime/blob/main/.codecov.yml) if you want a reference - I think the yaml file must be in the root directory, though.

To be honest, this is really quite fiddly and it would be simpler to bump the test coverage up than the change the settings.

petrelharp:
> To be honest, this is really quite fiddly and it would be simpler to bump the test coverage up than the change the settings.

lol.

andrewkern:
ugh okay! closing this PR

Issue #873, opened 2021-04-08
QC for Drosophila sechellia


jradrion:
**PR for new species:** {https://github.com/popsim-consortium/stdpopsim/pull/872}

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [ ] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [ ] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [ ] Population size.
- [ ] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [ ] Genome reference assembly.
- [ ] Mutation rate.
- [ ] Recombination rate.
- [ ] Recombination map(s) (if relevant).
- [ ] Population size.
- [ ] Generation time.

jeromekelleher:
Note that because the PR for this species used data from manually drawn NCBI, the test file is not stubbed out correctly and will need to be manually updated to include the correct chromosomes.

vitorpavinato:
I can do this if anyone hasn't picked it yet.

petrelharp:
Related: #885.

petrelharp:
Hi, @vitorpavinato - @percyfal is around and willing to help if you haven't got to this yet?

percyfal:
[...]

PR #1077, opened 2021-10-29
QC for Drosophila Sec adding tests

izabelcavassim:
Hi @jradrion, thanks for adding Drosophila sechellia to the catalog. 
I have just done the QC for this species #873, and we only disagree on the recombination map. 
I got my results from this:
https://petrov.stanford.edu/cgi-bin/recombination-rates_updateR5.pl

I know you uploaded some reference that has the recombination map for Drosophila Melanogaster: https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1002905
Can you tell me how you reached your values from this source? 

Thank you very much.

### DIFF
--- a/tests/test_DroSec.py
+++ b/tests/test_DroSec.py
@@ -21,27 +21,44 @@ class TestSpeciesData(test_species.SpeciesTestBase):
     # independently referring to the citations provided in the
     # species definition, filling in the appropriate values
     # and deleting the pytest "skip" annotations.
-    @pytest.mark.skip("Population size QC not done yet")
     def test_qc_population_size(self):
-        assert self.species.population_size == -1
+        assert self.species.population_size == 100000
 
     @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 20
 
 
 class TestGenomeData(test_species.GenomeTestBase):
 
     genome = stdpopsim.get_species("DroSec").genome
 
-    @pytest.mark.skip("Recombination rate QC not done yet")
-    @pytest.mark.parametrize(["name", "rate"], {}.items())
+    @pytest.mark.parametrize(
+        ["name", "rate"],
+        {
+            "2L": 2.39,
+            "2R": 2.66,
+            "3L": 1.79,
+            "3R": 1.96,
+            "X": 2.95,
+            "4": 0,
+        }.items(),
+    )
     def test_recombination_rate(self, name, rate):
         assert rate == pytest.approx(
             self.genome.get_chromosome(name).recombination_rate
         )
 
-    @pytest.mark.skip("Mutation rate QC not done yet")
-    @pytest.mark.parametrize(["name", "rate"], {}.items())
+    @pytest.mark.parametrize(
+        ["name", "rate"],
+        {
+            "2L": 1.5e-09,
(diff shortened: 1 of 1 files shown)


izabelcavassim:
- [ ]  fill out the test stubs
- [ ]  delete the pytest.mark.skip lines that make the tests not run
- [ ]  make sure they pass, talking to the original author to figure out discrepancies

izabelcavassim:
Hi @jradrion, it would be nice to get some input on the comments above since we want to finalize the QC of species we already have in the catalog. Thank you!

petrelharp:
Jeff may not be available for this as he's got a different job now; @andrewkern or @mufernando  do you have any insight here?

jradrion:
Hi @izabelcavassim, yes the values I got were derived from the Dmel values in Comeron et al. 2012, but IIRC they were lifted over to dm6. I can dig up the code I used to get them and figure out what I did, but I won't be able to get to it today. I should be able to find that for you by Wed-Thurs. The RRC tool you linked seems fine, if a bit dated, but if there is a preference to use the rates from that source, I say go for it!

izabelcavassim:
> Hi @izabelcavassim, yes the values I got were derived from the Dmel values in Comeron et al. 2012, but IIRC they were lifted over to dm6. I can dig up the code I used to get them and figure out what I did, but I won't be able to get to it today. I should be able to find that for you by Wed-Thurs. The RRC tool you linked seems fine, if a bit dated, but if there is a preference to use the rates from that source, I say go for it!

Hey @jradrion thanks for answering back, I think it would be important to have the values as visible as possible, so that is why I think using the link I provide and connect it to the Comeron article is a good way to go (so combining both our references).

I am not an expert in drosophila genomic architecture, but as far as I understand chromosome 4 does not recombine (such as anything besides the PAR in human Y chrom: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5721948/ ). Only big inversions seem to be observed across species (https://link.springer.com/content/pdf/10.1007/s004120100151.pdf) So for me, taking the average from other chroms would not be the correct approach (I think?). But happy to hear other people's thoughts! Thank you again.

Yes, just share the code you used and I will set chromosome 4 to zero, if you and others agree. 

Thanks

jradrion:
HI @izabelcavassim,

Sorry, I'm still trying to piece together what's going on here from memory. Looking back through my files it does not seem like I used any code specific to D. sechellia to arrive at the map. I think the idea was to simply use the same rates that were used for the Dmel map. The code I used to get those rates is [https://github.com/popsim-consortium/stdpopsim/blob/main/maintenance/liftOver_comeron2012.py](https://github.com/popsim-consortium/stdpopsim/blob/main/maintenance/liftOver_comeron2012.py). There is still an open PR where there was some discussion about various liftover concerns [https://github.com/popsim-consortium/stdpopsim/pull/885](https://github.com/popsim-consortium/stdpopsim/pull/885). I guess my only concern here is that if we say we are using D. melanogaster rates as a proxy for D. sechellia, we should make sure that the rates are the same between both species in our catalog. I also agree that linking directly to a website that displays recombination values is more transparent than linking to some liftover code.

izabelcavassim:
Hi @percyfal, thanks for the input, I have not touched upon this yet, but will do it this week. I will let you know once I do, I might need some help! I think I just need to go through what has already been done just so we don't waste what you guys have done in the past! Cheers and thanks!

PR #1092, opened 2021-11-04
Chromosome 4 recombination set to 0

izabelcavassim:


andrewkern:
ha looks like there is a test that included chr4 as a recombining chromosome. should get swapped over to the non recombining group here:

https://github.com/popsim-consortium/stdpopsim/blob/9ae648f664cb88cd5d5c8a9a9fc30f267b0a9f4e/tests/test_DroMel.py#L28-L36

izabelcavassim:
I think it will pass now. We need to fix the printing of the rates to allow scientific annotation, I will make an issue for that.

PR #1098, opened 2021-11-09
Reason for citation

izabelcavassim:


izabelcavassim:
sorry I thought we had corrected that before, the fix is in the CLI PR #1052

andrewkern:
closing this @izabelcavassim because i think you also have this code in #1052 but if i'm wrong we can reopen

PR #1103, opened 2021-11-12
moving annotations to ints, requiring ints in intervals

### LINKED ISSUE #1079: make intervals and annotations code consistent and make all intervals integer-valued
There's a number of operations on intervals in `utils.py` that are public but mostly undocumented. These are:
- `check_intervals_array_shape`
- `check_intervals_validity`
- `build_intervals_array`
- `mask_intervals`

Of these, I think that `check_intervals_array_shape` and probably `check_intervals_validity` should be private, and we should document the others.

Also, right now it's easy to get very confusing errors trying to pass in interval arrays having floats rather than ints. I think we should require that intervals are *always* of integer type, and give informative errors if they are not. We should also make sure we give an informative error in the case where nested lists are passed in instead of a numpy array. 

I'm of two minds whether we want to convert, say `[[0, 2.0], [4.0, 10]]` to the integer-valued numpy array.

andrewkern:
starting a PR to resolve #1079 

~~one outstanding item from that issue is that I haven't changed the method names of `check_intervals_validity` and `check_intervals_array_shape` to be e.g. `__check_intervals_validity` so that these are private methods~~

### DIFF
--- a/stdpopsim/utils.py
+++ b/stdpopsim/utils.py
@@ -9,7 +9,6 @@ import shutil
 import tarfile
 import contextlib
 import numpy as np
-import tskit
 
 
 def is_valid_demographic_model_id(model_id):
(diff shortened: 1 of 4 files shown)


stdpopsim/utils.py
```diff
@@ -184,19 +184,22 @@ def add_if_unique(chrom, syn):
...
 
-def check_intervals_validity(intervals, start=0, end=np.inf):
-    check_intervals_array_shape(intervals)
+def _check_intervals_validity(intervals, start=0, end=np.inf):
+    # is numpy?
+    if not isinstance(intervals, np.ndarray):
+        raise ValueError("Intervals must be a numpy array, did you try to pass a list?")
+    # are the intervals ints?
+    if (intervals.dtype != "int32") and (intervals.dtype != "int64"):
+        raise ValueError("intervals must be integer typed")
```
petrelharp, line 201:
```suggestion
        raise ValueError("Intervals must be arrays of type integer (i.e., with dtype='int').")
```

petrelharp:
LGTM - we just need some tests for what happens if you pass in a float array? (The tests should make sure the error that happens is the informative one, by doing like:
```
with pytest.raises(ValueError, match=".*integer type.*"):
     # code with float array 
```

andrewkern:
I added that test here https://github.com/andrewkern/stdpopsim/blob/d30d9ba997a120c64c8c3f4a0bbca0c1639c49b4/tests/test_utils.py#L480

andrewkern:
hahaha the suggested commit didn't due a pre-commit linting test!

andrewkern:
codecov is complaining but i think this is GTG @petrelharp

stdpopsim/utils.py
```diff
@@ -184,19 +183,24 @@ def add_if_unique(chrom, syn):
...
-    if (
-        (not isinstance(intervals, np.ndarray))
-        or (len(intervals.shape) != 2)
-        or (intervals.shape[1] < 2)
-    ):
+def _check_intervals_array_shape(intervals):
+    if (len(intervals.shape) != 2) or (intervals.shape[1] < 2):
         raise ValueError(
-            "Intervals must be 2D objects with at least 2 columns " "[left, right)."
+            "Intervals must be 2D numpy arrays with at least 2 columns "
```
petrelharp, line 189:
```suggestion
            "Intervals must be 2D numpy arrays with 2 columns "
```

petrelharp (approved):
LGTM - codecov is all good - but one small error in the docstring, I think?

andrewkern:
good catch!

Issue #846, opened 2021-04-08
include gene conversion feature of msprime/slim into stdpopsim 


fbaumdicker:
When we created stdpopsim, msprime did not have the gene conversion feature.
As gc is now possible in msprime as well as slim and gene conversion is a major factor of microbial evolution, it would be worth to include this as a possible parameter to the species models.

For E.Coli there might already be a commented part in the code that relates to the gene conversion of E.Coli.

Gene conversion can basically be handled like recombination. So if the parameter is not set, then gc is zero.
Otherwise, we have to specify a genome wide average for the gene conversion initiation rate and the mean tract length.

jeromekelleher:
I propose we add a ``gene_conversion_rate`` and ``gene_conversion_tract_length`` parameters to the Chromosome class, which default to zero. This can then be picked up in the contig and then passed to the simulation engine.

PR #1106, opened 2021-11-15
adding gene conversion support

### LINKED ISSUE #847: Add gene conversion parameters to E. coli. species model
When gene conversion can be handled as discussed in #846 we should add the corresponding parameters from Lapierre etl al. into the existing E.coli. species.

depends on #846

fbaumdicker:
This adds a mean gene conversion rate and gene conversion tract length to the chromosome class, as discussed in #846 .
In addition, the corresponding gene conversion parameters for EscCol are added which closes #847 .

It should now be possible to define the gene conversion rates and tract lengths for other species.
So far I did not notice any complications with the existing species and models.

However, the ploidy of the msprime engine is currently hardcoded and set to 2, which doubles the population size of EscCol.
But as integrating the ploidy into stdpopsim is much more entangled with the other species and demographic models, I did not touch the ploidy in this PR. I will open another issue for the ploidy generalization.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -248,6 +248,14 @@ def get_species_help(species_id):
     species_text += (
         f"Recombination rate: {species.genome.mean_recombination_rate:.4g}\n"
     )
+    species_text += (
+        f"Gene conversion rate: {species.genome.mean_gene_conversion_rate:.4g}\n"
+    )
+    species_text += (
+        f"Gene conversion tract length: "
+        f"{species.genome.range_gene_conversion_lengths}\n"
+    )
+
     return species_text
 
 
--- a/stdpopsim/species.py
+++ b/stdpopsim/species.py
@@ -170,6 +170,9 @@ class Species:
         length_multiplier=1,
         length=None,
         mutation_rate=None,
+        use_species_gene_conversion=False,
+        gene_conversion_rate=None,
+        gene_conversion_length=None,
         inclusion_mask=None,
         exclusion_mask=None,
     ):
@@ -195,6 +198,19 @@ class Species:
             ``genetic_map`` argument.
         :param float mutation_rate: The per-base mutation rate. If none is given,
             the mutation rate defaults to the rate specified by species chromosomes.
+        :param bool use_species_gene_conversion: If set to True the parameters for gene
+            conversion of the species chromosome are used if available. For "generic"
+            contigs the gene conversion rate is given by the (mean) rate specified by
+            species chromosomes and the gene conversion length defaults to the mean
+            length specified by species chromosomes.
+        :param float gene_conversion_rate: The per-base gene conversion rate. Can only
+            be set together with gene_conversion_length and if
+            use_species_gene_conversion is False; If none is given, and
+            use_species_gene_conversion is True the gene conversion rate specified by
+            species chromosomes is used. If none and use_species_gene_conversion is
+            False gene conversion is disabled.
+        :param float gene_conversion_length: The mean gene conversion length.
+            May only be provided if gene_conversion_rate is also specified.
         :param inclusion_mask: If specified, simulated genomes are subset to only
             inlude regions given by the mask. The mask can be specified by the
             path and file name of a bed file or as a list or array of intervals
@@ -217,6 +233,9 @@ class Species:
             length_multiplier=length_multiplier,
             length=length,
             mutation_rate=mutation_rate,
+            use_species_gene_conversion=use_species_gene_conversion,
+            gene_conversion_rate=gene_conversion_rate,
+            gene_conversion_length=gene_conversion_length,
             inclusion_mask=inclusion_mask,
             exclusion_mask=exclusion_mask,
         )
--- a/tests/test_EscCol.py
+++ b/tests/test_EscCol.py
@@ -26,3 +26,12 @@ class TestGenome(test_species.GenomeTestBase):
 
     def test_basic_attributes(self):
         assert len(self.genome.chromosomes) == 1
+
+    def test_mutation_rate(self):
+        assert self.genome.get_chromosome("Chromosome").mutation_rate == 8.9e-11
+
+    def test_gene_conversion_rate(self):
+        assert self.genome.get_chromosome("Chromosome").gene_conversion_rate == 8.9e-11
+
+    def test_gene_conversion_length(self):
+        assert self.genome.get_chromosome("Chromosome").gene_conversion_length == 345
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -254,7 +254,7 @@ class TestEndToEnd:
         self.verify(cmd, num_samples=7)
 
     def test_lapierre_constant(self):
-        cmd = "EscCol -L 100 2"
+        cmd = "EscCol -L 1000 2"
         self.verify(cmd, num_samples=2)
 
 
@@ -985,10 +985,10 @@ class TestNoQCWarning:
                 capture_output(stdpopsim.cli.stdpopsim_main, cmd.split())
 
     def test_noQC_warning(self):
-        self.verify_noQC_warning("EscCol -d FakeModel -D 10 -L 10")
+        self.verify_noQC_warning("EscCol -d FakeModel -D 10 -L 1000")
 
     def test_noQC_warning_quiet(self):
-        self.verify_noQC_warning("-q EscCol -d FakeModel -D 10 -L 10")
+        self.verify_noQC_warning("-q EscCol -d FakeModel -D 10 -L 1000")
 
     def verify_noQC_citations_not_written(self, cmd, caplog):
         # Non-QCed models shouldn't be used in publications, so citations
@@ -1011,14 +1011,14 @@ class TestNoQCWarning:
     @pytest.mark.usefixtures("caplog")
     def test_noQC_citations_not_written(self, caplog):
         self.verify_noQC_citations_not_written(
-            "EscCol -d FakeModel -D 10 -L 10", caplog
+            "EscCol -d FakeModel -D 10 -L 1000", caplog
         )
 
     @pytest.mark.filterwarnings("ignore::stdpopsim.QCMissingWarning")
     @pytest.mark.usefixtures("caplog")
     def test_noQC_citations_not_written_verbose(self, caplog):
         self.verify_noQC_citations_not_written(
-            "-vv EscCol -d FakeModel -D 10 -L 10", caplog
+            "-vv EscCol -d FakeModel -D 10 -L 1000", caplog
         )
 
 
--- a/tests/test_species.py
+++ b/tests/test_species.py
@@ -151,6 +151,15 @@ class SpeciesTestBase:
     def test_population_size_defined(self):
         assert self.species.population_size > 0
 
+    def test_default_gc(self):
+        for chrom in self.species.genome.chromosomes:
+            contig = self.species.get_contig(chrom.id)
+            assert contig.gene_conversion_rate is None
+            assert contig.gene_conversion_length is None
+            contig = self.species.get_contig(chrom.id, use_species_gene_conversion=True)
+            assert chrom.gene_conversion_rate == contig.gene_conversion_rate
+            assert chrom.gene_conversion_length == contig.gene_conversion_length
+
 
 class GenomeTestBase:
     """
@@ -253,38 +262,66 @@ class TestGetContig:
         assert isinstance(contig.recombination_map, msprime.RateMap)
 
     def test_contig_options(self):
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Cannot use genetic map"):
             # cannot use genetic map with generic contig
             self.species.get_contig(genetic_map="ABC")
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Cannot use length multiplier"):
             # cannot use length multiplier with generic contig
             self.species.get_contig(length_multiplier=0.1)
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Must specify sequence length"):
             # must specify length with generic contig or give chromosome name
             self.species.get_contig()
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Cannot specify sequence length"):
             # cannot specify length with named chromosome
             self.species.get_contig("chr1", length=1e6)
-        with pytest.raises(ValueError):
-            # cannot specify mask for generic contig
+        with pytest.raises(ValueError, match="Cannot use mask"):
+            # cannot specify inclusion mask for generic contig
             self.species.get_contig(length=1e4, inclusion_mask=[(0, 100)])
-        with pytest.raises(ValueError):
-            # connot specify mask for generic contig
+        with pytest.raises(ValueError, match="Cannot use mask"):
+            # cannot specify exclusion mask for generic contig
             self.species.get_contig(length=1e4, exclusion_mask=[(0, 100)])
-        with pytest.raises(ValueError):
-            # cannot use length multiplier with mask
+        with pytest.raises(ValueError, match="Cannot use length multiplier"):
+            # cannot use length multiplier with inclusion mask
             self.species.get_contig(
                 "chr22", inclusion_mask=[(0, 100)], length_multiplier=0.1
             )
-        with pytest.raises(ValueError):
-            # cannot use length multiplier with mask
+        with pytest.raises(ValueError, match="Cannot use length multiplier"):
+            # cannot use length multiplier with exclusion mask
             self.species.get_contig(
                 "chr22", exclusion_mask=[(0, 100)], length_multiplier=0.1
+            )
+        with pytest.raises(
+            ValueError, match="Cannot use species gene conversion rates"
+        ):
+            # cannot specify use_species_gene_conversion and custom gc rate
+            self.species.get_contig(
+                "chr1", use_species_gene_conversion=True, gene_conversion_rate=0.1
+            )
+        with pytest.raises(ValueError, match="without setting gene conversion length"):
+            # cannot specify custom gene conversion rate without gene conversion length
+            self.species.get_contig("chr1", gene_conversion_rate=0.1)
+        with pytest.raises(ValueError, match="without setting gene conversion rate"):
+            # cannot specify custom gene conversion length without gene conversion rate
+            self.species.get_contig("chr1", gene_conversion_length=1)
+
+    def test_use_species_gene_conversion(self):
+        contig = self.species.get_contig("chr22", use_species_gene_conversion=True)
+        if self.species.genome.get_chromosome("chr22").gene_conversion_rate is None:
+            assert contig.gene_conversion_rate is None
+            assert contig.gene_conversion_length is None
+        else:
+            assert (
+                contig.gene_conversion_rate
+                == self.species.genome.get_chromosome("chr22").gene_conversion_rate
+            )
+            assert (
+                contig.gene_conversion_length
+                == self.species.genome.get_chromosome("chr22").gene_conversion_length
             )
 
     def test_generic_contig(self):
         L = 1e6
-        contig = self.species.get_contig(length=L)
+        contig = self.species.get_contig(length=L, use_species_gene_conversion=True)
         assert contig.recombination_map.sequence_length == L
 
         chrom_ids = np.arange(1, 23).astype("str")
@@ -299,6 +336,22 @@ class TestGetContig:
             for c in self.species.genome.chromosomes
             if c.id in chrom_ids
         ]
+        gcs = [
+            c.gene_conversion_rate if c.gene_conversion_rate is not None else 0
+            for c in self.species.genome.chromosomes
+            if c.id in chrom_ids
+        ]
+        gcls = [
+            c.gene_conversion_length if c.gene_conversion_length is not None else 0
+            for c in self.species.genome.chromosomes
+            if c.id in chrom_ids
+        ]
 
         assert contig.mutation_rate == np.average(us, weights=Ls)
         assert contig.recombination_map.mean_rate == np.average(rs, weights=Ls)
+        if np.average(gcs, weights=Ls) == 0 or np.average(gcls, weights=Ls) == 0:
+            assert contig.gene_conversion_rate is None
+            assert contig.gene_conversion_length is None
+        else:
+            assert contig.gene_conversion_rate == np.average(gcs, weights=Ls)
+            assert contig.gene_conversion_length == np.average(gcls, weights=Ls)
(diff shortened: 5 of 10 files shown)


jeromekelleher:
Thanks @fbaumdicker, this is a great addition! There will be a quite a few tricky things to work out though, I fear.

Depending on release plans, I wonder if this should be postponed until after the next release? @petrelharp, @izabelcavassim, @grahamgower , what are the current plans for getting 0.2.0 out the door?

izabelcavassim:
I don't know what the current plans for when to release 0.2.0 are. But I would suggest that we only do that when we have:

- all already added species QCed
- finalized selection implementation

Are we willing to release it before the paper is written up, or?
Thanks for your contributions @fbaumdicker, I also like the idea of thinking about ploidy levels.

petrelharp:
As Izabel says, we haven't made release plans. But I think we'll be ready to do that in a month or two (ish)?

izabelcavassim:
I think a month is a good goal!

fbaumdicker:
At the hackathon @jeanrjc and I discussed with @petrelharp and @jeromekelleher that we first need the gene conversion feature in stdpopsim to include microbial species in a meaningful way. Otherwise, we are basically simulating one single tree. So what I am trying to say is that almost all microbial species can/should only pass a real QC, when the gene conversion rates are included. Thus, I tried to implement this before it is too late for the adding species project.
However, if you think we should postpone the addition of gc for a while that can of course easily be done. :wink:

petrelharp:
I'm happy to support getting it in! But I may not be thinking of the pitfalls that @jeromekelleher is?

fbaumdicker:
Cool, thanks @petrelharp.
Yes, @jeromekelleher will likely identify more tricky parts than I would.
I think here we should focus on the issues that might arise from gene conversion. I hope those are not too many.
Is there something particular you are concerned about @jeromekelleher?

For the possibly more involved ploidy addition, I opened a separate issue here #1111 .

jeromekelleher:
> I think here we should focus on the issues that might arise from gene conversion. I hope those are not too many.
Is there something particular you are concerned about @jeromekelleher?

Nothing specific, but there will surely be things we haven't considered and it feels like this release is already very big. What happens in the SLiM engine, for example - does that support ploidy?

grahamgower:
> What happens in the SLiM engine, for example - does that support ploidy?

Nope. This is possible in SLiM, but it's a bit of a hack.

fbaumdicker:
> > What happens in the SLiM engine, for example - does that support ploidy?
> 
> Nope. This is possible in SLiM, but it's a bit of a hack.

I think this is what @jeanrjc and @bhaller added to SLiM in this paper https://www.biorxiv.org/content/10.1101/2020.09.28.316869v5.full, but yes the SLiM engine in stdpopsim would still need to be adapted.

I also think that the selection parts and gc are not very related to each other, such that having them in different releases makes sense. On the other hand the adding species part would benefit from the gc part so having them in the same release would be nice.

bhaller:
> > What happens in the SLiM engine, for example - does that support ploidy?
> 
> Nope. This is possible in SLiM, but it's a bit of a hack.

In SLiM 3.7 (soon to be released) it is much less of a hack than it used to be, but it is still not simply a "flip of a switch" setting to say "this model is haploid rather than diploid".  But now, with just a line or two of custom script to enforce the chosen ploidy during reproduction (assuming you are using a nonWF model), you can make a haploid model extremely easily, and modeling things like alternation of generations or haplodiploidy is also quite easy.  Modeling ploidy above diploid is still a horrendous hack.  :->  I can share the relevant new bits of the draft manual for SLiM 3.7 if that would be helpful.

petrelharp:
From the meeting just now - open an issue to add gene conversion rates to the SLiM engine? Also perhaps open an issue to add GC rates to diploids (and we'll need to do some thinking there about the API, e.g., is it on *all* the time? or is it optional?)

petrelharp:
And, ping us on slack if you want some help figuring out the test failures?

fbaumdicker:
I hope I have now identified all the problems in the tests. However, there is still one problem that I would like to hear your opinion on before I fix it.

The remaining error occurs in 
```python
def test_species_simulation(species_id):
    cmd = f"-q {species_id} -L 1 --seed 1234 10"
    # Just check to see if the simulation runs
    with mock.patch("sys.stdout", autospec=True) as stdout:
        stdout.buffer = open(os.devnull, "wb")
stdpopsim.cli.stdpopsim_main(cmd.split())
```
The -L 1 parameter will cause msprime to terminate because the genome/conitg length is less than the mean gene conversion tract length for EscCol. But this error will also occur for the other species in the catalog if we add further gene conversion parameters.

In my opinion, there are now three possible ways to solve this problem:

A) We could simply increase the contig length here above the maximal gene conversion tract length. I guess -L 2000 would suffice. But this likely increases the running time of all tests, as it enables recombination into multiple trees.

B) We could change how msprime handles input when the length is below the mean gene conversion tract length.
We could e.g. only print a warning and simply proceed in msprime although length<gc tract length.
I don't believe there should be any problems in doing so. But we would have to check this in msprime of course and we would have to update the requirements of stdpopsim to the latest msprime version.

C) We could overwrite the gene conversion rate and tract length in the cmd line above. I did not check whether this is already possible or whether we would need to adapt the cli. I have something like:
```python
cmd = f"-q {species_id} -L 1 -c 0 0 --seed 1234 10"
```
in mind, where -c 0 0 would overwrite the gene conversion rate and length to 0.

Currently, I prefer the third option, but I would like to hear your opinion @jeromekelleher and @petrelharp, before I proceed.

jeromekelleher:
Thanks @fbaumdicker. Here's my thoughts:

A) is likely a non-starter as it would make the test require a very long time
B) Would take a long time to do, and I'm not sure it's the right approach in msprime anyway (we do think it's an error, right?)
C) We could do that, but that undermines the point of the test, which is supposed to check if the default simulations will run.

What if we use the ``-l/length-multiplier`` option rather than the fixed size L? I think that was probably what it was doing before, so perhaps that won't work either.

We could probably just delete the test I'd imagine, it's not doing anything particularly useful anyway and is going to be harder and harder to make work as more things get added to the catalog.

fbaumdicker:
I finally managed to come back to this. There is now only one failing test in the slim environment for gene conversion. Would you mind looking at the corresponding parts @bhaller? I guess, I need to go along the route you showed in the slim manual under 6.1.3. But in slim the gene conversion seems to work as a subset of the recombinations using the local recombination map, whereas in msprime gc occurs always at a constant rate along the chromosome even if a recombination map has been specified.

bhaller:
Hi @fbaumdicker.  The SLiM gene conversion model is described in detail in section 1.5.6 of the SLiM manual.  Yes, the rate for it follows the recombination rate map; my impression is that that is biologically accurate, but I'm very far from an expert in gene conversion.  :->  If you need a variable recombination rate along the chromosome, but a constant rate of gene conversion along the chromosome, that would be quite hard to set up in SLiM, I think; it is not a supported mode of operation (since, as I understand it, that is not how the biology works).

Section 6.1.3 is the starting point for using gene conversion in a model, yes.  It's a simple example of how to set up gene conversion in SLiM.

I don't know how to look into the problem you are seeing with the tests you have set up; I'm not a stdpopsim user at all.  But if you post a complete SLiM model here that runs under SLiMgui (i.e., does not require additional command-line parameters), and describe to me in what way it fails, I'd be happy to look at it and see what I can see.  :->  But is the failure you're getting a runtime error of some kind, or is the failure that the model does not produce the same results as msprime?  If the latter, then it sounds like maybe the recombination rate map difference is responsible, and I'm not sure what you ought to do about that.  :-O

petrelharp:
Awesome! Let's see. From Codecov, it looks to me like the only thing not being hit by the tests is the case when the `gene_conversion_rate` argument to `Contig( )` is *not* `None` (yellow colors on those if statements):
Screenshot from 2022-02-18 10-50-12
Looking at the diff, I can't quite tell *why* that's not being hit, though - but, perhaps we should add a very simple test in `test_genomes.py` that creates an anonymous Contig with gene conversion? And, maybe an addition to `test_EscCol.py`?

tests/test_cli.py
```diff
@@ -1011,23 +1011,12 @@ def verify_noQC_citations_not_written(self, cmd, caplog):
...
 
     @pytest.mark.filterwarnings("ignore::stdpopsim.QCMissingWarning")
     @pytest.mark.usefixtures("caplog")
     def test_noQC_citations_not_written_verbose(self, caplog):
         self.verify_noQC_citations_not_written(
-            "-vv EscCol -d FakeModel -D 10 -L 10", caplog
+            "-vv EscCol -d FakeModel -D 10 -L 1000", caplog
         )
-
-
```
petrelharp, line 1024: why are you removing this test, btw?

bhaller:
> Thanks for the details of SLiM @bhaller. If I understood it correctly I would say:
> As bacteria do not recombine and only use gene conversion, we could still use SLiM for prokaryotes with gene conversion. For eukaryotes, only using the same recombination maps for gene conversion and recombination will work. When talking to biologists I got the feeling that most of them think of gene conversion initiation anyway this way.
> However as a side note:
> In https://doi.org/10.1371/journal.pgen.1002905 for Drosophila the recombination map in Figure 1 and the gene conversion initiation rate along the chromosome in Figure 5 are clearly different. So in the long term the usage of two different "recombination maps" in SLiM, one for cross-over recombination and another one for gene conversion, might be worth looking at.

Hi @fbaumdicker.  If what you want to model is horizontal gene transfer (HGT) in bacteria (I don't like using the term "gene conversion" for that, since – as I understand it – it is a totally different mechanism), see section 16.14 of the SLiM manual.  You don't model HGT with SLiM's gene conversion feature since that feature only relates to diploids (as discussed in section 1.5.6; it roughly models the way Holliday junctions work during meiosis, I guess, although my understanding of all this stuff is rudimentary at best).  Your plan for eukaryotes sounds good to me.  Re: Drosophila and rate maps, interesting.  You are the first to bring this up; if you think it will matter to you guys then please open a GitHub issue for it on SLiM.  Otherwise I will wait for someone else to request it as a feature; I've got enough to do without looking for trouble :->.

But I think it would be pretty difficult to implement anyway, so it would not be likely to happen any time soon; and of course it's always important to keep in mind that perfect biological realism is an unattainable goal, so you always want to consider whether a particular aspect of that realism is really essential to one's model or not (but maybe it is!).

bhaller:
For a more advanced version of section 16.14's recipe, that actually models a circular chromosome, see https://peercommunityjournal.org/articles/10.24072/pcjournal.72/, by the way.

tests/test_cli.py
fbaumdicker: Normally, I wouldn't dare to remove a test 😉. But above @jeromekelleher suggested:
> We could probably just delete the test I'd imagine, it's not doing anything particularly useful anyway and is going to be harder and harder to make work as more things get added to the catalog.

fbaumdicker:
> One question - I haven't quite figured out how the default works, as compared to recombination rate. For `Contig` anyhow, the default rate is `None` - what does this do, and is it the same as recombination rate?

I'm not sure I understood the point of your question @petrelharp.
If a species has a Contig with a gene conversion rate that is None, the simulation will not include gene conversion, regardless of the recombination_rate. Only when a gene conversion rate and a gene conversion length are set, the simulation will include gc. Again, completely independent of the recombination parameters.

So when the gene conversion feature in stdpopsim is ready, we will have to add the corresponding default mean gene conversion rates and lengths from the literature to all species in the catalog where we consider this important.

petrelharp:
Sorry - I was trying to review the code, and had a hard time tracking down whether `None` would lead to no gene conversion or the default gene conversion rate - the behavior looks different [here](https://github.com/popsim-consortium/stdpopsim/pull/1106/files#diff-4be8425825d01514a6443535a13867b234ed9d2ba64b90f7012a794cf8233042R375) and [here](https://github.com/popsim-consortium/stdpopsim/pull/1106/files#diff-4be8425825d01514a6443535a13867b234ed9d2ba64b90f7012a794cf8233042R106). Probably everything is fine, I don't have the big picture here. (and, sorry I didn't see this earlier)

petrelharp:
Bigger-picture, do we just need some more tests here?

fbaumdicker:
> Sorry - I was trying to review the code, and had a hard time tracking down whether `None` would lead to no gene conversion or the default gene conversion rate - the behavior looks different [here](https://github.com/popsim-consortium/stdpopsim/pull/1106/files#diff-4be8425825d01514a6443535a13867b234ed9d2ba64b90f7012a794cf8233042R375) and [here](https://github.com/popsim-consortium/stdpopsim/pull/1106/files#diff-4be8425825d01514a6443535a13867b234ed9d2ba64b90f7012a794cf8233042R106). Probably everything is fine, I don't have the big picture here. (and, sorry I didn't see this earlier)

The difference is intentional @petrelharp.

In one case, we want for example create a generic unnamed contig that inherits the mean recombination and mutation rates -- and now the mean gene conversion rates of all named chromosomes in the species. So to calculate the mean gene conversion rate for chromosomes with and without gc, I need to consider a rate of 0 if the gene_conversion_rate for a chromosome is None.

In the other case, the gene conversion rate is None, i.e. not specified, when generating a modified named contig with the species_contig() function. In this case, I want to inherit the gene_conversion_rate from the named contig.

fbaumdicker:
I have now included more tests, mainly in the EscCol species, that cover the missing cases.
From my side, this would be ready to merge.

stdpopsim/cli.py
```diff
@@ -481,9 +489,7 @@ def add_simulate_species_parser(parser, species):
             help=(
                 "Specify a particular genetic map. By default, a chromosome-specific "
                 "uniform recombination rate is used. These default rates are listed in "
-                "the catalog: "
-                "<https://popsim-consortium.github.io/stdpopsim-docs/"
-                "stable/catalog.html> "
+                "the catalog: <https://stdpopsim.readthedocs.io/en/latest/catalog.html> "
```
petrelharp, line 492: whoops - it looks like you're reverting some changes? We're using the github.io link for docs now.

petrelharp (approved):
LGTM!

stdpopsim/cli.py
petrelharp, line 492:
```suggestion
                "the catalog: "
                "<https://popsim-consortium.github.io/stdpopsim-docs/"
                "stable/catalog.html> "
```

stdpopsim/cli.py
```diff
@@ -790,8 +796,8 @@ def run_simulation(args):
                     "and the model described in the original publication. "
                     "More information about the QC process can be found in "
                     "the developer documentation. "
-                    "https://popsim-consortium.github.io/stdpopsim-docs/"
-                    "latest/development.html#demographic-model-review-process"
+                    "https://stdpopsim.readthedocs.io/en/latest/development.html"
+                    "#demographic-model-review-process"
```
petrelharp, line 800:
```suggestion
                    "https://popsim-consortium.github.io/stdpopsim-docs/"
                    "latest/development.html#demographic-model-review-process"
```

petrelharp:
I guess I have one other thing that I think should be in here - could you put in the docstring in the appropriate method(s) for `gene_conversion_rate` something like "if None (the default) then the species mean gene conversion rate will be used; set to 0.0 to disable gene conversion"?

petrelharp:
On slack there was discussion about making "no GC" the default. I can think of some arguments both against and in favor of this - so that everyone's models don't suddenly start taking lots longer to run; because we don't have GC in SLiM models yet (#1114). How's it currently work, and what do you think, @fbaumdicker?

grahamgower (approved):
LGTM too.

grahamgower:
I think GC should be the default. GC support can be added to the SLiM engine without too much difficulty. On slack, I recall there being some worry that for some species we aren't confident about the quality of GC rate/length estimates---I say we just don't include GC values for such species.

fbaumdicker:
On slack, I argued just as @grahamgower that at least for bacteria GC should be the default and we should only include gc parameters into the catalog when we have high-quality estimates.

A compromise could be to deactivate gc by default when a non-constant recombination map is used and activate gc by default when the recombination map is constant, which is typically the case for bacteria. But that sounds more confusing to me.

fbaumdicker:
I have now changed the behavior as follows:
If no parameters or options are specified, no gene conversion is simulated.
If use_species_gene_conversion is set to True, the conversion rate and length from the species catalog are used.
If gene_conversion_rate **and** gene_conversion_length are set to custom values and use_species_gene_conversion is False, the custom parameters will be used.
Anything else will cause an error.

So this is ready to merge from my side :)

Translated with www.DeepL.com/Translator (free version)

petrelharp:
Great! For the record: this is the plan we discussed over in #1115.

stdpopsim/species.py
```diff
@@ -197,12 +198,22 @@ def get_contig(
...
+            conversion of the species chromosome are used if available. For "generic"
+            contigs the gene conversion rate is given by the (mean) rate specified by
+            species chromosomes and the gene conversion length defaults to the mean
+            length specified by species chromosomes.
+        :param float gene_conversion_rate: The per-base gene conversion rate. Can only
+            be set together with gene_conversion_length and if
+            use_species_gene_conversion is False; If none is given, and
+            use_species_gene_conversion is True the gene conversion rate specified by
+            species chromosomes is used. If none and use_species_gene_conversion is
+            False gene conversion is disabled.
```
petrelharp, line 211: A bit more concise?
```suggestion
        :param float gene_conversion_rate: The per-base gene conversion rate.
            If unspecified (and use_species_gene_conversion is False) then no gene
            conversion will occur. If specified, gene_conversion_length must also be given.
```

stdpopsim/species.py
```diff
@@ -197,12 +198,22 @@ def get_contig(
...
+        :param float gene_conversion_length: The mean gene conversion length. Can only
+            be set together with gene_conversion_rate and if use_species_gene_conversion
+            is False; If none is given, and use_species_gene_conversion is True the
+            tract length specified by species chromosomes is used. If none and
+            use_species_gene_conversion is False gene conversion is disabled.
```
petrelharp, line 216:
```suggestion
        :param float gene_conversion_length: The mean gene conversion length.
            May only be provided if gene_conversion_rate is also specified.
```

tests/test_EscCol.py
```diff
@@ -18,18 +19,38 @@ def test_basic_attributes(self):
...
-        assert genome.gene_conversion_rate == 0
+        assert genome.gene_conversion_rate == 0.1
         assert genome.gene_conversion_length == 1
 
+    def test_use_species_gene_conversion_unnamed_contig(self):
+        genome = self.species.get_contig(length=42, use_species_gene_conversion=True)
+        assert genome.gene_conversion_rate == 8.9e-11
+        assert genome.gene_conversion_length == 345
+
+    def test_use_species_gene_conversion_unnamed_contig_undefined_gc(self):
```
petrelharp, line 46: Oh good - I'm glad you're testing this, but it should probably be a test in test_species.py? (see suggestion below)

tests/test_species.py
```diff
@@ -281,6 +281,17 @@ def test_contig_options(self):
             self.species.get_contig(
                 "chr22", exclusion_mask=[(0, 100)], length_multiplier=0.1
             )
```
petrelharp, line 283: Shouldn't we also add a test here that if we `get_contig` with `use_species_gene_conversion` then the resulting rate and length agree with the species property (including if they are None)? That'll remove the need for the deepcopy in test_EscCol, I think?

petrelharp:
Awesome! Two minor comments. I'm getting ready to head out on vacation for next week, else I'd have a go at them.

tests/test_species.py
petrelharp: Ah, right - I got confused and thought this was testing all the contigs. I guess that could be a test up in `GenomeTestBase`? (or, is it `SpeciesTestBase`?)

petrelharp: Hm - I added this test, and turned up an error (I think) - unless I'm wrong, I think that we're not actually disabling gene conversion by default. (we need additional logic in `species_contig( )`, I believe)

tests/test_species.py
```diff
@@ -151,6 +151,15 @@ def test_generation_time_defined(self):
     def test_population_size_defined(self):
         assert self.species.population_size > 0
 
+    def test_default_gc(self):
+        for chrom in self.species.genome.chromosomes:
+            contig = self.species.get_contig(chrom.id)
+            assert chrom.gene_conversion_rate is None
```
fbaumdicker, line 157: I think this line should be 
`assert contig.gene_conversion_rate is None`
and
`assert contig.gene_conversion_length is None`
and this should fix the test error.

tests/test_species.py
fbaumdicker: Thanks for adding this. If I am not mistaken, the error appeared only because of a small bug, see my other comment.

tests/test_species.py
petrelharp: hah, good catch!

tests/test_EscCol.py
```diff
@@ -16,6 +17,40 @@ def test_basic_attributes(self):
         generation_time = 1.0 / (525600 / 20)
         assert round(abs(self.species.generation_time - generation_time), 7) == 0
```
petrelharp, line 19: Terribly sorry, but one more thing: these next five tests are still bugging me. I think they're good, but they should be in `test_genome.py`, not in `test_EscCol.py`. The reason for this is that when someone else comes along to implement a new bacterial species, they'll look at EscCol and cpoy what's there, thinking they need to do all these tests, which are really tests of the basic machinery. Conversely, when someone goes and looks in test_genome.py to see how the basic machinery is being tested, they won't see it. So - how about just move these over there? They can be unchanged, just need a `species = stdpopsim.get_species("EscCol")` at the start?

fbaumdicker: Don't worry, you are right, the best place for this is `test_genomes.py`.
I have moved the tests there now. It took me only a few minutes to do that. :)

petrelharp: Thanks! I would have done it, but was traveling...

petrelharp:
I'm going to rebase, squash, and merge this!!!

petrelharp:
Ok - it was a big rebase; perhaps you can take a look to see if I messed anything up, @fbaumdicker?

fbaumdicker:
looks good to me :)

tests/test_cli.py
```diff
@@ -1011,23 +1011,12 @@ def verify_noQC_citations_not_written(self, cmd, caplog):
...
         )
 
     @pytest.mark.filterwarnings("ignore::stdpopsim.QCMissingWarning")
     @pytest.mark.usefixtures("caplog")
     def test_noQC_citations_not_written_verbose(self, caplog):
         self.verify_noQC_citations_not_written(
-            "-vv EscCol -d FakeModel -D 10 -L 10", caplog
+            "-vv EscCol -d FakeModel -D 10 -L 1000", caplog
         )
-
```
petrelharp, line 1023: whoops! Didn't mean to remove these lines!

petrelharp:
Whoops, codecov made me notice some errors I'd introduced; glad I caught those. If the tests pass now, I'll merge.

PR #1110, opened 2021-11-15
git rebase instructions update

petrelharp:
After a painful attempt at a rebase over at #1052 with @izabelcavassim I realized that the "what if git rebase goes wrong" instructions could be improved. As written, it gets the diff to `upstream/main`. **However**, if followed literally, this will *undo* all changes made to `upstream/main` since the PR diverged from it! So, the right thing to do (I think) is to first squash the commits, and *then* merge them in. I've rewritten the docs to reflect this (although note that [the git docs](https://git-scm.com/docs/git-merge-base) suggest an alternative workflow, like:
```
fork_point=$(git merge-base --fork-point origin/master topic)
git rebase --onto origin/master $fork_point topic
```

I've left the current recommendation to do a patch and commit because it seems... less magical? But we could rewrite it to be all git.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -446,13 +446,39 @@ and have successfully squashed and rebased.
 When rebasing goes wrong
 ------------------------
 
-Sometimes rebasing goes wrong, and you end up in a frustrating loop of making and
-undoing the same changes over and over again. In this case, it can be simplest to
-make a diff of your current changes, and apply these in a single commit. First
-we take the diff between the current state of the files in our branch and
-``upstream/main`` and save it as a patch::
+Sometimes rebasing goes wrong, and you end up in a frustrating loop of making
+and undoing the same changes over and over again. First, here's an explanation
+of what's going on. Let's say that the branch we're working on (and trying to
+rebase) is called ``topic_branch``, and it branched off from ``upstream/main``
+at some point in the past::
 
-    $ git diff upstream/main > changes.patch
+         A1---A2---A3  (topic_branch)
+        /
+    ---M---o---o---o---o---B  (upstream/main)
+
+So, what we'd really like to do is to take the commits ``A1``, ``A2``, and
+``A3`` and apply them to the current state of the ``upstream/main`` branch,
+i.e., on top of commit ``B``. If we just do ``git rebase upstream/main``
+then git will try to first apply ``A1``; then ``A2``; and finally ``A3``.
+If there's conflicts, this is painful, so we might want to *first* squash
+the three commits together into one commit, and then rebase that single commit.
+Then we'll only have to resolve conflicts once. Said another way: we often
+use ``git rebase -i upstream/main`` to both squash *and* rebase; but
+it may be easier to squash first then rebase after.
+
+We'll be doing irreversible changes, so first we should make a backup copy of
+the branch::
+
+    $ git checkout topic_branch  # make sure we're on the right branch
+    $ git checkout -b topic_backup # make the backup
+    $ git checkout topic_branch  # go back to the topic branch
+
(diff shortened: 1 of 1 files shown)


petrelharp:
Note: I'm a bit confused on this, because I've used those instructions before and not had anything bad happen. Now I don't understand why not.  But maybe that was because I hadn't updated `upstream/main`, so that `upstream/main` was still pointing at the MRCA of my branch and upstream's `main`?

andrewkern:
so if we have a network like this 
```
	 o---o---o---B
	/
---X---o---o---o---o---A
```
as we assume `A` is the main branch and that `B` the topic branch, wouldn't we instead want to create a patch using:

```
git diff —merge-base A > my.patch
git checkout -b test_branch upstream/main
patch -p1 < my.patch
```

wouldn't that get us to a place where all of the changes in the topic branch are now in `A`?

petrelharp:
After consulting with @andrewkern I added somre more explanation. Maybe this is helpful?

petrelharp:
Ah, I didn't know about `git diff --merge-base`
```
           If --merge-base is given, instead of using <commit>, use the merge base of <commit> and HEAD.  git diff --merge-base A is equivalent to git diff $(git merge-base A HEAD).
```

petrelharp:
So that'll do the same thing as what I've got in this PR, except when "merge base" and "fork point" are different, which they *shouldn't* be in our workflow, IIUC. Should I switch it out, do you think?

andrewkern:
yeah i think you are right that they are equivalent in our use case, but yes i also think you should swap in the `--merge-base` usage cause it's clean. i've tried to add it via a code suggestion.

docs/development.rst
```diff
@@ -446,13 +446,39 @@ and have successfully squashed and rebased.
...
+
+    $ git checkout topic_branch  # make sure we're on the right branch
+    $ git checkout -b topic_backup # make the backup
+    $ git checkout topic_branch  # go back to the topic branch
+
+Next, we take the diff between the current state of the files and the place
+where your changes last diverged from ``upstream/main`` (the commit labelled
+``M`` in the diagram above), and save it as a patch::
+
+    $ fork_point=$(git merge-base --fork-point upstream/main topic_branch)
```
andrewkern, line 480:
```suggestion
git diff --merge-base A > changes.patch
git checkout -b test_branch upstream/main
patch -p1 < changes.patch
```

petrelharp: that should be `upstream/main` not `A` so I'm not merging this suggestion, but yes

grahamgower:
Hmm. I guess the new instructions here are better, at least in some circumstances. Presumably folks are getting into difficulties because of trying to follow the diff approach when there are conflicts between the topic branch and the latest upstream main? In any event there are lots of ways one can end up with inconsistencies and my instict would be to tell folks to take a backup of the topic branch while referring them to more authoritative git docs.

jeromekelleher (approved):
LGTM - nice to have the explanation, and definitely an improvement. But as @grahamgower said there's probably an infinite number of ways that things can go wrong and it's not just dealing with conflicts we're avoiding by taking the patch.

docs/development.rst
petrelharp: I don't think this is right either - pushing an updated version now.

petrelharp:
Ok, I think this is good to go. There's two changes here: 
1. additional explanation
2. changing `git diff upstream/main` to `git diff --merge-base upstream/main`.

The second thing is the only *functional* change. But this was I think a real bug in the previous instructions.

jeromekelleher:
Yeah ``git diff --merge-base upstream/main`` is a great addition!

PR #1116, opened 2021-11-17
DFE CLI accepting bed file

### LINKED ISSUE #1112: Add more ways of specifying DFE intervals in CLI
We have just implemented the --dfe-intervals in #1052 and would like to make this feature even more flexible

So the user will only be loading up a single contig worth of annotation using the custom bed feature 
That means that we can just say
intervals = np.loadtxt("foo.bed",usecols=[1,2])
and then intervals will have shape (n, 2)
that’s all there is, and then we just have to pass that to add_dfe
@andrewkern's comments above. 

I think we have a lot of names already in the CLI, so maybe we can just use --dfe-interval as a feature for both string "10000,200000", and for a bed name: "foo.bed"?
We then check whether the file is parsable by a comma, if so it is because it is only one string interval ("10000,200000"), if it is a .bed format, maybe it would be good that the user uses ".bed" in the file's name?

izabelcavassim:
closes #1112

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -633,11 +633,25 @@ def add_simulate_species_parser(parser, species):
             )
         if dfe is not None:
             dfe = species.get_dfe(args.dfe)
-            left, right = dfe_interval.split(",")
-            contig.add_dfe(
-                intervals=np.array([[int(left), int(right)]]),
-                DFE=dfe,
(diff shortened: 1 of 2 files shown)


izabelcavassim:
Closing because I didn't want to bother working on the conflicts, wrote a new PR instead.

PR #1117, opened 2021-11-17
setting up dfe annotations for CLI

andrewkern:
Starting to draft the CL interface we will want for using annotations with DFEs. following the current convention, i've introduced a `--dfe-annotation` option, that allows the user to give one of the specified annotations and apply it chromosome-wide.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -104,6 +104,13 @@ def get_genetic_map_wrapper(species, genetic_map_id):
         exit(str(ve))
 
 
+def get_annotation_wrapper(species, annotation_id):
+    try:
+        return species.get_annotations(annotation_id)
+    except ValueError as ve:
+        exit(str(ve))
+
+
 def get_dfe_wrapper(species, dfe_id):
     try:
         return species.get_dfe(dfe_id)
@@ -242,6 +249,42 @@ def get_species_help(species_id):
         f"Recombination rate: {species.genome.mean_recombination_rate:.4g}\n"
     )
     return species_text
+
+
+class HelpAnnotations(argparse.Action):
+    """
+    Action used to produce genetic map help text.
+    """
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        help_text = get_annotations_help(namespace.species, values)
+        print(help_text, file=sys.stderr)
+        parser.exit()
+
+
+def get_annotations_help(species_id, annotation_id):
+    """
+    Generate help text for the given annotation. If annotation_id is None, generate
+    help for all annotations. Otherwise, it must be a string with a valid
+    annotation ID.
+    """
+    species = stdpopsim.get_species(species_id)
+    if annotation_id is None:
+        annotation_text = f"\nAll annotations for {species.name}\n\n"
+        annotations = [annotation.id for annotation in species.annotations]
+    else:
+        annotations = [annotation_id]
+        annotation_text = "\nAnnotation  description\n\n"
+
+    indent = " " * 4
+    wrapper = textwrap.TextWrapper(initial_indent=indent, subsequent_indent=indent)
+    for an_id in annotations:
+        annot = get_annotation_wrapper(species, an_id)
+        annotation_text += f"{annot.id}\n"
+        annotation_text += wrapper.fill(textwrap.dedent(annot.description))
+        annotation_text += "\n\n"
+
+    return annotation_text
 
 
 def get_environment():
@@ -385,7 +428,7 @@ def add_simulate_species_parser(parser, species):
         help=f"Run simulations for {species.name}.",
     )
     species_parser.set_defaults(species=species.id)
-    species_parser.set_defaults(genetic_map=None)
+    species_parser.set_defaults(genetic_map=None, dfe_annotation=None)
     species_parser.add_argument(
         "--help-models",
         action=HelpModels,
@@ -566,6 +609,32 @@ def add_simulate_species_parser(parser, species):
         help=interval_help,
     )
 
+    annot_choices = [gm.id for gm in species.annotations]
+    if len(species.annotations) > 0:
+        species_parser.add_argument(
+            "--help-annotations",
+            action=HelpAnnotations,
+            nargs="?",
+            help=(
+                "Print list of annotations and exit. If a annotation ID is "
+                "given as an argument, show help for this annotation. Otherwise show "
+                "help for all available annotations"
+            ),
+        )
+    if len(species.annotations) > 0:
+        species_parser.add_argument(
+            "--dfe-annotation",
+            choices=annot_choices,
+            metavar="",
+            default=None,
+            help=(
+                "Specify the intervals over which a given DFE acts "
+                "using a genomic annotation track "
+                "Available maps: "
+                f"{', '.join(annot_choices)}. "
+            ),
+        )
+
     species_parser.add_argument(
         "-o",
         "--output",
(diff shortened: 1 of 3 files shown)


stdpopsim/cli.py
```diff
@@ -625,22 +694,49 @@ def run_simulation(args):
...
-                map(str, [0, int(contig.recombination_map.sequence_length)])
-            )
-        if dfe is not None:
+        # DFE assignment
+        dfe = None
+        dfe_summary_str = None
+        if args.dfe is None:
+            if args.dfe_interval is not None:
+                exit(
+                    "A DFE interval has been assigned without a DFE. "
```
izabelcavassim, line 703: I like that you send informative messages, we did not do that. We need to make sure we test these errors.

stdpopsim/cli.py
```diff
@@ -649,7 +745,7 @@ def run_simulation(args):
             contig=contig,
             samples=samples,
             dfe=args.dfe,
-            dfe_interval=dfe_interval,
+            dfe_interval=dfe_summary_str,  # not sure what we want to print
```
izabelcavassim, line 748: me neither, here we would want to print I guess:

- if the string "1000,2000", then just: "[1000,2000)"
- if a bed file, then np.min(array), np.max(array)
- if interval is not defined then contig length

stdpopsim/cli.py
```diff
@@ -732,7 +828,7 @@ def write_simulation_summary(
     dry_run_text += f"{indent}Genetic map: {gmap}\n"
     if dfe is not None:
         dry_run_text += f"{indent}DFE: {dfe}\n"
-        dry_run_text += f"{indent}DFE in interval: {dfe_interval}\n"
+        dry_run_text += f"{indent}DFE applied to: {dfe_interval}\n"
```
izabelcavassim, line 831: see comment above

tests/test_cli.py
```diff
@@ -835,6 +839,16 @@ def test_bad_dfe(self):
...
 
+    def test_bad_annotation(self):
+        species = stdpopsim.get_species("HomSap")
+        with mock.patch("stdpopsim.cli.exit", autospec=True) as mocked_exit:
+            cli.get_annotation_wrapper(species, "666_foo_666")
+            available_annots = [dm.id for dm in species.annotations]
+            avail_annots_str = ", ".join(available_annots)
+            mocked_exit.assert_called_once_with(
+                f"Annotations 'HomSap/666_foo_666' not in catalog ({avail_annots_str})"
+            )
```
izabelcavassim, line 850: add testes that would lead to the error messages you have included here.
But as you suggested to me last time, we can add most of the tests into test_slim_engine.py

izabelcavassim:
I just created a PR about the bed file thing #1116. They are conflicted, and I still don't exactly know what features we exactly would want to have on this.
@grahamgower thinks that is not useful to add a string interval (e.g "100,2000") as we do in #1052, 
@petrelharp likes the feature and me too. 
I suggested that both a string setting interval and a bed file name could be added with --dfe-interval, @grahamgower  thinks it is prone to errors. Can we all agree on something?

stdpopsim/cli.py
andrewkern: think this is all set, however i'm not dealing with the bed file in this PR.... not sure how to stage those merges

tests/test_cli.py
andrewkern: 👍

stdpopsim/cli.py
```diff
@@ -625,22 +694,49 @@ def run_simulation(args):
...
-            left, right = dfe_interval.split(",")
             contig.add_dfe(
-                intervals=np.array([[int(left), int(right)]]),
+                intervals=intervals,
                 DFE=dfe,
             )
             logger.info(
                 f"Applying selection under the DFE model {dfe.id} "
-                f"in interval [{left}, {right})."
+                f"in intervals {intervals}."
```
petrelharp, line 739: Isn't this going to print out something absolutely huge when we're using an Annotation? Would it be better to put the summary string here?

stdpopsim/cli.py
```diff
@@ -625,22 +694,49 @@ def run_simulation(args):
...
+        else:
+            if args.dfe_interval is not None:
+                if args.dfe_annotation is not None:
+                    exit(
+                        "A DFE annotation and a DFE interval have been "
+                        "selected. Please only use one "
+                    )
+                left, right = args.dfe_interval.split(",")
+                intervals = np.array([[int(left), int(right)]])
+                dfe_summary_str = f"[{left}, {right})"
```
petrelharp, line 720: This is really the `intervals_summary_str`, not the `dfe_summary_str`.

stdpopsim/cli.py
petrelharp: I think what you're printing is good?

tests/test_cli.py
petrelharp: Hm - why are those CLI tests in test_slim.py, not test_cli.py?

tests/test_slim_engine.py
```diff
@@ -373,6 +373,52 @@ def count_mut_types(ts):
...
+        assert n_mut_types[1] > 0
+        # test DFE with interval and no dfe
+        with tempfile.NamedTemporaryFile(mode="w") as f:
+            cmd = (
+                f"-q -e slim --slim-scaling-factor 20 --slim-path {slim_path} "
+                f"HomSap -c chr22 -l0.01 -o {f.name} "
+                "-d OutOfAfrica_3G09 "
+                "--dfe-interval 1000,100000 10"
+            ).split()
+            with pytest.raises(SystemExit):
```
petrelharp, line 399: If you do
```suggestion
           with pytest.raises(SystemExit, match="interval has been assigned without a DFE"):
```
then that'll test that the right error is called.

stdpopsim/cli.py
```diff
@@ -625,22 +694,49 @@ def run_simulation(args):
...
+                exit(
+                    "A DFE annotation has been assigned without a DFE. "
+                    "Please specify a DFE "
+                )
+        else:
+            if args.dfe_interval is not None:
+                if args.dfe_annotation is not None:
+                    exit(
+                        "A DFE annotation and a DFE interval have been "
+                        "selected. Please only use one "
```
petrelharp, line 716:
```suggestion
                        "selected. Please only use one."
```

stdpopsim/cli.py
```diff
@@ -625,22 +694,49 @@ def run_simulation(args):
...
+        if args.dfe is None:
+            if args.dfe_interval is not None:
+                exit(
+                    "A DFE interval has been assigned without a DFE. "
+                    "Please specify a DFE "
+                )
+            elif args.dfe_annotation is not None:
+                exit(
+                    "A DFE annotation has been assigned without a DFE. "
+                    "Please specify a DFE "
```
petrelharp, line 709:
```suggestion
                    "Please specify a DFE."
```

stdpopsim/cli.py
```diff
@@ -625,22 +694,49 @@ def run_simulation(args):
...
-            )
-        if dfe is not None:
+        # DFE assignment
+        dfe = None
+        dfe_summary_str = None
+        if args.dfe is None:
+            if args.dfe_interval is not None:
+                exit(
+                    "A DFE interval has been assigned without a DFE. "
+                    "Please specify a DFE "
```
petrelharp, line 704:
```suggestion
                    "Please specify a DFE."
```

petrelharp (approved):
Some minor suggestions.

petrelharp:
Also TODO on this topic (not for this PR):
- write an example for the docs of using an Annotation
- make it so you can pass the name of an annotation to `contig.add_DFE( )`?

tests/test_cli.py
andrewkern: good question, but `test_slim.py` has it's own `TestCLI` class. let's refactor that in a separate PR

stdpopsim/cli.py
andrewkern: yep

tests/test_slim_engine.py
andrewkern: nice

Issue #1123, opened 2021-11-27
Correct labels of populations on demesdraw figures?


sunyatin:
Dear all,

I feel like the ordering of population labels for several of the Homo sapiens demographic models (eg. PapuansOutOfAfrica_10J19, OutOfAfricaArchaicAdmixture_5R19) that are represented in the [doc](https://popsim-consortium.github.io/stdpopsim-docs/main/catalog.html#sec_catalog_HomSap_models) are suspicious... I feel like they correspond to the order of the populations_configuration, but not to the actual ordering that demesdraw performs. For instance, for [OutOfAfricaArchaicAdmixture_5R19](https://popsim-consortium.github.io/stdpopsim-docs/main/_images/sec_catalog_homsap_models_outofafricaarchaicadmixture_5r19.png), the pops are ordered in the figure, from left to right as:

CHB, YRI, Neanderthal, ArchaicAFR, CEU

But I would rather see:

ArchaicAFR, YRI, CEU, CHB, Neanderthal

What do you think?

Cheers,
Rémi

grahamgower:
Yes I think you're right. Ping @aabiddanda.

aabiddanda:
Great spot @sunyatin - it was a remarkably simple fix and an oversight on my part. I've put in a PR to fix this: #1124 just now. If you have any comments on the catalog or the images feel free to let me know!

grahamgower:
Closed by #1124.

PR #1124, opened 2021-11-28
Pushed fix for error in #1123

aabiddanda:
This resolves the error described in #1123 and now corresponds to the exact same `xticklabels` used by `demesdraw`.


grahamgower:
Thanks @aabiddanda!

PR #1130, opened 2021-12-01
Fix broken link to status badge

Chris1221:
Sorry to make a whole PR here, I just noticed that at some point the links to the CI badges got broken (and the link to the CircleCI page lead to the project landing page rather than the SVG of the badge so it wouldn't render). Feel free to reject and just fix whenever next updated, just wanted to throw in a quick fix if no one noticed.

Chris


grahamgower:
Thanks @Chris1221.

Issue #1149, opened 2021-12-15
don't simulate neutral mutations in SLiM


petrelharp:
@mufernando points out that it looks like we are actually simulating neutral mutations twice, once with SLiM and once with msprime. The fix should be to change [this line](https://github.com/popsim-consortium/stdpopsim/blob/1c489090a6dde9414dc1b8825cdb2d0d540dbe3c/stdpopsim/slim_engine.py#L946) to something like:
```
        mut_props = ", ".join([str(prop) if mt.is_neutral() else 0.0 for prop, mt in d.proportions])
```

We also should have a test that catches this, by doing a SLiM simulation without recapitation and checking if any actual mutations are of the neutral types.

Ah - the test could be added to [this one](https://github.com/popsim-consortium/stdpopsim/blob/1c489090a6dde9414dc1b8825cdb2d0d540dbe3c/tests/test_slim_engine.py#L987), checking that all neutral mutation types have 0.0 proportion. If so, then we could go ahead and recapitate and still run the test.

mufernando:
test that neutral mutations have proportion 0 in SLiM, similar to this: https://github.com/popsim-consortium/stdpopsim/blob/1c489090a6dde9414dc1b8825cdb2d0d540dbe3c/tests/test_slim_engine.py#L987

mufernando:
Peter, it looks like this test you linked is doing the right thing. But none of these tests were being run because we were waiting for SLiM 3.7. See how you were computing the proportion of neutral mutations and excluding them from the mutation rate here: https://github.com/popsim-consortium/stdpopsim/blob/1c489090a6dde9414dc1b8825cdb2d0d540dbe3c/tests/test_slim_engine.py#L997

PR #1151, opened 2021-12-16
fixing neutral mutations

mufernando:
Fix to #1149

We were simulating neutral mutations within SLiM before. The fix was not as straightforward as @petrelharp pointed in the issue because we cannot have a genomic element type with all 0.0 proportions in SLiM. I added tests to this behavior (one of which is redundant, but is more obviously correct so I thought it wouldn't hurt? Also made some minor cosmetic changes throughout...

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -623,11 +623,11 @@ def get_msp_and_slim_mutation_rate_maps(contig):
             sum(
                 [
                     p
-                    for p, mt in zip(g.proportions, g.mutation_types)
-                    if not mt.is_neutral
+                    for p, mt in zip(d.proportions, d.mutation_types)
+                    if (not mt.is_neutral)
                 ]
             )
-            for g in contig.dfe_list
+            for d in contig.dfe_list
         ]
         + [0]
     )  # append 0 for the -1 labels
@@ -637,7 +637,6 @@ def get_msp_and_slim_mutation_rate_maps(contig):
     msp_mutation_rate_map = msprime.RateMap(
         position=breaks, rate=contig.mutation_rate * (1 - slim_fractions[dfe_labels])
     )
-
     return (msp_mutation_rate_map, (slim_breaks, slim_rates))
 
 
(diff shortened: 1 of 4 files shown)


stdpopsim/slim_engine.py
```diff
@@ -943,7 +943,14 @@ def matrix2str(
                 # T is the default for WF simulations.
                 printsc(f"    m{mid}.convertToSubstitution = F;")
         mut_types = ", ".join([str(mt) for mt in mut_type_list])
-        mut_props = ", ".join([str(prop) for prop in d.proportions])
+        mut_props_list = [
+            prop if (not mt.is_neutral) else 0.0
+            for prop, mt in zip(d.proportions, d.mutation_types)
+        ]
+        mut_props = ", ".join(map(str, mut_props_list))
+        # SLiM does not let you define a GET with all 0 proportions
```
mufernando, line 951: need to test this behavior

tests/test_slim_engine.py
```diff
@@ -1003,7 +1004,7 @@ def verify_mutation_rates(self, contig, ts):
             self.verify_rates_match(mut_rate, intervals, slim_rates, slim_ends)
 
         # also verify any bits not covered by DFEs have no mutations
-        empty_intervals = np.empty((0, 2))
+        empty_intervals = np.array([[0, int(contig.length)]])
```
mufernando, line 1007: Idk why I did this, but at one point the `_check_intervals_validity` was failing on an empty numpy array. either way, this seems more natural to me?

tests/test_slim_engine.py
```diff
@@ -1473,6 +1492,7 @@ def test_drawn_mutation_meets_AF_threshold(self):
                 seed=seed,
             )
             assert ts.num_mutations == 1
+            ts = pyslim.SlimTreeSequence(ts)
```
mufernando, line 1495: we only need pyslim treeseq in this test because it uses `individuals_alive_at`.

mufernando:
The only test failing is because of pyslim.

tests/test_slim_engine.py
petrelharp: Hm - well, this is not the same thing and the error was because `np.empty( )` needed `dtype='int'`... but, I think you're right! This was in error, and your array is the correct one.

Issue #1148, opened 2021-12-15
store annotation information in metadata


petrelharp:
Proposal (with @mufernando):
In the SLiM script we know what mutation type goes in each genomic region, but we don't use that information later when adding neutral mutations with msprime.
So, we should in fact store that information in top-level metadata, and use the information to tell msprime which mutation types to add.

To do this, we need to:

1. store in metadata, for each DFE, (a) the intervals it applies to, (b) and the mutation types it has, including the SLiM ID and whether it's being simulated by SLiM
2. 
2. when we add mutations with msprime, we no longer have to re-compute the msp_rate_map, instead we loop over the mutation types within each dfe, and, if it's neutral, we add the mutations with msprime.

nspope:
Closing as I think we have this in the metadata (it's used during recapitation). Although, we may not want to document as it could well change.

PR #1152, opened 2021-12-20
Dfes in ts metadata

mufernando:
The idea of this PR is to add metadata to tree sequences produced by the SLiM engine. I'm focusing on adding information about the DFEs onto the tree sequence and this will have two main uses:

- It will allow us to streamline the way we overlay neutral mutations with `sim_mutations`. The metadata in the ts will encode information about which mutation types SLiM did not simulate and we can simply iterate over those and add them with `sim_mutations`. Importantly, this will make so that the slim mutation type ids are the same as the ones in the original `Contig`, so we won't have to create new mutation types in the "recapitate and mutate" step.
- The metadata is also useful for downstream analyses.  For instance, this will allow people to figure out which intervals have which mutation types and at what proportions/rates. In #1148 @petrelharp describes our most pressing use case, that @xin-huang needs to know which mutations are neutral within exons to benchmark the DFE inference software.

PS: note this PR is based off of #1157, so some of the changes displayed here are from the other PR. as soon as that is merged I can rebase onto main.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -57,6 +57,7 @@ import numpy as np
 import msprime
 import pyslim
 import tskit
+import json
 
 logger = logging.getLogger(__name__)
 
@@ -602,12 +603,62 @@ _slim_debug_output = """
 
 """
 
+_raw_stdpopsim_top_level_schema = {
+    "stdpopsim": {
+        "description": "Top-level metadata for a tree sequence produced by stdposim using the SLiM engine",  # noqa: E501
+        "type": "object",
+        "properties": {
+            "DFEs": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "id": {"type": "string"},
+                        "description": {"type": "string"},
+                        "long_description": {"type": "string"},
+                        "mutation_types": {
+                            "type": "array",
+                            "items": {
+                                "type": "object",
+                                "properties": {
+                                    "dominance_coeff": {"type": "number"},
+                                    "distribution_type": {"type": "string"},
+                                    "distribution_args": {
+                                        "type": "array",
+                                        "items": {"type": ["number", "string"]},
+                                    },
+                                    "convert_to_substitution": {"type": "boolean"},
+                                    "Q_scaled_index": {
+                                        "type": "array",
+                                        "items": {"type": "number"},
+                                    },
+                                    "slim_mutation_type_id": {"type": "integer"},
+                                    "is_neutral": {"type": "boolean"},
+                                },
+                            },
+                        },
+                        "proportions": {"type": "array", "items": {"type": "number"}},
+                        "citations": {
+                            "type": "array",
+                            "items": {"type": ["string", "object"]},
+                        },
+                        "intervals": {
+                            "type": "array",
+                            "items": {"type": "array", "minItems": 2, "maxItems": 2},
+                        },
+                        "start_time": {"type": "number"},
+                        "end_time": {"type": "number"},
+                    },
+                },
+            }
+        },
+    }
+}
 
-def get_msp_and_slim_mutation_rate_maps(contig):
+
+def get_slim_mutation_rate_map(contig):
     """
-    Returns a :class:`msprime.RateMap` with the mutation rate map for overlay
-    of neutral mutations and a tuple with the breakpoints and rates of mutations
-    for SLiM.
+    Returns a tuple with the breakpoints and rates of mutations for SLiM.
     """
     breaks, dfe_labels = contig.dfe_breakpoints()  # beware -1 labels
     slim_fractions = np.array(
@@ -625,11 +676,7 @@ def get_msp_and_slim_mutation_rate_maps(contig):
     )  # append 0 for the -1 labels
     slim_rates = contig.mutation_rate * slim_fractions[dfe_labels]
     slim_breaks = breaks[1:] - 1  # SLiM has inclusive right endpoints
-
-    msp_mutation_rate_map = msprime.RateMap(
-        position=breaks, rate=contig.mutation_rate * (1 - slim_fractions[dfe_labels])
-    )
-    return (msp_mutation_rate_map, (slim_breaks, slim_rates))
+    return (slim_breaks, slim_rates)
 
 
 def slim_array_string(iterable, indent, width=80):
@@ -646,6 +693,74 @@ def slim_array_string(iterable, indent, width=80):
         )
         + ")"
     )
+
+
+def _enum_dfe_and_intervals(contig):
+    """
+    Returns an iterator which, for every `DFE` in the `contig`, returns a tuple
+    containing a numeric id for said DFE, the DFE object and the list of
+    intervals associated with the DFE. We need to assign each DFE a numeric id
+    both in the slim_makescript and in the _add_dfes_to_metadata, so with this
+    function we can ensure these two steps assigns numeric ids to the DFEs in the
+    same way.
+    """
+    assert len(contig.dfe_list) == len(contig.interval_list)
+    for i, (d, ints) in enumerate(zip(contig.dfe_list, contig.interval_list)):
+        yield i, d, ints
+
+
+def _dfe_to_mtypes(contig):
+    """
+    Assigns a mutation type id to each of the mutation types contained in this
+    `Contig`. This function will return a dictionary with `len(contig.dfe_list)`
+    elements, in which the position of the DFE in `contig.dfe_list` is the key.
+    For each DFE, the dictionary holds a list of tuples that contain the an
+    assigned mutation type id (used in  SLiM) and the `MutationType` object.
+    This is necessary so that we use the same numeric ids to the mutation types
+    in the _add_dfes_to_metadata and slim_makescript steps.
+    """
+    mid = 0
+    dfe_to_mtypes = {}
+    for i, d, _ in _enum_dfe_and_intervals(contig):
+        dfe_to_mtypes[i] = []
+        for mt in d.mutation_types:
+            dfe_to_mtypes[i].append((mid, mt))
+            mid += 1
+    return dfe_to_mtypes
+
+
+def _add_dfes_to_metadata(ts, contig):
+    """
+    Adds the DFEs to the top-level metadata of a tree sequence using
+    information in the `contig`.
+    """
+
+    def _get_json(obj):
+        # recursively flattens an object to a dictionary.
+        return json.loads(
+            json.dumps(obj, default=lambda o: getattr(o, "__dict__", str(o)))
+        )
+
+    tables = ts.dump_tables()
+    schema = tables.metadata_schema.asdict()
+    metadata = tables.metadata
+    schema["properties"].update(_raw_stdpopsim_top_level_schema)
+    dfes = _get_json(contig.dfe_list)
+    dfe_to_mtypes = _dfe_to_mtypes(contig)
+    for i, d, ints in _enum_dfe_and_intervals(contig):
+        dfes[i]["intervals"] = ints.tolist()
+        dfes[i]["proportions"] = d.proportions
+        for j, (mid, mt) in enumerate(dfe_to_mtypes[i]):
+            add = {
+                "slim_mutation_type_id": mid,
+                "is_neutral": mt.is_neutral,
+            }
+            dfes[i]["mutation_types"][j].update(add)
+    new_metadata = {"stdpopsim": {"DFEs": dfes}}
+    metadata.update(new_metadata)
+    tables.metadata_schema = tskit.MetadataSchema(schema)
+    tables.metadata = metadata
+    return tables.tree_sequence()
 
 
 def msprime_rm_to_slim_rm(recombination_map):
@@ -671,7 +786,6 @@ def slim_makescript(
     slim_rate_map,
 ):
 
-    mutation_types = contig.mutation_types()
     pop_names = [pop.name for pop in demographic_model.model.populations]
     # Use copies of these so that the time frobbing below doesn't have
     # side-effects in the caller's model.
@@ -780,14 +894,17 @@ def slim_makescript(
     fitness_callbacks = []
     condition_on_allele_frequency = []
     op_id = stdpopsim.ext.ConditionOnAlleleFrequency.op_id
+    slim_mutation_ids = []
+    for x in _dfe_to_mtypes(contig).values():
+        slim_mutation_ids.extend([u[0] for u in x])
     for ee in extended_events:
         if hasattr(ee, "mutation_type_id"):
             mt_id = getattr(ee, "mutation_type_id")
             cls_name = ee.__class__.__name__
-            if not (0 <= mt_id < len(mutation_types)):
+            if mt_id not in slim_mutation_ids:
                 raise ValueError(
-                    f"Invalid {cls_name} {ee.mutation_type_id} {len(mutation_types)} "
-                    f"event with mutation type id {mt_id}.{contig}"
+                    f"Invalid {cls_name} event with mutation type id "
+                    f"{mt_id} on {contig} with valid IDs {slim_mutation_ids}."
                 )
         if hasattr(ee, "start_time") and hasattr(ee, "end_time"):
             # Now that GenerationAfter times have been accounted for, we can
@@ -919,11 +1036,12 @@ def slim_makescript(
         return "".join(s)
 
     # Genomic element types.
-    next_mutation_type_id = 0
-    for j, d in enumerate(contig.dfe_list):
+    dfe_mtypes = _dfe_to_mtypes(contig)
+    for j, d, ints in _enum_dfe_and_intervals(contig):
         # Mutation types.
         mut_type_list = []
-        for m in d.mutation_types:
+        assert len(dfe_mtypes[j]) == len(d.mutation_types)
+        for mid, m in dfe_mtypes[j]:
             if len(m.Q_scaled_index) >= 1:
                 distrib_args = [str(arg) for arg in m.distribution_args]
                 for k in m.Q_scaled_index:
@@ -934,8 +1052,6 @@ def slim_makescript(
             # dealing with distributions given by "s" Eidos script,
             else:
                 distrib_args = "; ".join([f'"{a}"' for a in m.distribution_args])
-            mid = next_mutation_type_id
-            next_mutation_type_id += 1
             mut_type_list.append(mid)
             printsc(
                 f"    initializeMutationType({mid}, {m.dominance_coeff}, "
@@ -945,6 +1061,12 @@ def slim_makescript(
                 # T is the default for WF simulations.
                 printsc(f"    m{mid}.convertToSubstitution = F;")
         mut_types = ", ".join([str(mt) for mt in mut_type_list])
+        # We will assign a proportion of 0.0 to mutation types that are neutral
+        # unless all the mutation types in a DFE are neutral. In such case,
+        # SLiM would not allow all mutation types in a genomic element type to
+        # have 0.0 proportion. And the proportions in SLiM won't matter anyway,
+        # because all the intervals in this fully neutral DFE will have 0.0
+        # mutation rate anyway.
         mut_props_list = [
             prop if (not mt.is_neutral) or d.is_neutral else 0.0
             for prop, mt in zip(d.proportions, d.mutation_types)
@@ -958,11 +1080,11 @@ def slim_makescript(
                 f"    // Note: genomic element type {j} is entirely neutral,"
                 " so will not be simulated by SLiM."
             )
-        if contig.interval_list[j].shape[0] > 0:
-            element_starts = slim_array_string(contig.interval_list[j][:, 0], indent)
+        if ints.shape[0] > 0:
+            element_starts = slim_array_string(ints[:, 0], indent)
             # stdpopsim intervals are 0-based left inclusive, right exclusive, but
             # SLiM intervals are right inclusive
-            element_ends = slim_array_string(contig.interval_list[j][:, 1] - 1, indent)
+            element_ends = slim_array_string(ints[:, 1] - 1, indent)
             printsc(
                 f"    initializeGenomicElement({j}, {element_starts}, {element_ends});"
             )
@@ -1190,6 +1312,7 @@ class _SLiMEngine(stdpopsim.Engine):
         slim_burn_in=10.0,
         dry_run=False,
         verbosity=None,
+        _recap_and_rescale=True,
     ):
         """
         Simulate the demographic model using SLiM.
@@ -1239,7 +1362,7 @@ class _SLiMEngine(stdpopsim.Engine):
 
         # figuring out mutations in SLiM and neutral mutations to be added
         # by msprime later
-        msp_rate_map, slim_rate_map = get_msp_and_slim_mutation_rate_maps(contig)
+        slim_rate_map = get_slim_mutation_rate_map(contig)
 
         # TODO: remove this after a release or two. See #745.
         self._warn_zigzag(demographic_model)
@@ -1289,9 +1412,11 @@ class _SLiMEngine(stdpopsim.Engine):
 
             ts = tskit.load(ts_file.name)
 
-        ts = self._recap_and_rescale(
-            ts, seed, recap_epoch, contig, slim_scaling_factor, msp_rate_map
-        )
+        ts = _add_dfes_to_metadata(ts, contig)
+        if _recap_and_rescale:
+            ts = self._recap_and_rescale(
+                ts, seed, recap_epoch, contig, slim_scaling_factor
+            )
 
         if contig.inclusion_mask is not None:
             ts = stdpopsim.utils.mask_tree_sequence(ts, contig.inclusion_mask, False)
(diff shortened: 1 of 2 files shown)


stdpopsim/slim_engine.py
```diff
@@ -599,6 +599,47 @@
 
 """
 
+_raw_stdpopsim_top_level_schema = {
```
mufernando, line 602: Can you give this schema a quick look @petrelharp ?

petrelharp:
We need to change `ts.reference_sequence` to `ts.reference_sequence.data`.

mufernando:
Note some of the changes here are from my previous PR.

stdpopsim/slim_engine.py
petrelharp: Let's also stick in an `id`, since [all DFEs have IDs](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/dfe.py#L176) and something we'd really like to get out of the metadata is which mutation type goes with which DFE?

petrelharp: And - how about the mutation type entries look exactly like what we get if we do `mut_type.asdict()`?

stdpopsim/slim_engine.py
```diff
@@ -943,7 +982,11 @@ def matrix2str(
...
         mut_types = ", ".join([str(mt) for mt in mut_type_list])
-        mut_props = ", ".join([str(prop) for prop in d.proportions])
+        mut_props_list = [
+            prop if (not mt.is_neutral) or d.is_neutral else 0.0
+            for prop, mt in zip(d.proportions, d.mutation_types)
+        ]
+        mut_props = ", ".join(map(str, mut_props_list))
         printsc(
             f"    initializeGenomicElementType({j}, c({mut_types}), c({mut_props}));"
         )
```
petrelharp, line 992: How about append here:
```
if d.is_neutral:
    printsc(f"    // Note: genomic element type {j} is entirely neutral, so will not be simulated by SLiM.")
```

tests/test_slim_engine.py
```diff
@@ -1020,12 +1020,31 @@ def verify_genomic_elements(self, contig, ts):
         ge_types = self.slim_metadata_key0(
             ts.metadata["SLiM"]["user_metadata"], "genomicElementTypes"
         )
+        # getting the ids for the neutral mutation types
+        neutral_ids = [
+            mtdict["id"]
+            for mtdict in contig.mutation_types()
```
petrelharp, line 1026: Now that we put this information into metadata, we can (and I think *should* entirely remove this [`Contig.mutation_types()`](https://popsim-consortium.github.io/stdpopsim-docs/latest/api.html#stdpopsim.Contig.mutation_types) method - so, can we use something else here?

tests/test_slim_engine.py
```diff
@@ -1020,12 +1020,31 @@ def verify_genomic_elements(self, contig, ts):
...
         assert len(contig.dfe_list) == len(ge_types)
         for j, (dfe, intervals) in enumerate(
             zip(contig.dfe_list, contig.interval_list)
         ):
             assert str(j) in ge_types
             ge = self.slim_metadata_key0(ge_types, str(j))
+            # figuring out if all mutation types are neutral in this GET
+            all_neutral = all(
+                [mtype_id in neutral_ids for mtype_id in ge["mutationTypes"]]
+            )
```
petrelharp, line 1039:
```suggestion
            all_neutral = all(
                [mt.is_neutral for mt in dfe.mutation_types]
            )
```

tests/test_slim_engine.py
```diff
@@ -1020,12 +1020,31 @@ def verify_genomic_elements(self, contig, ts):
...
+            # checking that the neutral mutations have 0.0 proportion unless
+            # all the mutations are neutral in this dfe
+            for mtype_id, prop in zip(ge["mutationTypes"], ge["mutationFractions"]):
+                if mtype_id in neutral_ids:
+                    if all_neutral:
+                        assert prop > 0.0
+                    else:
+                        assert np.isclose(prop, 0.0)
```
petrelharp, line 1047: One might have an "all neutral" DFE with a proportion that is 0.0, so this isn't quite right. How about:
```suggestion
            for mt, prop in zip(dfe.mutation_types, ge["mutationFractions"]):
                if mt.is_neutral or all_neutral:
                         assert np.isclose(prop, 0.0)
```

tests/test_slim_engine.py
```diff
@@ -1003,7 +1003,7 @@ def verify_mutation_rates(self, contig, ts):
             self.verify_rates_match(mut_rate, intervals, slim_rates, slim_ends)
 
         # also verify any bits not covered by DFEs have no mutations
-        empty_intervals = np.empty((0, 2))
+        empty_intervals = np.array([[0, int(contig.length)]])
```
petrelharp, line 1006: This is not an empty interval any more? We're needing to check that anything *not covered* by an interval works right, and if we start with these intervals we won't have any empty intervals.

petrelharp:
Looks good to me, after a few fixes!

petrelharp:
I think this is ready to go now - I've rebased it, even - but, have a look, and let's see about the tests.

stdpopsim/slim_engine.py
mufernando: yes, good call! maybe also add the description to the DFE?

there is no `asdict` method to `MutationType`. But the schema is basically the same as `mut_type.__dict__` with three more properties? `slim_mutation_type_id`, `is_neutral`, `simulated_in_SLiM` which are needed for Xin's analyses.

tests/test_slim_engine.py
mufernando: No, that would trigger an error in SLiM actually (you cannot have a genomic element type with all the proportions equal to 0). So I should add a check to the `slim_engine` to catch that.

stdpopsim/slim_engine.py
mufernando: note that the `mutation_types` are nested within each DFE in this schema, so you can tell which mt is with each DFE.

tests/test_slim_engine.py
petrelharp: That's not what I mean - I mean an all neutral DFE where some of the mutation types have nonzero proportion and others have a zero proportion. (e.g., the proportions are `[1.0, 0.0]`).

petrelharp:
Sorry - did I mess things up?

mufernando:
No, I fucked up in merging your commits I think. Ideas to test the metadata?

petrelharp:
We're testing for consistency between metadata and contig already - so, just add a few more `assert`s to that same code?

stdpopsim/slim_engine.py
```diff
@@ -704,6 +697,79 @@ def slim_array_string(iterable, indent, width=80):
...
+    have 0.0 mutation rate anyway, as computed by
+    `get_msp_and_slim_mutation_rate_maps`.
+    """
+    return [
+        prop if _simulate_mutation_type_in_slim(mt) or d.is_neutral else 0.0
+        for prop, mt in zip(d.proportions, d.mutation_types)
+    ]
+
+
+def add_top_level_metadata(ts, contig):
```
petrelharp, line 723: Wait, what's the idea here? We want SLiM itself to be storing this information, because then we don't have to coordinate between the python and SLiM code so much?

mufernando:
I need input on these two things for now:

- The metadata schema in `_raw_stdpopsim_top_level_schema`
- the function `add_top_level_metadata`
- Note I created some internal use functions (`_func_name`) that allow us to link the `slim_makescript` step with the post-SLiMulation step that will add the metadata. I don't like that the way we define the mutation type ids in `add_top_level_metadata` and `slim_makescript` are independent from one another. One alternative here would be to use `Contig.mutation_types()` in both? But that method is kinda messy and @petrelharp wanted to get rid of it.

stdpopsim/slim_engine.py
```diff
@@ -648,6 +697,68 @@ def slim_array_string(iterable, indent, width=80):
     )
 
 
+def _simulate_mutation_type_in_slim(mt):
```
petrelharp, line 700: I don't think this function helps?

petrelharp:
> The metadata schema in _raw_stdpopsim_top_level_schema

This looks good!

petrelharp:
Oh, wait - that metadata schema looks good... but, do we want it? We don't *need* to set a metadata schema; the default schema says "anything goes"; should we change it?

petrelharp:
Hm, ok - we've two options:
1. (what you've done) Put metadata in `ts.metadata["stdpopsim"]` by editing the tree sequence after we get it back from SLiM.
2. (what I was imagining) Put metadata in `ts.metadata["SLiM"]["user_metadata"]` by asking SLiM to put the relevant information there.

Clearly, `ts.metadata["stdpopsim"]` is a better place for the metadata. And, we might want to start storing other, non-SLiM metadata. However, it's real nice to just have SLiM tell us exactly what it did rather than have to reconstruct what we think it did.  I suppose we could ask SLiM to write out the information, then move it over to `ts.metadata["stdpopsim"]`? Shall I code up what I mean?

petrelharp:
We've just discussed this issue in the coding session, whether to ask SLiM to put the information in metadata or to do it ourselves. Asking SLiM to do it is a little more elegant... *but* then it would end up under `ts.metadata["SLiM"]["user_metadata"]`, instead of under `ts.metadata["stdpopsim"]` (as in this PR). Since we'll want to put other non-SLiM-specific things in metadata as well, the latter is clearly better. So, I'll plan on fixing up and getting this PR in.

mufernando:
lmk how to help @petrelharp.

mufernando:
> I need input on these two things for now:
> 
> * The metadata schema in `_raw_stdpopsim_top_level_schema`
> * the function `add_top_level_metadata`
> * Note I created some internal use functions (`_func_name`) that allow us to link the `slim_makescript` step with the post-SLiMulation step that will add the metadata. I don't like that the way we define the mutation type ids in `add_top_level_metadata` and `slim_makescript` are independent from one another. One alternative here would be to use `Contig.mutation_types()` in both? But that method is kinda messy and @petrelharp wanted to get rid of it.

I think these notes need to be answered for us to move forward!

petrelharp:
Good timing, I'm just getting to this also.  Here's my thoughts:

- the metadata schema looks good.
- The method of adding top-level metadata is good, but the method should be not be part of the public API.
- Good idea for factoring out things like "is this simulated in SLiM" and "proportions".  
 
So, I think having the private functions that we call in two places (script generation and metadata construction) is a good way to go, but thinkng a bit more, how about we get the mutationIDs (the trickiest bit) in a slightly different way: ask SLiM to store the information in it's own metadata (i.e., under `ts.metadata['SLiM']['user_metadata']`), but do *not* document this, and use that to fill out our own metadata (under `ts.metadata['stdpopsim']`). It'd be easy, it's not a big deal the information is in there twice (esp if we don't put in the `intervals`), and would guard against future changes in SLiM.

I'm happy to have a crack at that if you want.

mufernando:
Ok, so I finished the code part (no tests yet). I did not do what you suggested @petrelharp (to use the SLiM metadata to get at the mutation type ids) because it felt a bit wasteful to me? That would work for the mutation type ids, but we also need to know the intervals. So it felt weird to use the SLiM user metadata to retrieve the mutation type ids but not the intervals? Also if we did things this way then there would be some leftover/undocumented metadata under `["SLiM"]["user_metadata"]` that would be a partial duplicate of what the final ts would have under `["stdpopsim"]["DFEs"]`.

My solution is far from pretty though. A bunch of hidden functions that deal with all the nastiness...

stdpopsim/slim_engine.py
mufernando: maybe not? it felt better to me to have this explicitly somewher and then if in the future we decide to change this behavior then we will have to change it just once?

stdpopsim/slim_engine.py
```diff
@@ -648,6 +697,92 @@ def slim_array_string(iterable, indent, width=80):
...
+    """
+    for i, (d, ints) in enumerate(zip(contig.dfe_list, contig.interval_list)):
+        yield i, d, ints
+
+
+def _dfe_to_mtypes_dict(contig):
+    """
+    Assigns a mutation type id to each of the mutation types contained in this
+    Contig. This function will return a dictionary with `len(contig.dfe_list)`
+    elements with the DFE ids as keys. For each DFE id, the dictionary holds a
```
mufernando, line 721: keys are not dfe ids, but the position of the dfe in the list.

petrelharp: should be list, not a dict?

stdpopsim/slim_engine.py
```diff
@@ -648,6 +700,92 @@ def slim_array_string(iterable, indent, width=80):
...
+    Returns true if this `stdpopsim.MutationType` is to be simulated by SLiM.
+    """
+    return not mt.is_neutral
+
+
+def _enum_dfe_and_intervals(contig):
+    """
+    Returns an iterator which, for every `DFE` in the `contig`, returns a tuple
+    containing a numeric id for said DFE, the DFE object and the list of
+    intervals associated with the DFE.
```
petrelharp, line 714: explain why this is useful

stdpopsim/slim_engine.py
```diff
@@ -648,6 +700,92 @@ def slim_array_string(iterable, indent, width=80):
...
+    tables = ts.dump_tables()
+    schema = tables.metadata_schema.asdict()
+    metadata = tables.metadata
+    schema["properties"].update(_raw_stdpopsim_top_level_schema)
+    dfes = _get_json(contig.dfe_list)
+    dfe_to_mtypes = _dfe_to_mtypes_dict(contig)
+    for i, d, ints in _enum_dfe_and_intervals(contig):
+        print(i, d)
+        dfes[i]["intervals"] = ints.tolist()
+        dfes[i]["proportions"] = _mutation_types_proportions_slim(d)
```
petrelharp, line 774: this should be DFE proportions, not SLiM proportions

stdpopsim/slim_engine.py
```diff
@@ -648,6 +700,92 @@ def slim_array_string(iterable, indent, width=80):
...
+        )
+
+    tables = ts.dump_tables()
+    schema = tables.metadata_schema.asdict()
+    metadata = tables.metadata
+    schema["properties"].update(_raw_stdpopsim_top_level_schema)
+    dfes = _get_json(contig.dfe_list)
+    dfe_to_mtypes = _dfe_to_mtypes_dict(contig)
+    for i, d, ints in _enum_dfe_and_intervals(contig):
+        print(i, d)
```
mufernando, line 772: remove print

stdpopsim/slim_engine.py
```diff
@@ -1422,7 +1561,7 @@ def _recap_and_rescale(
         # Creating a SLiM mutation model
         # TODO: fix this up to do the right types? or, share them?
         model = msprime.SLiMMutationModel(
-            type=len(contig.mutation_types()), next_id=max_id + 1
```
mufernando, line 1425: need to fix this. find which mutation type wasn't simulated by SLiM and add them here!

stdpopsim/slim_engine.py
```diff
@@ -1422,7 +1561,7 @@ def _recap_and_rescale(
         # Creating a SLiM mutation model
         # TODO: fix this up to do the right types? or, share them?
```
petrelharp, line 1562: do something like
```
for d in dfe_list:
  for mt in d.mutation_types:
     if not mt.simulated_in_slim:
       # add it, with type mt.mutation_type
```

stdpopsim/slim_engine.py
mufernando: I think to be consistent with the other steps that use `_enum_dfe_and_intervals` we should leave this as a dictionary. In practice it does the same thing a list would do, but in the future if we decide to change the way we index DFEs it would be easier to change.

stdpopsim/slim_engine.py
mufernando: fixed!

stdpopsim/slim_engine.py
petrelharp: Ok, I guess? It seems confusing to have a dict indexed by integers. At least, the docstring needs to be fixed.

mufernando:
I think this is ready for review. I added one test to check the metadata also.

stdpopsim/slim_engine.py
```diff
@@ -648,6 +706,98 @@ def slim_array_string(iterable, indent, width=80):
...
+def _enum_dfe_and_intervals(contig):
+    """
+    Returns an iterator which, for every `DFE` in the `contig`, returns a tuple
+    containing a numeric id for said DFE, the DFE object and the list of
+    intervals associated with the DFE. We need to assign each DFE a numeric id
+    both in the slim_makescript and in the _add_dfes_to_metadata, so with this
+    function we can ensure these two steps assigns numeric ids to the DFEs in the
+    same way.
+    """
+    for i, (d, ints) in enumerate(zip(contig.dfe_list, contig.interval_list)):
```
petrelharp, line 725:
```suggestion
    assert len(contig_dfe_list) == len(contig.interval_list)
    for i, (d, ints) in enumerate(zip(contig.dfe_list, contig.interval_list)):
```

stdpopsim/slim_engine.py
```diff
@@ -602,6 +603,63 @@
...
+                                    "is_neutral": {"type": "boolean"},
+                                    "simulated_in_SLiM": {"type": "boolean"},
+                                },
+                            },
+                        },
+                        "proportions": {"type": "array", "items": {"type": "number"}},
+                        "slim_proportions": {
+                            "type": "array",
+                            "items": {"type": "number"},
+                        },
```
petrelharp, line 645:
```suggestion
```
This seems like an internal detail that shouldn't be in metadata?

stdpopsim/slim_engine.py
```diff
@@ -648,6 +706,98 @@ def slim_array_string(iterable, indent, width=80):
...
+    tables = ts.dump_tables()
+    schema = tables.metadata_schema.asdict()
+    metadata = tables.metadata
+    schema["properties"].update(_raw_stdpopsim_top_level_schema)
+    dfes = _get_json(contig.dfe_list)
+    dfe_to_mtypes = _dfe_to_mtypes(contig)
+    for i, d, ints in _enum_dfe_and_intervals(contig):
+        dfes[i]["intervals"] = ints.tolist()
+        dfes[i]["proportions"] = d.proportions
+        dfes[i]["slim_proportions"] = _mutation_types_proportions_slim(d)
```
petrelharp, line 786: If we don't put these in metadata then we don't need `_mutation_types_proprotions_slim` to be factored out.

stdpopsim/slim_engine.py
```diff
@@ -780,14 +929,16 @@ def fix_time(event):
     fitness_callbacks = []
     condition_on_allele_frequency = []
     op_id = stdpopsim.ext.ConditionOnAlleleFrequency.op_id
+    num_mutation_types = sum([len(d.mutation_types) for d in contig.dfe_list])
```
petrelharp, line 932:
```suggestion
    slim_mutation_ids = []
    for x in _dfe_to_mtypes(contig).values():
        slim_mutation_ids.extend([u[0] for u in x])
```

stdpopsim/slim_engine.py
```diff
@@ -780,14 +929,16 @@ def fix_time(event):
...
     for ee in extended_events:
         if hasattr(ee, "mutation_type_id"):
             mt_id = getattr(ee, "mutation_type_id")
             cls_name = ee.__class__.__name__
-            if not (0 <= mt_id < len(mutation_types)):
+            if not (0 <= mt_id < num_mutation_types):
```
petrelharp, line 937:
```suggestion
            if not mt_id in slim_mutation_ids:
```

stdpopsim/slim_engine.py
```diff
@@ -780,14 +929,16 @@ def fix_time(event):
...
                 raise ValueError(
-                    f"Invalid {cls_name} {ee.mutation_type_id} {len(mutation_types)} "
-                    f"event with mutation type id {mt_id}.{contig}"
+                    f"Invalid {cls_name} {ee.mutation_type_id} "
+                    f"{num_mutation_types} event with mutation type id "
+                    f"{mt_id}.{contig}"
```
petrelharp, line 941:
```suggestion
                    f"Invalid {cls_name} event with mutation type id "
                    f"{mt_id} on {contig} with valid IDs {slim_mutation_ids}."
```

stdpopsim/slim_engine.py
```diff
@@ -934,8 +1087,9 @@ def matrix2str(
             # dealing with distributions given by "s" Eidos script,
             else:
                 distrib_args = "; ".join([f'"{a}"' for a in m.distribution_args])
-            mid = next_mutation_type_id
-            next_mutation_type_id += 1
+            # ensuring mutation type ids are unique
+            assert mid != last_mid
+            last_mid = mid
```
petrelharp, line 1092: We don't need to do this - if we were to check this somewhere, it should be in ` _dfe_to_mtypes`, but there it's totally obvious.
```suggestion
```

stdpopsim/slim_engine.py
```diff
@@ -919,11 +1070,13 @@ def matrix2str(
...
     # Genomic element types.
-    next_mutation_type_id = 0
-    for j, d in enumerate(contig.dfe_list):
+    dfe_mtypes = _dfe_to_mtypes(contig)
+    for j, d, ints in _enum_dfe_and_intervals(contig):
         # Mutation types.
         mut_type_list = []
-        for m in d.mutation_types:
+        assert len(dfe_mtypes[j]) == len(d.mutation_types)
+        last_mid = -1
```
petrelharp, line 1078:
```suggestion
```

stdpopsim/slim_engine.py
```diff
@@ -1376,20 +1530,46 @@ def _recap_and_rescale(
         Apply post-SLiM transformations to ``ts``. This rescales node times,
         does recapitation, simplification, and adds neutral mutations.
         """
+        # Adjusting metadata (rescaling times and adding recap DFE
         # Node times come from SLiM generation numbers, which may have been
         # divided by a scaling factor for computational tractability.
```
petrelharp, line 1535:
```suggestion
        # Times come from SLiM generation numbers, which may have been
        # divided by a scaling factor for computational tractability.
```

stdpopsim/slim_engine.py
```diff
@@ -1189,6 +1340,7 @@ def simulate(
         slim_scaling_factor=1.0,
         slim_burn_in=10.0,
         dry_run=False,
+        recap_and_rescale=True,
```
petrelharp, line 1343: I see this is useful in testing, but how about we make it an undocumented option (`_recap_and_rescale`) so we don't have to maintain it?

stdpopsim/slim_engine.py
```diff
@@ -1376,20 +1530,46 @@ def _recap_and_rescale(
...
+                    "slim_mutation_type_id": max_id + 1,
+                    "is_neutral": True,
+                    "simulated_in_SLiM": False,
+                }
+            ],
+            "proportions": [1.0],
+            "citations": [],
+            "intervals": [[0, ts.sequence_length]],
+        }
+        metadata["stdpopsim"]["DFEs"].append(recap_dfe)
```
petrelharp, line 1567: nicely done

petrelharp:
Looks good, besides those suggestions!

stdpopsim/slim_engine.py
mufernando: I also thought this at first, but this makes our lives so much easier in the tests (see line 992 of test_slim_engine.py). But I could replicate the behavior independently and get the slim proportions there...

stdpopsim/slim_engine.py
mufernando: good catch!

stdpopsim/slim_engine.py
petrelharp: If [that's](https://github.com/popsim-consortium/stdpopsim/pull/1152/files#diff-b201385bdd9318912fe29722d750c1d4007aa5fa1454b0d8328eb6cba68dcdadR992) the place we need it then I think replicating is the way to go - currently that bit is only testing our ability to store stuff in metadata, not that we've got the proportions right.

stdpopsim/slim_engine.py
mufernando: yeah, I used this while debugging but it is not needed!

stdpopsim/slim_engine.py
mufernando: yeah, agreed! now I replicated the behavior.

stdpopsim/slim_engine.py
mufernando: done

mufernando:
Ah, I fixed the conflict and rebased/squashed.

petrelharp:
You've still got two commits?

petrelharp (approved):
Looks good! Awesome, thanks for the hard work here.

petrelharp:
I went ahead and squashed those two commits.

petrelharp:
That was weird, though - it seemed like there were two, identical commits??

mufernando:
oops, thanks!

PR #1157, opened 2021-12-22
Do not simulate neutral mutations

### LINKED ISSUE #1149: don't simulate neutral mutations in SLiM

petrelharp:
This is a merge of #1151 and #1152 by @mufernando, with minor changes. Closes #1149.

Here's where we're at:

This PR makes it so SLiM does not simulate neutral mutation types, leaving those for msprime.
This is verified, in part, by saving information about mutation types to top-level metadata, when `verbosity=3`, in `ts.metadata['SLiM']['user_metadata']`.

We would *also* like to have stdpopsim store information in top-level metadata (e.g., under `ts.metadata['stdpopsim']`); for instance, which mutation types go with which DFE? @mufernando made a start at this in #1152. But, that's *separate* because (1) we'll be sticking that information in a different slot in metadata (SLiM can only put stuff in `["SLiM"]["user_metadata"]`); (2) we want to put *more* information than what we have easily available within SLiM (e.g., names of DFEs); (3) our use of metadata here is us asking SLiM to tell us exactly what it's doing, for unit testing, and that's just a different thing than the goal of putting useful information for downstream users in metadata. (But - the SLiM-produced metadata will totally be useful for testing the stdpopsim metadata, when we get that in.)

So - I think we should merge this, and build on #1152 to fix up #1148. (And, I think this was @mufernando's intention?)

Could you have a look at the justification and the code, @mufernando?

### DIFF
--- a/stdpopsim/dfe.py
+++ b/stdpopsim/dfe.py
@@ -224,6 +224,10 @@ class DFE:
                     "mutation_types must be a list of MutationType objects."
                 )
 
+    @property
+    def is_neutral(self):
+        return all([m.is_neutral for m in self.mutation_types])
+
     def __str__(self):
         long_desc_lines = [
             line.strip()
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -603,17 +603,6 @@ _slim_debug_output = """
 """
 
 
-def msprime_rm_to_slim_rm(recombination_map):
-    """
-    Convert recombination map from start position coords to end position coords.
-    """
-    rates = recombination_map.rate.copy()
-    # replace missing values with 0 recombination rate
-    rates[recombination_map.missing] = 0
-    ends = [int(pos) - 1 for pos in recombination_map.position]
-    return rates, ends[1:]
-
-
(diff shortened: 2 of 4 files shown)


mufernando:
thanks for organizing this!! looks good to me heh

mufernando (approved):

PR #1166, opened 2022-01-12
Test recombination map using metadata

### LINKED ISSUE #1164: pass recombination map to recapitation
Currently, we aren't, as asked by @vsbuffalo

petrelharp:
<s>This will close #1164. As a first step to this, I've stored the recombination map in metadata. In the process, I found (I think) and fixed (I think) an off-by-one bug in how we got recombination maps into SLiM. It just meant that the boundaries of recombination map regions were off by one.</s>

*Edit:* all of that was wrong, lol. What this PR does is:
- provides a test that we use in SLiM the recombination map we think
- adjusts how recombination maps are rescaled using `Q`

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -79,8 +79,11 @@ initialize() {
     defineConstant("pop_names", $pop_names);
 
     _recombination_rates = $recombination_rates;
+    if (Q != 1) {
+        _recombination_rates = (1-(1-2*_recombination_rates)^Q)/2;
+    }
     _recombination_ends = $recombination_ends;
-    defineConstant("recombination_rates", (1-(1-2*_recombination_rates)^Q)/2);
+    defineConstant("recombination_rates", _recombination_rates);
     defineConstant("recombination_ends", _recombination_ends);
     // whatever is in this dictionary will be saved out in the .trees file
     defineConstant("metadata", Dictionary("Q", Q));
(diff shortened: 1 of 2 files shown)


stdpopsim/slim_engine.py
```diff
@@ -78,8 +78,11 @@
     defineConstant("pop_names", $pop_names);
 
     _recombination_rates = $recombination_rates;
+    if (Q != 1) {
```
petrelharp, line 81: I'm putting the conditional to avoid floating-point error in the Q=1 case.

stdpopsim/slim_engine.py
```diff
@@ -651,11 +657,24 @@ def slim_array_string(iterable, indent, width=80):
...
+    (i.e., positions 0, ..., L-1) will have sequence length equal to L, because
+    intervals in tskit are open on the right, so the interval [0, L) does not
+    include L. Note that this implies that the recombination rate that a msprime
+    RateMap assigns to the interval [L-1, L) has no effect in a discrete msprime
+    simulation.
     """
     rates = recombination_map.rate.copy()
     # replace missing values with 0 recombination rate
     rates[recombination_map.missing] = 0
-    ends = [int(pos) - 1 for pos in recombination_map.position]
```
petrelharp, line 658: This  `-1` was, I believe, an error.

petrelharp: Whoops! No, it was correct before.

petrelharp:
I was wrong about there being an off-by-one error (unsurprisingly), and have written some more msprime docs to make this more clear in the future (hopefully): https://github.com/tskit-dev/msprime/pull/1991

stdpopsim/slim_engine.py
petrelharp: nope!

Issue #1162, opened 2022-01-05
Use a fixed seed in SLiM engine tests.


grahamgower:
Maybe this is an issue with seed propagation, but not having a fixed seed can lead to intermittent test failures, e.g. 

```python
__________________________ TestCLI.test_simulate_dfe ___________________________

self = <tests.test_slim_engine.TestCLI object at 0x7fdc0d89e450>

    def test_simulate_dfe(self):
        slim_path = os.environ.get("SLIM", "slim")
    
        def count_mut_types(ts):
            mut_info = {}
            for mut in ts.mutations():
                for j, md in zip(
                    mut.derived_state.split(","), mut.metadata["mutation_list"]
                ):
                    if j not in mut_info:
                        mut_info[int(j)] = md
            num_neutral = sum([mut_info[j]["selection_coeff"] == 0.0 for j in mut_info])
            return [num_neutral, abs(len(mut_info) - num_neutral)]
    
        # test a DFE without demography
        with tempfile.NamedTemporaryFile(mode="w") as f:
            cmd = (
                f"-q -e slim --slim-scaling-factor 20 --slim-path {slim_path} "
                f"HomSap -c chr22 -l0.1 -o {f.name} --dfe Gamma_K17 10"
            ).split()
            capture_output(stdpopsim.cli.stdpopsim_main, cmd)
            ts = tskit.load(f.name)
        assert ts.num_samples == 10
        assert all(tree.num_roots == 1 for tree in ts.trees())
        n_mut_types = count_mut_types(ts)
        assert n_mut_types[0] > 0
        assert n_mut_types[1] > 0
    
        # test a DFE without demography with interval
        with tempfile.NamedTemporaryFile(mode="w") as f:
            cmd = (
                f"-q -e slim --slim-scaling-factor 20 --slim-path {slim_path} "
                f"HomSap -c chr22 -l0.01 -o {f.name} --dfe Gamma_K17 "
                f"--dfe-interval 1000,100000 10"
            ).split()
            capture_output(stdpopsim.cli.stdpopsim_main, cmd)
            ts = tskit.load(f.name)
        assert ts.num_samples == 10
        assert all(tree.num_roots == 1 for tree in ts.trees())
        n_mut_types = count_mut_types(ts)
        assert n_mut_types[0] > 0
        assert n_mut_types[1] > 0
    
        # test DFE with demography and interval
        with tempfile.NamedTemporaryFile(mode="w") as f:
            cmd = (
[...]

PR #1169, opened 2022-01-14
refactor of DFE CLI tests

petrelharp:
I was working on tracking down #1162, and in the process refactored the DFE CLI tests (so I could isolate where it was happening). I would also like to make this test faster, but I cannot figure out how to do it - [`test_dfe_annotation`](https://github.com/popsim-consortium/stdpopsim/pull/1169/files#diff-b201385bdd9318912fe29722d750c1d4007aa5fa1454b0d8328eb6cba68dcdadR366) needs to simulate a whole chromosome (to use an annotation), and the scaling factor can't be bigger than about 20 (or else we hit the #1162 error). Unless perhaps we can use
```
  -i INCLUSION_MASK, --inclusion-mask INCLUSION_MASK
                        Path to inclusion mask specified in bed format.
```
to speed things up - I haven't tried yet. Does anyone else know, or have good ideas?

### DIFF
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -20,6 +20,7 @@ import stdpopsim.cli
 from .test_cli import capture_output
 
 IS_WINDOWS = sys.platform.startswith("win")
+slim_path = os.environ.get("SLIM", "slim")
 
 
 def slim_simulate_no_recap(seed=1234, verbosity=2, **kwargs):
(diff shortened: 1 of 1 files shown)


grahamgower:
I guess we could add an (undocumented) species just for testing this kind of thing. But maybe we just need annotations for more species, particularly one with really short chromosomes?

petrelharp:
Well, we do kinda want to be able to test the annotations of long chromosomes, sometimes? If we had an easy way to run a simulation with a very small `N`, that would work as well?

grahamgower:
Sure. That could also be accomplished by adding a species that's just for tests.

petrelharp:
Using the inclusion mask does *not* speed things up.

petrelharp:
Could you have a look at this now, @izabelcavassim? I have not solved the problem of making the tests run faster, but I'd like to merge this much in anyhow.

(Note: I also tried chr4 from DroMel... but it takes FOREVER even though it's so small...)

izabelcavassim:
Thinking here, when we use the DFE annotation, do we really need to use the whole chromosome? Because we could also say I want annotation ranging from X-Y in that given contig. By doing this slice would we be losing the main feature of the DFE-annotation? @petrelharp ?

petrelharp:
> Thinking here, when we use the DFE annotation feature, do we really need to use the whole chromosome? Because we could also say I want annotation ranging from X-Y in that given contig. By doing this slice would we be losing the main feature of the DFE-annotation?

This is what I was going for in trying the inclusion mask. However, I didn't track down what exactly `--inclusion-mask` is doing: it seems like (from runtime) it is simulating the whole chromosome (I remember we discussed this but don't remember the conclusion).

grahamgower:
From memory, the mask is applied by chopping into the tree sequence post simulation. But @apragsdale will remember the details.

petrelharp:
Have you had a look at the code, @izabelcavassim - can we merge this?

izabelcavassim:
> Have you had a look at the code, @izabelcavassim - can we merge this?

Yes, from what I see it is all fine! LGTM!

PR #1178, opened 2022-01-26
Add Ragsdale DroMel DFE

apragsdale:
This adds the D melanogaster DFE inferred in [Ragsdale et al, 2016](https://doi.org/10.1534/genetics.115.184812) - see Table S1. The inferred DFE assumed negatively selected mutations are drawn from a log-normal with a given proportion, and positively selected mutations all have a fixed selection coefficient with the remaining proportion.

The re-scaling of the log-normal scale parameter should be double-checked, since the mean of the log-normal is given as 2Ne*s. Pretty sure I just need to divide the scale by 2Ne, but this is based on the scipy definition of the log-normal (which should be standard), and I'm not 100% sure if there are differences in the implementation here.

As an aside, since this was fit on top of an inferred single-population demographic history (using the Zambian DPGP3 data), it might make sense to also add that demographic model in a separate PR. (Edit: which I am happy to do.)

### DIFF
--- a/stdpopsim/catalog/DroMel/dfes.py
+++ b/stdpopsim/catalog/DroMel/dfes.py
@@ -48,3 +48,71 @@ def _HuberDFE():
 
 
 _species.add_dfe(_HuberDFE())
+
+
+def _RagsdaleDFE():
+    id = "LognormalPlusPositive_R16"
+    description = "Deleterious log-normal and beneficial mixed DFE"
+    long_description = """
+    Return deleterious and beneficial MutationType()s representing a drosophila DFE
+    from Ragsdale et al. (2016), https://doi.org/10.1534/genetics.115.184812.
+    DFE parameters are given in Table S1, with deleterious mutations drawn from a
+    log-normal distribution and a point mass of positive selection. The DFE was
+    inferred assuming synonymous variants are neutral and a relative mutation
+    rate ratio of 2.5 nonsynonymous to 1 synonymous mutation. Results are given as
+    scaled parameters, so that S = 2*N*s, so an estimate of Ne is required to
+    convert the scaled selection coefficients to unscaled coefficients. Because the
+    original study did not report an estimated effective population size, we use
+    the estimate from Huber et al. (2017) of Ne=2.8e6, which was estimated from
+    observed genome-wide synonymous mutation diversity and assuming a mutation rate
+    of 3e-9.
+    """
+    citations = [
+        stdpopsim.Citation(
+            author="Ragsdale et al.",
+            year=2016,
+            doi="https://doi.org/10.1534/genetics.115.184812",
+            reasons={stdpopsim.CiteReason.DFE},
+        )
+    ]
+    # the effective population size is used to rescale genetic to physical units
+    # and Ne here comes from Huber et al (2017), which used the same dataset to
+    # estimate Ne and a different parameterization of the DFE
+    Ne = 2.8e6
+
+    # mutation proportions
+    p_syn = 1 / (1 + 2.5)  # proportion of neutral
+    p_non = 1 - p_syn
+    p_positive = 0.0079 * p_non
+    p_negative = (1 - 0.0079) * p_non
+
+    # selection strengths
+    S_positive = 39.9
+    s_positive = S_positive / 2 / Ne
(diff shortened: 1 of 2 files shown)


apragsdale:
Maybe there was a good reason to have the DFE ID strictly tested as `regex = re.compile(r"[A-Z][a-z]*_[A-Z]*\d\d")`, but the other class IDs were less restrictive, such as allowing camel case. I made that change here, to mirror the pattern of demographic model ID naming.

petrelharp:
Very nice! I am not so sure about your scaling by Ne choice, though. Talking out loud here: I think we have two options:

1. As you have done (cleverly!), scale selection coefficients by the `species.Ne`. This would then produce the same "effective" strength of selection in another species as what was observed in this study; where "effective" would mean the same strength of selection relative to drift, and roughly the same probability of fixation of new, beneficial alleles, but would not have the same influx of genetic variance, for instance.
2. Take some estimate of Ne for the samples used in this study and use that to fix `s`. This would then produce the same biological strength of selection (in terms of differences to number of offspring or survival, etcetera) as what was inferred in this study.

I vote for the second one, as it's closer to the biology, and because of problematic issues around "what's Ne anyhow" in a non-constant demography.  Is there a reasonable choice for `Ne` to translate the `s` over?

apragsdale:
Those are good questions. My thought is that if the DFE for a species is typically assumed to be used in demographic models for that same species, and if we trust the Ne that we have for that species, it should be ok. But there's nothing stopping someone from taking the DFE from one species and a demographic model from another, and mixing them together (in which case, option 2 is probably better, I agree).

> Is there a reasonable choice for `Ne` to translate the `s` over?

From the paper itself, we avoided having to estimate Ne directly, so that parameters were only ever reported in scaled units (we were interested in downstream analyses that didn't depend on an estimate of Ne). The dataset used is the 197 Zambian fly genomes, which has been pretty widely used, and I'm sure some papers out there have estimated Ne for this population. That would be pulling information from independent studies though, and may end up being a bit of a rabbit hole. What do you think?

izabelcavassim:
I would think that the best strategy here is to give the scaled selection coefficient, and that would imply that the given DFE can only be used by the species in question. And that the Ne used for scaling s, is the same used in the publication of the specific DFE that we are designing. So I would be favoring option 1. That is how we do Kim DFE at least. We should explicitly document what Ne we have used to scale s. No?

petrelharp:
Ok, this is a good comparison. In this PR, @apragsdale is dividing `s` [by whatever the species Ne is](https://github.com/popsim-consortium/stdpopsim/pull/1178/files#diff-08d75f656012e518eecc3573dbca968c71007b7610108da9db67799ddc0fcdaeR74), programatically.
On the other hand, for the Kim DFE, the mean `s` [is hardcoded](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/catalog/HomSap/dfes.py#L29), so I think that you have done option (2), not option (1) there, @izabelcavassim?

What's currently in this PR is nearly equivalent to sticking in, hardcoded, for Ne our current default DroMel value for Ne. (Hardcoding it would be my option 2.) This is arguably not the best Ne since it's not coming from the Zambian genomes, but using it seems Good Enough and lets us avoid a rabbit hole.  But, if we leave it the way it is, then if in the future DroMel's default Ne changes then so will this DFE, which seems odd, and it also makes it harder to track down exactly what the DFE is.  It seems to me that everything would be simpler and more transparent if we make explicit our choice of Ne.

apragsdale:
Ok, thanks for the input all! Peter, you've convinced me that we should hard-code it. The question is what Ne to use now. The Huber17 DFE for melanogaster (the only other DFE in the dromel catalog) used the same dataset of 197 Zambian flies to infer the DFE*. How about we use the Ne that they used in that paper to rescale Ne*s to s? I'll do some digging to find out what that was.

* maybe that's a bit funny to have the only two DFEs coming from different inferences on the same dataset, but they are at least two different functional parameterizations of the DFE.

apragsdale:
If this looks reasonable to both of you @izabelcavassim @petrelharp I can squash all of this.

petrelharp (approved):
Awesome, thanks!

izabelcavassim:
Can you insert this one in @petrelharp?

PR #1188, opened 2022-02-08
Fix test.yml workflow.

### LINKED ISSUE #1187: test action is broken
This line is trying to clone a pr that has been closed (and thus doesn't exist).
https://github.com/popsim-consortium/stdpopsim/blob/a74b25c514a68165013a71cd04cee39031d8bad4/.github/workflows/tests.yml#L87

E.g. failure here.
https://github.com/popsim-consortium/stdpopsim/runs/5099954537?check_suite_focus=true

grahamgower:
Closes #1187.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,7 @@ jobs:
         uses: actions/cache@v2
         env:
           # Increase this to reset the cache if the key hasn't changed.
-          CACHE_NUM: 1
+          CACHE_NUM: 2
         with:
           path: |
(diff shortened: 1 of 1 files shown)


grahamgower:
An additional problem problem that arose here was that the built slim binary was not put into the cache, but the slim binary was only built when the cache was invalidated. We could put the slim binary into the cache, but given this build is a temporary measure (we'll use the conda slim package in the future), that seemed like overkill. So this change avoids that complexity and just always builds SLiM (except on windows) regardless of the cache status.

jeromekelleher (commented):
LGTM - building from git is always messy

PR #1199, opened 2022-02-17
document rate units (closes #1197)

### LINKED ISSUE #1197: clarify use of generation time and units of rates
In checking a QC, I just got mixed up about whether `mutation_rate` and `recombination_rate` are in units of "per year" or "per generation". As far as I can tell:

- mutation and recombination rates are in units of "per generation"
- time units in the "parameters" of demographic models say whether they are in years or generations, helpfully
- the msprime engine simulates things with time units of generations
- and so demographic models are *stored* with time units of generations ([e.g., here](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/catalog/HomSap/demographic_models.py#L125))

However, the [SLiM engine does a bunch of things with the generation time](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/slim_engine.py#L828), and so it's not actually clear to me that the SLiM engine is doing the right thing here, given that time units coming in are in units of generations?

At least, this needs to be documented better (i.e., to say the units of these rates in the places we mention them, and clarify in the "adding a species" docs); I'm happy to do this. But, do you remember what's going on with the SLiM engine, @grahamgower?

petrelharp:
I think I've said this in enough places now. (In particular, it now shows up under the "genome properties" table in the Catalog).

### DIFF
--- a/docs/_ext/speciescatalog.py
+++ b/docs/_ext/speciescatalog.py
@@ -342,7 +342,6 @@ class SpeciesCatalogDirective(SphinxDirective):
             entry += nodes.paragraph(text="{:g}".format(chrom.mutation_rate))
             row += entry
 
-            # TODO add mutation/recombination rate.
             rows.append(row)
         tbody = nodes.tbody()
         tbody.extend(rows)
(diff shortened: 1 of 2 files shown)


grahamgower (approved):
LGTM

jeromekelleher (commented):
LGTM, minor comment

PR #1200, opened 2022-02-17
Add a logfile for fitness

petrelharp:
This would add two arguments for the SLiM engine to `.simulate( )`, that writes out to `logfile` a log with columns `generation` and, for each subpopulation, `mean_fitness_pX` and `sd_fitness_pX` (where `X` is the ID of the subpop), the mean and SD of fitness. By default it does this every 1000 generations, but this is controllable with `logfile_interval`. The default is `logfile=None`, which means "don't log". The testing is minimal but sufficient, I think.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -526,6 +526,32 @@ _slim_main = """
 """
 
 
+_slim_logfile = """
+// Optional logfile output:
+// Fitness values are only available in early(),
+// so logging happens in that stage.
+1 early () {
+    defineConstant("log", sim.createLogFile("$logfile", logInterval=NULL));
+    log.addGeneration();
+    for (pop in sim.subpopulations.id) {
+        log.addCustomColumn(
+            "mean_fitness_p" + pop,
+            "mean(p" + pop + ".cachedFitness(NULL));"
+        );
+        log.addCustomColumn(
+            "sd_fitness_p" + pop,
+            "sd(p" + pop + ".cachedFitness(NULL));"
+        );
+    }
+}
+
+1: early() {
+    if (sim.generation % $loginterval == 1)
+        log.logRow();
+}
+"""
+
+
 _slim_debug_output = """
 
 ///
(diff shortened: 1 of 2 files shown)


petrelharp:
@izabelcavassim, want to give this a try to see if it explains the weird results from last meeting? (reminder of how to [check out my PR](https://tskit.dev/tskit/docs/stable/development.html#checking-out-someone-else-s-pull-request), or with `gh`)

petrelharp:
Just as we suspected!  We should show these plots in Tuesday's meeting and figure out what to do about it (either don't do genome-wide selection, or do it and report that it doesn't make sense).

petrelharp:
Could you have a look at the code? If it looks good to you, let's hit the "Rebase and merge" button.

mufernando:
wow!

stdpopsim/slim_engine.py
```diff
@@ -1370,6 +1406,11 @@ def simulate(
         :param dry_run: If True, run the first generation setup and then end the
             simulation.
         :type dry_run: bool
+        :param logfile: Name of file to write a log of summary statistics
+            (currently, only mean and SD of fitness values per population).
+            Defaults to None, meaning "do not log".
+        :param logfile_interval: Interval at which to write out logfile
```
izabelcavassim, line 1412: Interval (in generations)

stdpopsim/slim_engine.py
```diff
@@ -1370,6 +1406,11 @@ def simulate(
         :param dry_run: If True, run the first generation setup and then end the
             simulation.
         :type dry_run: bool
+        :param logfile: Name of file to write a log of summary statistics
+            (currently, only mean and SD of fitness values per population).
+            Defaults to None, meaning "do not log".
+        :param logfile_interval: The number of generations (i.e., SLiM time steps)
+            between each writing out to the logfile (if specified).
```
petrelharp, line 1413: How's that?

petrelharp: Wait, I can do better.

izabelcavassim:
awesome! Thanks!

Issue #579, opened 2020-08-25
QC for {_inferred_1_M_13} ({BosTaurus})


iechevarriaz:
**PR for new model:** https://github.com/popsim-consortium/stdpopsim/pull/558
**Original paper:** https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3748359/
**Parameter values:** {reference to where parameter values can be found}
**Population indices:** {description of population indices}

**Potential issues that might lead to differences between the model implementations:**
{description of any potential issues}

**QC'er requests:** {tag potential developers to QC the model}

gregorgorjanc:
Just a note here that I have "eyeballed" with @iechevarriaz the effective population sizes at different times from this paper https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3748359/ Figure 4b. These are the best estimates, from my knowledge.

iechevarriaz:
Good morning! There are values for Ne and generations ago in the following supplementary material https://academic.oup.com/mbe/article/30/9/2209/1002512#supplementary-data

agladstein:
I will work on this QC!

grahamgower:
Any updates here @agladstein? It would be great to get Bos taurus in. Should we split the species and the demographic model into two commits?

grahamgower:
I've taken a look at the demographic model while making pr #600, and it looks like the model implementation is not piecwise-constant, as in MacLeod et al. It's probably better to stick with the publication here, unless there's a good reason not to?

I.e. at the moment, the model has:
```
        demographic_events=[
            msprime.PopulationParametersChange(
                time=1,
                initial_size=90,
                growth_rate=-0.095894024,
                population_id=0,
            ),  # Ne 90 to 120
            msprime.PopulationParametersChange(
                time=4, growth_rate=-0.24465639, population_id=0
            ),  # Ne 120 to 250
            msprime.PopulationParametersChange(
                time=7, growth_rate=-0.0560787, population_id=0
            ),  # Ne 250 to 350
```

But MacLeod et al. actually shows something more like:
```
        demographic_events=[
            msprime.PopulationParametersChange(
                time=1,
                initial_size=90,
                growth_rate=0,
                population_id=0,
            ),  # Ne = 90
            msprime.PopulationParametersChange(
                time=4, initial_size=120, population_id=0
            ),  # Ne = 120
            msprime.PopulationParametersChange(
                time=7, initial_size=250, population_id=0
            ),  # Ne = 250
```

[...]

PR #1214, opened 2022-04-11
Updated BosTau Holstein demography mut rate #579

gregorgorjanc:
Updated BosTau Holstein demography mut rate #579 to match the supplement https://academic.oup.com/mbe/article/30/9/2209/1002512 value since we use supplement Ne estimates - the supplement mutation rate is meant to be the sequence error anyway, which is what the simulation gives.

@igronau It would be great if you can merge this in. Thanks!


grahamgower (approved):
LGTM, thanks @gregorgorjanc

PR #1224, opened 2022-04-22
Update black in pre-commit.

### LINKED ISSUE #1223: update black
Earlier this year, [black 22.1.0](https://github.com/psf/black/releases/tag/22.1.0) made a minor change to how it formats space around the power operator (e.g., `two_to_the_power_of_five = 2 ** 5` is now written as `two_to_the_power_of_five = 2**5`). This means we'll start to get some drift between what the CI accepts and what devs have installed. We should just update black in CI/pre-commit and force devs to update.

grahamgower:
Closes #1223.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
         language_version: python3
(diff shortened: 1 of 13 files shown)


grahamgower:
I'm not sure what the process is for developers after this gets merged. Maybe the installed pre-commit sees this change and autoupdates accordingly? In the worst case, something like ` pre-commit clean && pre-commit install -f` might be needed by a developer (but possibly not until they change/add a line that uses the power operator!). @jeromekelleher do you know?

jeromekelleher (approved):
LGTM

The update is pretty painless I think. I didn't notice any changes when it went through tskit/msprime, so I think it's probably fine to go ahead.

grahamgower:
Cool, thanks Jerome.

Issue #544, opened 2020-05-26
QC for PonAbe


grahamgower:
The **PonAbe** species definition has not yet been QCed!

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [ ] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [ ] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [x] Population size.
- [ ] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [ ] Genome reference assembly.
- [x] Mutation rate.
- [x] Recombination rate.
- [ ] Recombination map(s) (if relevant).
- [x] Population size.
- [ ] Generation time.

grahamgower:
Some notes on the PonAbe parameters:
 * Nater et al. 2017 use mu=1.5e-8 (hoping that the value is similar to humans and chimps) and 25 year generations (citing [this book](https://oxford.universitypressscholarship.com/view/10.1093/acprof:oso/9780199213276.001.0001/acprof-9780199213276)).
 * Locke et al. 2011 assumed a mutation rate of 2.0e-8 and 20 years per generation, both without citation.

The current stdpopsim code uses mu=1.5e-8 and 20 years per generation, and the only demographic model implemented is from Locke et al.. Probably it would be better if we were more consistent here. Thoughs @apragsdale?

Ensembl only include an old assembly, btw. http://rest.ensembl.org/info/assembly/Pongo%20abelii
[...]

Issue #539, opened 2020-05-26
QC for AraTha


grahamgower:
The **AraTha** species definition has not yet been QCed!

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

- [x] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [x] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [x] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [x] Population size.
- [x] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [x] Genome reference assembly.
- [x] Mutation rate.
- [x] Recombination rate.
- [x] Recombination map(s) (if relevant).
- [x] Population size.
- [x] Generation time.

grahamgower:
I've recently reviewed some items on this list, and thus they're already checked.

jeromekelleher:
Do you think we can push this back to version 0.1.3 @grahamgower? It'll take a while to get all these QC'd, and I'd like to get 0.1.2 out ASAP.

grahamgower:
Yeah, I guess so. But we really should try to get some eyes on everything soon, so that if we find anything concerning we can push out 0.1.3 in a hurry. Note that there are still comments like this in the code: "Lengths from TAIR 10 although its unclear what reference the genetic map used   -- follow up on this with Salome 2012 authors".

jeromekelleher:
I agree it's not ideal, but we don't *know* there are bugs in the species definitions. We do know that there are bugs in the currently released models, so we shouldn't delay getting a fix out for the models just to be sure that there aren't other problems. As you say, we can push out more 0.1 releases as we need.

grahamgower:
[...]

PR #1241, opened 2022-05-11
added qc-ed species to the catalog

petrelharp:
I've gone through the `tests/test_SpeciesName.py` files to see which had been QC'ed (is there a better way to do this?), and have added those QC'ed species to the catalog.

Along the way, I noticed that a few more of the older species are not as QC'ed as the current standard (mostly just that recombination rates aren't checked) beyond AraTha (#539) and PonAbe (#544):
- [CanFam](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_CanFam.py)
- [DroMel](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_DroMel.py#L36) - note it just tests if rates are nonzero

### DIFF
--- a/docs/catalog.rst
+++ b/docs/catalog.rst
@@ -40,9 +40,17 @@ page to learn about the process for adding new items to the catalog.
 Then, if you feel ready, make an issue on our
 `GitHub page <https://github.com/popgensims/stdpopsim/issues>`_.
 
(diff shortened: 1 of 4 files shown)


grahamgower (approved):
LGTM. Thanks @petrelharp!

Issue #733, opened 2021-01-12
QC for OutOfAfrica_4J17 (HomSap)


grahamgower:
**PR for new model:** #726
**Original paper:** https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5500150/
**Parameter values:** table 4 in the main text.
**Population indices:** YRI=0, CEU=1, CHB=2, JPT=3

**Potential issues that might lead to differences between the model implementations:**
Nothing known apriori.

jeffspence:
I'll work on QCing this model.  I'll shoot to be done by 5/16/22

jeffspence:
I'll QC this model.  I'll get to it on or before 5/16/22

PR #1246, opened 2022-05-13
QC for OutOfAfrica_4J17

jeffspence:
QC for https://github.com/popsim-consortium/stdpopsim/issues/733 (@grahamgower)
 
This is my first time doing QC -- so let me know if I'm doing anything wrong.  I was able to (blind) replicate the model.  The only issues I ran into were:

1. The times in generations as currently implemented have all been cast to integers.  e.g., in the notation of Table 4 [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5500150/) T_{CH-JP} is 9kya.  The generation time is 29y so I originally implemented this as `9e3/29` but that caused the tests to fail.  I could get the tests to pass if I changed this to `int(9e3/29)`.  The models are inferred by `moments` which assumes discrete (hence integer) generations so I guess that makes sense, but since they only report to the precision of 9kya, it's not clear whether it makes sense to convert to an integer or not.
2. I initially used the newer `msprime` API to specify the population splits (`add_population_split`) but that caused one of the populations to become "inactive" which caused the tests to fail.  I could get around this by switching to the old style of specifying the equivalent population splits as mass migrations, plus setting migration rates to zero.  I guess this is maybe a personal preference kind of thing, so I'm happy to go either way.

otherwise my initial implementation matched the current implementation.

### DIFF
--- a/stdpopsim/qc/HomSap.py
+++ b/stdpopsim/qc/HomSap.py
@@ -1258,6 +1258,97 @@ def ZigZag():
 _species.get_demographic_model("Zigzag_1S14").register_qc(ZigZag())
 
 
+def JouganousOOA2017():
+    """
+    Four population OOA model from Jouganous et al. 2017, fit to the
+    SFS using moments. The populations are YRI, CEU, CHB, and JPT.
+    The model is shown in Figure 3 and the parameters are taken from
+    table 4.
+    The model includes population splits, directional migration and
+    exponential growth in the various populations. The model was fit
+    using data from 20 samples from each population and assumes a
+    mutation rate of 1.44e-8 per generation and a generation time
+    of 29 years.
+    """
+    id = 'QC-OutOfAfrica_4J17'
+
+    N_A = 11293
+    N_AF = 23721
+    N_B = 2831
+    N_EU0 = 2512
+    r_EU = 0.0016
+    N_AS0 = 1019
+    r_AS = 0.0026
+    N_JP0 = 4384
+    r_JP = 0.0129
+    m_AF_B = 16.8e-5
+    m_AF_EU = 1.14e-5
+    m_AF_AS = 0.56e-5
+    m_EU_AS = 4.75e-5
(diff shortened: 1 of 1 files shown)


jeffspence:
I should also say that my first commit on this branch is the model as I original implemented it, and the second commit is the tweaks I mentioned above (switching generation times to integers, and switching to `add_mass_migration`)

grahamgower (approved):
LGTM, thanks @jeffspence.

grahamgower:
> QC for #733 (@grahamgower)
> 
> This is my first time doing QC -- so let me know if I'm doing anything wrong. I was able to (blind) replicate the model. The only issues I ran into were:
> 
> 1. The times in generations as currently implemented have all been cast to integers.  e.g., in the notation of Table 4 [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5500150/) T_{CH-JP} is 9kya.  The generation time is 29y so I originally implemented this as `9e3/29` but that caused the tests to fail.  I could get the tests to pass if I changed this to `int(9e3/29)`.  The models are inferred by `moments` which assumes discrete (hence integer) generations so I guess that makes sense, but since they only report to the precision of 9kya, it's not clear whether it makes sense to convert to an integer or not.

Conversion to intergers is fine here. It makes sense in the context, and there will be no practical difference in simulation output as a result.

> 2. I initially used the newer `msprime` API to specify the population splits (`add_population_split`) but that caused one of the populations to become "inactive" which caused the tests to fail.  I could get around this by switching to the old style of specifying the equivalent population splits as mass migrations, plus setting migration rates to zero.  I guess this is maybe a personal preference kind of thing, so I'm happy to go either way.
> 

This can be dealt with by using the `population_id_map` parameter to DemographicModel. But its undocumented, so don't worry about it.

> otherwise my initial implementation matched the current implementation.

Great! I'm going to merge this without squashing, because your first model version could be useful. Thanks!

PR #1250, opened 2022-05-16
QC sunflower

xin-huang:
In this pr, I QC'ed sunflower https://github.com/popsim-consortium/stdpopsim/issues/1219

I also updated https://github.com/popsim-consortium/stdpopsim/blob/main/environment.yml by adding gcc
because gcc is needed when creating the virtual environment using `conda env create -f environment.yml`


petrelharp:
Nice, thanks! Mind squashing those two commits?

PR #1259, opened 2022-05-16
qc OutOfAfricaExtendedNeandertalAdmixturePulse_3I21

awohns:
Hi @LeonardoIasi, I read through the paper and supplement and QC'ed the model based on Supplementary Figure 1a and notes in the main text/materials. Great paper by the way, I really enjoyed reading it!
There appear to be some discrepancies between my QC and your implementation:

1. There are two models specified in your supplementary figure 1, I think the description of the model should make clear that stdpopsim only implements the simpler model (the one in Supp. Figure 1a).
2. The only generation time I see specified in the paper or supplement is 29 years (in the discussion) vs. 25 years in your parameter file. Additionally, the model implementation divides this by 1000 (`25 / 1000`) so that years can be specified in kya. I would suggest keeping everything in years so that `generation_time` is consistent with other models.
3. The split times are listed in generations in Supplemental Figure 1a. Multiplying these by 29 years (rather than 25 years) matches the split times (in kya) in your parameter file, so it seems that 29 years is what should be used in stdpopsim.
4. Supplemental Figure 1a gives a Neanderthal Admixture percentage of 0.03, rather than 0.029. I don't see 0.029 anywhere in the paper or the [Github repo referenced by the paper](https://github.com/LeonardoIasi/Extended_Admixture_Pulse). Is this the integral of the gamma distribution with the given parameters?
5. Supplemental Figure 1a specifies a mutation rate of 2e-8, but it doesn't appear that your implementation includes a mutation rate?
6. Very minor comment, but perhaps use "AFR" and "OOA" as population descriptions instead of "YRI" and "CEU" since they're described as "African" and "non-African" in Supp. Fig. 1?

### DIFF
--- a/docs/parameter_tables/HomSap/OutOfAfricaExtendedNeandertalAdmixturePulse_3I21.csv
+++ b/docs/parameter_tables/HomSap/OutOfAfricaExtendedNeandertalAdmixturePulse_3I21.csv
@@ -6,4 +6,5 @@ Time (kya),290,Neandertal Human split
 Time (kya),73.95,Time of OOA event
 Time (kya),50,Neandertal migrations starts
 Time (kya),30,Neandertal migrations end
-Generation time (yrs.),25,Generation time
+Generation time (yrs.),29,Generation time
+Mutation rate,2e-8,Per-base per-generation mutation rate
--- a/stdpopsim/catalog/HomSap/demographic_models.py
+++ b/stdpopsim/catalog/HomSap/demographic_models.py
@@ -71,12 +71,11 @@ def _ooa_nea_extended_pulse():
         """
         tm = (migration_stop - migration_start) / 2 + migration_start
         k = 1 / ((migration_stop - migration_start) / (4 * tm)) ** 2
-        m = [stats.gamma.pdf(x=range(split_time + 1), a=k, loc=0, scale=tm / k)]
+        m = stats.gamma.pdf(x=range(split_time + 1), a=k, loc=0, scale=tm / k)
 
         """
         Scaling the distribution by the total migration rate.
         """
-        m = m[0]
         m[abs(m) < migration_cutoff] = 0
         m_scaled = m * total_migration_rate / sum(m)
 
@@ -86,7 +85,7 @@ def _ooa_nea_extended_pulse():
         for x in range(split_time + 1):
 
             """
-            Writing gene flow events which are inside the set time boarders and
+            Writing gene flow events which are inside the set time borders and
             over the set migration cutoff. They will be included in the m(t)
             distribution.
             """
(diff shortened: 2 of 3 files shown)


LeonardoIasi:
Hello @awohns,
Thanks you so much for QC'ing the model. Glad you liked the paper. 
1. Yes the model is indeed the simple model from Supp. Figure 1a.
2. I agree keeping it in years is probably better.
3. Yes 29 years per generation should be used. Sorry that was my mistake.
4. When the Gamma model is implemented in the simulation I am using a per generation migration cutoff of 1e-5. Hence the tails of Gamma distribution is are slightly cutoff. Integrating over this truncated Gamma with the given parameters results in an overall migration rate of 0.029 (rounded) and not exactly 0.03.
5. Yes in the paper we used 2e-8, sorry I did not include the mutation rate parameter in the model.
6. I agree calling them AFR and OOA makes more sense.

petrelharp:
Is the "nonzero migration rate" error in the failing test from your model, @awohns?

awohns:
It looked like that appeared when I removed the lines in the implementation which set the migration rate to zero. I've added those back and looks like all the tests related to the QC passed, but I got a failure related to a recombination map? I rebased and now there's `ERROR: No matching distribution found for demes==0.2.2`. Any idea what's going on here @petrelharp?

awohns:
Also, I've made some of the changes suggested in the conversation with @LeonardoIasi above to the original implementation, but this just changes the scaling factor of the times, removes a few unnecessary lines of code, and uses predefined populations, so low chance of breaking the model I hope!

petrelharp:
I re-ran that job and it passed this time! =) With that, is there anything else you want to do here, @awohns?

awohns:
> I re-ran that job and it passed this time! =)

Great! Thanks @petrelharp. I think I'm done here if all looks ok to you and the changes to the original model are acceptable

petrelharp:
How's this look to you, @LeonardoIasi?

petrelharp:
It looks good to me - but, IIRC it was not obvious to you which model from the paper was used exactly - perhaps you could edit [the description](https://github.com/popsim-consortium/stdpopsim/blob/dc2c51c646961c3a1aacba529a3844c64b7968a0/stdpopsim/catalog/HomSap/demographic_models.py#L223) to make this clear (e.g. "the simple model of Fig X with an extended gamma-shaped pulse").

LeonardoIasi:
Thanks you very much @awohns  and @petrelharp, it looks good. I checked it and the only thing to change would be to the parameter: tm_NHc_stop = int(50000 / generation_time) to tm_NHc_stop = int(53200 / generation_time). Since the generation time is now 29 years, otherwise the pulse would not be 800 generations long (which is what we used in the paper). In years the extended pulse is then 23200 years long.

awohns:
Thanks both, just made the two changes

PR #1268, opened 2022-05-17
QC for honeybee

manolofperez:
> * [x]  Recombination rate.
>   
>   * This might be genome-wide, or per-chromosome. Both are fine.
>   * Check there's a comment describing where it came from, and/or how calculated.
>   * From a publication? Check the value(s) match the publication.
>   * Calculated somehow? Average over a recombination map? Redo the calculation.
> * [x]  Mutation rate.
>   
>   * This might be genome-wide, or per-chromosome. Both are fine.
>   * Check there's a comment describing where it came from, and/or how calculated.
>   * From a publication? Check the value(s) match the publication.
>   * Calculated somehow? Redo the calculation.
> * [x]  Recombination map (if present).
>   
>   * Does it match the assembly? Liftover is fine, if clearly stated.
>   * Is the description/long_description a good summary of how the map was created?
> * [x]  Population size.
> * [x]  Generation time.
> 
> For each citation, check:
> 
> * Doi link.
> * Is publication a preprint? Is there a peer-reviewed publication instead?
> * Is the year correct.
> * Is the author correct (spelling, accents/ligatures/etc.)
> 
> Citations are required for:
> 
> * [x]  Genome reference assembly.
> * [x]  Mutation rate.
> * [x]  Recombination rate.
> * [x]  Recombination map(s) (if relevant).
> * [x]  Population size.
> * [x]  Generation time.
> 
> The final PR should:
> 
> * [x]  fill out the test stubs
> * [x]  delete the `pytest.mark.skip` lines that make the tests not run
> * [x]  make sure they pass, talking to the original author to figure out discrepancies

### DIFF
--- a/tests/test_ApiMel.py
+++ b/tests/test_ApiMel.py
@@ -21,67 +21,65 @@ class TestSpeciesData(test_species.SpeciesTestBase):
     # independently referring to the citations provided in the
     # species definition, filling in the appropriate values
     # and deleting the pytest "skip" annotations.
-    @pytest.mark.skip("Population size QC not done yet")
     def test_qc_population_size(self):
-        assert self.species.population_size == -1
+        assert self.species.population_size == 2e05
 
-    @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 2
 
 
 class TestGenomeData(test_species.GenomeTestBase):
 
     genome = stdpopsim.get_species("ApiMel").genome
 
-    @pytest.mark.skip("Recombination rate QC not done yet")
     @pytest.mark.parametrize(
         ["name", "rate"],
         {
-            "CM009931.2": -1,
-            "CM009932.2": -1,
-            "CM009933.2": -1,
-            "CM009934.2": -1,
-            "CM009935.2": -1,
-            "CM009936.2": -1,
-            "CM009937.2": -1,
-            "CM009938.2": -1,
-            "CM009939.2": -1,
-            "CM009940.2": -1,
-            "CM009941.2": -1,
-            "CM009942.2": -1,
-            "CM009943.2": -1,
-            "CM009944.2": -1,
-            "CM009945.2": -1,
-            "CM009946.2": -1,
-            "CM009947.2": -1,
+            "CM009931.2": 23.9e-8,
+            "CM009932.2": 24.6e-8,
+            "CM009933.2": 24.1e-8,
+            "CM009934.2": 27.6e-8,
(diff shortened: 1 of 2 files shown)


tests/test_ApiMel.py
```diff
@@ -21,66 +21,62 @@ def test_common_name(self):
...
+            "CM009942.2": 21.2e-8,
+            "CM009943.2": 23.4e-8,
+            "CM009944.2": 24.6e-8,
+            "CM009945.2": 22.1e-8,
+            "CM009946.2": 22.8e-8,
+            "CM009947.2": 0,
         }.items(),
     )
     def test_recombination_rate(self, name, rate):
         assert pytest.approx(rate, self.genome.get_chromosome(name).recombination_rate)
```
petrelharp, line 58: This is actually a pre-existing bug, it should be this:
```suggestion
        assert rate == pytest.approx(self.genome.get_chromosome(name).recombination_rate)
```

tests/test_ApiMel.py
```diff
@@ -21,66 +21,62 @@ def test_common_name(self):
...
+            "CM009941.2": 3.4e-9,
+            "CM009942.2": 3.4e-9,
+            "CM009943.2": 3.4e-9,
+            "CM009944.2": 3.4e-9,
+            "CM009945.2": 3.4e-9,
+            "CM009946.2": 3.4e-9,
+            "CM009947.2": 3.4e-9,
         }.items(),
     )
     def test_mutation_rate(self, name, rate):
```
petrelharp, line 82: same change on next line

petrelharp:
Sorry, I should have noticed the problem with `approx( )` before - can you make the suggested change and update?

tests/test_ApiMel.py
manolofperez: Oh! Sorry for not noticing this. I will fix it.

manolofperez (commented):
I changed the files.

manolofperez (commented):
Updated the test file.

petrelharp:
Hm - looks like your pre-commit isn't set up. To fix this, you could [make the change suggested](https://github.com/popsim-consortium/stdpopsim/runs/6478752771?check_suite_focus=true), or you could [install the pre-commit hooks](https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#github-workflow) and it'll do it for you (note however you have to do `git commit` twice). And, then squash these commits down to one?  Or, I'm happy to do this if it all sounds too frustrating - it's very easy once set up.

petrelharp:
No problem! These things take practice.  And it was a bit tricky - I did it [this way](https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#when-rebasing-goes-wrong).

PR #1269, opened 2022-05-18
QC for BosTau

### LINKED ISSUE #1239: QC for BosTau
**PR for new species:** #600 #809 

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

The QC reviewer should start a pull request that fills out the test stubs
with independently obtained values. The reviewer may look at the python code
for rationale provided in comments, but should ignore the actual code
as much as possible - comments in the code should give enough information
that it's obvious how to get the correct value from the provided references.
(In particular, we shouldn't copy-paste the value from the code into the test!)

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [ ] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [ ] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [ ] Population size.
- [ ] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [ ] Genome reference assembly.
- [ ] Mutation rate.
- [ ] Recombination rate.
- [ ] Recombination map(s) (if relevant).
- [ ] Population size.
- [ ] Generation time.

The final PR should:
- [ ] fill out the test stubs
- [ ] delete the `pytest.mark.skip` lines that make the tests not run
- [ ] make sure they pass, talking to the original author to figure out discrepancies

gtsambos:
Closes #1239 when done. 
To do still:

- [ ] mutation rate check
- [ ] citation check (update with proper genome assembly link)

See #1239 for some issues to do with the parameter choices that I've noticed.

@igronau @grahamgower @petrelharp


gtsambos:
One more issue: there's are a few different mutation rates in the text of [the citation](https://doi.org/10.1101/079863) than the one coded here, see lines 174-184:

To me it sounds like we want the second one (1.17x10-8) but instead we have the first one (1.2x10-8). BosTau experts are welcome to chime in here

gtsambos:
Okay, I think I've done what I can here on my own, now I just need to hear back from others about the parameter choices. Can someone help me interpret the pre-commit error here, also?

gtsambos:
Here's a summary of the remaining questions:

 - **population size**: should we use the present-day value of 90, or the ancestral value of 62000?
 - **generation time**: this is not actually estimated in the Macleod et al. citation, but rather a heuristic chosen based on other analyses which they link to -- would it be better to use the rate from one these other analyses and cite it directly?
 - **mutation rate**: the Halder et al citation provides multiple mutation rates, have we chosen the right one?
 - **pre-commit**: what's the actual error here?

grahamgower:
> * **pre-commit**: what's the actual error here?

Pre-commit ran `black` and something changed. You'll need to make those same changes and commit them. If you have `pre-commit` set up this should happen automatically when you commit. You might be able to fix this manually by running `black` on that file yourself.

gtsambos:
> Pre-commit ran black and something changed. You'll need to make those same changes and commit them.

cool thanks Graham, will fix after dinner

gtsambos:
ps, it would be great if the developer documentation talked about how to get these pre-commits working

petrelharp:
> ps, it would be great if the developer documentation talked about how to get these pre-commits working

[They do](https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#github-workflow), but I think only in the `latest` version of the docs, not `stable`.

petrelharp:
Your rationale in general makes sense to me but I'm hoping @gregorgorjanc will have more informed opinions.

>   generation time: this is not actually estimated in the Macleod et al. citation. Their value is just a heuristic choice based on other analyses, which they link to in the paper -- would it be better to use the rate from one these other analyses and cite it directly?

I think a presumably well-thought-out heuristic choice that we can cite is a great way to go.

gregorgorjanc:
> Your rationale in general makes sense to me but I'm hoping @gregorgorjanc will have more informed opinions.
> 
> > generation time: this is not actually estimated in the Macleod et al. citation. Their value is just a heuristic choice based on other analyses, which they link to in the paper -- would it be better to use the rate from one these other analyses and cite it directly?
> 
> I think a presumably well-thought-out heuristic choice that we can cite is a great way to go.

I am happy to cite a more specific paper, but for simplicity stuck with the Macleod et al. citation since I used her work extensively.

gregorgorjanc:
> One more issue: there are a few different mutation rates in the text of [the citation](https://doi.org/10.1101/079863), see lines 174-184:
> 
> To me it sounds like we want the second one (1.17x10-8) but instead we have the first one (1.2x10-8). BosTau experts are welcome to chime in here

@gtsambos these checks are awesome!

I was not sure which estimate was supposed to be more relevant (1.17... or 1.2...), but since 1.17... would round up to 1.2... I just went with 1.2... Considering an error of estimation, I fell this should be fine.

gtsambos:
ah, I've spotted one error here in my own code -- the mitochondrial recombination rate should be 0

tests/test_BosTau.py
```diff
@@ -32,83 +33,127 @@ def test_basic_attributes(self):
...
         }.items(),
     )
     def test_recombination_rate(self, name, rate):
         assert rate == pytest.approx(
-            self.genome.get_chromosome(name).recombination_rate
+            self.genome.get_chromosome(name).recombination_rate, abs=0.001
         )
 
-    @pytest.mark.skip("Mutation rate QC not done yet")
+    @pytest.mark.skip("Mutation rate QC currently fails due to discrepancy")
```
petrelharp, line 122: Having failing tests here is good, it lets us see where the issues still are.
```suggestion
```

tests/test_BosTau.py
```diff
@@ -32,83 +33,127 @@ def test_basic_attributes(self):
...
+            "28": avg_rate,
+            "29": avg_rate,
+            "X": avg_rate,
+            "MT": avg_rate,
         }.items(),
     )
     def test_recombination_rate(self, name, rate):
         assert rate == pytest.approx(
-            self.genome.get_chromosome(name).recombination_rate
+            self.genome.get_chromosome(name).recombination_rate, abs=0.001
```
petrelharp, line 119: I think as written this will always pass, because recombination rates are like 1e-8, so anything is equal to within +/- 0.001.
```suggestion
            self.genome.get_chromosome(name).recombination_rate, rel=0.001
```

tests/test_BosTau.py
```diff
@@ -32,83 +32,126 @@ def test_basic_attributes(self):
...
+            "22": 1.17e-8,
+            "23": 1.17e-8,
+            "24": 1.17e-8,
+            "25": 1.17e-8,
+            "26": 1.17e-8,
+            "27": 1.17e-8,
+            "28": 1.17e-8,
+            "29": 1.17e-8,
+            "X": 1.17e-8,
+            "MT": 1.17e-8,
```
petrelharp, line 154:
```suggestion
            "MT": 0.0,
```

tests/test_BosTau.py
```diff
@@ -32,83 +32,126 @@ def test_basic_attributes(self):
...
+            "22": avg_rate,
+            "23": avg_rate,
+            "24": avg_rate,
+            "25": avg_rate,
+            "26": avg_rate,
+            "27": avg_rate,
+            "28": avg_rate,
+            "29": avg_rate,
+            "X": avg_rate,
+            "MT": avg_rate,
```
petrelharp, line 113:
```suggestion
            "MT": 0.0,
```

gregorgorjanc:
Works for me! I am somewhat worried seeing 60K Ne since most people will
know this is way off, but I also agree that 90 on its own is useless too!
All good. Let’s move onwards with this large Ne.

With regards,
  Gregor

gtsambos:
Thanks @petrelharp! There's one extra thing  -- I think we should reinstate the Rosen et al citation thats's been commented out, since this is the actual genome assembly paper, as discussed in #1239 . The one that I replaced it with is the ensembl paper that verified the chromosome lengths. I hope we can have two citations for this 'reason'?

gtsambos:
>  I am somewhat worried seeing 60K Ne since most people will
know this is way off, but I also agree that 90 on its own is useless too!

For BosTau and other domesticated species, it would be good to include some prominent NOTE in the catalog iwhere we strongly recommend that users simulate with a defined demographic model

igronau:
> For BosTau and other domesticated species, it would be good to include some prominent NOTE in the catalog iwhere we strongly recommend that users simulate with a defined demographic model

@gtsambos - you're right, but this note is relevant for many species, e.g., ones that underwent severe bottlenecks and expansions. If we decide to add a cautionary note about simulating without a demographic model, then it should be to all species that underwent extreme changes in Ne, and not just domesticated species. Humans probably also apply here. It'll be difficult IMO to know where to draw the line. The default Ne captures the best fit under a simplifying assumption of a single panmictic population with constant size. This approximation is going to be worse for species with more fluctuating Ne's. This is something that people should be generally aware of when simulating their data. Maybe the best solution is to highlight this point in the tutorials.

petrelharp:
Thanks! I've uncommented the Rosen citation.

I'm kinda with @igronau here - I don't really think we need the note - but, I'd be happy to add it later. I'd like to merge this in now, though.

PR #1273, opened 2022-05-19
adding last species to the catalog

petrelharp:
I'll accumulate the last few species in the QC queue here.


petrelharp:
The decrease in coverage is expected - it's due to the AnoGam model. I'm merging this anyhow.

PR #1279, opened 2022-05-25
QC for GAS_1A17 (AnoGam) 

### LINKED ISSUE #1278: QC for GAS_1A17 (AnoGam)
**PR for new model:** This was added along with the species itself in #856
**Original paper:** https://doi.org/10.1038/nature24995 
**Parameter values:** 
**Population indices:** it's a one-population model, for the Gambia sampling location of A. gambiae

**Potential issues that might lead to differences between the model implementations:**
The reported values are with the intermediate (3.5e-9) mutation rate

**QC'er requests:**

petrelharp:
Closes #1278.

I've renamed the model `GabonAg1000G_1A17`, since the string is supposed to be "descriptive", and "GAS" is just the internal code used for Gabon. I've also decided not to put all 110 epoch times and demographic sizes in the summary table. I ran into no problems!


petrelharp:
So - I got the same model, but *then* I tried to remove the "epochs" in the model in which nothing changes - i.e., the zero-length epochs and the ones where the population size is the same as before. Now, things don't pass, and someone else ought to fix up the original model. Might you be able to have a go at this, @andrewkern?

petrelharp:
Never mind! I got it! Simplified the model in the code by some careful hand editing, and it agrees with the programatic solution in QC. I think tihs is ready to go!

andrewkern:
right on. i'm going to merge this

petrelharp:
wait let me squash it first

PR #1287, opened 2022-06-06
Update species.py

gregorgorjanc:
Adding Rosen et al. ref to Genome.

@gtsambos is this OK?


petrelharp:
Ah, whoops - I should have caught that one; yes, Rosen et al has "citation reason" of Assembly, so should be in this list. Thanks!

PR #1291, opened 2022-06-15
QC for chimpanzee

janaobsteter:
Edited the species.py and test.py through communication with the original authors


petrelharp:
Oh wait, I was silly and forgot to check whether the @pytest.mark.skips were deleted. I'll fix that up elsewhere.

PR #1292, opened 2022-06-18
Dfe qc

xin-huang:
In this pr,

I added `register_qc` in `dfe.py` for DFE QC.
I also added `test_dfes.py` for testing `dfe.py`
I also added a QCed DFE `Huber2017` in `qc/HomSap.ypy`


tests/test_dfes.py
```diff
@@ -0,0 +1,167 @@
...
+        for i in range(len(mt1)):
+            assert mt1[i].dominance_coeff == mt2[i].dominance_coeff
+            assert mt1[i].distribution_type == mt2[i].distribution_type
+            assert mt1[i].distribution_args == mt2[i].distribution_args
+            assert mt1[i].convert_to_substitution ==  mt2[i].convert_to_substitution
+
+    def test_proporitions_match(self):
+        p1 = self.dfe.proportions
+        p2 = self.dfe.qc_dfe.proportions
+        assert p1 == p2
```
petrelharp, line 147: Shouldn't this be `assert np.all(p1 == p2)`? Or, `assert np.allclose(p1 == p2)`?

petrelharp:
This looks *great*, @xin-huang! I have one comment, and I see one commented-out test (for the parameter tables). Otherwise, it looks good to go - are you missing anything?

xin-huang:
Hi @petrelharp 

I have questions about the DFE parameter tables. Will users also create and put their DFE parameter tables in https://github.com/popsim-consortium/stdpopsim/tree/main/docs/parameter_tables ?

Or we just create these parameter tables automatically with the codes in https://github.com/popsim-consortium/stdpopsim/blob/main/docs/_ext/speciescatalog.py ? Then we don't need to test here?

And I think we also display DFE parameters, such as shape and scale parameters, in `stdpopsim-docs`. Currently, it just lists DFE names.

stdpopsim/qc/HomSap.py
```diff
@@ -1565,3 +1565,27 @@ def extended_pulse(
...
+    return stdpopsim.DFE(
+        id=id,
+        description=id,
+        long_description=id,
+        mutation_types=[neutral, negative],
+        proportions=[0.3, 0.7],
+    )
+
+
+_species.get_dfe("Gamma_H17").register_qc(Huber2017())
```
xin-huang, line 1591: Hi @petrelharp 

Should I remove these codes and open another pr for QC this DFE and other DFEs?

petrelharp: I think it's important to leave this QC in here, because we should be able to see that your QC framework works.

petrelharp:
Good question - I agree we should put the parameters in the documentation. And, i guess that DFEs are standard enough we can write automatic code to generate these tables - good thought. That seems better/easier than writing and parseing parameter tables like for demographic models. Want to open an issue for that, and we'll get this PR sorted out first?

tests/test_dfes.py
xin-huang: changed to `assert np.allclose(p1, p2)`

xin-huang:
Hi @petrelharp 

I added more tests to test the whole `dfe.py` and also modified those error messages to return more accurate information.

stdpopsim/dfe.py
```diff
@@ -45,81 +45,100 @@ class MutationType(object):
...
-            ):
+            if len(self.distribution_args) != 2:
                 raise ValueError(
                     "Normally-distributed sel. coefs. (distribution_type='n') "
-                    "use a (mean, sd) parameterisation, requiring sd > 0."
+                    "use a (mean, sd) parameterisation."
+                )
+            if self.distribution_args[1] <= 0:
+                raise ValueError(
+                    "The sd parameter must be positive."
```
petrelharp, line 112:
```suggestion
            if self.distribution_args[1] < 0:
                raise ValueError(
                    "The sd parameter must be nonnegative."
```

stdpopsim/dfe.py
```diff
@@ -45,81 +45,100 @@ class MutationType(object):
...
-            ):
+            if len(self.distribution_args) != 2:
                 raise ValueError(
                     "Lognormally-distributed sel. coefs. (distribution_type='l') "
                     "use a (meanlog, sdlog) parameterisation, requiring sdlog > 0."
                 )
+            if self.distribution_args[1] <= 0:
+                raise ValueError(
+                    "The sdlog parameter must be positive."
+                )
```
petrelharp, line 141:
```suggestion
            if self.distribution_args[1] < 0:
                raise ValueError(
                    "The sdlog parameter must be nonnegative."
                )
```

stdpopsim/dfe.py
```diff
@@ -204,9 +223,26 @@ def __attrs_post_init__(self):
...
+        ):
+            raise ValueError(
+                "mutation_types must be a list."
+            )
+
+        if (len(self.proportions) > 0) and (len(self.mutation_types) == 0):
+            raise ValueError(
+                "mutation_types is an empty list, while proportions is not."
+            )
+
```
petrelharp, line 243: Isn't this caught by the next test (that tests if they are the same length)?
```suggestion
```

stdpopsim/dfe.py
```diff
@@ -215,9 +251,11 @@ def __attrs_post_init__(self):
...
             if not isinstance(p, (float, int)) or p < 0:
                 raise ValueError("proportions must be nonnegative numbers.")
-        sum_p = sum(self.proportions)
-        if not np.isclose(sum_p, 1):
-            raise ValueError("proportions must sum 1.0.")
+
+        if (len(self.proportions) > 0) and (len(self.mutation_types) > 0):
+            sum_p = sum(self.proportions)
+            if not np.isclose(sum_p, 1):
+                raise ValueError("proportions must sum 1.0.")
```
petrelharp, line 258:
```suggestion
                raise ValueError("proportions must sum to 1.0.")
```

tests/test_dfes.py
```diff
@@ -5,10 +5,219 @@
...
+    """
+    Tests for creating a MutationType instance.
+    """
+
+    def test_default_mutation_type(self):
+        mt = dfe.MutationType()
+        assert mt.dominance_coeff == 0.5
+        assert mt.distribution_type == "f"
+        assert mt.distribution_args == [0]
+        assert mt.convert_to_substitution == True
```
petrelharp, line 23:
```suggestion
        assert mt.convert_to_substitution is True
```

tests/test_dfes.py
```diff
@@ -5,10 +5,219 @@
...
+             )
+        assert mt.Q_scaled_index == []
+
+    def test_create_bad_mutation_type(self):
+        # dominance_coeff must be float
+        with pytest.raises(ValueError) as e_info:
+            mt = dfe.MutationType(
+                     dominance_coeff=True,
+                 )
+        assert str(e_info.value) == "dominance_coeff must be float."
```
petrelharp, line 67:
```suggestion
        assert str(e_info.value) == "dominance_coeff must be a float."
```

tests/test_dfes.py
```diff
@@ -5,10 +5,219 @@
...
+             )
+        assert mt.Q_scaled_index == [0]
+
+        mt = dfe.MutationType(
+                 distribution_type="l",
+                 distribution_args=[0.5, 1],
+             )
+        assert mt.Q_scaled_index == []
+
+    def test_create_bad_mutation_type(self):
```
petrelharp, line 61: These are great, and no need to change these, but FYI, a nicer way to do this is like
```
with pytest.raises(ValueError, match="dominance_coeff"):
    mt = dfe.MutationType(dominance_coeff=True)
```

The code is a bit shorter, and I think it's nice to test on a good substring rather than exact equality of the error string (since what if we do minor edits to the error string in the future).

tests/test_dfes.py
```diff
@@ -67,6 +276,181 @@ def test_wrap_long_lines(self):
...
+                long_description="test",
+                mutation_types=[dfe.MutationType()], 
+            )
+        assert d.proportions == [1]
+
+    def test_create_bad_dfes(self):
+        with pytest.raises(TypeError) as e_info:
+            # id, description, long_description are required
+            d = dfe.DFE()
+        assert str(e_info.value) == "DFE.__init__() missing 3 required keyword-only arguments: 'id', 'description', and 'long_description'"
```
petrelharp, line 340:
```suggestion
        with pytest.raises(TypeError, match="required .* arguments"):
            # id, description, long_description are required
            d = dfe.DFE()
```

petrelharp: Although we don't really need to test this behavior - this is part of the fundamental behavior of python.

tests/test_dfes.py
```diff
@@ -67,6 +276,181 @@ def test_wrap_long_lines(self):
...
+                    proportions=1)
+        assert str(e_info.value) == "proportions must be a list."
+            
+        with pytest.raises(ValueError) as e_info:
+            # mutation_types must be a list
+            d = dfe.DFE(
+                    id="test",
+                    description="test",
+                    long_description="test",
+                    mutation_types=1,
```
petrelharp, line 357:
```suggestion
                    mutation_types=dfe.MutationType(),
```

tests/test_dfes.py
```diff
@@ -67,6 +276,181 @@ def test_wrap_long_lines(self):
...
+            # mutation_types must be a list of MutationType objects
+            d = dfe.DFE(
+                    id="test",
+                    description="test",
+                    long_description="test",
+                    mutation_types=["neutral"],
+                )
+        assert str(e_info.value) == "mutation_types must be a list of MutationType objects."
+
+    def test_dfe_is_neutral(self):
```
petrelharp, line 425: Let's also test that the DFE with no mutation types is neutral?

petrelharp:
This is *great*! I've just a few minor comments - can you see about those?

petrelharp:
> And I have a suggestion. I think we should use dictionaries for mutation_types and proportions with the keys neutral, negative, and positive, because different people may implement neutral, negative, and positive mutations in different orders if using lists. And we should provide functions to return proportions for neutral, negative, and positive mutations. Then we don't need to hard-code these proportions in the analysis2 pipelines.

Hm, interesting idea. I don't think we can do just as you suggested, because for instance some mutation types can be both positive and negative (e.g., type "n" with mean 0 and SD 1). Can you point to the place in the analysis2 code where this is being awkward?

xin-huang:
Hi @petrelharp 

I updated the codes based on your comments.

The analysis2 code is here, for example,
https://github.com/popsim-consortium/analysis2/blob/main/workflows/dfe.snake#L85

Can we provide functions to return these proportions for different mutation types?

petrelharp (approved):
LGTM, thanks!

petrelharp:
Ok, let's squash and rebase this, and I'll merge it after checking the rebase.

And, about your idea - it looks like you'd like to get information about the DFEs from the tree sequence? Like, out of metadata? (We should probably move this to a different issue.)

xin-huang:
I moved those DFE tests from `test_genomes.py` to `test_dfes.py`.
I can build the documents in my computer, but it failed here.

stdpopsim/dfe.py
```diff
@@ -39,87 +38,94 @@ class MutationType(object):
     :vartype convert_to_substitution: bool
     """
 
-    dominance_coeff = attr.ib(default=0.5, type=float)
+    dominance_coeff = attr.ib(default=0.5, type=[float, int])
```
petrelharp, line 41: I think this is why the docs aren't building (when you build, do you at least get a warning?). The error says `error while formatting arguments for stdpopsim.MutationType: unhashable type: 'list'`; perhaps you should use a tuple instead?

OH WAIT: I think that dominance coefficient should be a float - why int? I think attrs does not do anything with this anyhow? (From the docs for `attr.ib( )`:
Screenshot from 2022-07-21 22-03-41

stdpopsim/dfe.py
```diff
@@ -196,27 +202,32 @@ class DFE:
...
         self.citations = [] if self.citations is None else self.citations
         if self.proportions == [] and len(self.mutation_types) == 1:
             self.proportions = [1]
 
-        if not (
-            isinstance(self.proportions, collections.abc.Collection)
-            and isinstance(self.mutation_types, collections.abc.Collection)
-            and len(self.proportions) == len(self.mutation_types)
-        ):
+        if not (isinstance(self.proportions, list)):
```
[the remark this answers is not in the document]
petrelharp: OH WAIT: the reason for this was that `np.array([0.2, 0.8])`, for instance is *not* a list. And we certainly want to allow numpy arrays for `proportions`. Maybe `isinstance(proportions, (collections.abc.Sequence, np.ndarray))`?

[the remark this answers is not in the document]
petrelharp: (since an ndarray is not a Sequence, for some reason...)

stdpopsim/dfe.py
```diff
@@ -196,27 +202,32 @@ class DFE:
...
             )
 
         for p in self.proportions:
             if not isinstance(p, (float, int)) or p < 0:
                 raise ValueError("proportions must be nonnegative numbers.")
-        sum_p = sum(self.proportions)
-        if not np.isclose(sum_p, 1):
-            raise ValueError("proportions must sum 1.0.")
+
+        if (len(self.proportions) > 0) and (len(self.mutation_types) > 0):
```
petrelharp, line 227: We don't need to check the length of `mutation_types` here (we already know it's the same length as proporitons).
```suggestion
        if len(self.proportions) > 0):
```

petrelharp:
Looks good! Just two very minor things (apologies!)

stdpopsim/dfe.py
xin-huang: I didn't get any warning about `error while formatting arguments for stdpopsim.MutationType: unhashable type: 'list'` in my computer, only got `RuntimeWarning: overflow encountered in power`.

tests/test_dfes.py
```diff
@@ -0,0 +1,776 @@
...
+        mt = dfe.MutationType()
+        assert mt.convert_to_substitution is True
+        for c in (True, False):
+            mt = dfe.MutationType(convert_to_substitution=c)
+            assert mt.convert_to_substitution == c
+
+    def test_dominance_coeff(self):
+        mt = dfe.MutationType()
+        assert mt.dominance_coeff == 0.5
+        for dominance_coeff in (-10, 0, 0.5, 1, 50):
```
xin-huang, line 265: @petrelharp because I found the test codes allows `dominance_coeff` to be `int`. So we should change this line into `for dominance_coeff in (-10.0, 0.0, 0.5, 1.0, 50.0)` and only allow `dominance_coeff` as `float`?

petrelharp: If you pass in an int to dominance_coeff do we get an error? We could put a `float( )` around `dominance_coeff`?

xin-huang: No error now, because I changed the condition from `if not isinstance(self.dominance_coeff, float):` to `if not isinstance(self.dominance_coeff, (float, int)):`

petrelharp: No harm in saying "float" in the docs, though.

petrelharp:
Are you all done here? Should I take a quick skim through and merge this?

xin-huang:
> Are you all done here? Should I take a quick skim through and merge this?

Yes, please.

petrelharp (commented):
I've made a few more nitpicks, mostly related to cleaning up the code you moved over from the other file. And, I'm fixing them up (see the next commit)!

petrelharp:
OK - I've dealt with all those things above. Want to have a look, @xin-huang and see if my changes look okay? If so, go ahead and squash, then I'll merge these in (supposing it passes tests).

petrelharp:
The tests pass - codecov isn't happy but that's because of #1308; we can ignore that here.

xin-huang:
LGTM
squashed

Issue #1082, opened 2021-11-02
Align selective sweeps API with DFEs


petrelharp:
The [API for incorporating selective sweeps](https://popsim-consortium.github.io/stdpopsim-docs/main/tutorial.html#selective-sweep) works directly with mutation type IDs (i.e., they need to know what the SLiM ID of the mutation type will be once it is simulated). Here's the rewrite of that example using the `add_DFE` way of doing things from #1078:
```
species = stdpopsim.get_species("DroMel")
model = stdpopsim.PiecewiseConstantSize(100000)
samples = model.get_samples(100)
contig = species.get_contig("2L", length_multiplier=0.01)

### start modified stuff
mt = stdpopsim.MutationType(
        distribution_type="f",
        dominance_coeff=1.0,
        distribution_args=[0.5],
        convert_to_substitution=False,
)
dfe = stdpopsim.DFE(
    id="new_mutation",
    mutation_types=[mt],
    proportions=[1.0],
    description="added mutation",
    long_description="mutation type to be added",
)
contig.add_DFE(
    intervals=np.empty((0,2), dtype='int'),
    DFE=dfe,
)
for mt_info in contig.mutation_types():
    if mt_info["dfe_id"] == dfe.id:
        break
                                                  
mut_id = mt_info["id"]
### end modified stuff

coordinate = int(contig.recombination_map.sequence_length / 2)
T_mut = 1000
extended_events = [
    stdpopsim.ext.DrawMutation(
        time=T_mut,
        mutation_type_id=mut_id,
        population_id=0,
        coordinate=coordinate,
        save=True,
    )
]
```

This works, but to make it easier, I propose making a new `Contig` method that does the (a) make a mutation type (b) make a DFE with that mutation type  (c) add it with no intervals to the Contig steps, and (d) return the mutation type ID. (i.e., everything in the `###` above) We could call this `Contig.add_single_mutation_type`? Or, something more evocative? What do you think, @grahamgower?

grahamgower:
Yeah, we definitely need a wrapper function for this to avoid confusion. It seems a bit awkward that the mutation type needs to have a dummy DFE, but if this simplifies the implementation then I guess it's a win. Perhaps the name of the function could be more explicit, e.g. `Contig.add_single_site_mutation_type`?

petrelharp:
[...]

PR #1306, opened 2022-07-15
Use string `mutation_id` in place of integer `mutation_type_id` in extended events 

nspope:
Following the rationale/outline [here](https://github.com/popsim-consortium/stdpopsim/issues/1082#issuecomment-966730838):

- Replace `mutation_type_id` with string `mutation_id` in various extended events (`ext.DrawMutation`, etc)
- Add `Contig.add_single_site_mutation_type` to create a named, interval-less DFE representing a point mutation
- Modify `slim_makescript` to match `mutation_id` to `DFE.id`; check for uniqueness of name, lack of interval, etc when parsing extended events
- Update tests accordingly

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1224,65 +1224,54 @@ of at least 0.8 by the end. Next, we'll walk through the steps required to do th
     this conditioning works by re-running the simulation until the condition is
     achieved, a nearly impossible condition will result in very long run times.
 
-First, we need to define the mutation type for the selected mutation.
-So we can refer to it later, we need its "mutation type ID". This is just
-the index of the new mutation type in the contig's list of mutation types.
+First, we need to add a single site mutation in the middle of the contig.
+This represents a mutation of a single base pair, with associated fitness and
+dominance coefficients.  The single site mutation must have a unique string ID
+so that we can refer to it later.
 
 .. code-block:: python
 
-    mt = stdpopsim.MutationType(
-        distribution_type="f",
-        dominance_coeff=1.0,
-        distribution_args=[0.5],
-        convert_to_substitution=False,
+    mut_id = "hard sweep"
+    coordinate = round(contig.recombination_map.sequence_length / 2)
+    contig.add_single_site(
+        id=mut_id,
+        coordinate=coordinate,
+        fitness_coeff=0.1,
+        dominance_coeff=0.5,
     )
-    dfe = stdpopsim.DFE(
-        id="new_mutation",
-        mutation_types=[mt],
-        proportions=[1.0],
-        description="added mutation",
-        long_description="mutation type to be added",
-    )
-    contig.add_dfe(
-        intervals=np.empty((0, 2), dtype="int"),
-        DFE=dfe,
-    )
-    for mt_info in contig.mutation_types():
-        if mt_info["dfe_id"] == dfe.id:
-            break
 
-    mut_id = mt_info["id"]
+.. note::
+
+    Note that single site mutations are internally stored as DFEs, and DFEs
+    cannot overlap. As a consequence, if a DFE is added to the contig over an
+    interval that contains an existing single site mutation, the latter will be
+    "overwritten" and an error will be raised by the simulation engine.
 
 Next, we will set up the "extended events" which will modify the demography.
-The first extended event is placing of the selected mutation,
-which will occur in a random individual from the first population (id 0),
-in the middle of the contig, 1000 generations ago.
-We specify ``save=True`` to ``stdpopsim.ext.DrawMutation``
-so that the simulation can restart from that point if the mutation is lost.
+The first extended event is the origination of the selected mutation, which
+will occur in a random individual from the first population (id 0), 1000
+generations ago.  We specify ``save=True`` to ``stdpopsim.ext.DrawMutation`` so
+that the simulation can restart from that point if the mutation is lost.
 
 .. code-block:: python
 
-    coordinate = round(contig.recombination_map.sequence_length / 2)
     T_mut = 1000
     extended_events = [
         stdpopsim.ext.DrawMutation(
             time=T_mut,
-            mutation_type_id=mut_id,
+            single_site_id=mut_id,
             population_id=0,
-            coordinate=coordinate,
             save=True,
         )
     ]
 
-Next, we condition on the mutation not being lost.
-Since in the next step we condition on the mutation being at 80% frequency at
-the end, this is redundant, but it allows the simulation to immediately restart
-from any generation in which the mutation is lost, rather than waiting until the end.
-Note that this conditioning must start one
-generation after the mutation is placed, for which we use
-``stdpopsim.ext.GenerationAfter(T_mut)``.
-We cannot simply specify ``T_mut - 1`` if rescaling is present,
-otherwise the conditioning would start
+Next, we condition on the mutation not being lost.  Since in the next step we
+condition on the mutation being at 80% frequency at the end, this is redundant,
+but it allows the simulation to immediately restart from any generation in
+which the mutation is lost, rather than waiting until the end.  Note that this
+conditioning must start one generation after the mutation is placed, for which
+we use ``stdpopsim.ext.GenerationAfter(T_mut)``.  We cannot simply specify
+``T_mut - 1`` if rescaling is present, otherwise the conditioning would start
 at the same generation when the mutation is placed.
 
 .. code-block:: python
@@ -1291,7 +1280,7 @@ at the same generation when the mutation is placed.
         stdpopsim.ext.ConditionOnAlleleFrequency(
             start_time=stdpopsim.ext.GenerationAfter(T_mut),
             end_time=0,
-            mutation_type_id=mut_id,
+            single_site_id=mut_id,
             population_id=0,
             op=">",
             allele_frequency=0.0,
@@ -1307,7 +1296,7 @@ Finally, we condition on the mutation being above 80% at the end of the simulati
         stdpopsim.ext.ConditionOnAlleleFrequency(
             start_time=0,
             end_time=0,
-            mutation_type_id=mut_id,
+            single_site_id=mut_id,
             population_id=0,
             op=">=",
             allele_frequency=0.8,
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1224,10 +1224,12 @@ of at least 0.8 by the end. Next, we'll walk through the steps required to do th
     this conditioning works by re-running the simulation until the condition is
     achieved, a nearly impossible condition will result in very long run times.
 
-First, we need to add a single site mutation in the middle of the contig.
-This represents a mutation of a single base pair, with associated fitness and
-dominance coefficients.  The single site mutation must have a unique string ID
-so that we can refer to it later.
+To set things up, we need to first add the site at which the selected mutation
+will occur.  This is like adding a DFE, except to a single site -- we're saying
+that there is a potential mutation at a particular site with defined fitness
+consequences. So that we can refer to the single site later, we give it a
+unique string ID. Here, we'll add the site in the middle of the contig, with ID
+"hard sweep".
 
 .. code-block:: python
 
@@ -1242,10 +1244,11 @@ so that we can refer to it later.
 
 .. note::
 
-    Note that single site mutations are internally stored as DFEs, and DFEs
-    cannot overlap. As a consequence, if a DFE is added to the contig over an
-    interval that contains an existing single site mutation, the latter will be
-    "overwritten" and an error will be raised by the simulation engine.
+    Note that single site mutations are internally stored as DFEs, and more
+    than one DFE cannot apply to the same segment of genome. As a consequence,
+    if another DFE is added to the contig over an interval that already
+    contains a single site mutation, the single site mutation will be
+    "overwritten" and an error will be raised in simulation.
 
 Next, we will set up the "extended events" which will modify the demography.
 The first extended event is the origination of the selected mutation, which
@@ -1261,7 +1264,6 @@ that the simulation can restart from that point if the mutation is lost.
             time=T_mut,
             single_site_id=mut_id,
             population_id=0,
-            save=True,
         )
     ]
 
--- a/stdpopsim/ext/selection.py
+++ b/stdpopsim/ext/selection.py
@@ -76,7 +76,7 @@ class DrawMutation(ExtendedEvent):
     """
 
     time = attr.ib(type=float)
-    mutation_type_id = attr.ib(type=int)
+    mutation_id = attr.ib(type=str)
     population_id = attr.ib(type=int)
     coordinate = attr.ib(type=int)
     save = attr.ib(type=bool, default=False)
@@ -100,7 +100,7 @@ class ChangeMutationFitness(ExtendedEvent):
 
     start_time = attr.ib(type=float)
     end_time = attr.ib(type=float)
-    mutation_type_id = attr.ib(type=int)
+    mutation_id = attr.ib(type=str)
     population_id = attr.ib(type=int)
     selection_coeff = attr.ib(type=float)
     dominance_coeff = attr.ib(type=float)
@@ -127,7 +127,7 @@ class ConditionOnAlleleFrequency(ExtendedEvent):
 
     start_time = attr.ib(type=float)
     end_time = attr.ib(type=float)
-    mutation_type_id = attr.ib(type=int)
+    mutation_id = attr.ib(type=str)
     population_id = attr.ib(type=int)
     op = attr.ib(type=str, default=None)
     allele_frequency = attr.ib(type=float, default=None)
--- a/stdpopsim/ext/selection.py
+++ b/stdpopsim/ext/selection.py
@@ -76,9 +76,8 @@ class DrawMutation(ExtendedEvent):
     """
 
     time = attr.ib(type=float)
-    mutation_id = attr.ib(type=str)
+    single_site_id = attr.ib(type=str)
     population_id = attr.ib(type=int)
-    coordinate = attr.ib(type=int)
     save = attr.ib(type=bool, default=False)
 
     def __attrs_post_init__(self):
@@ -100,7 +99,7 @@ class ChangeMutationFitness(ExtendedEvent):
 
     start_time = attr.ib(type=float)
     end_time = attr.ib(type=float)
-    mutation_id = attr.ib(type=str)
+    single_site_id = attr.ib(type=str)
     population_id = attr.ib(type=int)
     selection_coeff = attr.ib(type=float)
     dominance_coeff = attr.ib(type=float)
@@ -127,7 +126,7 @@ class ConditionOnAlleleFrequency(ExtendedEvent):
 
     start_time = attr.ib(type=float)
     end_time = attr.ib(type=float)
-    mutation_id = attr.ib(type=str)
+    single_site_id = attr.ib(type=str)
     population_id = attr.ib(type=int)
     op = attr.ib(type=str, default=None)
     allele_frequency = attr.ib(type=float, default=None)
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -462,7 +462,7 @@ class Contig:
             long_description=long_description,
         )
         self.add_dfe(
-            intervals=np.array([[coordinate - 1, coordinate]], dtype="int"),
+            intervals=np.array([[coordinate, coordinate + 1]], dtype="int"),
             DFE=dfe,
         )
 
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -386,6 +386,64 @@ class Contig:
         self.dfe_list.append(DFE)
         self.interval_list.append(intervals)
 
+    def add_single_site_mutation_type(self, id, selection_coeff=0.0, dominance_coeff=1.0):
+        """
+        Adds a DFE representing a mutation at a single site, contain a single
+        :class:`MutationType` with the provided parameters. The string ID of
+        the mutation may be referenced by extended events passed to the
+        simulation engine.
+
+        For instance, to simulate a selective sweep,
+        ```
+        contig.add_single_site_mutation_type(
+            id = "sweep_variant", 
+            selection_coef=0.1,
+        )
+        extended_events = [
+    	    stdpopsim.ext.DrawMutation(
+    	        time=mutation_time,
+    	        mutation_id="sweep_variant",
+    	        population_id=0,
+    	        coordinate=mutation_coordinate,
+    	        save=True,
+    	    ),
+    	    stdpopsim.ext.ConditionOnAlleleFrequency(
+    	        start_time=stdpopsim.ext.GenerationAfter(mutation_time),
+    	        end_time=0,
+    	        mutation_id="sweep_variant",
+    	        population_id=0,
+    	        op=">",
+    	        allele_frequency=0.0,
+    	    )
+        ]
+        engine = stdpopsim.get_engine("slim")
+        ts_sweep = engine.simulate(...,
+            extended_events=extended_events,
+        )
+        ```
+
+        :param str id: A unique identifier for the DFE representing the mutation.
+        :param float selection_coeff: The selection coefficient of the mutation, see :class:`MutationType`.
+        :param float dominance_coeff: The dominance coefficient of the mutation, see :class:`MutationType`.
+        """
+        mut_type = stdpopsim.MutationType(
+                distribution_type="f",
+                dominance_coeff=dominance_coeff,
+                distribution_args=[selection_coeff],
+                convert_to_substitution=False,
+        )
+        dfe = stdpopsim.DFE(
+            id=id,
+            mutation_types=[mut_type],
+            proportions=[1.0],
+            description="Single site mutation",
+            long_description="Added by 'Contig.add_single_site_mutation_type'",
+        )
+        self.add_dfe(
+            intervals=np.empty((0,2), dtype='int'), 
+            DFE=dfe,
+        )
+
     @property
     def is_neutral(self):
         """
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -462,7 +462,7 @@ class Contig:
             long_description=long_description,
         )
         self.add_dfe(
-            intervals=np.array([[coordinate, coordinate + 1]], dtype="int"),
+            intervals=np.array([[coordinate - 1, coordinate]], dtype="int"),
             DFE=dfe,
         )
 
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -386,35 +386,40 @@ class Contig:
         self.dfe_list.append(DFE)
         self.interval_list.append(intervals)
 
-    def add_single_site_mutation_type(
-        self, id, selection_coeff=0.0, dominance_coeff=1.0
+    def add_single_site(
+        self,
+        id,
+        coordinate,
+        selection_coeff=0.0,
+        dominance_coeff=0.5,
+        description=None,
+        long_description=None,
     ):
         """
-        Adds a DFE representing a mutation at a single site, contain a single
-        :class:`MutationType` with the provided parameters. The string ID of
-        the mutation may be referenced by extended events passed to the
-        simulation engine.
+        Adds a model of a single site mutation at the provided coordinate.
+        The string label of the single site may be referenced by extended
+        events passed to the simulation engine.
 
         For instance, to simulate a selective sweep,
 
         .. code-block:: python
 
-            contig.add_single_site_mutation_type(
+            contig.add_single_site(
                 id="sweep_variant",
-                selection_coef=0.1,
+                selection_coeff=0.1,
+                coordinate=mutation_coordinate,
             )
             extended_events = [
                 stdpopsim.ext.DrawMutation(
                     time=mutation_time,
-                    mutation_id="sweep_variant",
+                    single_site_id="sweep_variant",
                     population_id=0,
-                    coordinate=mutation_coordinate,
                     save=True,
                 ),
                 stdpopsim.ext.ConditionOnAlleleFrequency(
                     start_time=stdpopsim.ext.GenerationAfter(mutation_time),
                     end_time=0,
-                    mutation_id="sweep_variant",
+                    single_site_id="sweep_variant",
                     population_id=0,
                     op=">",
                     allele_frequency=0.0,
@@ -426,12 +431,23 @@ class Contig:
                 extended_events=extended_events,
             )
 
-        :param str id: A unique identifier for the DFE representing the mutation.
+        :param str id: A unique identifier for the single site.
+        :param int coordinate: The coordinate of the site on the contig.
         :param float selection_coeff: The selection coefficient of the mutation,
             see :class:`MutationType` for details.
         :param float dominance_coeff: The dominance coefficient of the mutation,
             see :class:`MutationType` for details.
+        :param str description: A short description of the single site mutation
+            model as it would be used in written text, e.g., "Strong selective
+            sweep".
+        :param str long_description: A concise, but detailed, summary of the
+            single site mutation model.
         """
+        if description is None:
+            description = "Single site mutation"
+        if long_description is None:
+            long_description = "Added by Contig.add_single_site"
+
         mut_type = stdpopsim.MutationType(
             distribution_type="f",
             dominance_coeff=dominance_coeff,
@@ -442,11 +458,11 @@ class Contig:
             id=id,
             mutation_types=[mut_type],
             proportions=[1.0],
-            description="Single site mutation",
-            long_description="Added by 'Contig.add_single_site_mutation_type'",
+            description=description,
+            long_description=long_description,
         )
         self.add_dfe(
-            intervals=np.empty((0, 2), dtype="int"),
+            intervals=np.array([[coordinate, coordinate + 1]], dtype="int"),
             DFE=dfe,
         )
 
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -386,7 +386,9 @@ class Contig:
         self.dfe_list.append(DFE)
         self.interval_list.append(intervals)
 
-    def add_single_site_mutation_type(self, id, selection_coeff=0.0, dominance_coeff=1.0):
+    def add_single_site_mutation_type(
+        self, id, selection_coeff=0.0, dominance_coeff=1.0
+    ):
         """
         Adds a DFE representing a mutation at a single site, contain a single
         :class:`MutationType` with the provided parameters. The string ID of
@@ -394,43 +396,47 @@ class Contig:
         simulation engine.
 
         For instance, to simulate a selective sweep,
-        ```
-        contig.add_single_site_mutation_type(
-            id = "sweep_variant", 
-            selection_coef=0.1,
-        )
-        extended_events = [
-    	    stdpopsim.ext.DrawMutation(
-    	        time=mutation_time,
-    	        mutation_id="sweep_variant",
-    	        population_id=0,
-    	        coordinate=mutation_coordinate,
-    	        save=True,
-    	    ),
-    	    stdpopsim.ext.ConditionOnAlleleFrequency(
-    	        start_time=stdpopsim.ext.GenerationAfter(mutation_time),
-    	        end_time=0,
-    	        mutation_id="sweep_variant",
-    	        population_id=0,
-    	        op=">",
-    	        allele_frequency=0.0,
-    	    )
-        ]
-        engine = stdpopsim.get_engine("slim")
-        ts_sweep = engine.simulate(...,
-            extended_events=extended_events,
-        )
-        ```
+
+        .. code-block:: python
+
+            contig.add_single_site_mutation_type(
+                id="sweep_variant",
+                selection_coef=0.1,
+            )
+            extended_events = [
+                stdpopsim.ext.DrawMutation(
+                    time=mutation_time,
+                    mutation_id="sweep_variant",
+                    population_id=0,
+                    coordinate=mutation_coordinate,
+                    save=True,
+                ),
+                stdpopsim.ext.ConditionOnAlleleFrequency(
+                    start_time=stdpopsim.ext.GenerationAfter(mutation_time),
+                    end_time=0,
+                    mutation_id="sweep_variant",
+                    population_id=0,
+                    op=">",
+                    allele_frequency=0.0,
+                ),
+            ]
+            engine = stdpopsim.get_engine("slim")
+            ts_sweep = engine.simulate(
+                ...,
+                extended_events=extended_events,
+            )
 
         :param str id: A unique identifier for the DFE representing the mutation.
-        :param float selection_coeff: The selection coefficient of the mutation, see :class:`MutationType`.
-        :param float dominance_coeff: The dominance coefficient of the mutation, see :class:`MutationType`.
+        :param float selection_coeff: The selection coefficient of the mutation,
+            see :class:`MutationType` for details.
+        :param float dominance_coeff: The dominance coefficient of the mutation,
+            see :class:`MutationType` for details.
         """
         mut_type = stdpopsim.MutationType(
-                distribution_type="f",
-                dominance_coeff=dominance_coeff,
-                distribution_args=[selection_coeff],
-                convert_to_substitution=False,
+            distribution_type="f",
+            dominance_coeff=dominance_coeff,
+            distribution_args=[selection_coeff],
+            convert_to_substitution=False,
         )
         dfe = stdpopsim.DFE(
             id=id,
@@ -440,7 +446,7 @@ class Contig:
             long_description="Added by 'Contig.add_single_site_mutation_type'",
         )
         self.add_dfe(
-            intervals=np.empty((0,2), dtype='int'), 
+            intervals=np.empty((0, 2), dtype="int"),
             DFE=dfe,
         )
 
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -405,21 +405,21 @@ class Contig:
         .. code-block:: python
 
             contig.add_single_site(
-                id="sweep_variant",
+                id="hard_sweep",
                 selection_coeff=0.1,
                 coordinate=mutation_coordinate,
             )
             extended_events = [
                 stdpopsim.ext.DrawMutation(
                     time=mutation_time,
-                    single_site_id="sweep_variant",
+                    single_site_id="hard_sweep",
                     population_id=0,
                     save=True,
                 ),
                 stdpopsim.ext.ConditionOnAlleleFrequency(
                     start_time=stdpopsim.ext.GenerationAfter(mutation_time),
                     end_time=0,
-                    single_site_id="sweep_variant",
+                    single_site_id="hard_sweep",
                     population_id=0,
                     op=">",
                     allele_frequency=0.0,
@@ -440,8 +440,7 @@ class Contig:
         :param str description: A short description of the single site mutation
             model as it would be used in written text, e.g., "Strong selective
             sweep".
-        :param str long_description: A concise, but detailed, summary of the
-            single site mutation model.
+        :param str long_description: If necessary, a more detailed summary.
         """
         if description is None:
             description = "Single site mutation"
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -950,19 +950,20 @@ def slim_makescript(
     slim_mutation_ids = []
     slim_mutation_dfe = []
     for i, x in _dfe_to_mtypes(contig).items():
-      slim_mutation_ids.extend([u[0] for u in x])
-      slim_mutation_dfe.extend([i for u in x])
+        slim_mutation_ids.extend([u[0] for u in x])
+        slim_mutation_dfe.extend([i for u in x])
     for ee in extended_events:
         if hasattr(ee, "mutation_id"):
             mut_id = getattr(ee, "mutation_id")
             cls_name = ee.__class__.__name__
             mtype_idx = [
-                j for j,i in enumerate(slim_mutation_dfe) 
+                j
+                for j, i in enumerate(slim_mutation_dfe)
                 if contig.dfe_list[i].id == mut_id
             ]
             n_mtypes = len(mtype_idx)
             # TODO: are these checks too restrictive? For example, these will
-            # prevent using DrawMutation with an arbitrary MutationType in an 
+            # prevent using DrawMutation with an arbitrary MutationType in an
             # arbitary DFE.
             if n_mtypes != 1:
                 raise ValueError(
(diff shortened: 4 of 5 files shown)


stdpopsim/genomes.py
```diff
@@ -386,6 +386,70 @@ def add_dfe(self, intervals, DFE):
...
+                    population_id=0,
+                    op=">",
+                    allele_frequency=0.0,
+                ),
+            ]
+            engine = stdpopsim.get_engine("slim")
+            ts_sweep = engine.simulate(
+                ...,
+                extended_events=extended_events,
+            )
```
andrewkern, line 427: so I guess next step is to write a helper function for this code block to ease the sweep simulation interface?

stdpopsim/ext/selection.py
```diff
@@ -76,7 +76,7 @@ class DrawMutation(ExtendedEvent):
     """
 
     time = attr.ib(type=float)
-    mutation_type_id = attr.ib(type=int)
+    mutation_id = attr.ib(type=str)
```
andrewkern, line 79: this change looks fine to me. I'd want to check with @mufernando, who i believe agonized a bit over `mutation_type_id`s before, but everything seems to be working on my end

andrewkern (commented):
This looks good to me @nspope. I do wonder if the checks are too restrictive? Can we simulate an arbitrary DFE _and_ add a single mutation? that's the goal

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
+            mut_id = getattr(ee, "mutation_id")
+            cls_name = ee.__class__.__name__
+            mtype_idx = [
+                j
+                for j, i in enumerate(slim_mutation_dfe)
+                if contig.dfe_list[i].id == mut_id
+            ]
+            n_mtypes = len(mtype_idx)
+            # TODO: are these checks too restrictive? For example, these will
+            # prevent using DrawMutation with an arbitrary MutationType in an
```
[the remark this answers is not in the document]
nspope: Yup, that should be possible, because the check is just ensuring that the mutation ID in each extended event maps to a single mutation (so the DFE id is synonymous with the mutation id in SLiM). These have no associated intervals, so should not "intersect" with any other DFEs. So, these point mutations can coexist with an arbitrary number of other DFEs with multiple mutation types and non-unique ids.

mufernando:
Nate, this is looking good!

We also use the mutation type ids in the top-level metadata that we add to the tree sequences. It is necessary that the mutation type ids (int) be kept there because SLiM does not have a way to define the mutation types with strings, I believe. But, it would be nice to add the string name to the metadata as well.

https://github.com/popsim-consortium/stdpopsim/blob/6d9c1d67b6214185396bde2ddf81aef238ec744a/stdpopsim/slim_engine.py#L639

nspope:
Thanks @mufernando! Numeric mutation type ids are still added to the metadata as before. For extended events, the string ids that reference particular mutations are in reality just DFE ids -- where the DFE is a "dummy" with no interval and a single fixed mutation type. So I think these would be clear enough in the existing schema?

andrewkern:
hey @petrelharp -- might taking a look at this PR when you get a chance?

stdpopsim/genomes.py
```diff
@@ -386,6 +386,70 @@ def add_dfe(self, intervals, DFE):
         self.dfe_list.append(DFE)
         self.interval_list.append(intervals)
 
+    def add_single_site_mutation_type(
+        self, id, selection_coeff=0.0, dominance_coeff=1.0
```
petrelharp, line 390: not sure about this default - should be the same as the usual default (which is 0.5: https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/dfe.py#L42 )

stdpopsim/genomes.py
```diff
@@ -386,6 +386,70 @@ def add_dfe(self, intervals, DFE):
...
+    ):
+        """
+        Adds a DFE representing a mutation at a single site, contain a single
```
petrelharp, line 393: It's a bit confusing to call this a DFE ("distribution of fitness effects"), even though that's what it is under the hood. I'm not sure how else to explain this right now, though?

stdpopsim/genomes.py
```diff
@@ -386,6 +386,70 @@ def add_dfe(self, intervals, DFE):
...
+            distribution_type="f",
+            dominance_coeff=dominance_coeff,
+            distribution_args=[selection_coeff],
+            convert_to_substitution=False,
+        )
+        dfe = stdpopsim.DFE(
+            id=id,
+            mutation_types=[mut_type],
+            proportions=[1.0],
+            description="Single site mutation",
```
petrelharp, line 445: the description and long description should be arguments, too? But, defaulting to these values? (could do `description=None` and then `if description is None: description = "Single site mutation"`)

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
+                        f"{cls_name} event with mutation_id '{mut_id}' is "
+                        f"a single site mutation, but is associated with "
+                        f"interval(s) {dfe_intervals}."
+                    )
+
         if hasattr(ee, "mutation_type_id"):
+            # TODO: this is a bit redundant now that ee.mutation_type
+            # replaces ee.mutation_type_id in the API, but should
+            # allow setattr(ee, "mutation_type_id", <int>) to work for
+            # arbitrary DFEs, if ee.mutation_type is None.
```
petrelharp, line 990: Not sure I'm following - are you saying that leaving this bit of code in here would allow people do do the undocumented workflow
```
ee = ExtendedEvent(mutation_type=None, ...)
setattr(ee, "mutation_type_id", 1)
slim_makescript( ..., extended_events=[ee])
```
... ?

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
+            cls_name = ee.__class__.__name__
+            mtype_idx = [
+                j
+                for j, i in enumerate(slim_mutation_dfe)
+                if contig.dfe_list[i].id == mut_id
+            ]
+            n_mtypes = len(mtype_idx)
+            # TODO: are these checks too restrictive? For example, these will
+            # prevent using DrawMutation with an arbitrary MutationType in an
+            # arbitary DFE.
```
petrelharp, line 967: IIRC, the machinery for conditioning on frequencies under the hood in SLiM requires that the mutation in question be the only one of its mutation type, meaning that this does have to be a requirement.

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
+            if n_mtypes != 1:
+                raise ValueError(
+                    f"{cls_name} event with mutation_id '{mut_id}' must "
+                    f"map to a single MutationType instance on {contig}, "
+                    f"but has {n_mtypes} associated MutationType instances."
```
petrelharp, line 972: I think this could be said more clearly? Something like "the mutation id associated with this event must be unique, but there are X other mutation types on this contig with the same mutation id."?

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
+                )
+            else:
```
petrelharp, line 974: don't need an `else:` after a `raise`

stdpopsim/slim_engine.py
petrelharp: So, I guess I think this code block should be part of the previous `if hasattr(ee, "mutation_id"):` code block. (We shouldn't be supporting that undocumented workflow).

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
+                        f"interval(s) {dfe_intervals}."
+                    )
+
         if hasattr(ee, "mutation_type_id"):
+            # TODO: this is a bit redundant now that ee.mutation_type
+            # replaces ee.mutation_type_id in the API, but should
+            # allow setattr(ee, "mutation_type_id", <int>) to work for
+            # arbitrary DFEs, if ee.mutation_type is None.
             mt_id = getattr(ee, "mutation_type_id")
             cls_name = ee.__class__.__name__
```
petrelharp, line 992: Is `cls_name` used outside of the `raise` statement?  If not, just do `{ee.__class__.__name__}` instead of defining the variable?

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,54 @@ def fix_time(event):
...
             if mt_id not in slim_mutation_ids:
```
petrelharp, line 993: I think this can now be an `assert` as we should have tested that things all worked above?

tests/test_slim_engine.py
```diff
@@ -1420,17 +1422,118 @@ def test_no_mutation_types_defined(self):
...
+            proportions=[0.5, 0.5],
+            description="test test",
+            long_description="test test test",
+        )
+        contig.add_dfe(
+            intervals=np.empty((0, 2), dtype="int"),
+            DFE=dfe,
+        )
+        engine = stdpopsim.get_engine("slim")
+        with pytest.raises(ValueError):
```
petrelharp, line 1453: We should be checking we're actually throwing the right ValueError here, like:
```suggestion
        with pytest.raises(ValueError, match=<some substring of the error message we're trying to hit>):
```

petrelharp:
It's a bit weird this is called `mutation_id` when actually it's the ID of a DFE? But then, it's also weird to talk about the "DFE" of a single mutation? Maybe we could call it `single_site_mutation_type_id`, or something similar but shorter, matching where it's specified (which is in `add_single_site_mutation_type`)?

If there's higher-level API planned then maybe this is redundant... but, in that case it'd be nice to sketch out what that higher-level API is now?

stdpopsim/slim_engine.py
nspope: I was thinking in terms of pickled extended events created using the old `mutation_type_id` API being passed to the simulation engine. But you're right, if it's undocumented it should go. I could add a check that `mutation_type_id` is unset a bit earlier in that code block.

tests/test_slim_engine.py
```diff
@@ -1420,17 +1422,118 @@ def test_no_mutation_types_defined(self):
...
+            engine.simulate(
+                demographic_model=self.model,
+                contig=contig,
+                samples=self.samples,
+                extended_events=extended_events,
+                dry_run=True,
+            )
+
+    def test_mutation_with_oor_mutation_type_id(self):
```
petrelharp, line 1506: We don't want to make this part of the API, so shouldn't test it, I think.

tests/test_slim_engine.py
```diff
@@ -1420,17 +1422,118 @@ def test_no_mutation_types_defined(self):
...
+                demographic_model=self.model,
+                contig=contig,
+                samples=self.samples,
+                extended_events=[draw_mut],
+                dry_run=True,
+            )
+
     def test_bad_time(self):
         for time in (-1,):
             with pytest.raises(ValueError):
```
petrelharp, line 1528: here we really need a `match=XXX` because this could *also* throw the error that there's no such mutation_id.

petrelharp:
I've got some minor comments (sorry there's so many of them). In general, this is definitely a win - the examples in the test code are simpler!

Also note, we'll need to fix up the examples in the documentation, I think? (or, maybe this stuff is undocumented yet?)

nspope:
Thanks @petrelharp that's great feedback. I'll update the docs at the end once we sort out naming.

What @andrewkern suggested for a higher-level API were convenience functions along the lines of,
```
contig.add_single_site_mutation_type(id, fitness_effect)
sweep_events = selective_sweep(id, coordinate, population, starting_time, starting_frequency)
engine.simulate(model, contig, extended_events=sweep_events, ...)
```
which is great, but I think the lower-level stuff should also be exposed so that people can do more complicated things with conditioning?

With regards to the naming convention, are `single_site_dfe_id` and `contig.add_single_site_dfe` too garbled? Given that the string id is stored within the DFE, not the mutation type it contains. Also, what about moving `coordinate` from `DrawMutation` into the DFE creation (it could be stored as a single interval with zero span)? So something like,
```
contig.add_single_site_dfe(id="sweep", coordinate=100, fitness_coeff=0.1)
sweep_events = [
  DrawMutation(single_site_dfe_id=id, population_id=0, time=T_mut),
  ...
]
```

stdpopsim/slim_engine.py
petrelharp: We don't need to worry about this (or even backwards compatibility), as this hasn't been part of the released API.

petrelharp: And, good thought, I really wouldn't worry about pickled old-style extended events. Anyone who's doing that is on their own here (and can deal with it).

petrelharp:
> I think the lower-level stuff should also be exposed so that people can do more complicated things with conditioning?

Ok, that sounds good - I agree.

> With regards to the naming convention, are `single_site_dfe_id` and `contig.add_single_site_dfe` too garbled? Given that the string id is stored within the DFE, not the mutation type it contains. 

That seems OK, and not really worse than `mutation_id`? Let's see - or, how about just `single_site_id`? Or, `dfe_id`?

> Also, what about moving `coordinate` from `DrawMutation` into the DFE creation (it could be stored as a single interval with zero span)? So something like,

Hm, good thought. Let's see, I'm wondering if there's a work flow that wouldn't want this getting in the way. Under the hood, here's where the [conditioning on allele frequency happens](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/slim_engine.py#L322) and it uses [this code](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/slim_engine.py#L173), which does indeed identify the allele in question as being the only one of the given mutation type. So, it does totally make sense to restrict it to a given location.

Thinking about this more: we're baking this restriction based on how the underlying SLiM implementation works into the API, which is something we really shouldn't do. The user should not have to know that SLiM requires a separate mutation type, and hence a separate DFE. However, it is reasonable that the  user say something like "I'm declaring there will be a mutation at this position that I'm going to explicitly manipulate", which is basically what `add_single_site_mutation_type` is doing. But maybe it should just be called `contig.add_single_site`?

It'd help to think all this through on the whiteboard for a minute, next time you're in?

nspope:
We talked about moving `coordinate` to an argument of `Contig.add_single_site` and storing an associated interval that is `[coordinate, coordinate+1]` for the associated dummy DFE. This would then be used by `slim_makescript` to give positions to extended events like `DrawMutation` that reference the single site mutation. 

However, if a DFE is subsequently added that overlaps the coordinate of the "single site DFE", then the latter ends up with an empty interval in `contig.interval_list` (because of the interval subtraction that occurs when a new DFE is added). We could just raise an informative error in `slim_makescript` if this happens, something like "The single site with label {...} has no associated coordinate; it may have been deleted by the addition of an overlapping DFE". 

The only other ways to deal with this that I can think of, are to modify behavior of `Contig.add_dfe` and/or `utils.mask_intervals` to give "priority" to single sites (seems hacky) or make a new class for single site mutations (seems like bloat).

nspope:
Hey @andrewkern and @petrelharp, I think this is good to go if you'd like to take another look.

docs/tutorial.rst
```diff
@@ -1224,65 +1224,54 @@ of at least 0.8 by the end. Next, we'll walk through the steps required to do th
     this conditioning works by re-running the simulation until the condition is
     achieved, a nearly impossible condition will result in very long run times.
 
-First, we need to define the mutation type for the selected mutation.
-So we can refer to it later, we need its "mutation type ID". This is just
-the index of the new mutation type in the contig's list of mutation types.
+First, we need to add a single site mutation in the middle of the contig.
+This represents a mutation of a single base pair, with associated fitness and
+dominance coefficients.  The single site mutation must have a unique string ID
+so that we can refer to it later.
```
petrelharp, line 1230: This doesn't actually add the mutation, it just adds, like, the 'potential mutation' or something. How about
```suggestion
To set things up, we need to first add the site at which the selected mutation
will occur. This is like adding a DFE, except to a single site - we're saying that
there's a potential mutation at the given site that has these fitness consequences.
So that we can refer to it later, we give it unique string ID. Here, we'll add the site
in the middle of the contig, with ID "hard sweep".
```

docs/tutorial.rst
```diff
@@ -1224,65 +1224,54 @@ of at least 0.8 by the end. Next, we'll walk through the steps required to do th
...
-        if mt_info["dfe_id"] == dfe.id:
-            break
 
-    mut_id = mt_info["id"]
+.. note::
+
+    Note that single site mutations are internally stored as DFEs, and DFEs
+    cannot overlap. As a consequence, if a DFE is added to the contig over an
+    interval that contains an existing single site mutation, the latter will be
+    "overwritten" and an error will be raised by the simulation engine.
```
petrelharp, line 1248: Are you sure it will raise an error? I don't think that it should - it should just remove the effects of any previous DFEs at that site (and, so maybe it's worth noting you can't put more than one of these single site things at the same position).

docs/tutorial.rst
```diff
@@ -1224,65 +1224,54 @@ of at least 0.8 by the end. Next, we'll walk through the steps required to do th
...
-    coordinate = round(contig.recombination_map.sequence_length / 2)
     T_mut = 1000
     extended_events = [
         stdpopsim.ext.DrawMutation(
             time=T_mut,
-            mutation_type_id=mut_id,
+            single_site_id=mut_id,
             population_id=0,
-            coordinate=coordinate,
             save=True,
```
petrelharp, line 1264: Shouldn't `save` always be `True`?  When would we not want it to be?

stdpopsim/genomes.py
```diff
@@ -386,6 +386,86 @@ def add_dfe(self, intervals, DFE):
...
+        :param int coordinate: The coordinate of the site on the contig.
+        :param float selection_coeff: The selection coefficient of the mutation,
+            see :class:`MutationType` for details.
+        :param float dominance_coeff: The dominance coefficient of the mutation,
+            see :class:`MutationType` for details.
+        :param str description: A short description of the single site mutation
+            model as it would be used in written text, e.g., "Strong selective
+            sweep".
+        :param str long_description: A concise, but detailed, summary of the
+            single site mutation model.
```
petrelharp, line 444:
```suggestion
        :param str long_description: If necessary, a more detailed summary.
```

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,58 @@ def fix_time(event):
...
-    for x in _dfe_to_mtypes(contig).values():
+    slim_mutation_dfe = []
+    for i, x in _dfe_to_mtypes(contig).items():
         slim_mutation_ids.extend([u[0] for u in x])
+        slim_mutation_dfe.extend([i for u in x])
     for ee in extended_events:
-        if hasattr(ee, "mutation_type_id"):
-            mt_id = getattr(ee, "mutation_type_id")
+        if hasattr(ee, "single_site_id"):
+            mut_id = getattr(ee, "single_site_id")
```
petrelharp, line 957: Can we not call this `mut_id`, since it's not a mutation ID (it's a DFE ID); maybe `single_site_id`? It's so hard for me to keep this straight already.

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,58 @@ def fix_time(event):
...
             cls_name = ee.__class__.__name__
-            if mt_id not in slim_mutation_ids:
```
petrelharp, line 957: Wouldn't it be cleaner to check if there's a unique DFE with this ID, and then to check if that DFE has a single mutation type?

docs/tutorial.rst
petrelharp: Oh, I see what error you're talking about, sorry. I read this the wrong way around. Maybe just replace "the latter" with "the single site mutation", to make it harder to get mixed up. (I had read it as saying you couldn't put a single site over the top of an existing DFE.)

petrelharp, line 1248:
```suggestion
    Note that single site mutations are internally stored as DFEs, and more than one DFE
    cannot apply to the same segment of genome. As a consequence, if another DFE is added to the contig over an
    interval that already contains an existing single site mutation, the single site mutation will be
    "overwritten" and an error will be raised in simulation.
```

petrelharp:
I've got some minor comments - in general, looks good.

stdpopsim/genomes.py
```diff
@@ -386,6 +386,86 @@ def add_dfe(self, intervals, DFE):
...
+        Adds a model of a single site mutation at the provided coordinate.
+        The string label of the single site may be referenced by extended
+        events passed to the simulation engine.
+
+        For instance, to simulate a selective sweep,
+
+        .. code-block:: python
+
+            contig.add_single_site(
+                id="sweep_variant",
```
andrewkern, line 408:
```suggestion
                id="hard_sweep",
```

maybe change this name to match above?

stdpopsim/genomes.py
```diff
@@ -386,6 +386,86 @@ def add_dfe(self, intervals, DFE):
...
+                selection_coeff=0.1,
+                coordinate=mutation_coordinate,
+            )
+            extended_events = [
+                stdpopsim.ext.DrawMutation(
+                    time=mutation_time,
+                    single_site_id="sweep_variant",
```
andrewkern, line 415:
```suggestion
                    single_site_id="hard_sweep",
```

stdpopsim/genomes.py
```diff
@@ -386,6 +386,86 @@ def add_dfe(self, intervals, DFE):
...
+                    population_id=0,
+                    save=True,
+                ),
+                stdpopsim.ext.ConditionOnAlleleFrequency(
+                    start_time=stdpopsim.ext.GenerationAfter(mutation_time),
+                    end_time=0,
+                    single_site_id="sweep_variant",
```
andrewkern, line 422:
```suggestion
                    single_site_id="hard_sweep",
```

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,58 @@ def fix_time(event):
...
+            mtype_idx = mtype_idx[0]
+            setattr(ee, "mutation_type_id", slim_mutation_ids[mtype_idx])
+            dfe_idx = slim_mutation_dfe[mtype_idx]
+            dfe_intervals = contig.interval_list[dfe_idx]
+            if dfe_intervals.shape[0] == 0:
+                raise ValueError(
+                    f"The single site id '{mut_id}' referenced by {cls_name} "
+                    f"has no coordinate: it may have been removed by the addition "
+                    f"of an overlapping DFE or a single site with an "
+                    f"identical coordinate."
```
andrewkern, line 981:
```suggestion
                    f"has no coordinate: it may have been removed by the addition "
                    f"of an overlapping DFE or a single site with an "
                    f"identical coordinate after instatiation of '{cls_name}'."
```

andrewkern (approved):
looks good to me @nspope. I've left some minor suggestions.

docs/tutorial.rst
nspope: Then I'll go ahead and remove this option; on the slim side of things, too.

nspope: We should always save state for allele frequency conditioning, right? Currently there is a `save` option in `ext.ConditionOnAlleleFrequency`, that I can also remove

stdpopsim/slim_engine.py
nspope: Hm, not sure I understand-- instantiation of `DrawMutation` won't check that the single site exists/has a coordinate -- this is done in `slim_makescript`. So one could add a single site, add a DFE overlapping the single site, then setup the extended events; and the simulation would still raise an error.

docs/tutorial.rst
petrelharp: Well, there's a disk space/speed trade-off here: I think the way it works is that when it does the conditioning it jumps back to the most recently saved time; if it's saved more often that will be more efficient. However, it seems like unnecessary complexity for the API here - just saving all the time should be fine, unless someone is doing some overly complicated conditioning. I guess it's worth checking, though - since conditioning can act over and *interval* of time (right?), at what point in that interval does the save happen?

stdpopsim/slim_engine.py
petrelharp: IIUC, Andy is just trying to make the error message more clear. Currently, a user might think from the error message that first adding the overlapping DFE and then adding the single site would produce this error, so he's clarifying that it's only a problem if the overlapping DFE is added to the contig *after* the single site.

docs/tutorial.rst
nspope: The save occurs just once, at the beginning of the interval. There is a speed cost associated with saving because of writing out the tree sequence, but I don't think there should be extra use of disk space -- because the slim function `save` is always writing to the same file (the global `trees_file`) ... unless I'm missing something? With complicated conditioning, it seems to me that the rejection sampling would have difficultly completing *without* saving frequently. But, I'm happy to punt this to another issue if there's any question about removing the `save` toggle from `ext.ConditionOnAlleleFrequency`.

stdpopsim/slim_engine.py
andrewkern: yes indeed @petrelharp -- thanks for clarifying my cryptic stuff

docs/tutorial.rst
petrelharp: Ok, we've looked at the SLiM code and we think that there's not a big enough benefit to providing the flexibility to optionally save or not save at these conditioning events to warrant the more complicated interface. Does this seem OK to you, @grahamgower? (tldr; we want to remove the `save=` option from the API for these events, since it seems like unnecessary complexity)

stdpopsim/slim_engine.py
```diff
@@ -948,17 +948,58 @@ def fix_time(event):
...
+            assert len(contig.dfe_list[dfe_idx].mutation_types) == 1
+            mt = contig.dfe_list[dfe_idx].mutation_types[0]
+            if not mt.distribution_type == "f":
+                raise ValueError(
+                    f"The single site id '{mut_id}' referenced by "
+                    f"{cls_name} has a mutation type with fitness "
+                    f"distribution '{mt.distribution_type}', instead "
+                    f"of a fixed fitness coefficient."
+                )
+
```
petrelharp, line 1002: we think that `coordinate` and `mutation_type_id` can be local variables in this loop over extended events, rather than an attribute

docs/tutorial.rst
grahamgower: > With complicated conditioning, it seems to me that the rejection sampling would have difficultly completing _without_ saving frequently.

Yes, this was the main reason for having the `save=` option. The initial selection implementation was strongly motivated by the desire to model adaptive introgression as in https://github.com/popsim-consortium/stdpopsim/blob/main/docs/selection_example.py (and be able to also do this on top of a DFE).

I'm not sure if there's a simpler way to achieve this?

grahamgower: I guess that example should still work with `save=True` in all conditioning events. So probably its ok to remove the option and always do that. I almost certainly chose to store prior to the introduction of the selected mutation because I was following a SLiM recipe (and had about 5 mins experience with SLiM at the time), so it would be a good idea to also check the standard SLiM recipe(s) to see if there's any reasoning that we're missing here.

nspope: That makes perfect sense to me @grahamgower -- what we were thinking was to always save the state when a mutation is drawn or at the beginning of a frequency conditioning interval; and drop the `save` toggle from the API. Does this seem OK, or can you think of a case when `save=False` would make sense?

nspope: Sounds good!

grahamgower: Thinking about this a bit more... if one conditions on the allele not being lost and saves the simulation state, the allele could be at very low frequency; if one waits `T` generations then conditions on the allele at high frequency, then the chance of the allele having rised in frequency when `T` is small is negligible and so the simulation enters into an endless save/restore loop. Clearly one must be careful about AF conditioning. But, having the option for `save=False` allows one to change the AF conditioning, without having to worry about this problem. This might be desirable when `T` is not known in advance, but is drawn from a distribution (as I was doing to generate training data for a CNN).

nspope: Currently, conditioning events have fixed start/end times and require the associated `DrawMutation` event to have `save=True` -- so simulations would still end up in an endless restore cycle if the AF condition is unrealistic. To avoid this in the way you describe, the random AF conditioning start time would have to be drawn on the SLiM side of things, if I'm understanding correctly? Alternatively, there could be an option forcing SLiM to error out if a simulation fails an AF condition, so that a rejection sampler w/ random times could be implemented via a `try` block on the python side. Either way, sounds like enough infrastructure to warrant a separate PR.

grahamgower: The complete working code is here: https://github.com/grahamgower/genomatnn/blob/main/genomatnn/sim.py#L645

I've needed to have `genomatnn` depend on a stdpopsim branch/fork which is pinned at a post-stdpopsim-0.1.2 version that worked when I wrote this, as I was uncertain what the final stdpopsim API would look like when we finally push out the next stdpopsim release. If the `save=` option is removed I would have a hard time updating the genomatnn code to use the next stdpopsim release. Although this wouldn't be the end of the world, as updating that code isn't a high priority for me anyhow.

nspope: Ah, I think I understand, but let me spell it out to be sure. Say you're simulating a frequency trajectory $f$ from generation $t$ until the present, conditional on lower bounds $x_a, x_b$ at times $b > a > t$. So, drawing a sample from $$\mathrm{Pr}(f_{t ... 0} \mid f_{a...b-1} > x_a, f_{b...0} > x_b) = \mathrm{Pr}(f_{t ... a-1}) \times \mathrm{Pr}(f_{a ... b-1} \mid f_{a...b-1} > x_a, f_{a-1}) \times \mathrm{Pr}(f_{b ... 0} \mid f_{b...0} > x_b, f_{b-1})$$ If `save=True` universally, then we're rejection sampling sequentially from each of the terms on the right hand side. The problem is that the rejection rate might effectively be 1 when sampling from $\mathrm{Pr}(f_{b ... 0} \mid f_{b...0} > x_b, f_{b-1})$ if $f_{b-1} \ll x_b$. So instead of sampling from $\mathrm{Pr}(f_{a ... b-1} \mid f_{a...b-1} > x_a, f_{a-1})$ followed by $\mathrm{Pr}(f_{b ... 0} \mid f_{b ... 0} > x_b, f_{b-1})$, it might be much more efficient to just rejection sample once from the joint distribution $\mathrm{Pr}(f_{a ... b-1}, f_{b ... 0} \mid f_{a ... b-1} > x_a, f_{b ... 0} > x_b, f_{a-1})$ -- in fact this might be the only way to get the sampler to reliably finish in a reasonable amount of time.

This does seem useful -- especially if we're ultimately going to de-emphasize this lower-level API in favor of convenience functions for specific models (so that more complexity in the lower-level interface isn't so much an issue). So, how about we keep `save` for `ConditionOnAlleleFrequency` and get rid of `save` for `DrawMutation`. What do you think, @petrelharp and @andrewkern?

andrewkern: Yeah this is a good argument in favor of keeping `save` with `ConditionOnAlleleFrequency` -- agreed. 

Sounds good to me @nspope

petrelharp: Um, hold on here - maybe this is what you're saying already, but: even though $f$ is Markov,
$$
  \mathbb{P}(f_t > x | f_{s_i} > x_i \text{ for } 1 \le i \le k)
  \neq
  \mathbb{P}(f_{s_1} > x_1) \mathbb{P}(f_{s_2} > x_2 | f_{s_1} > x_1} \cdots \mathbb{P}(f_t > x | f_{s_k} > x_k} ,
$$
and so besides any issue of efficiency of the rejection sampling, there's actually an issue of *correctness* of the conditioning. To do the conditioning correctly, we've got to restart all the way back to the first event.

So, I think that we need `save=True` for when the mutation is introduced, but we should *not* have it for any subsequent events (unfortunately)? I'd want to think through this more carefully to be sure, but I currently don't see how doing `save=True` at any subsequent events will actually give you a correctly conditioned simulation.

grahamgower: Maybe I'm not following, but surely this is ok if we first condition on AF>0, then condition on (say) AF>0.2. There's no other way to obtain AF>0.2 except if AF>0 in every proceeding generation.

nspope: I might be being dense here, @petrelharp -- but if $f$ is Markov, then 
$$\mathrm{Pr}(f_{i}\~\mathrm{for}\~i\~\mathrm{in}\~1 \leq i \leq k) = \mathrm{Pr}(f_{1}) \times \mathrm{Pr}(f_{2} \mid f_{1}) \times \cdots \times \mathrm{Pr}(f_{k} \mid f_{k-1})$$ And,
$$\mathrm{Pr}(f_{1 \leq i \leq k} \mid f_{i} > x_i\~\mathrm{for}\~i\~\mathrm{in}\~1 \leq i \leq k) = \mathrm{Pr}(f_{1}) \mathbb{I}[f_{1} > x_1] \times \mathrm{Pr}(f_{2} \mid f_{1}) \mathbb{I}[f_{2} > x_2] \times \cdots \times \mathrm{Pr}(f_{k} \mid f_{k-1}) \mathbb{I}[f_{k} > x_k] $$ So why wouldn't sequentially rejection sampling each of the $\mathrm{Pr}(\cdot \mid \cdot) \mathbb{I}[\cdot]$ terms -- that is, conditioning on $f_i$ for $i < j$ to rejection sample $f_j$ -- produce a draw from the right distribution?

petrelharp: Ok, I was totally wrong, you all are right. This is a general fact about stochastic processes, and here's the basic calculation:
calcs


In "words", if X is a random variable and Y is X|A, then X|(A&B) and Y|B have the same distribution. So, let X be the random frequency trajectory, let A be the event that the frequency at time t1 is in some interval, and B be the event that the frequency at some later time t2 is in some other interval -- conditioning first on the A then on B is the same as conditioning on both at the same time.

nspope: In conclusion, `save=True` is correct in multiple `ConditionOnAlleleFrequency`, but may not always be desirable for efficient rejection sampling. However, `save=True` should always be the case for `DrawMutation`. So, we'll remove `save` from `DrawMutation` and retain it for `ConditionOnAlleleFrequency`. For the future, higher-level interface for sweeps in `stdpopsim`, the models will be simple enough that I don't think we'll have to expose `save`.

grahamgower: Thanks for spending the time to think this all through @petrelharp and @nspope!

The only reason I can think of to want to set `save=False` in `DrawMutation()` is if we one day support multiple selected mutations in draw mutation. E.g. a drawn mutation in each of two different populations, potentially drawn at different times. A use case like this ought not take much work to support (probably just fix the AF calculation in the SLiM code). Is removing `save` from DrawMutation a big win in terms of code simplification?

nspope: I'm ambivalent @grahamgower -- I think the eventual goal is to have a higher level interface for simple sweep scenarios that is cleaner, but that produces lists of these extended event classes under the hood. If you see an advantage to keeping `save` arguments in all the events (in terms of flexibility and future extensions and so on), then that's good enough for me! We might want to document some of these nuances, though, in case this comes up again.

petrelharp: Um, after some more consideration, I'm worried about the entire rejection sampling set-up? I'm probably missing something silly, but: consider the case where you introduce a mutation at time 0 and condition on it being not lost at both 100 and 200 generations later. As Andy pointed out yesterday, trajectories that are not lost at both times will tend to be higher at time 100 then trajectories that are not lost at just time 100. In other words: let f(t) be the frequency at time t, then
$$\mathbb{P}(f(100) > x | f(100) > 0 \text{ and } f(200) > 0) > \mathbb{P}(f(100) > x | f(100) > 0)$$
However, if we rejection sample back to time 100, then the distribution of frequencies at time 100 will be given by the expression on the right, when we want the expression on the left.

The problem with the argument on the whiteboard was that the rejection sampling isn't actually sampling from that conditional distribution.

And, here's some R code that demonstrates that this intermediate rejection sampling doesn't get the right thing. Consider $Z = X + Y$; say we want to sample from $Z$ given $Z > 1$. This is equivalent to conditioning on $Z$ given $Z > 1$ and $X > -N$ for large enough $N$, which amounts to no conditioning on $N$. So, if our rejection sampling approach worked, then we should be able to (a) first sample $X$, and then (b) repeatedly sample $Y$ until we get $X + Y > 1$. However, that's clearly (in some cases) not the same as repeatedly sampling $X+Y$ until we get $X+Y > 1$.
```
f <- function (n, a, b, x) {
    out <- c()
    while (length(out) < n) {
        z <- a(n) + b(n)
        out <- c(out, z[z > x])
    }
    return(out[1:n])
}

g <- function (n, a, b, x) {
    out <- u <- a(n)
    free <- rep(TRUE, n)
    while (any(free)) {
        u <- out + b(n)
        ut <- free & (u > x)
        out[ut] <- u[ut]
        free[ut] <- FALSE
    }
    return(out)
}

a <- rnorm
b <- rnorm
x <- 1
n <- 1000
fn <- f(n, a, b, x)
gn <- g(n, a, b, x)

layout(1:2)
hist(fn, breaks=40, main=expression(paste("x + y | ", x + y > 1)))
hist(gn, breaks=40, main=expression(paste("x + (y | ", x + y > 1, ")")))
```
which gets
Screenshot from 2022-08-09 08-44-02

grahamgower: > $\mathbb{P}(f(100) > x | f(100) > 0 \text{ and } f(200) > 0) > \mathbb{P}(f(100) > x | f(100) > 0)$
> However, if we rejection sample back to time 100, then the distribution of frequencies at time 100 will be given by the expression on the right, when we want the expression on the left.

I'm confused by the setup here. If you take only simulations that finish with $f(200) > 0$, then surely you have the expression on the left? My assumption when rejection sampling is that $\mathbb{P}(f(200) > x | f(100) > 0) = \mathbb{P}(f(200) > 0)$---is there something I'm missing here?

nspope: I'm confused as well. 

1. **Single rejection step**: Sample a trajectory $f(t)$ from time 0 until time 200, accept the trajectory if $f(100) > 0 \land f(200) > 0$, and store the value of $u = f(100)$. This will be a random sample from $\mathrm{Pr}(f(100) \mid f(100) > 0 \land f(200) > 0)$.
2. **Sequential rejection steps**: 
    - Sample a trajectory $f(t)$ from time 0 until time 100, accept the (incomplete) trajectory if $f(100) > 0$, and store $u = f(100)$. This will be a sample from $\mathrm{Pr}(f(100) \mid f(100) > 0)$. 
    - Then simulate a conditioned trajectory $\tilde{f}(t)$ from time 100 to time 200, conditional on $\tilde{f}(100) = u$. Accept  if $\tilde{f}(200) > 0$, and store $v = \tilde{f}(100)$. This will be a sample from $\mathrm{Pr}(f(100) \mid f(100) > 0 \land f(200) > 0)$. 
    - So the first rejection step generates RVs from a distribution of trajectories with a single condition. The second rejection step filters these original RVs to get new RVs that are from a distribution of trajectories with two conditions.

To try to give a simulated example that more closely reflects this process: let $X \sim \mathrm{Uniform}(-1, 1)$. Let $Y = 0$ with probability 0.5 and $Y \sim \mathrm{Uniform}(0, 1)$ with probability 0.5. Let $Z = X Y$. We want samples from $\mathrm{Pr}(X \mid X > 0 \land Z > 0)$ which is just a uniform distribution over [0, 1]. 

Here is R code using a single rejection step (simulate $X, Y$ and accept if $X > 0 \land XY > 0$, versus sequential rejection steps (e.g. first simulate $X$ and accept if $X > 0$, then simulate $Y$ until $XY > 0$).
```r
f <- function(n, draw_X, draw_Y, lb_X, lb_Z) {
  out <- c()
  while (length(out) < n) {
    X <- draw_X()
    Y <- draw_Y()
    Z <- X * Y
    if (X > lb_X & Z > lb_Z) {
      out <- c(out, Y)
    }
  }
  return(out)
}

g <- function(n, draw_X, draw_Y, lb_X, lb_Z) {
  out <- c()
  while (length(out) < n) {
    X <- draw_X()
    if (X > lb_X) {
      while (TRUE) {
        Y <- draw_Y()
        Z <- X * Y
        if (Z > lb_Z) {
          out <- c(out, Y)
          break
        }
      }
    }
  }
  return(out)
}

lb_X <- lb_Z <- 0
n <- 10000
proposal_X <- function() runif(1, -1, 1)
proposal_Y <- function() if(runif(1, 0, 1) < 0.5) 0 else runif(1, 0, 1)
fn <- f(n, proposal_X, proposal_Y, lb_X, lb_Z)
gn <- g(n, proposal_X, proposal_Y, lb_X, lb_Z)
ecdf <- function(x) data.frame(z=sort(x), p=1:length(x)/length(x))

library(ggplot2)
df <- cbind(rbind(ecdf(fn),ecdf(gn)), 
    Algorithm=rep(c("Reject X&Z simultaneously","Reject X&Z sequentially"),each=n)
)
ggplot(df) + 
  geom_step(aes(x=z, y=p, color=Algorithm)) + 
  xlab("Y") + ylab("ECDF") + theme(legend.position=c(0.30,0.8))
```


This seems fine ... what am I missing?

petrelharp: > If you take only simulations that finish with $f(200) > 0$, then surely you have the expression on the left? 

Well, the problem is that the procedure of (simulate back to t=100 and then rejection sample back to t=100 until you are above 0 at 200) is *not* the same thing as simulating $f$ in the first place. You do indeed end up with only trajectories that stay above zero, but not having the right distribution.

> My assumption when rejection sampling is that $\mathbb{P}(f(200) > x | f(100) > 0) = \mathbb{P}(f(200) > x)$ ---is there something I'm missing here?

I think you mean that $\mathbb{P}(f(200) > x \text{ and } f(100) > 0) = \mathbb{P}(f(200) > x)$? That I do agree with; however, it's not true with the conditional.

And, one reason this is so tricky to reason about is that rejection sampling back to an intermediate time makes it not at all obvious what process we're actually simulating. There's two processes:
1. $f$, conditioned on $f(200) > 0$
2. $f$ run until time 100 and conditioned on $f(100) > 0$; and then after that time point conditioned on $f(200) > 0$.

The reason the second process is different is because the process "doesn't know" about the conditioning at $t=200$ for the first 100 time steps.

petrelharp: Also note that you could make this correct by using importance weights - if in a particular simulation you have to rejction sample $n$ times until you get a good one, and give that simulation weight $1/n$, then you *do* get the right distribution.

petrelharp: Also also, I am **very** sorry to have not noticed this issue much earlier.

nspope: After talking it out, I think Peter is totally correct -- my counterexample above really doesn't address the problem. Here's my understanding of the issue:

Let $f_a$ be the allele frequency at time $a$, which occurs with probability $\mathrm{Pr}(f_a)$ when we run the simulator from time 0 to $a$. If we subsequently run the simulator from $a$ until a future time $b$ where we check if $f_b > 0$, then the entire trajectory will get accepted with probability $\mathrm{Pr}(f_b > 0 \mid f_a)$ -- so ultimately, our samples of $f_a$ will have the distribution $$\mathrm{Pr}(f_a \mid f_b > 0) = \frac{\mathrm{Pr}(f_a) \mathrm{Pr}(f_b > 0 \mid f_a)}{\mathrm{Pr}(f_b > 0)}$$ where the denominator is just a normalizing constant to make it integrate/sum to 1.

This will be the case when every candidate trajectory is simulated from the beginning. If we're checkpointing, then if some candidate trajectory doesn't satisfy $f_b > 0$, we'd rerun the simulator from the same value of $f_a$ until we finally get a simulation where $f_b > 0$. Thus, the checkpointed $f_a$ gets accepted with probability 1, because we're forcing the rejection sampler to return a trajectory that includes it! In this case, even though the sampled trajectories satisfy $f_b > 0$, they will produce $f_a$ that are distributed according to $\mathrm{Pr}(f_a)$ instead of the conditional distribution $\mathrm{Pr}(f_a \mid f_b > 0)$ (which is what would be produced by the "un-checkpointed" simulator). So the allele frequency trajectories (and thus genealogies) produced by checkpointing won't come from the correct stochastic process.

Does this make sense @grahamgower? I could do some simple simulations to verify that this is indeed the case, if that would help.

andrewkern: here's a simple simulation of a random walk on a lattice that demonstrates the issue with multiple conditioning.

in this case i'm conditioning on the process being at some state `x > 0` at `t1` and/or `x > 0.1` at `t2`. Below I plot the random walk and then a boxplot of the final state of the random walk for an unconditioned process, one which we condition on the state at either of the times, or finally when we condition on the state at both of the times.  

```python
import numpy as np
from matplotlib import pyplot as plt

reps = 500
gens = 1000
t1 = gens / 8
t2 = gens / 2
x = np.random.random((gens,reps)) > 0.5
x = x.astype(np.float16) - 0.5
xc = x.cumsum(axis=0)
cond1 = xc[int(t1),:] > 0
cond2 = xc[int(t2), :] > 0.1
cond_joint = np.logical_and(cond1, cond2)
fig, axs = plt.subplots(2,1)
p = axs[0].plot(xc)
labels = ['unconditional', 'given x > 0 at T1', 'given x > 0.1 at T2', 'given x > 0 at T1 and x > 0.1 at T2']
plt.boxplot([xc[-1, :], xc[-1, cond1], xc[-1, cond2], xc[-1, cond_joint]], vert=False, labels=labels)
plt.show()
```
output looks like 
image

so multiple conditioning isn't equivalent to single conditioning, at least in this really simple setup

grahamgower: Ok, I get it now. Thanks everyone! I think my confusion stemmed from thinking about the allele frequency at the end of the rejection sampling process, rather than thinking about the *state of the simulation*.

So, this is a problem for all save/restore rejection sampling in SLiM then, right? If so, then this affects a lot of people and a lot of studies!

grahamgower: It would be nice if we could convert this whole conversation to an issue, or a discussion. Anyone know if this is possible?

nspope: @grahamgower It may be OK to checkpoint the state when a mutation is drawn, because in this case there's only one possible allele frequency (1/N) so nothing really would be influenced by the future conditioning. But, if multiple mutations are drawn on the same contig at different times, and are all checkpointed, then I think that would lead to the same issues as discussed above.

I'll go ahead and summarise the setup & explanations in this thread to post a new issue.

petrelharp: I agree - to be extra clear, the problem occurs if there is a checkpointed state that is after the mutation has been introduced that might be returned to from a later point - in other words, if there's an event that has `save=True`, other than the initial mutation's introduction, followed by a rejection-sampling-based conditioning later in time.

And the reason that it *is* ok to do `save` at the mutation's introduction is that *that* is exactly what we want to condition on - the trajectory from that point forward. The way it works is that if we have `save` points at times `t0, t1, ..., tn`, then the trajectory is conditioned *separately* on each interval $[t_i, t_{i+1})$.

nspope: Alrighty I opened an issue at https://github.com/popsim-consortium/stdpopsim/issues/1322. Maybe we should fix this in a separate PR, to keep the "single site string ID" as a separate addition ... I can see us going back-and-forth on some of the implementation details wrt to allowing checkpointing for `DrawMutation` events.

nspope:
I think this is good @andrewkern and @petrelharp. When this one is merged I'll make a separate PR removing `save`; the issue tracking that bug is #1322 .

petrelharp:
Looks good, thanks!!

petrelharp:
Ah, do you mind squashing this down to one commit? Then I'll merge it in.

andrewkern:
LGTM too @nspope

petrelharp:
nm, I can just hit "squash and merge"

Issue #1308, opened 2022-07-24
log-normal DFE doesn't allow negative values


petrelharp:
Over in #1292 I turned on the test that actually simualted from all of the defined DFEs. The [`DroMel.LognormalPlusPositive_R16`](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/catalog/DroMel/dfes.py#L53) model uses a lognormal distribution for deleterious mutations. However, the simulation run errored because the selection coefficients that it produced were positive, and sometimes very very large. The problem is that we're trying to use something like `rlnorm(meanlog=-3, sdlog=2)` to get negative numbers. That's in fact how many distributions work in eidos: passing a negative mean for a positive distribution gets you numbers drawn from that distribution multipied by -1. However, the arguments to `rlnorm( )` are the mean and SD on a log scale, i.e., `mean(log( rlnorm(10000, -2, 4) ))` will be close to -2. (and so `mean(rlnorm(10000, -2, 4))` will be close to `exp(-2)` which is `>0`)

So, we need to fix up our underlying log-normal implementation somehow to allow this possibility.

I've also commented out the DFE so it doesn't error (in that PR). I think there are other issues with it (for instance, in the code it calls the meanlog and sdlog parameters the "shape" and "scale", which they are not).

As for the fix: recall that for most distributions we can use SLiM's built-in distribution types, e.g. `"e"` for exponential; however, it doesn't have a built-in log-normal, so we provide that one [in eidos code](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/dfe.py#L111).  I propose *removing" the `"l"` distribution type (which is what we used for "log-normal", and replacing it with "lp" for "positive log-normal" and "ln" for "negative lognormal".

andrewkern:
> As for the fix: recall that for most distributions we can use SLiM's built-in distribution types, e.g. `"e"` for exponential; however, it doesn't have a built-in log-normal, so we provide that one [in eidos code](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/dfe.py#L111). I propose *removing" the `"l"` distribution type (which is what we used for "log-normal", and replacing it with "lp" for "positive log-normal" and "ln" for "negative lognormal".

seems like a perfectly fine fix to me. in practice, are people fitting models of this sort?

petrelharp:
Well, I ran into this because Ragsdale et al used a (negative) log-normal? So, yes?

[...]

PR #1309, opened 2022-07-24
Lognormal fix

petrelharp:
Here's the fix I proposed in #1308.

### DIFF
--- a/stdpopsim/catalog/DroMel/dfes.py
+++ b/stdpopsim/catalog/DroMel/dfes.py
@@ -89,6 +89,9 @@ def _RagsdaleDFE():
     # selection strengths
     S_positive = 39.9
     s_positive = S_positive / 2 / Ne
+    # TODO: these are *not* scale and shape parameters:
+    # the arguments to lognorm are meanlog and sdlog.
+    # Rename these after checking!
     scale_negative = 5.42 / 2 / Ne
     shape_negative = 3.36
     h = 0.5
@@ -96,8 +99,8 @@ def _RagsdaleDFE():
     synonymous = stdpopsim.MutationType()
     negative = stdpopsim.MutationType(
         dominance_coeff=h,
-        distribution_type="l",  # log-normal distribution
-        distribution_args=[-scale_negative, shape_negative],
+        distribution_type="ln",  # negative log-normal distribution
+        distribution_args=[scale_negative, shape_negative],
     )
     positive = stdpopsim.MutationType(
         dominance_coeff=h,
@@ -115,6 +118,4 @@ def _RagsdaleDFE():
     )
 
 
-# This DFE is NOT correct due to a bug/deficiency: the log-normal implementation
-# does not allow negative values.
-# _species.add_dfe(_RagsdaleDFE())
+_species.add_dfe(_RagsdaleDFE())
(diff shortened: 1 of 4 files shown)


stdpopsim/catalog/DroMel/dfes.py
```diff
@@ -89,15 +89,18 @@ def _RagsdaleDFE():
...
     scale_negative = 5.42 / 2 / Ne
     shape_negative = 3.36
     h = 0.5
 
     synonymous = stdpopsim.MutationType()
     negative = stdpopsim.MutationType(
         dominance_coeff=h,
-        distribution_type="l",  # log-normal distribution
-        distribution_args=[-scale_negative, shape_negative],
+        distribution_type="ln",  # log-normal distribution
```
andrewkern, line 102:
```suggestion
        distribution_type="ln",  # negative log-normal distribution
```

andrewkern: or maybe "log-normal negative..."

stdpopsim/dfe.py
```diff
@@ -23,7 +23,12 @@ class MutationType(object):
     - ``g``: gamma, two parameters (mean, shape)
     - ``n``: normal, two parameters (mean, SD)
     - ``w``: Weibull, two parameters (scale, shape)
-    - ``l``: logNormal, two parameters (mean and sd on log scale; see rlnorm)
+    - ``lp``: positive logNormal, two parameters (mean and sd on log scale; see rlnorm)
+    - ``ln``: negative logNormal, two parameters (mean and sd on log scale; see rlnorm)
+
+    Type "lp" is always positive, and type "ln" is the same as "lp", multiplied by -1.
```
andrewkern, line 29:
```suggestion
    Type "lp" is always positive, and type "ln" is always negative. Both are the same log-normal, just reflected about zero
```

andrewkern: trying to unpack this a bit

andrewkern (changes requested):
is this still missing the `lp` distribution definition? I don't seem to see it here?

petrelharp:
> is this still missing the `lp` distribution definition? I don't seem to see it here?

Not sure what you mean - it's implemented [here](https://github.com/popsim-consortium/stdpopsim/pull/1309/files#diff-cb24342fdfba499cf83efa2df3a903540da15727db67aafe2d87bf28af607bbdR125)?

stdpopsim/dfe.py
petrelharp: I've changed it; see what you think?

petrelharp:
Heh, found a bug that I wasn't actually doing `lp`; now testing that `lp` is >0 and `ln` is <0.

stdpopsim/dfe.py
andrewkern: nice. much better.

andrewkern (approved):
LGTM @petrelharp

andrewkern:
huh what's this failed test?

petrelharp:
> huh what's this failed test? thinking it's just stochastic from that one slimulation

Hmph, yes, probably but I want to fix that test - that's why I found the last bug. (and! it shouldn't be stochastic!)

petrelharp:
Ah, I hadn't set a seed on that one.

petrelharp:
Ok, this should be ready to merge.

Issue #1321, opened 2022-08-09
AraTha DFE


chriscrsmith:
**Short description of what you'd like added:**
The catalog would benefit from a DFE(s) for Arabidopsis.

**Rationale and considerations:**
In particular, we'd like to have this for the analysis2 pipeline and paper.

chriscrsmith:
Opened a QC issue for this: #1344

petrelharp:
Closed in #1324.

PR #1324, opened 2022-08-15
adding DFE for Arabidopsis

chriscrsmith:
Addressing #1321 

Using estimated gamma shape and mean from https://doi.org/10.1038/s41467-018-05281-7. This study estimates dominance in addition to the DFE, however I think we decided to roll with the additive-only model because it is easier to implement.

Note that I plugged in [0, 1.0] for the proportions of neutral and selected loci.

Also I put a minus sign on the gamma mean which seems to be convention.


petrelharp:
One thing (and the reason codecov is complaining): you need also to put
```
from . import dfes  # noqa somethingsomething
```
in `__init__.py`.

petrelharp:
And: this DFE is estimated without dominance?

stdpopsim/catalog/AraTha/dfes.py
```diff
@@ -0,0 +1,52 @@
...
+    From Huber et al. (2018), https://doi.org/10.1038/s41467-018-05281-7.
+    Gamma parameters are based on the additive-only model for A. thaliana described in
+    Supplementary Table 4.
+    """
+    citations = [
+        stdpopsim.Citation(
+            author="Huber et al.",
+            year=2018,
+            doi="https://doi.org/10.1038/s41467-018-05281-7",
+            reasons="to be defined",  # include the dfe_model reason
```
petrelharp, line 26:
```suggestion
            reasons=stdpopsim.CiteReason.DFE,  # include the dfe_model reason
```

petrelharp: hm, I see you copied this from `catalog/HomSap/dfes.py`; could you fix this there, also?

petrelharp:
> Note that I plugged in [0, 1.0] for the proportions of neutral and selected loci.

Is this right, or was it not clear from the paper?

petrelharp:
> Also I put a minus sign on the gamma mean which seems to be convention.

Yes, good - that is how you specify deleterious `s` in SLiM.

stdpopsim/catalog/AraTha/dfes.py
chriscrsmith: got it

chriscrsmith:
> > Note that I plugged in [0, 1.0] for the proportions of neutral and selected loci.
> 
> Is this right, or was it not clear from the paper?

I am pretty sure this is right. In the supp mat table with the gamma parameters I used, it does not list any proportion neutral. And the main text makes it seem to me that neutral mutations were not included in most DFEs estimates.

chriscrsmith:
> One thing (and the reason codecov is complaining): you need also to put

Doh. I must have gotten tired of waiting for the codecov to finish and I forgot to come back to this.

chriscrsmith:
What does the comment "noqa ..." mean/do?

grahamgower:
> noqa

This tells `flake8` to ignore the line. It can be used to ignore specific types of flake8 errors, e.g.
```
import foo  # noqa: F401
```
Means that flake8 won't produce the "imported but not used" error. The import may have desirable side effects, but flake8 doesn't know about this and just sees that the variable `foo` isn't used in the current scope.

stdpopsim/catalog/AraTha/__init__.py
```diff
@@ -4,4 +4,5 @@
 from . import species  # noqa: F401
 from . import genetic_maps  # noqa: F401
 from . import demographic_models  # noqa: F401
-from . import annotations  # noqa: F401
+from . import dfes  # noqa: F401
+from . import annotations  # noqa: F401
```
petrelharp, line 8:
```suggestion
from . import annotations  # noqa: F401

```

chriscrsmith:
> > noqa
> 
> This tells `flake8` to ignore the line. It can be used to ignore specific types of flake8 errors, e.g.
> 
> ```
> import foo  # noqa: F401
> ```
> 
> Means that flake8 won't produce the "imported but not used" error. The import may have desirable side effects, but flake8 doesn't know about this and just sees that the variable `foo` isn't used in the current scope.

Got it! Thanks

petrelharp:
Awesome, thanks! I"ll merge this; could you open a QC issue for it, @chriscrsmith?

chriscrsmith:
QC issue #1344

PR #1326, opened 2022-08-16
port to slim v4

petrelharp:
With these changes, tests now pass using the new version of SLiM on my computer at least. Mostly this just required changing "generation" to "tick"; there was only one thing that was hard to track down.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -7,7 +7,7 @@ described in the SLiM reference manual. In our generated SLiM script, all the
 demographic model parameters are defined in multi-dimensional arrays at the
 top of the script, in the `initialize()` block. These arrays define the event
 generations, and event blocks are subsequently constructed programmatically
-using `sim.registerLateEvent()`, rather than writing out the blocks verbatim.
+using `community.registerLateEvent()`, rather than writing out the blocks verbatim.
 This design is intended to permit modification of demographic parameters in
 the generated SLiM script, without needing to directly convert event times in
 the past into forwards-time generations.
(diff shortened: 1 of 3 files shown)


petrelharp:
AFAICT, codecov is just being confused here. The only lines it says are not being covered now are lines like "don't test this on windows".

andrewkern:
okay, well to review this properly i'm going to have to start by reading the new [time units](https://tskit.dev/pyslim/docs/latest/time_units.html) section!

stdpopsim/slim_engine.py
```diff
@@ -503,20 +503,22 @@
...
+                " ("+pop_names[pop]+") at tick "+g+", but only "+
                 N_g+" individuals will be alive.");
         }
 
-        sim.registerLateEvent(NULL,
-            "{dbg(self.source); " +
-            "inds=p"+pop+".sampleIndividuals("+n+"); " +
-            "sim.treeSeqRememberIndividuals(inds);}",
-            g, g);
+        if (n > 0) {
```
andrewkern, line 510: what happens if n <= 0?

andrewkern (approved):
looks like a simple enough conversion to v4, but there is definitely going to be breakage. 

should average generation time be recorded in the metadata (as in the pyslim examples)? currently everything is WF so perhaps not?

petrelharp:
> looks like a simple enough conversion to v4, but there is definitely going to be breakage.

Where?

> should average generation time be recorded in the metadata (as in the pyslim examples)? currently everything is WF so perhaps not?

No, everything is WF, so one tick = one generation.  (well, actually it equals `Q` generations...) It's something we'll have to deal with if/when we provide nonWF recipes, but that's a hurdle for the future.

stdpopsim/slim_engine.py
petrelharp: https://github.com/MesserLab/SLiM/issues/345 happens

andrewkern:
okay sounds good @petrelharp!

PR #1335, opened 2022-08-23
Remove SLiM checkpointing except for at first DrawMutation

### LINKED ISSUE #1328: Fix checkpointing behavior in SLiM rejection sampling
As described in #1325, checkpointed rejection sampling can easily produce incorrect simulations. The only situation where checkpointing is "safe" is when a mutation is drawn and there are no other extant mutations that are being conditioned on.

I think the following would fix the issue:

- Remove `save` option from python API (extended event python classes)
- Remove save mechanics from SLiM engine for `ConditionOnAlleleFrequency` events
- In `slim_makescript`, sort `DrawMutation` events by time and set `save` flag only at the first event

nspope:
Fixes #1328.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -296,20 +296,18 @@ _slim_main = """
         n_checkpoints = 0;
         for (i in 0:(ncol(drawn_mutations)-1)) {
             save = drawn_mutations[4,i] == 1;
-            if (!save) {
-                next;
-            }
+            if (save) {
+                // Saving the state at more than one timepoint can can cause
+                // incorrect conditioning in the rejection samples.
+                if (n_checkpoints > 0) {
+                    err("Attempt to save state at more than one checkpoint");
+                }
+                n_checkpoints = n_checkpoints + 1;
 
-            // Saving the state at more than one timepoint can can cause
-            // incorrect conditioning in the rejection samples.
-            if (n_checkpoints > 0) {
-                err("Attempt to save state at more than one checkpoint");
+                // Unconditionally save the state before the mutation is drawn.
+                g = G_start + gdiff(T_start, drawn_mutations[0,i]);
+                community.registerLateEvent(NULL, "{save();}", g, g);
             }
-            n_checkpoints = n_checkpoints + 1;
-
-            // Unconditionally save the state before the mutation is drawn.
-            g = G_start + gdiff(T_start, drawn_mutations[0,i]);
-            community.registerLateEvent(NULL, "{save();}", g, g);
         }
     }
     if (length(condition_on_allele_frequency) > 0) {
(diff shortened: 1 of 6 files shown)


nspope:
Hey @grahamgower, if you have a minute could you please take a look? As this would impact the training sims for your CNNs.

stdpopsim/slim_engine.py
```diff
@@ -293,13 +293,22 @@
...
             save = drawn_mutations[4,i] == 1;
             if (!save) {
                 next;
             }
-            g = G_start + gdiff(T_start, drawn_mutations[0,i]);
+
+            // Saving the state at more than one timepoint can can cause
+            // incorrect conditioning in the rejection samples.
+            if (n_checkpoints > 0) {
+                err("Attempt to save state at more than one checkpoint");
```
petrelharp, line 306: Doesn't this just error if `drawn_mutations` has more than one column?

stdpopsim/slim_engine.py
```diff
@@ -1052,7 +1046,14 @@ def fix_time(event):
     # Check that drawn mutations exist for extended events that need them.
     for single_site_id in referenced_single_site_ids:
         if single_site_id not in drawn_single_site_ids:
-            raise ValueError(f"No drawn mutation for single site id {single_site_id}.")
+            raise ValueError(
+                f"No drawn mutation exists for single site id {single_site_id}."
```
petrelharp, line 1050: Maybe I'm wrong here but shouldn't this be
```suggestion
                f"Single site id {single_site_id} is referenced but does not exist."
```

stdpopsim/slim_engine.py
petrelharp: oh I see, never mind!

stdpopsim/slim_engine.py
```diff
@@ -293,13 +293,22 @@
 
     // Save/restore events. These should come before all other events.
     if (length(drawn_mutations) > 0) {
+        n_checkpoints = 0;
         for (i in 0:(ncol(drawn_mutations)-1)) {
             save = drawn_mutations[4,i] == 1;
             if (!save) {
```
petrelharp, line 299: Gee, this part would be clearer if we did
```
if (save):
    ...
```

stdpopsim/slim_engine.py
nspope: No, because of the "next" a few lines before. Multiple drawn mutations are tested here:

https://github.com/popsim-consortium/stdpopsim/blob/a0da48882a7eb69a77778a620e78447ef775f3ce/tests/test_slim_engine.py#L1378

petrelharp (commented):
LGTM: some minor suggestions.

stdpopsim/slim_engine.py
nspope: Nope, this is checking that a single site with a `ChangeMutationFitness` or `ConditionOnAlleleFrequency` event also has a `DrawMutation`  event.

The check that the single site exists happens earlier, here:
https://github.com/popsim-consortium/stdpopsim/blob/a0da48882a7eb69a77778a620e78447ef775f3ce/stdpopsim/slim_engine.py#L960

petrelharp (approved):
Thanks!!

PR #1338, opened 2022-08-26
import PapAnu genetic map

petrelharp:
Looks like a simple whoopsie, forgot to import the genetic map! Odd that codecov didn't pick up on this?

PR #1340, opened 2022-08-26
Fix the error in PapAnu genetic map

saurabhbelsare:
This is in response to https://github.com/popsim-consortium/stdpopsim/pull/1338.
 Fixed the error in the genetic map -> Missing sha, incorrect file_pattern, missing import.

### DIFF
--- a/stdpopsim/catalog/PapAnu/__init__.py
+++ b/stdpopsim/catalog/PapAnu/__init__.py
@@ -3,3 +3,4 @@ Catalog definitions for PapAnu (Ensembl ID='papio_anubis')
 """
 from . import species  # noqa: F401
 from . import demographic_models  # noqa: F401
+from . import genetic_maps  # noqa: F401
--- a/stdpopsim/catalog/PapAnu/genetic_maps.py
+++ b/stdpopsim/catalog/PapAnu/genetic_maps.py
@@ -5,7 +5,7 @@ _species = stdpopsim.get_species("PapAnu")
 
 _gm = stdpopsim.GeneticMap(
     species=_species,
-    id="Pyrho_PAnubis1.0",
+    id="Pyrho_PAnubis1_0",
     description="Pyrho inferred genetic map for Papio Anubis",
     long_description="""
         These estimates were obtained from a sample of Papio Anubis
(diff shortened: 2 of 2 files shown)


saurabhbelsare:
I found what the problem here is. The genetic maps uploaded are inside a directory inside the tarball, and they shouldn't be. [Here](https://drive.google.com/file/d/10pwLq7LeP1FPPYuiZ53deWYUibczV56V/view?usp=sharing) is a new file which fixes this problem. Uploading this tarball to the same URL as the existing PAnubis genetic map tarball should fix the error. @andrewkern could you change this file whenever you get the chance? Sorry for the trouble.

petrelharp:
Thanks - we just need to re-run the tests when that gets uploaded; ping me to do that if you can't.

petrelharp:
ping @saurabhbelsare

saurabhbelsare:
Hi @andrewkern, have you updated the genetic map tarball with the new one I linked above? If so, I'll rerun the tests and hopefully they should all now pass.

andrewkern:
okay i've gone ahead and uploaded this tarball to AWS. i'm going to try to retrigger the tests and see if they pass.

andrewkern:
looks like the sha256 hash is now out of sync @saurabhbelsare. can you push the new ones?

saurabhbelsare:
There's something else off here. I'm fixing it all and will push again asap.

andrewkern:
okay please ping me when it's ready

saurabhbelsare:
Hi Andy, Sorry about this. I fixed the sha256 hash, but then it ran into another error because one of the files inside the tarball seems to have become corrupted somehow. I've fixed everything now (I think). I ran all the tests locally, and they have all completed without any error. [Here is the new (hopefully final) tarball](https://drive.google.com/file/d/15KzJ3HoqdjKhLmPsOKVH3Rvlb3zmEXO_/view?usp=sharing). I have pushed the latest genetic_maps.py with the updated sha256 has for this new tarball that I've just shared here. So the tests will fail at my first push, but should complete when you upload this new tarball and rerun the tests. Sorry for the errors.

andrewkern:
okay i've uploaded the new file. i will restart the tests next.

andrewkern:
@saurabhbelsare this is still failing tests. it looks like you've now changed the filename from 
`papio_anubis_genetic_map.tar.gz` to `papio_anubis_genetic_map.tar.gz`

saurabhbelsare:
Hi Andy, I think this error might be coming from the file not being at that URL, so the tests can't find it to run? Because when I just run 
`wget https://stdpopsim.s3-us-west-2.amazonaws.com/genetic_maps/PapAnu/papio_anubis_genetic_map.tar.gz` on the command line directly, it gives `Error 404: Not found.` Has the file uploaded properly?

andrewkern:
it's uploaded but you changed the name of the tar ball you sent me.

saurabhbelsare:
Oh I see it was missing the .tar suffix. Sorry. I changed it in PapAnu/genetic_maps.py, and now the ubuntu and windows tests seem to have completed. Not sure why the macos tests didn't run at all.

andrewkern:
okay looks like this is all set. thanks @saurabhbelsare, i'm going to squash and merge

Issue #656, opened 2020-10-27
rename stdpopsim.ext.ExtendedEvents to ...SLiMEvents


grahamgower:
Also rename `extended_events` variables to `slim_events`. This more properly reflects what the events are and how they're used. Suggested by @petrelharp.

jeromekelleher:
Agreed!

nspope:
@grahamgower and @petrelharp, do you still prefer this naming scheme? I think we should settle on one or the other before the upcoming release.

petrelharp:
Sounds good to me - but, I'm also voting to keep these out of the public API anyhow.

nspope:
Decided to keep these as "ExtendedEvents" in #1341. The rationale is that we may support sweeps down the road in the msprime engine, but would likely still be using these "Event" classes to define single site mutation models.

Issue #608, opened 2020-10-03
fix allele frequency calculation in slim code, or disallow multiple drawn mutations per MutationType


grahamgower:
The allele frequency calculation in the slim engine (used for allele frequency conditioning) currently assumes there is only one mutation per `MutationType`.
https://github.com/popsim-consortium/stdpopsim/blob/973d5b05f853f5412fceff5b48e12e3c5573eb2d/stdpopsim/slim_engine.py#L444-L452

But a user could easily break this by drawing two mutations of the same mutation type. See note here: https://github.com/popsim-consortium/stdpopsim/blob/973d5b05f853f5412fceff5b48e12e3c5573eb2d/stdpopsim/ext/selection.py#L115-L133

nspope:
Fixed by #1341. We're opting to disallow multiple `DrawMutation` events at the same site for the time being.

PR #1341, opened 2022-08-27
Add API for selective sweeps

### LINKED ISSUE #1329: Convenience functions for single site selection models
In #1082, @andrewkern suggested making the single site selection API more user-friendly. The idea is to have convenience functions that set up lists of extended events for frequently-used models like hard sweeps.

This could look something like:
```
contig.add_single_site(id="sweep_locus", coordinate=10000, fitness_coeff=0.1)
sweep_events = stdpopsim.ext.selective_sweep(
  single_site_id="sweep_locus",
  population_id=0, 
  start_time=1000, 
  end_time=0, 
  min_end_frequency=1.0
)
```
which would condition on a mutation originating in population 0 at time 1000 and reaching fixation by the end of the simulation.

Some questions are:
- What other models would be useful to support-- soft sweeps and adaptive introgression, probably? Should these be separate convenience functions, or combined into one function with more arguments?
- How much flexibility to allow in terms of allele frequency conditioning; e.g. for the hard sweep model above, should it be possible to condition on ending allele frequencies in multiple populations, rather than just the population where the mutation originated?
- To what degree should the lower-level API (lists of user-created extended events) be documented?
- Should `stdpopsim.ext` be removed?

### LINKED ISSUE #1336: Docstrings needed for `stdpopsim.ext` extended event classes
Docstrings are needed for the extended event classes in `stdpopsim.ext`.

### LINKED ISSUE #1337: Allele frequency conditioning with multiple `DrawMutation` events at the same locus
Relates to #608.

If multiple `DrawMutation` events with the same single site id are supplied to the SLiM engine, then multiple mutations of the same type will be generated. If these mutations make it into the sampled individuals, they will ultimately end up as different alleles (integer codes) in the outputted tree sequence.

This creates a bit of a dilemma for `ConditionOnAlleleFrequency`. What exactly should be conditioned on? Some options are:

1. The simulation could abort if there's more than one `DrawMutation` event at the same site. This is *de facto* what happens currently -- if there's more than one extant mutation at a site when the AF condition is evaluated, then SLiM will exit with a rather ambiguous error message. So we'd just need to add a check for overlapping `DrawMutation` events in `slim_makescript` and exit with an informative error.
2. The allele frequency condition could apply to the sum of frequencies of derived alleles.
3. The allele frequency condition could apply independently to all derived alleles. This seems useful in some cases, like for conditioning on the survival of all derived alleles. However we'd need some additional checks to ensure that the condition is possible to satisfy.
4. Separate conditions could be applied to separate alleles. I think the only reasonable way to allow this would be to permit multiple mutation types at the same "single site" (in which case alleles could have separate fitness coefficients) -- but this would require more infrastructure, testing, etc.

My vote is for (1) as this will require very little work to implement, and should cover many use cases (?). Ultimately it'd be nice to extend this to (4).

nspope:
- Fixes #1329 and #1082.
- Fixes a bug in the SLiM engine at https://github.com/popsim-consortium/stdpopsim/blob/ec4fedd9859dc23b7b12013d69172d851a5d43d2/stdpopsim/slim_engine.py#L217 that caused fitness calculations to be disabled for one generation after a restore. This probably had very little impact on simulation correctness, but did reduce efficiency of the rejection sampler.
- Fixes a bug where fitness logging would be disabled if the logging interval was set to every generation.
- Removes `selection_coeff` and `dominance_coeff` as arguments from `Contig.add_single_site`, as these are now specified in `ext.selective_sweep`. For the lower-level API this means that these parameters have to be set by a `ChangeMutationFitness` event.
- Fixes #1336.
- Fixes #1337 and #608.
- "Fixes" #656. As these classes may ultimately also be used for sweep specification in msprime, we'll keep calling them "extended events".

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1189,7 +1189,7 @@ but nonetheless there is lower diversity in exons than outside of them:
     is preliminary, and subject to change!
 
 You may be interested in simulating and tracking a single mutation. To illustrate
-this scenario, let's simulate a selective sweep until it reaches an abitrary
+this scenario, let's simulate a selective sweep until it reaches an arbitrary
 allele frequency.
 
 First, let's define a contig and a demographic model; here, we are simulating a
@@ -1212,34 +1212,21 @@ We must also decide the time the mutation will be added, when selection will
 start and at what frequency we want our selected mutation to be at the end of
 the simulation.
 
-Let's assume the mutation appeared 1000 generations ago, it has a positive
-effect on fitness (s=0.5). Also, we want the mutation to have reached a frequency
-of at least 0.8 by the end. Next, we'll walk through the steps required to do this:
-
-.. note::
-
-    Note that because we are doing a forward-in-time simulation, you should be
-    careful with your conditioning. For example, even a strongly selected mutation
-    would not be able to reach 80% frequency in just a few generations. Since
-    this conditioning works by re-running the simulation until the condition is
-    achieved, a nearly impossible condition will result in very long run times.
-
-To set things up, we need to first add the site at which the selected mutation
-will occur.  This is like adding a DFE, except to a single site -- we're saying
-that there is a potential mutation at a particular site with defined fitness
-consequences. So that we can refer to the single site later, we give it a
-unique string ID. Here, we'll add the site in the middle of the contig, with ID
-"hard sweep".
+First, we need to add the site at which the selected mutation will occur.  This
+is like adding a DFE, except to a single site -- we're saying that there is a
+potential mutation at a particular site with defined fitness consequences. So
+that we can refer to the single site later, we give it a unique string ID.
+Here, we'll add the site in the middle of the contig with ID "hard sweep",
+so named because we will imagine this beneficial mutation originates at
+frequency :math:`1 / 2N`.
 
 .. code-block:: python
 
-    mut_id = "hard sweep"
+    locus_id = "hard sweep"
     coordinate = round(contig.recombination_map.sequence_length / 2)
     contig.add_single_site(
-        id=mut_id,
+        id=locus_id,
         coordinate=coordinate,
-        fitness_coeff=0.1,
-        dominance_coeff=0.5,
     )
 
 .. note::
@@ -1251,62 +1238,34 @@ unique string ID. Here, we'll add the site in the middle of the contig, with ID
     "overwritten" and an error will be raised in simulation.
 
 Next, we will set up the "extended events" which will modify the demography.
-The first extended event is the origination of the selected mutation, which
-will occur in a random individual from the first population (id 0), 1000
-generations ago.
+This is done through :func:`stdpopsim.ext.selective_sweep`, which represents a
+general model for a mutation that is beneficial within a single population.  We
+specify that the mutation should originate 1000 generations ago in a random
+individual from the first population (named "pop_0" by default); that the
+selection coefficient for the mutation should be 0.5; and that the frequency of
+the mutation in the present day (e.g. at the end of the sweep) should be
+greater than 0.8.
 
 .. code-block:: python
 
-    T_mut = 1000
-    extended_events = [
-        stdpopsim.ext.DrawMutation(
-            time=T_mut,
-            single_site_id=mut_id,
-            population_id=0,
-        )
-    ]
-
-Next, we condition on the mutation not being lost.  Since in the next step we
-condition on the mutation being at 80% frequency at the end, this is redundant,
-but it allows the simulation to immediately restart from any generation in
-which the mutation is lost, rather than waiting until the end.  Note that this
-conditioning must start one generation after the mutation is placed, for which
-we use ``stdpopsim.ext.GenerationAfter(T_mut)``.  We cannot simply specify
-``T_mut - 1`` if rescaling is present, otherwise the conditioning would start
-at the same generation when the mutation is placed.
-
-.. code-block:: python
-
-    extended_events.append(
-        stdpopsim.ext.ConditionOnAlleleFrequency(
-            start_time=stdpopsim.ext.GenerationAfter(T_mut),
-            end_time=0,
-            single_site_id=mut_id,
-            population_id=0,
-            op=">",
-            allele_frequency=0.0,
-        )
+    extended_events = stdpopsim.ext.selective_sweep(
+        single_site_id=locus_id,
+        population="pop_0",
+        selection_coeff=0.5,
+        mutation_generation_ago=1000,
+        min_freq_at_end=0.8,
     )
 
-Finally, we condition on the mutation being above 80% at the end of the simulation.
-(The "end" is at time 0, since "time" is in generations before the end of the simulation.)
+.. note::
 
-.. code-block:: python
+    Note that because we are doing a forward-in-time simulation, you should be
+    careful with your conditioning. For example, even a strongly selected mutation
+    would not be able to reach 80% frequency in just a few generations. Since
+    this conditioning works by re-running the simulation until the condition is
+    achieved, a nearly impossible condition will result in very long run times.
 
-    extended_events.append(
-        stdpopsim.ext.ConditionOnAlleleFrequency(
-            start_time=0,
-            end_time=0,
-            single_site_id=mut_id,
-            population_id=0,
-            op=">=",
-            allele_frequency=0.8,
-        )
-    )
-
-Now we can simulate, using SLiM of course.
-For comparison, we will run the same simulation
-without selection - i.e., without the "extended events":
+Now we can simulate, using SLiM of course.  For comparison, we will run the
+same simulation without selection - i.e., without the "extended events":
 
 .. code-block:: python
 
--- a/stdpopsim/ext/selection.py
+++ b/stdpopsim/ext/selection.py
@@ -48,8 +48,9 @@ def validate_time_range(start_time, end_time):
 class ExtendedEvent(object):
     """
     ExtendedEvent is analogous to msprime.DemographicEvent, but is used here
-    for SLiM things that msprime doesn't support. Times are all in units of
-    generations before the present, just like for msprime.DemographicEvent.
+    for explicitly manipulating mutations. Currently, these are only used by
+    the SLiM engine. Times are all in units of generations before the present,
+    just like for msprime.DemographicEvent.
     """
 
     pass
@@ -58,23 +59,25 @@ class ExtendedEvent(object):
 @attr.s(kw_only=True)
 class DrawMutation(ExtendedEvent):
     """
-    TODO: docstring.
+    Introduce a new mutation with the given mutation type in the given
+    population at a given single site. The mutation will be added to one
+    randomly chosen chromosome in the given population. Only one mutation
+    may be drawn per single site.
 
-    Draw a new mutation with the given mutation type in the given population
-    at the given genomic coordinate. The mutation will be added to one randomly
-    chosen chromosome in the given population.
-
-    FIXME: Drawing multiple mutations using the same mutation type causes
-           erroneous allele frequency calculation in the SLiM code!
-           Multiple drawn mutations should be disallowed, or the allele frequency
-           calculation should somehow be tied to specific drawn mutations.
-
-    See SLiM documentation for addNewDrawnMutation().
+    :ivar time: The number of generations in the past at which to introduce the
+        mutation.
+    :vartype start_time: float
+    :ivar single_site_id: The string ID of the single site at which to
+        introduce the mutation, see :meth:`Contig.add_single_site`.
+    :vartype single_site_id: str
+    :ivar population: The name of the population in which to introduce
+        the mutation.
+    :vartype population: str
     """
 
     time = attr.ib(type=float)
     single_site_id = attr.ib(type=str)
-    population_id = attr.ib(type=int)
+    population = attr.ib(type=str)
 
     def __attrs_post_init__(self):
         validate_time(self.time)
@@ -83,20 +86,38 @@ class DrawMutation(ExtendedEvent):
 @attr.s(kw_only=True)
 class ChangeMutationFitness(ExtendedEvent):
     """
-    TODO: docstring.
-
     Change the selection and/or dominance coefficients for the given mutation
     type in the given population. The mutation this applies to need not exist
     in the population for all generations between start_time and end_time.
     E.g. if the mutation could be transferred to the population via migration.
+    This requires that a `DrawMutation` event with the same `single_site_id` is
+    also supplied to the simulation engine.
 
     See SLiM documentation for registerFitnessCallback().
+
+    :ivar start_time: The number of generations in the past at which the change
+        in mutation fitness begins.
+    :vartype start_time: float
+    :ivar end_time: The number of generations in the past at which the change
+        in mutation fitness ends.
+    :vartype end_time: float
+    :ivar single_site_id: The string ID of the single site at which to change
+        the mutation fitness, see :meth:`Contig.add_single_site`.
+    :vartype single_site_id: str
+    :ivar population: The name of the population in which to change
+        the mutation fitness. If `None` (the default), then all populations are
+        affected.
+    :vartype population: str
+    :ivar selection_coeff: The new selection coefficient of the mutation.
+    :vartype selection_coeff: float
+    :ivar dominance_coeff: The new dominance coefficient of the mutation.
+    :vartype dominance_coeff: float
     """
 
     start_time = attr.ib(type=float)
     end_time = attr.ib(type=float)
     single_site_id = attr.ib(type=str)
-    population_id = attr.ib(type=int)
+    population = attr.ib(type=str, default=None)
     selection_coeff = attr.ib(type=float)
     dominance_coeff = attr.ib(type=float)
 
@@ -107,21 +128,35 @@ class ChangeMutationFitness(ExtendedEvent):
 @attr.s(kw_only=True)
 class ConditionOnAlleleFrequency(ExtendedEvent):
     """
-    TODO: docstring.
-
     Condition on the allele frequency of a drawn mutation with the given
     mutation type in the given population. The mutation need not have been
-    drawn in the population being conditioned upon.
+    drawn in the population being conditioned upon.  This requires that a
+    `DrawMutation` event for the same single site is also supplied to the
+    simulation engine.
 
-    This uses a save/restore mechanism in the SLiM code to do rejection
-    sampling on a simulation (when the condition is not met) without throwing
-    away the entire simulation.
+    :ivar start_time: The number of generations in the past at which to start
+        checking the allele frequency condition.
+    :vartype start_time: float
+    :ivar end_time: The number of generations in the past at which to stop
+        checking the allele frequency condition.
+    :vartype end_time: float
+    :ivar single_site_id: The string ID of the single site at which to check
+        the allele frequency condition, see :meth:`Contig.add_single_site`.
+    :vartype single_site_id: str
+    :ivar population: The name of the population in which to check
+        the allele frequency condition.
+    :vartype population: str
+    :ivar op: The comparison operator used for the allele frequency condition,
+        one of `(">", ">=", "<", "<=")`.
+    :vartype op: str
+    :ivar allele_frequency: The allele frequency to compare against.
+    :vartype op: float
     """
 
     start_time = attr.ib(type=float)
     end_time = attr.ib(type=float)
     single_site_id = attr.ib(type=str)
-    population_id = attr.ib(type=int)
+    population = attr.ib(type=str, default=None)
     op = attr.ib(type=str, default=None)
     allele_frequency = attr.ib(type=float, default=None)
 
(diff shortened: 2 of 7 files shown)


nspope:
I think we need to store information about the sweep in the metadata, but I'm not sure yet what that should look like (as mutation fitness can change through time). Currently, single sites will show up in the metadata as strictly neutral DFEs that span a length-1 interval on the contig.

petrelharp:
What do you mean "we need to"? We have not yet formalized how we want to store anything in metadata; currently I think the only thing we're always storing there is `Q`. And, getting into the weeds of trying to store enough information there to completely recreate everything is a giant can of worms - there's some kind of happy balance to be struck that we haven't gotten at yet.

However, storing stuff in metadata *for testing purposes* is very useful; maybe this is what you meant? (i.e., in the "if debug > x" block)

nspope:
What I mean is that `_recap_and_rescale` walks over the DFEs in the metadata here
https://github.com/popsim-consortium/stdpopsim/blob/3166eadd31097771c6b1dc4dd2ce8e0792f02220/stdpopsim/slim_engine.py#L1668-L1671 

To simulate mutations over strictly neutral DFEs. Single sites are currently stored as strictly neutral DFEs, so there's a small chance they could end up with a neutral mutation. So, I think we should store something about the sweep in the metadata for its parent DFE, like the fitness coefficient.

petrelharp:
Oh, good catch - I forgot about that. Yes, i agree - should put something distinguishing in the metadata. I've not got a good suggestion at the moment, though.

stdpopsim/ext/selection.py
```diff
@@ -56,20 +56,28 @@ class ExtendedEvent(object):
...
+    Introduce a new mutation with the given mutation type in the given
+    population at a given single site. The mutation will be added to one
+    randomly chosen chromosome in the given population.
 
     FIXME: Drawing multiple mutations using the same mutation type causes
            erroneous allele frequency calculation in the SLiM code!
            Multiple drawn mutations should be disallowed, or the allele frequency
            calculation should somehow be tied to specific drawn mutations.
 
     See SLiM documentation for addNewDrawnMutation().
```
petrelharp, line 70:
```suggestion
```

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+            population `population_id` is below `min_freq_at_end` in this
+            generation, the simulation will restart from `mutation_time`.
+
+    If the mutation is lost at any point between `mutation_time` and
+    `end_time`, the simulation will restart from `mutation_time`.
+
+    During the sweep, the fitnesses of individuals in population
+    `population_id` that are homozygous or heterozygous for the beneficial
+    mutation are `1 + selection_coeff` and  `1 + selection_coeff *
+    dominance_coeff`, respectively. Fitnesses of individuals in other
```
petrelharp, line 230: Are you sure? The way it works in SLiM is that genotypes `aa`, `Aa`, and `AA` have fitnesses `1`, `1+hs` and `1+s`, respectively.

petrelharp: oh wait that's what this said, just not in the order I expected, lol =)

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+        to introduce the mutation.
+    :param float selection_coeff: The selection coefficient of the beneficial
+        mutation during the sweep in population `population_id`.  Must be greater
+        than or equal to zero.
+    :param float dominance_coeff: The dominance coefficient of the beneficial
+        mutation during the sweep in population `population_id`.  Must be between
+        zero and one. Defaults to `0.5`.
+    :param float start_time: The number of generations in the past at which the
+        sweep starts and the mutation switches from neutral to beneficial. Defaults to
+        `mutation_time` (e.g. a "hard" sweep).
```
petrelharp, line 254:
```suggestion
        `mutation_time` (i.e., selection directly on  a new mutation).
```
"hard" versus "soft" refers to the *ending* condition (fixed or not); the starting condition is "standing variation versus new mutation".

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+    :param float end_time: The number of generations in the past at which the
+        sweep ends and the mutation switches from beneficial to neutral. Defaults to
+        `0` (e.g. the end of the simulation).
```
petrelharp, line 257:
```suggestion
        `0` (i.e., the end of the simulation).
```

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+            `population_id` is below `min_freq_at_start` in this generation, the
+            simulation will restart from `mutation_time`.
+
+        (3) At time `end_time`, the mutation reverts to neutrality in
+            population `population_id`. If the frequency of the mutation in
+            population `population_id` is below `min_freq_at_end` in this
+            generation, the simulation will restart from `mutation_time`.
+
+    If the mutation is lost at any point between `mutation_time` and
+    `end_time`, the simulation will restart from `mutation_time`.
```
petrelharp, line 225: I think we should say "conditioned on" somewhere, eg:
```suggestion
    `end_time`, the simulation will restart from `mutation_time`.
    The effect of the "restarts" is that the resulting simulation is
    drawn from the distribution of simulations conditioned on the mutation
    being introduced at time ``mutation_time``, being above ``min_freq_at_start``
    at ``start_time``, and above ``min_freq_at_and`` at ``end_time``.
```

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+    example, if `start_time` is very close to `mutation_time` and
+    `min_freq_at_start` is substantially greater than zero, the vast majority
+    of proposed allele frequency trajectories will be rejected and the
+    simulation may take an infeasibly long time to complete.
+
+    :param str single_site_id: The string ID of the single site at which to introduce
+        the mutation, see :meth:`Contig.add_single_site`.
+    :param int population_id: The integer ID of the population in which the mutation is
+        introduced and the sweep occurs.
+    :param float mutation_time: The number of generations in the past at which
```
petrelharp, line 244: it's nice to call these "time ago" eg `mutation_time_ago`; although if other `time` arguments don't we could not, for consistency.

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+            single_site_id=single_site_id,
+            start_time=GenerationAfter(mutation_time),
+            end_time=end_time,
+            population_id=population_id,
+            allele_frequency=0.0,
+            op=">",
+        ),
+    ]
+
+    # Neutral period between mutation introduction and onset of sweep.
```
petrelharp, line 296: why do you need to change to `s=0`? isn't that what `DrawMutation` does?

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+                end_time=end_time,
+                population_id=population_id,
+                selection_coeff=selection_coeff,
+                dominance_coeff=dominance_coeff,
+            ),
+        ]
+    )
+
+    if min_freq_at_start is not None:
+        if not 0 < min_freq_at_start <= 1:
```
petrelharp, line 327: it should be allowed to set `min_freq_at_start = 0.0`? I'd do this like
```
if min_freq_at_start is None:
    min_freq_at_start = 0.0
if min_freq_at_start > 0:
   # ...
```

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+            raise ValueError(
+                "If specified, the allele frequency at the start of the sweep "
+                "must be in (0, 1]."
```
petrelharp, line 330:
```suggestion
                "must be in [0, 1]."
```

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+                    end_time=GenerationAfter(start_time),
+                    population_id=population_id,
+                    allele_frequency=min_freq_at_start,
+                    op=">=",
+                ),
+            ]
+        )
+
+    if min_freq_at_end is not None:
+        if not 0 < min_freq_at_end <= 1:
```
petrelharp, line 351: same

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+        ]
+    )
+
+    if min_freq_at_start is not None:
+        if not 0 < min_freq_at_start <= 1:
+            raise ValueError(
+                "If specified, the allele frequency at the start of the sweep "
+                "must be in (0, 1]."
+            )
+        if start_time == mutation_time:
```
petrelharp, line 332: Put this inside the `if min_freq_at_start > 0:` block and we're good. (I think it should be allowed to have `min_freq_at_start = 0.0` and `start_time == mutation_time`, since after all the docs say that's the defautls.

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+            raise ValueError(
+                "If specified, the allele frequency at the end of the sweep "
+                "must be in (0, 1]."
+            )
+        slim_events.extend(
+            [
+                ConditionOnAlleleFrequency(
+                    single_site_id=single_site_id,
+                    start_time=end_time,
+                    end_time=end_time,
```
petrelharp, line 361: above we had `GenerationAfter(start_time)`; should this also have `GenerationAfter( )`?

stdpopsim/ext/selection.py
```diff
@@ -151,3 +190,195 @@ def __attrs_post_init__(self):
...
+                    op=">=",
+                ),
+            ]
+        )
+
+    # Neutral period following sweep.
+    if end_time > 0:
+        slim_events.extend(
+            [
+                ChangeMutationFitness(
```
petrelharp, line 373: Do we ned this one (since doesn't it revert back to neutral)?

stdpopsim/genomes.py
```diff
@@ -542,10 +528,14 @@ def add_single_site(
...
             distribution_type="f",
-            dominance_coeff=dominance_coeff,
-            distribution_args=[selection_coeff],
+            dominance_coeff=0.5,
+            distribution_args=[0.0],
             convert_to_substitution=False,
         )
+
+        # TODO: Is there a need to safeguard against `_recap_and_rescale` simulating
+        # mutations at the single site, because it is recorded as strictly neutral?
```
petrelharp, line 537: annoyingly, yes, we should probably do this

petrelharp:
A thought: since msprime actually has (less sophisticated) ability to simluate sweeps; maybe we don't want to call these slim_events, as maybe in the future we'll do 'em with msprime also? (what do you think, @andrewkern?)

petrelharp:
This is great! I have minor suggestions, above. @andrewkern should probably look at the proposed API at least to see if it's the right amount of flexible?

stdpopsim/ext/selection.py
nspope: Wait-- for example in the abstract for [this article](https://academic.oup.com/genetics/article/210/4/1429/5931099) they define soft sweeps as those that start from standing variation (so that multiple haplotypes are carried to high frequency instead of just one). I thought the ending condition of the allele -- fixation or not -- was referred to as a "complete" or "partial" sweep. Do I have this backwards?

stdpopsim/ext/selection.py
nspope: I don't think so, because `end_time` can be 0. But you're right: when I look at when the mutation is increasing fitness, it's from `(start_time-1)` through `end_time`, and this doesn't align with the docstring.

stdpopsim/genomes.py
nspope: OK, I'll figure out a way to stick the selection/dominance coefficients into the metadata for the parent DFE.

docs/tutorial.rst
```diff
@@ -1212,34 +1212,19 @@ We must also decide the time the mutation will be added, when selection will
...
-will occur.  This is like adding a DFE, except to a single site -- we're saying
-that there is a potential mutation at a particular site with defined fitness
-consequences. So that we can refer to the single site later, we give it a
-unique string ID. Here, we'll add the site in the middle of the contig, with ID
-"hard sweep".
+First, we need to add the site at which the selected mutation will occur.  This
+is like adding a DFE, except to a single site -- we're saying that there is a
+potential mutation at a particular site with defined fitness consequences. So
+that we can refer to the single site later, we give it a unique string ID.
+Here, we'll add the site in the middle of the contig, with ID "hard sweep".
```
andrewkern, line 1219:
```suggestion
Here, we'll add the site in the middle of the contig, with ID "hard sweep",
so named because we will imagine this beneficial mutation originates at 
frequency 1 / 2N.
```

maybe we could spell this out for users?

docs/tutorial.rst
```diff
@@ -1250,63 +1235,34 @@ unique string ID. Here, we'll add the site in the middle of the contig, with ID
     contains a single site mutation, the single site mutation will be
     "overwritten" and an error will be raised in simulation.
 
-Next, we will set up the "extended events" which will modify the demography.
-The first extended event is the origination of the selected mutation, which
-will occur in a random individual from the first population (id 0), 1000
-generations ago.
+Next, we will set up the "SLiM events" which will modify the demography.  This
+is done through :func:`stdpopsim.ext.selective_sweep`, which represents a general model
+for a mutation that is beneficial over a time interval within a single
```
andrewkern, line 1240:
```suggestion
for a mutation that is beneficial within a single
```
removing the time interval bit, because we aren't yet allowing for time inhomogenous selection and I think this is clearer

docs/tutorial.rst
```diff
@@ -1250,63 +1235,34 @@ unique string ID. Here, we'll add the site in the middle of the contig, with ID
...
+population.  We specify that the mutation should originate 1000 generations ago
+in a random individual from the first population (id 0); that the selection
+coefficient for the mutation should be 0.5; and that the frequency of the mutation
+in the present day (e.g. at the end of the sweep) should be 0.8.
```
andrewkern, line 1244:
```suggestion
in the present day (e.g. at the end of the sweep) should be greater than 0.8.
```

stdpopsim/ext/selection.py
andrewkern: yeah i agree with you nate-- hard vs soft is about the origin of the beneficial alleles, not whether it has fixed or not.

stdpopsim/genomes.py
```diff
@@ -484,52 +484,38 @@ def add_single_site(
...
         coordinate,
-        selection_coeff=0.0,
-        dominance_coeff=0.5,
         description=None,
         long_description=None,
     ):
         """
-        Adds a model of a single site mutation at the provided coordinate.
-        The string label of the single site may be referenced by extended
+        Adds a potential mutation at a single site at the provided coordinate.
```
andrewkern, line 491:
```suggestion
        Adds a single site mutation class at the provided coordinate.
```

andrewkern (approved):
looks good to me @nspope. some minor suggestions here.

stdpopsim/ext/selection.py
```diff
@@ -56,20 +56,28 @@ class ExtendedEvent(object):
...
-    TODO: docstring.
-
-    Draw a new mutation with the given mutation type in the given population
-    at the given genomic coordinate. The mutation will be added to one randomly
-    chosen chromosome in the given population.
+    Introduce a new mutation with the given mutation type in the given
+    population at a given single site. The mutation will be added to one
+    randomly chosen chromosome in the given population.
 
     FIXME: Drawing multiple mutations using the same mutation type causes
```
mufernando, line 65: Is this bug still here?

nspope: It's fixed, good catch

mufernando (approved):
made a few comments, but LGTM after you finish responding @petrelharp and @andrewkern points!

stdpopsim/ext/selection.py
petrelharp: lol, sorry, I got my wires crossed!

stdpopsim/ext/selection.py
nspope: I agree @mufernando, `min_freq_at_XXX` implies a closed interval and as the mutation can never be lost, it's gotta be >0. I'll change the docs instead.

stdpopsim/ext/selection.py
nspope: As @mufernando says above, a half-open interval makes sense given the argument name-- I'll change docstring to reflect this

stdpopsim/ext/selection.py
nspope: Changed to `xxx_generation_ago` -- this isn't consistent with the argument naming convention used by the extended events themselves, but those aren't exposed/documented so I think it's fine.

nspope:
This should be good to go. The only thing I'm unsure of is how far to go in modifying the metadata for single sites. What I've done is to find those single sites that are referenced by extended events, and mark the mutation type within the associated DFE as non-neutral. This ensures that neutral mutations won't be accidentally simulated on top of any single sites with drawn mutations.

But, there's no other information copied over (selection coefficients, etc). In general, there could be multiple `ChangeMutationFitness` events for a single mutation -- and in that case it's not clear to me which one should "donate" parameters to the metadata. Alternatively, information from all associated extended events could be copied over into `long description`. I think the decision on what to do here could be postponed until either msprime sweeps are supported, or an API for more complicated "sweep-like" scenarios is introduced.

stdpopsim/genomes.py
```diff
@@ -542,10 +528,14 @@ def add_single_site(
...
         mut_type = stdpopsim.MutationType(
             distribution_type="f",
-            dominance_coeff=dominance_coeff,
-            distribution_args=[selection_coeff],
+            dominance_coeff=0.5,
+            distribution_args=[0.0],
             convert_to_substitution=False,
         )
+
+        # TODO: Is there a need to safeguard against `_recap_and_rescale` simulating
```
petrelharp, line 536: still this TODO

stdpopsim/ext/selection.py
```diff
@@ -83,14 +86,31 @@ def __attrs_post_init__(self):
...
 class ChangeMutationFitness(ExtendedEvent):
     """
-    TODO: docstring.
-
     Change the selection and/or dominance coefficients for the given mutation
     type in the given population. The mutation this applies to need not exist
     in the population for all generations between start_time and end_time.
     E.g. if the mutation could be transferred to the population via migration.
+    This requires that a `DrawMutation` event for the same single site is also
+    supplied to the simulation engine.
```
petrelharp, line 94:
```suggestion
    This requires that a `DrawMutation` event with the same single_site_id is also
    supplied to the simulation engine.
```

stdpopsim/ext/selection.py
```diff
@@ -107,15 +127,29 @@ def __attrs_post_init__(self):
...
     Condition on the allele frequency of a drawn mutation with the given
     mutation type in the given population. The mutation need not have been
-    drawn in the population being conditioned upon.
-
-    This uses a save/restore mechanism in the SLiM code to do rejection
-    sampling on a simulation (when the condition is not met) without throwing
-    away the entire simulation.
+    drawn in the population being conditioned upon.  This requires that a
+    `DrawMutation` event for the same single site is also supplied to the
+    simulation engine.
```
petrelharp, line 134:
```suggestion
    `DrawMutation` event with the same single site id is also supplied to the
    simulation engine.
```

petrelharp:
Sorry to bring this up so late, but I'm just noticing that sweeps created by `selective_sweep` are *only* beneficial in `population_id`. I would have expected that a "sweep" unless further qualified would be globally beneficial. Sure, many (local) sweeps are only locally adaptive, but the generic "selective sweep" is globally adaptive, I would have thought? Maybe this is OK, but here are some other options:
1. call this `local_sweep( )` instead of `selective_sweep( )`
2. just have it globally beneficial
3. have a flag that switches it from "global" to "local"
4. have separate arguments for "originating population" and "populations where it's beneficial

I like option (3), currently - it seems to strike a good balance between simplicity and allowing the original behavior. I'd say that it should default to 'global'.

andrewkern:
ah this is a good point @petrelharp! ultimately we'd like to be able to have a bit of flexibility about the selection coefficient in space and time

petrelharp (commented):
Other than the API comment, looks great to me!

nspope:
- Extended events now use string `population` instead of integer `population_id`.
- Added flag for global sweep to `ext.selective_sweep` that's on by default. The `ChangeMutationFitness` event is modified so that if `population` is omitted, it applies to all populations. 
- Allele frequency conditioning only applies to the source population, regardless of this flag -- after talking it over with @petrelharp we decided that global frequency conditioning was a bad idea (eg would make the rejection sampling really inefficient in many cases). I think this behavior is made sufficiently clear in the documentation but see what you think.
- Do we want to migrate this stuff out of `ext`?

andrewkern:
this looks all good to me @nspope. wrt `ext` I think this can probably come out of there, but I vote we merge this first

stdpopsim/slim_engine.py
```diff
@@ -470,21 +470,36 @@
...
             /* We explicitly format() here to prevent integral-valued floats
              * from getting converted to integers during string interpolation
              * (this triggers a type error when the fitness callback runs). */
             f_hom = format("%e", 1 + selection_coeff);
             f_het = format("%e", 1 + selection_coeff * dominance_coeff);
-            sim.registerMutationEffectCallback(NULL,
-                "{if (homozygous) return "+f_hom+"; else return "+f_het+";}",
-                mut_type, pop_id, g_start, g_end);
+
+            /* "All populations" is encoded by a negative value of pop_id. */
```
petrelharp, line 479: Oh dear, is there a different way to do this? Like `None`? Here it would be easy to make a programming mistake? I realize that this is internal only, so not a big deal, though.

petrelharp:
I didn't see it - do you have a test for what happens if you pass in an integer for 'population' when it's now supposed to be a string? (ie the old behavior)

petrelharp:
I'm happy to merge this when you say it's ready.

stdpopsim/slim_engine.py
nspope: Yeah I agree, I think setting pop_id to NULL should do it (and would simplify that code blurb).

nspope: Ugh, using NULL in place of -1 is more complicated than I thought, because of how data vectors are assembled in `slim_makescript`. Let's pass on this for now.

andrewkern:
okay i'm going to squash and merge. thanks @nspope!!

Issue #1346, opened 2022-08-31
Add a way to retrieve a (possibly random) interval from a `Contig`


nspope:
As discussed in #670 and #401, and started in #402.

nspope:
Closed by #1348.

Issue #670, opened 2020-10-30
API for `Contig` class and `Species.get_contig()`.


grahamgower:
We've recently had a number of ideas that require or suggest changes to the `Contig` class or the `get_contig()` method. I've opened this issue so that we can think about the API more concretely. It looks like we're going to end up with lots of new parameters to `get_contig()`, so I'll start with describing some of the semantics for various use cases (these naturally extend to hypothetical CLI params too).

Currently possible:
* `get_contig("chr1")`. Simulate chromosome 1. Use a uniform recombination rate matching chr1.
* `get_contig("chr1", genetic_map="foo")`. Simulate chromosome 1 using the "foo" genetic map.
* `get_contig("chr1", length_multiplier=0.01)`. I don't care what it is, just simulate something. And be quick about it! Use a uniform recombination rate matching chr1.

Proposed:
* `get_contig(length=1e6)`. Simulate 1 Mb. Use a uniform recombination rate matching the genome-wide average. #668/#669
* `get_contig("chr1", genetic_map="foo", mask=unmappable_intervals)`. Simulate chromosome 1 using the "foo" genetic map, but mask out the regions in "unmappable_intervals" (this could be some array data structure, or maybe a .bed filename?). #664
* `get_contig("chr1", start=32e6, length=1e6, genetic_map="foo")`. Simulate 1Mb. Use the section of the recombination map from chr1:32Mb-33Mb. #401/#402
* `get_contig(length=1e6, genetic_map="foo")`. Simulate 1 Mb. Use a 1Mb chunk of the "foo" genetic map, sampled uniformly at random. #401/#402
* `get_contig(length=1e6, genetic_map="foo", mask=unmappable_intervals)`. Simulate 1 Mb. Use a 1Mb chunk of the "foo" genetic map, sampled uniformly at random from intervals not in the "unmappable_intervals". (proposed interaction of #668 and #401).

Invalid configurations:
* `length_multiplier` with any of: `genetic_map`,`mask`, or `length`.

---
In #644, @mufernando is adding new `mutation_types` and `genomic_element_types` attributes to the Contig class in order to handle distinct mutation types/rates on distinct genomic annotations. This should naturally combine with many of the above use cases.
* `get_contig(length=1e6)` with `some_BGS_model()`. Simulate 1 Mb, with 1Mb of annotations drawn uniformly at random from the annotations used by the given BGS model.
[...]

PR #1348, opened 2022-09-03
Add API to get contig from region of named chromosome

nspope:
Mostly addresses #1346, #670, #401 (and supersedes #402) by adding a way to get a contig from an arbitrary interval of a named chromosome.

Thanks to `msprime.RateMap.slice` this was pretty straightforward, except for the handling of coordinates when DFEs are added. I opted to use the same coordinate system as the chromosome from which the contig is extracted, so for example the following adds a DFE in the middle of the contig:
```python
contig = species.get_contig(chromosome="chr1", left=1e6, right=1.1e6)
contig.add_dfe(intervals=[[1.05e6, 1.06e6]], dfe=...)
```
IMO this is cleaner than requiring the user to manually shift the DFE coordinates (and is less confusing with DFE bed files, annotations, etc).

Previously, if an added DFE had an interval that fell outside the contig, a rather ambiguous SLiM error would get raised. Now, the added DFE intervals are silently clipped to the contig boundaries. If all the intervals fall outside of these boundaries, a warning is raised and a DFE with empty intervals is added.

The same goes for masks (previously an error would get thrown from tskit if a mask interval overlapped a contig boundary): these are now clipped, with a warning if all masked intervals fall outside of the contig boundaries.

### DIFF
--- a/stdpopsim/cli.py
+++ b/stdpopsim/cli.py
@@ -556,8 +556,22 @@ def add_simulate_species_parser(parser, species):
         "--length-multiplier",
         default=1,
         type=float,
-        help="Simulate a sequence of length l times the named chromosome's length, "
+        help="Simulate a sequence of length l times the named chromosome's length "
         "using the named chromosome's mutation and recombination rates.",
+    )
+    species_parser.add_argument(
+        "--left",
+        default=None,
+        type=float,
+        help="The leftmost coordinate (inclusive) of the contig on the named "
+        "chromosome. Defaults to 0 (the beginning of the chromosome).",
+    )
+    species_parser.add_argument(
+        "--right",
+        default=None,
+        type=float,
+        help="The rightmost coordinate (exclusive) of the contig on the named "
+        "chromosome. Defaults to the length of the chromosome.",
     )
     species_parser.add_argument(
         "-s",
@@ -602,11 +616,11 @@ def add_simulate_species_parser(parser, species):
     )
     interval_help = (
         "Specify the interval where selection (given a DFE) is simulated. "
-        "Anything outside of the interval is simulated as neutral "
+        "Anything outside of the interval is simulated as neutral. "
         "If no interval is specified, "
-        "selection is simulated across the entire region (contig). "
-        "Interval (for now) is writen as 'left,right' (separated by a comma, "
-        "with no space, for instance: --dfe-interval 1000,2000"
+        "selection is simulated across the entire contig. "
+        "The interval is written as 'start,end' (separated by a comma, "
+        "with no space, for instance: --dfe-interval 1000,2000. "
         # say something here about multiple dfes?
     )
     species_parser.add_argument(
@@ -618,7 +632,7 @@ def add_simulate_species_parser(parser, species):
     )
     bed_help = (
         "A bed file specifing the intervals where selection (given a DFE) is simulated. "
-        "Non-overlaping intervals belonging to the same chromosome are required. "
+        "Non-overlapping intervals belonging to the same chromosome are required. "
         "If no interval is specified, "
         "selection is simulated across the entire contig. "
         "See also --dfe-interval and --dfe-annotation."
@@ -709,6 +723,8 @@ def add_simulate_species_parser(parser, species):
             inclusion_mask=args.inclusion_mask,
             exclusion_mask=args.exclusion_mask,
             mutation_rate=model.mutation_rate,
+            left=args.left,
+            right=args.right,
         )
         engine = stdpopsim.get_engine(args.engine)
         logger.info(
@@ -765,9 +781,8 @@ def add_simulate_species_parser(parser, species):
                 intervals_summary_str = f"[{left}, {right})"
             if intervals_summary_str is None:
                 # case where no intervals specified but we have a DFE
-                intervals = np.array(
-                    [[0, int(contig.recombination_map.sequence_length)]]
-                )
+                _, left, right = contig.original_coordinates
+                intervals = np.array([[left, right]])
                 intervals_summary_str = f"[{intervals[0][0]}, {intervals[0][1]})"
 
             dfe = species.get_dfe(args.dfe)
@@ -861,7 +876,9 @@ def write_simulation_summary(
             f"using mutation rate from species contig ({contig.mutation_rate})"
         )
     contig_len = contig.recombination_map.sequence_length
+    contig_origin = contig.origin
     dry_run_text += "Contig Description:\n"
+    dry_run_text += f"{indent}Contig origin: {contig_origin}\n"
     dry_run_text += f"{indent}Contig length: {contig_len}\n"
     dry_run_text += f"{indent}Mean recombination rate: {mean_recomb_rate}\n"
     dry_run_text += f"{indent}Mean mutation rate: {mut_rate}\n"
--- a/stdpopsim/genetic_maps.py
+++ b/stdpopsim/genetic_maps.py
@@ -122,8 +122,10 @@ class GeneticMap:
             rates = np.append(recomb_map.rate, 0)
             recomb_map = msprime.RateMap(position=positions, rate=rates)
         elif map_length > chrom.length:
+            # TODO: consider making this an error, and deprecating genetic maps
+            # that do not match the assembly.
             warnings.warn(
                 f"Recombination map has length {map_length}, which is longer than"
-                f" chromosome length {chrom.length}. The former will be used."
+                f" chromosome length {chrom.length}. The latter will be used."
             )
         return recomb_map
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -189,6 +189,9 @@ class Contig:
     and ``interval_list``. These must be of the same length, and the k-th DFE
     applies to the k-th interval; see :meth:`.add_dfe` for more information.
 
+    The contig may be a segment of a named chromosome. If so, the original
+    coordinate system is used for :meth:`.add_dfe` and :meth:`add_single_site`.
+
     :ivar mutation_rate: The rate of mutation per base per generation.
     :vartype mutation_rate: float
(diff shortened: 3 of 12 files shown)


nspope:
Ah, there's a slight issue. If the recombination map is longer than the chromosome, the current behavior of stdpopsim is to use the recombination map length for simulations instead of the chromosome length. But, the behavior in this PR is to clip the recombination map to the size of the contig. I can modify the PR to keep the current behavior (when the contig is the entire chromosome and the recombination map is longer) but this seems like an odd edge case to document. Especially if the plan is to deprecate genetic maps that don't match the assembly. So I've kept the new "clip-to-contig" behavior.

There's no effect on any of the tests, aside from the occasional stochastic failure of `test_slim_engine.py::TestRecombinationMap::test_chr1` that uses "HapMapII_GRCh37" and explicitly checks that simulated tree sequence breakpoints fall within the recombination map length. Updating this test to use "HapMapII_GRCh38" eliminates the stochastic failure.

stdpopsim/genomes.py
```diff
@@ -352,20 +369,49 @@ def species_contig(
...
+                        f"Right coordinate {right} falls outside chromosome "
+                        f"{chrom.id} with length {chrom.length}."
+                    )
+
             if genetic_map is None:
                 logger.debug(f"Making flat chromosome {length_multiplier} * {chrom.id}")
                 gm = None
+                # TODO: is there a need for `length_multiplier`, given the ability
+                # to specify an arbitary interval? Combining both here results in
+                # a confusing coordinate system.
```
grahamgower, line 396: Agreed. I vote to deprecate the length multiplier. It might have been useful during initial testing and development, but it doesn't compose well with any of the newer Contig features.

stdpopsim/genomes.py
```diff
@@ -352,20 +369,49 @@ def species_contig(
...
                 logger.debug(f"Getting map for {chrom.id} from {genetic_map}")
                 gm = species.get_genetic_map(genetic_map)
                 recomb_map = gm.get_chromosome_map(chrom.id)
+                recomb_map = recomb_map.slice(left=left, right=right, trim=True)
 
+            # TODO: previously, a mask with intervals falling outside the
+            # contig would throw an error in tskit. Now, the mask is silently
+            # clipped to the contig boundaries.  Should a warning be raised if
+            # this happens? (A warning will be raised if ALL masked intervals
+            # are outside of the contig).
```
grahamgower, line 414: Is there a reason not to just explicitly raise an error in this case?

stdpopsim/genomes.py
```diff
@@ -410,7 +463,16 @@ def species_contig(
...
 
-    def dfe_breakpoints(self):
+    @property
+    def original_coordinates(self):
+        """
+        If the contig represents a chromosomal segment, this is gives the name
+        of the parent chromosome and the left and right coordinates of the
+        contig on this chromosome.
+        """
+        return stdpopsim.utils.parse_region_string(self.origin)
```
grahamgower, line 473: I think it would be better to store the original coordinates as Contig fields and produce the `origin` property from them, instead of the storing and parsing the origin string.

stdpopsim/utils.py
```diff
@@ -249,3 +250,54 @@ def mask_intervals(intervals, mask):
...
+    ``intervals`` should be a numpy array with two columns that are the
+    beginning and end coordinates, respectively. The coordinates are then
+    shifted such that ``left`` becomes 0.
+    """
+    intervals = np.array(intervals).astype(int)
+    _check_intervals_validity(intervals)
+    _check_intervals_validity(np.array([[left, right]], dtype=int))
+    out = []
+    for inter in intervals:
+        il, ir = inter
```
grahamgower, line 267:
```suggestion
    for (il, ir) in intervals:
```

tests/test_masking.py
```diff
@@ -14,18 +13,18 @@
 
 
 class TestMasking:
-    def test_load_intervals(self):
+    def test_load_intervals(self, tmp_path):
```
grahamgower, line 16:
```suggestion
    @pytest.usefixtures("tmp_path")
    def test_load_intervals(self, tmp_path):
```
I know this isn't required, but unless one is very familiar with pytest then reading tests with "magic" arguments is very confusing.

grahamgower (changes requested):
This is great, thanks @nspope!

I'm luke warm on all the clipping though. Can these just raise errors instead? Are there any reasonable use cases for mismatched masks, recombination maps, and genome coordinates?

stdpopsim/genomes.py
nspope: Hmm, say there's an available chromosome-wide mask and annotation, and the user wants to use these whilst simulating some small region. One option is that the user clips the mask/annotation to the region themselves, and stdpopsim raises errors if these don't all fall within the region. Another option is that stdpopsim does the clipping and warns if something extreme happens, like all mask/annotation intervals falling outside of the region (this is the current strategy in the PR). I could go either way -- not sure what the right trade-off is between ease-of-use and preventing miss-specification.

Another possibility is to go the latter route (stdpopsim does the clipping-to-region) but first check that the mask/annotation matches the chromosome boundaries and raise an error if not.

stdpopsim/genomes.py
nspope: I agree. So, leave the length multiplier feature in there (with a deprecation warning?) but don't try to accommodate it when adding new arguments/features.

nspope:
Thanks @grahamgower -- with regards to clipping, say there's an available chromosome-wide mask and annotation, and the user wants to use these whilst simulating some small region. One option is that the user clips the mask/annotation to the region themselves, and stdpopsim raises errors if any of the provided intervals fall outside the region. Another option is that stdpopsim does the clipping and warns if something extreme happens, like all mask/annotation intervals falling outside of the region (this is the current strategy in the PR). I could go either way -- not sure what is the right trade-off between ease-of-use and preventing miss-specification.

Another possibility is to go the latter route (stdpopsim does the clipping-to-region) but first check that the mask/annotation matches the chromosome boundaries and raise an error if not.

stdpopsim/genomes.py
grahamgower: Yeah, exactly. Don't worry about deprecating it in this pr though.

stdpopsim/genomes.py
grahamgower: Ok, I get it. In that case I think you've made the right decision.

nspope:
I think the last round of commits should address your comments, @grahamgower.

@andrewkern and/or @petrelharp, could you also take a look? It'd be good to get another opinion with regards to (1) keeping coordinate system of parent chromosome for adding DFEs; (2) clipping masks/maps/dfes to the contig boundaries.

grahamgower (approved):
LGTM!

andrewkern (commented):
just some minor changes here but looks good to me

andrewkern:
we'll also want to add some docs for this including a use case example

stdpopsim/cli.py
```diff
@@ -602,11 +616,14 @@ def add_simulate_species_parser(parser, species):
...
     interval_help = (
         "Specify the interval where selection (given a DFE) is simulated. "
-        "Anything outside of the interval is simulated as neutral "
+        "Anything outside of the interval is simulated as neutral. "
         "If no interval is specified, "
         "selection is simulated across the entire region (contig). "
         "Interval (for now) is writen as 'left,right' (separated by a comma, "
-        "with no space, for instance: --dfe-interval 1000,2000"
+        "with no space, for instance: --dfe-interval 1000,2000. "
+        "If --left and --right are used to specify a chromosomal segment "
```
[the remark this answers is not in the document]
nspope: yeah, they can. I'll remove that bit, everything is on the same coordinate system so I don't think we need to confuse things by talking about the coordinate system.

stdpopsim/genomes.py
```diff
@@ -238,10 +246,14 @@ class Contig:
...
     interval_list = attr.ib(factory=list)
+    original_coordinates = attr.ib(default=None, type=tuple)
 
     def __attrs_post_init__(self):
+        if self.original_coordinates is None:
+            self.original_coordinates = (None, 0, int(self.length))
+        _, left, right = self.original_coordinates
         self.add_dfe(
-            np.array([[0, int(self.length)]]),
+            [[left, right]],
```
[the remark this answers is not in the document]
nspope: It's all converted under the hood to numpy arrays now -- see this test (which was skipped previous to this PR):

https://github.com/popsim-consortium/stdpopsim/blob/2e042a7a04d97896bfa3f076df8f1963efdf3f5b/tests/test_genomes.py#L113

and _check_intervals_validity will complain if it's not the right shape. 

Or do you think we want to make the input format specific?

stdpopsim/genetic_maps.py
```diff
@@ -122,8 +122,10 @@ def get_chromosome_map(self, id):
             rates = np.append(recomb_map.rate, 0)
             recomb_map = msprime.RateMap(position=positions, rate=rates)
         elif map_length > chrom.length:
+            # TODO: Is there any reason to not make this an error, and
+            # deprecate genetic maps that do not match the assembly?
```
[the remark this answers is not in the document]
nspope: Fair enough-- but I do think we want to deprecate/discourage using old genetic maps so let's keep a TODO there?

nspope:
in addition to use case, we'll need to make it clear in the docs that simulating a region with selection is not the same as simulating a chromosome with selection and cropping to the region.

nspope:
@andrewkern do you mind taking a look at the additions to the tutorial?

Otherwise I think this is good to go!

andrewkern (approved):

andrewkern:
looks good to me @nspope! can you rebase and squash these changes before we merge?

petrelharp:
> rebase and squash these changes before we merge?

"Squash and merge" will do this for you.

PR #1354, opened 2022-09-05
update ubuntu in tests (closes #1353)

### LINKED ISSUE #1353: update ubuntu in tests
Github warns us about our tests now:
```
The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002
```
... I figure we should just change to [20.04 here](https://github.com/popsim-consortium/stdpopsim/blob/main/.github/workflows/tests.yml#L20).

petrelharp:
Let's see if this works. However, 18 is still supported [until next April](https://wiki.ubuntu.com/BionicBeaver/ReleaseNotes) so maybe I'm jumping the gun here.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04, macos-10.15, windows-latest]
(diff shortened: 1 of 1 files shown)


grahamgower (approved):
LGTM

PR #1355, opened 2022-09-08
revision to gene conversion API

### LINKED ISSUE #1332: Support for gene conversion in SLiM engine
With #1106 the msprime engine supports gene conversion, and the rate/mean tract length parameters are stored as attributes of `stdpopsim.Contig`. It'd be nice to also allow gene conversion with the SLiM engine. This would require:

1. Having `slim_makescript` add a line with `initializeGeneConversion(...);` to the `initialize()` block of the SLiM script.
2. Reconciling differences in parameterization between the SLiM and msprime GC models. If I'm understanding the documentation correctly, msprime independently samples the locations of simple crossovers and GC events, using separately provided rates (a constant rate in the case of GC). Whereas SLiM draws breakpoints according to rates in a supplied recombination map, then randomly assigns each event to be a simple crossover or GC according to a supplied probability. So rates of GC events can vary across the sequence in SLiM, but not in msprime (currently). 
3. Support for heteroduplex mismatch repair -- there are two additional parameters in SLiM that control the proportion of GC tracts that are complex and the nucleotide bias resulting from mismatch repair within these tracts. These don't have an influence on ancestry but do influence stored mutations (?)

Relevant docs for [msprime](https://tskit.dev/msprime/docs/latest/ancestry.html#gene-conversion) and [SLiM](http://benhaller.com/slim/SLiM_Manual.pdf#page=32).

petrelharp:
Closes #1332; see there for discussion. Summary:

- `bacterial_recombination` is a property of a Genome and of a Contig; if this is True then recombination is all GC, so
  * `gene_conversion_fraction` must *not* be set
  * `gene_conversion_length` *must* be set
  * the recombination rate is converted at simulate( )-time to a gene conversion rate
- otherwise (ie for 'crossover' recombination), "recombination rate" is crossover rate, so:
  * if total rate of crossover + gene conversion is `total_rate` then `recomb_rate = total_rate * (1 - gc_frac)`
  * so for msprime, we set gene conversion rate to `gc_frac * total_rate = recomb_rate * gc_frac / (1 - gc_frac)`, where `recomb_rate` is the *mean* rate for the recombination map (since this must be constant)
  * and for SLiM we set the recombination rate (which here means the total rate!) to `recomb_rate / (1 - gc_frac)` and the gene conversion fraction to `gc_frac`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -78,6 +78,73 @@ Demographic Models
 .. autoclass:: stdpopsim.Population()
     :members:
 
+
+.. _sec_api_gene_conversion:
+
+*******************************************
+Gene conversion and bacterial recombination
+*******************************************
+
+Some species have estimates of the process of gene conversion.
+However, by default gene conversion does *not* happen, because
+(unfortunately) it can make simulations take much longer,
+and result in much bigger files.
+This is because the rate of gene conversion is usually much higher than
+the rate of crossing-over, so enabling gene conversion effectively
+increases the rate of recombination, which can have a strong effect on runtimes.
+To enable gene conversion for a species that has estimates,
+pass the `use_species_gene_conversion=True` argument to
+:meth:`Engine.simulate`.
+
+Gene conversion attributes are visible through the `gene_conversion_fraction`
+and `gene_conversion_length` properties of the species' `genome`.
+The "gene conversion fraction" gives the (average) fraction of
+double-stranded breaks that resolve as gene conversion
+(a.k.a non-crossover) events,
+and the "gene conversion length" is the mean length of the gene conversion tracts.
+So, if the rate of crossing over (which is often referred to as
+"recombination rate") is `r` and the gene conversion fraction is `f`,
+then the *total* rate of double-stranded breaks is `r / (1-f)`,
+and the gene conversion rate is `r * f / (1-f)`.
+
+A consequence of this is that
+the `recombination_map` attribute of a :class:`Contig` is the rate of
+double-stranded break initiation, not the rate of crossovers. The terminology
+conflicts somewhat with the `recombination_rate` property of
+a :class:`Chromosome`, which specifies the rate of crossovers. So, creating
+a contig with `use_species_gene_conversion=True` will result in a Contig
+with larger "recombination rates" than otherwise, because these rates
+include gene conversion as well as crossing over:
+
+.. code-block:: python
+
+    species = stdpopsim.get_species("DroMel")
+    contig = species.get_contig("2L")
+    contig.recombination_map.mean_rate
+    # 2.40462600791e-08
+    contig = species.get_contig("2L", use_species_gene_conversion=True)
+    contig.recombination_map.mean_rate
+    # 1.414485887005882e-07
+
+
+Different engines implement gene conversion in different ways: at the time
+of writing, msprime only allows a *constant* rate of gene conversion,
+while SLiM only allows gene conversion to be a fixed fraction of double-stranded
+break events. So, in SLiM we use the recombination map for the local
+rate of double-stranded breaks along the genome, while in msprime
+we specify a constant rate of gene conversion so that the average total number
+is the same as we would get from SLiM.
+
+In principle, **bacterial recombination** is a completely different issue,
+since bacterial recombination is by horizontal transfer of genomic segments.
+However, bacterial recombination is at present implemented using the engine's
+gene conversion mechanisms. So, bacterial species have the `bacterial_recombination`
+flag set, and need `gene_conversion_length` to be defined. (It could also be called horizontal
+transfer or homologous recombination segment length, but that would add another
+option.) For such species, gene conversion does not happen, so `gene_conversion_fraction`
+must not be set, and the "recombination rate" is the rate of bacterial recombination.
+
+
 .. _sec_api_dfes:
 
 ******************
--- a/stdpopsim/catalog/DroMel/species.py
+++ b/stdpopsim/catalog/DroMel/species.py
@@ -33,6 +33,7 @@ _ComeronEtAl = stdpopsim.Citation(
     author="Comeron et al",
     doi="https://doi.org/10.1371/journal.pgen.1002905",
     year=2012,
+    reasons={stdpopsim.CiteReason.REC_RATE, stdpopsim.CiteReason.GENE_CONVERSION},
 )
 
 # Mean chromosomal rates, calculated by taking the
@@ -42,7 +43,7 @@ _ComeronEtAl = stdpopsim.Citation(
 # Chromosome 4 isn't in this map, so the average of
 # 2L, 2R, 3L and 3R weighted by their respective
 # chromosome lengths was used instead.
-_recombination_rate_data = {
+_recombination_rate = {
     "2L": 2.40462600791e-08,
     "2R": 2.23458641776e-08,
     "3L": 1.79660308862e-08,
@@ -54,27 +55,52 @@ _recombination_rate_data = {
 }
 
 
-_chromosomes = []
-for name, data in genome_data.data["chromosomes"].items():
-    _chromosomes.append(
-        stdpopsim.Chromosome(
-            id=name,
-            length=data["length"],
-            synonyms=data["synonyms"],
-            mutation_rate=5.49e-9,  # _SchriderEtAl de novo mutation rate
-            recombination_rate=_recombination_rate_data[name],
-        )
-    )
+# Comeron et al:
+# - GC avg track length = 518bp
+# - 83% of DSBs are resolved as Gene Conversions
+# - avg rate of GC is 1.25E-07 / bp / female meiosis
+# (not sure if the latter estimate agrees)
 
-_genome = stdpopsim.Genome(
-    chromosomes=_chromosomes,
-    assembly_name=genome_data.data["assembly_name"],
-    assembly_accession=genome_data.data["assembly_accession"],
+_gene_conversion_fraction = {
+    "2L": 0.83,
+    "2R": 0.83,
+    "3L": 0.83,
+    "3R": 0.83,
+    "4": 0,
+    "X": 0.83,
+    "Y": 0,
+    "mitochondrion_genome": 0,
+}
+
+_gene_conversion_length = {c: 518 for c in genome_data.data["chromosomes"]}
+
+_mutation_rate = {c: 5.49e-9 for c in genome_data.data["chromosomes"]}
+
+# _chromosomes = []
+# for name, data in genome_data.data["chromosomes"].items():
+#     _chromosomes.append(
+#         stdpopsim.Chromosome(
+#             id=name,
+#             length=data["length"],
+#             synonyms=data["synonyms"],
+#             mutation_rate=5.49e-9,  # _SchriderEtAl de novo mutation rate
+#             recombination_rate=_recombination_rate_data[name],
+#             gene_conversion_fraction=_gene_conversion_fraction_data[name],
+#             gene_conversion_length=_gene_conversion_length,
+#         )
+#     )
+
+_genome = stdpopsim.Genome.from_data(
+    genome_data.data,
+    recombination_rate=_recombination_rate,
+    mutation_rate=_mutation_rate,
+    gene_conversion_fraction=_gene_conversion_fraction,
+    gene_conversion_length=_gene_conversion_length,
     citations=[
         _SchriderEtAl.because(stdpopsim.CiteReason.MUT_RATE),
         _DosSantosEtAl,
         _HoskinsEtAl,
-        _ComeronEtAl.because(stdpopsim.CiteReason.REC_RATE),
+        _ComeronEtAl,
     ],
 )
 stdpopsim.utils.append_common_synonyms(_genome)
--- a/stdpopsim/catalog/EscCol/species.py
+++ b/stdpopsim/catalog/EscCol/species.py
@@ -32,8 +32,7 @@ for name, data in genome_data.data["chromosomes"].items():
             # Wielgoss et al. (2011) calculated for strain REL606,
             # from synonymous substitutions over 40,000 generations.
             mutation_rate=8.9e-11,
-            recombination_rate=0.0,
-            gene_conversion_rate=8.9e-11,
+            recombination_rate=8.9e-11,
             gene_conversion_length=345,
         )
     )
@@ -43,6 +42,7 @@ for name, data in genome_data.data["chromosomes"].items():
 
 _genome = stdpopsim.Genome(
     chromosomes=_chromosomes,
+    bacterial_recombination=True,
     assembly_name=genome_data.data["assembly_name"],
     assembly_accession=genome_data.data["assembly_accession"],
     citations=[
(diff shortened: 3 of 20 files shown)


petrelharp:
Thinking more on this, I don't think that people should be able to change `bacterial_recombination` at all - so, this shouldn't be an argument to `Species.get_contig( )`, and it should be fixed by the Species property. That'll clear up a fair bit of the confusing stuff in here.

tests/test_genomes.py
```diff
@@ -158,38 +186,52 @@ def test_is_neutral(self):
 
 
 class TestGeneConversion(object):
-    species = stdpopsim.get_species("EscCol")
+    dro_mel = stdpopsim.get_species("DroMel")
```
fbaumdicker, line 189: I don't see where this is used later.

stdpopsim/genomes.py
```diff
@@ -320,31 +356,34 @@ def species_contig(
...
                     u_tot += chrom_data.length * chrom_data.mutation_rate
-                    if chrom_data.gene_conversion_rate is None:
-                        local_gc_rate = 0
-                    else:
-                        local_gc_rate = chrom_data.gene_conversion_rate
+                    if chrom_data.gene_conversion_fraction is not None:
+                        gcf_tot += (
+                            chrom_data.length * chrom_data.gene_conversion_fraction
+                        )
+                    if chrom_data.gene_conversion_length is not None:
```
fbaumdicker, line 370: This might cause some confusing effects as we are using the mean of the gc fraction and not the actual gc rate.

Example:

Two chromosomes with the same length.
Chromosome A has a recombination rate of 10 and a gc_fraction of 0.1, i.e. a gc_rate of approx. 1.11
Chromosome B has a recombination rate of 1 and a gc_fraction of 0.5, i.e. a gc_rate of 1.
The generic contig will now have a recombination rate of 5.5 and a gc_fraction of 0.3 resulting in a gc_rate of approx. 2.35.

fbaumdicker: The straightforward solution would be to calculate the actual gc_rates for each chromosome take the mean and calculate the corresponding gc_fraction from this.

fbaumdicker (commented):
So far this looks good to me.
There are two points we should think about:
A): How do we generate the gene_conversion_fraction for a generic contig (see my other comment)
B): I did not see whether a gene_conversion_fraction of 1.0 is allowed or not. I would vote for allowing a gene_conversion_fraction of 1.0. In this case, we should set the gene_conversion_rate equal to the recombination rate and set the recombination rate to zero in msprime and simply use a gene_conversion_fraction of 1.0 in slim.

petrelharp:
OK, after discussing with @nspope and @andrewkern, a revised, simpler, proposal:

- `species.get_contig( )` should *not* have arguments for gc fraction or gc length. These are pulled from the species if `use_species_gene_conversion=True`, and not otherwise. If these are not defined for a species, then `use_species_gene_conversion=True` will throw an error.
- The Contig that is returned, however, *should* have these properties. Furthermore, if gene conversion is turned on, the Contig's recombination map has been scaled up so that it is the rate of double-stranded breaks (ie crossovers + gene conversions).

Consequences/rationale:
- If someone wants to modify the default GC rate by hand (eg to turn it on for a species that does not have it defined), instead of passing an argument to `species.get_contig( )` they'd modify the contig after the fact, like we [document here](https://popsim-consortium.github.io/stdpopsim-docs/latest/api.html#stdpopsim.Contig) for recombination maps already. They *would* be able to set gc fraction to 1.0.
- The recombination map and gene conversion fraction are ready to be passed in to SLiM, as-is.
- When we simulate with msprime, we will have to un-scale the recombination map, since msprime wants a crossover map. But it's just a scaling, so it's easy.
- Less of the machinery is exposed to the API, which is good, since (a) we're not trying to be a fully-optioned simulator, (b) more options are more confusing, and (c) I suspect we'd want to change it in the future (e.g., what if we wanted separate GC and crossover maps).

So - this is changing things, but I think it still does what you want, @fbaumdicker, since there's an easy way to specify GC yourself; it'd be something like:
```
gc_frac = 0.9
contig = species.get_contig("chr1")
contig.recombination_map.rates *= 1/(1 - gc_frac)
contig.gene_conversion_fraction = gc_frac
contig.gene_conversion_length = 345
engine.simulate(...)
```

stdpopsim/genomes.py
petrelharp: hah good call

docs/api.rst
```diff
@@ -78,6 +78,62 @@ Demographic Models
 .. autoclass:: stdpopsim.Population()
     :members:
 
+
```
petrelharp, line 81: I'm not sure this is the right place to put this?

petrelharp:
And, here's how it looks in the catalog:
Screenshot from 2022-09-13 12-08-44
Screenshot from 2022-09-13 12-20-51

stdpopsim/genomes.py
```diff
@@ -390,15 +447,27 @@ def species_contig(
...
+                gene_conversion_fraction = chrom.gene_conversion_fraction
                 gene_conversion_length = chrom.gene_conversion_length
+                if (
+                    gene_conversion_fraction is not None
+                    and gene_conversion_fraction > 0
+                ):
+                    recomb_map = msprime.RateMap(
+                        position=recomb_map.position,
+                        rate=recomb_map.rate / (1 - gene_conversion_fraction),
+                    )
```
andrewkern, line 463: this patterns comes up a number of times in the code now. I wonder if it's worth writing a couple utility funcs that go from recomb_map to gc adjusted recomb_map?

andrewkern (commented):
looks good Peter. A few comments here.

docs/api.rst
```diff
@@ -78,6 +78,62 @@ Demographic Models
...
+we specify a constant rate of gene conversion so that the average total number
+is the same as we would get from SLiM.
+
+In principle, **bacterial recombination** is a completely different issue,
+since bacterial recombination is by horizontal transfer of genomic segments.
+However, bacterial recombination is at present implemented using the engine's
+gene conversion mechanisms. So, bacterial species have the `bacterial_recombination`
+flag set, and need `gene_conversion_length` to be defined. (It would be more accurate
+to call this the "horizontal transfer segment length", but that would add another
+option.) For such species, gene conversion does not happen, so `gene_conversion_fraction`
```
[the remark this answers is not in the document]
fbaumdicker: Unfortunately, there is no non-confusing term here to describe this. Some call it (bacterial) homologous recombination, and some refer to it as horizontal gene transfer even if it occurs within the same taxa/species. Maybe the best solution is to simply delete the "more accurate" and replace it with:
"It could also be called horizontal transfer or homologous recombination segment length, but ... "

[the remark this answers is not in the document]
petrelharp: I was worried about this also but
Screenshot from 2022-09-13 17-06-20

stdpopsim/catalog/HomSap/species.py
```diff
@@ -67,6 +67,31 @@
...
+    reasons={stdpopsim.CiteReason.GC_RATE},
+)
+
+_li2019 = stdpopsim.Citation(
+    doi="http://dx.doi.org/10.1038/s41467-019-11675-y",
+    year=2019,
+    author="Li, Bitoun, Alternose, Davies, Davies, and Myers",
+    reasons={stdpopsim.CiteReason.GC_RATE},
+)
+
```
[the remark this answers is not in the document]
petrelharp: e. coli person like
Screenshot from 2022-09-13 17-07-56
?

docs/api.rst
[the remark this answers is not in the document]
andrewkern: 😐

petrelharp, line 133:
```suggestion
flag set, and need `gene_conversion_length` to be defined. (It could also be called horizontal
transfer or homologous recombination segment length, but that would add another
option.) For such species, gene conversion does not happen, so `gene_conversion_fraction`
```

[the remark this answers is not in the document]
petrelharp: whoops - for some reason I didn't see Franz's comment before I posted this!

stdpopsim/catalog/HomSap/species.py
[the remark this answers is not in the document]
petrelharp: um, but really, what do you mean?

stdpopsim/genomes.py
petrelharp: I've had a look, and I don't think we actually do the same thing twice?

stdpopsim/catalog/HomSap/species.py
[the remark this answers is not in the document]
andrewkern: sorry i didn't see these were in humans-- simply wondering if we want an expert in GC to look over the numbers we are choosing for each genome. this is fine for now ofc but should we seek help?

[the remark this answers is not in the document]
petrelharp: yes! This is the major question about this PR for me - whether to merge with these dubious numbers and get someone to do it right or get the right numbers in first (and then QC, I suppose).

andrewkern:
found some drosophila melanogaster numbers to add to this! [Josep Comeron's (2012) paper](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1002905) has nice estimates:
- GC avg track length = 518bp
- 83% of DSBs are resolved as Gene Conversions
- avg rate of GC is 1.25E-07 / bp / female meiosis

igronau:
I wrote this comment on the original GitHub issue (#1332), but this is probably a better place. I'm making (final) edits to the **adding species paper**, and realized after discussion with @fbaumdicker that we need to incorporate the changes you mention here in the manuscript text. Currently, the section on **Updates to stdpopsim and the species catalog** includes a short paragraph on gene conversion specifying how the parameterization is set (using Franz's original setup). I'm not sure that we want to dive into the detail of the different cases because this will distract readers from the main point, and might also change in the future. Currently, the only concrete example we give is for bacteria (*E. Coli*). So, I thought that we'd mostly focus on parameterization for bacteria and archea, which is fairly simple  -- with the `bacterial_recombination` flag set and the `gene_conversion_tract_length` set to the appropriate value. We could also mention that gene conversion may be used in other species together with crossover recombination, and parameterization there is more complex (without diving into the details). That leads to another question: do we currently have an example in the catalog for a species where we model gene conversion and crossover recombination?

petrelharp:
Hi, @igronau - I'd say if we're describing gene conversion here we should describe both types; after all, we've got it in (I think this is ready to merge) and we've spent a bunch of time thinking about the API, so I'm not worried it'll change (at least, not soon)? And, yes - we've got estimates of gene conversion + crossover for DroMel in (and it'd be nice to get in some more).

petrelharp:
OK - removed the HomSap values; added DroMel, adjusted tests... I think this is good to merge! Want to have another glance, @andrewkern?

tests/test_genomes.py
```diff
@@ -158,38 +203,51 @@ def test_is_neutral(self):
...
+        contig = esc_col.get_contig(chromosome="Chromosome")
+        assert contig.bacterial_recombination is True
+        assert contig.gene_conversion_fraction is None
+        assert contig.gene_conversion_length > 0
+
+        rm = msprime.RateMap(position=[0.0, contig.length], rate=[0.1])
+        contig.recombination_map = rm
+        contig.gene_conversion_length = 1
+        assert contig.recombination_map == rm
+        assert contig.gene_conversion_length == 1
```
andrewkern, line 235: maybe i'm missing it, but it seems like you should have a test that checks if recombination_map of a contig grows when the gene_conversion_fraction is > 0

andrewkern (commented):
looking really good. a few minor things and maybe a test to add?

stdpopsim/catalog/DroMel/species.py
```diff
@@ -54,27 +55,52 @@
...
+    recombination_rate=_recombination_rate,
+    mutation_rate=_mutation_rate,
+    gene_conversion_fraction=_gene_conversion_fraction,
+    gene_conversion_length=_gene_conversion_length,
     citations=[
         _SchriderEtAl.because(stdpopsim.CiteReason.MUT_RATE),
         _DosSantosEtAl,
         _HoskinsEtAl,
-        _ComeronEtAl.because(stdpopsim.CiteReason.REC_RATE),
+        _ComeronEtAl,
```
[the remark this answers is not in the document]
petrelharp: These are the reasons given in `_ComeronEtAl` above, so by my reading this wouldn't actually change anything? See the source for `Citation.because( )`: https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/citations.py#L61

tests/test_cli.py
```diff
@@ -1022,11 +1022,15 @@ def test_noQC_citations_not_written_verbose(self, caplog):
...
-    "species_id", [species.id for species in stdpopsim.all_species()]
-)
+@pytest.mark.parametrize("species_id", [s.id for s in stdpopsim.all_species()])
 def test_species_simulation(species_id):
-    cmd = f"-q {species_id} -L 1 --seed 1234 10"
+    # L must be longer than any gene conversion tract lengths
+    L = 1000
+    species = stdpopsim.get_species(species_id)
+    for chrom in species.genome.chromosomes:
+        if chrom.gene_conversion_length is not None:
```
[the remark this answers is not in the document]
petrelharp: Since the GC length for StrAga is 120Kb, I had to make this change or run this test with L > 120000 for all species, which seemed like it might take a while? I'll put an explanatory comment.

stdpopsim/catalog/EscCol/species.py
```diff
@@ -32,8 +32,7 @@
             # Wielgoss et al. (2011) calculated for strain REL606,
             # from synonymous substitutions over 40,000 generations.
             mutation_rate=8.9e-11,
-            recombination_rate=0.0,
-            gene_conversion_rate=8.9e-11,
+            recombination_rate=8.9e-11,
```
igronau, line 35: There is currently no citation associated with recombination rate in E. Coli. From discussion with @fbaumdicker , I understand that this can be attributed to Wielgoss + [Didelot et al. (2012) ](https://pubmed.ncbi.nlm.nih.gov/22712577/)

stdpopsim/catalog/EscCol/species.py
```diff
@@ -32,8 +32,7 @@
...
             gene_conversion_length=345,
         )
```
igronau, line 37: There is also no citation associated with the average tract length of gene conversion. @fbaumdicker referred me again to [Didelot et al. (2012)](https://pubmed.ncbi.nlm.nih.gov/22712577/), which quotes an average of 542. This should probably be re-checked.

igronau (commented):
Looks good to me. @fbaumdicker - do you agree with the change of the average tract length to 542?

andrewkern:
these darn macos tests keep failing across different repos

petrelharp:
I think that @fbaumdicker is busy at the moment - assuming that last test passes, I'd like to merge this and open some issues to QC the GC.

fbaumdicker:
Yes, I was busy and yes I am fine with the change to 542.

PR #1356, opened 2022-09-09
Convert SLiM allele codes to nucleotides

### LINKED ISSUE #1334: should SLiM follow msprime in doing Jukes-Cantor mutations?
Since msprime 1.0, `sim_mutations( )` has by default produced JC mutations (and hence, nucelotide mutations). Questions:

1. Do we want to do JC for msprime mutations? (I think yes?)
2. Do we want to do JC for *SLiM* mutations? (I also think yes?)
3. Do we want to convert the output of SLiM simulations so that the ancestral and derived states are nucleotides rather than lists of SLiM IDs (which produce technically invalid and at least weird VCFs)? We'd do this with [`pyslim.convert_alleles( )`](https://tskit.dev/pyslim/docs/latest/python_api.html#pyslim.convert_alleles). (I'm not sure? maybe by default but with an optional parameter?) This would mean that the tree sequences that get output *can't* be loaded back into SLiM (unless there's an optional parameter that is toggled).

nspope:
Fixes #1334.

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -27,3 +27,4 @@ biopython
 demes==0.2.0
 demesdraw==0.2.0
 boto3
+kastore
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -18,7 +18,7 @@ attrs
 appdirs
 humanize
 pre-commit
-pyslim>=0.600
+pyslim>=1.0
 pandas
 numpy
 scikit-allel
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -18,7 +18,7 @@ attrs
 appdirs
 humanize
 pre-commit
-pyslim>=1.0
+pyslim>=1.0.1
 pandas
 numpy
 scikit-allel
--- a/tests/test_slim_engine.py
+++ b/tests/test_slim_engine.py
@@ -27,8 +27,9 @@ def count_mut_types(ts):
     mut_info = {}
     for mut in ts.mutations():
         for j, md in zip(mut.derived_state.split(","), mut.metadata["mutation_list"]):
-            if j not in mut_info:
-                mut_info[int(j)] = md
+            uid = f"{str(mut.id)}_{j}"
+            if uid not in mut_info:
+                mut_info[uid] = md
     num_neutral = sum([mut_info[j]["selection_coeff"] == 0.0 for j in mut_info])
     return [num_neutral, abs(len(mut_info) - num_neutral)]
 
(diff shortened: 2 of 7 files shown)


nspope:
`pyslim.generate_nucleotides` is failing with `pyslim == 0.7.0` and SLiM v4,
```
>       self._slim_generation = ts.metadata['SLiM']['generation']
E       KeyError: 'generation'
```
so I'm upping to `pyslim>=1.0` in `requirements/development.txt`

petrelharp:
It's weird that pyslim version even worked before?

petrelharp:
I forget - did we decide not to do actual nucleotide models in the SLiM engine?

petrelharp:
I bumped the cache so it'll use the new pyslim here.

nspope:
This is ready to go @petrelharp; except I'm not sure what to bump the pyslim version to in `requirements.txt`. Locally, tests pass (quite briskly!) with the github HEAD of pyslim.

requirements/development.txt
```diff
@@ -18,7 +18,7 @@ attrs
 appdirs
 humanize
 pre-commit
-pyslim>=0.600
+pyslim>=1.0
```
petrelharp, line 21:
```suggestion
pyslim>=1.0.1
```

petrelharp: I mean, it should work with 1.0 but slowly, so why not

tests/test_slim_engine.py
```diff
@@ -281,6 +282,51 @@ def test_assert_min_version(self):
...
+        ts = engine.simulate(
+            demographic_model=model,
+            contig=contig,
+            samples=samples,
+            slim_scaling_factor=10,
+            slim_burn_in=0,
+        )
+        for mut in ts.mutations():
+            alleles = mut.derived_state.split(",")
+            assert all([x in "AGCT" for x in alleles])
```
petrelharp, line 316:
```suggestion
            assert mut.derived_state in "ACGT"
```

tests/test_slim_engine.py
```diff
@@ -27,8 +27,9 @@ def count_mut_types(ts):
     mut_info = {}
     for mut in ts.mutations():
         for j, md in zip(mut.derived_state.split(","), mut.metadata["mutation_list"]):
```
petrelharp, line 29: I think that all alleles should now be single nucleotides, so I'm surprised this works?

petrelharp:
Hm, I see the tests are passing, but I am surprised. I think so that we don't have multiple SLiM mutations stacked in a single tskit mutations, we ought to:

- make all mutation types in the same `mutationStackGroup`, and
- set `mutationStackPolicy="l"`

... but, maybe this is not necessary for some reason?

tests/test_slim_engine.py
nspope: yeah, AFAICT none of these tests actually generate multiple alleles per site. I modified the existing tests as little as possible to try to avoid introducing stochastic failures-- this code works with single or stacked alleles, coded as integers or nucleotides.

petrelharp: In case it wasn't clear, what I mean is that we might now have `mut.derived_state` equal to `"A"` while `mut.metadata["mutation_list"]` a list of length greater than one. The `zip` won't fail but the test probably would.

nspope:
Hmm, I see what you mean. I agree, we should set `mutationStackPolicy="l"` to mimic SLiM's nucleotide models. Is there any reason we would want to make this toggle-able, to get the old behavior? Like, should setting `keep_mutation_ids_as_alleles=True` also set `mutationStackPolicy="s"`.

It seems to me that setting the stacking policy to "l" across the board is what we want, but maybe I'm overlooking something wrt initializing simulations from tree sequences?

tests/test_slim_engine.py
nspope: you're totally right -- it'll fail with stacked mutations after conversion to nucleotides. Setting the stacking policy like you suggest should do the trick.

nspope: ... or, enumerating `mutation_list` rather than zipping it with `derived_state` makes these tests pass regardless of whether the derived state is a stack or a single allele.

nspope:
Turns out that it isn't possible to set the stacking policy consistently, as `msprime.SLiMMutationModel` doesn't allow for a policy aside from 's' (and is used for recapitation/neutral DFEs). 

But, a minor modification makes the tests pass regardless -- so we'll sort out the stacking policy down the road.

nspope:
Strange -- after bumping pyslim to 1.0.1, the CI is failing because it can't import `kastore`.

andrewkern:
huh. i can replicate this error on my laptop by bumping the `pyslim` version to `1.0.1` in the `requirements.txt`. did something move on the `pyslim` dependencies that `pip` is reading from? will go look there

andrewkern:
looking at `pyslim` and comparing releases, I'm wondering if this change is the issue? 
image

@petrelharp can we try adding `kastore` back in as a `pip` dependency to see if that will cure this?

petrelharp:
I think something else must be going on, since everything passed on other PRs with that same requirements file?

nspope:
Thanks @andrewkern -- those screen captures are of requirements.txt for pyslim, not stdpopsim, right? Earlier commits on this PR would have used pyslim 1.0, and passed. So you are saying that because the most recent commit in this PR bumps pyslim to 1.0.1, and because dependencies for pyslim shifted from 1.0 to 1.0.1, this may be causing the kastore error -- do I have this right?

petrelharp:
We looked at this - yes, the thing to do is to add `kastore` as a pip testing requirement here, in stdpopsim.

nspope:
@petrelharp, are you OK with leaving the `SLiMMutationModel`-stacking loose end hanging for now? As it sounds like the solution will involve post-processing in pyslim. If so I'll go ahead and merge.

petrelharp:
Yes, sounds good to me.

Issue #1232, opened 2022-04-28
Separate demographic models from the stdpopsim python package


grahamgower:
We plan to transition stdpopsim's demographic models to the Demes YAML format (#962). This means that the demographic models will become programming-language agnostic, and thus useful outside of the scope of the stdpopsim Python package. So in principle, any simulator that supports Demes could make use of the stdpopsim catalog, without depending on the stdpopsim Python package (the CLI/API that one gets with "pip install stdpopsim"). In fact, such a simulator need not even be written in Python. However, making this easy for some external simulator would require us to reorganise the stdpopsim project a little.

I've opened this issue to see what the community thinks about such a reorganisation. We had some discussion of this in the context of Demes a while ago (https://github.com/popsim-consortium/demes-python/issues/203#issuecomment-813391254), and @apragsdale raised it again recently in (#962).

To summarise, the main reasons a Python simulator may not want to depend on the stdpopsim Python package just to get the catalog models:
* stdpopsim is licensed under the GPL (stdpopsim is closely tied to msprime, which itself is GPLed), which may be undesirable for the downstream simulator
* stdpopsim has lots of dependencies

It would be possible for us to split stdpopsim into two Python packages, say `stdpopsim` and `stdpopsim-catalog`, where the former depends on the latter. `stdpopsim-catalog` would consist of just the YAML files, and maybe a helper function or two to query them. This would have a liberal license and no dependencies. The files themselves would still live inside the stdpopsim git repository, from which both python packages would be built.

I'm not sure about how to package things for non-Python software, but perhaps @apragsdale or @molpopgen have suggestions in this regard. At the very least, a non-python project could include the stdpopsim git repository as a git submodule.

PS: @jeromekelleher has also suggested we might one day move to a YAML-based format for the species info in the catalog (chromosome lengths, mutation rates, recombination maps, etc.)---sorry I couldn't find where this was posted. Of course, this would also be useful outside the context of a stdpopsim CLI/API.

andrewkern:
i'm not sure about this idea but at least part of that is my unfamiliarity with the current `demes` spec. 

A large portion of what makes stdpopsim work is associating species specific parameters with demographic models, and then QCing those implementations-- how would a `stdpopsim-catalog` module achieve this? `demes` models aren't carrying around params like mutation and recombination rates right? 

[...]

PR #1361, opened 2022-09-16
Add ploidy as attribute of species; rework samples specification

### LINKED ISSUE #1111: adding ploidy as an attribute of species
**using values for ploidy unequal to 2**
*Currently the ploidy is set to 2 for all species, when simulating with msprime.
Instead, we could add a ploidy attribute (with default 2) to the species class. Ploidy could then be passed down to the simulation engine.*

**Rationale and considerations:**
*Currently the following simulation will lead to too many trees, as the ploidy of EscCol is 1 and this simulates 4 instead of 2 haploids*

```python
import stdpopsim

species = stdpopsim.get_species("EscCol")
engine = stdpopsim.get_engine("msprime")
contig = species.get_contig("Chromosome")
ts = engine.simulate(model, contig, samples)
```
*If contig would carry the ploidy information we could pass that information to msprime and avoid too much breaking of existing species by setting the default ploidy to 2.
In the long run, this might not only enable adding haploid species, but maybe also polyploid species.*

*Admittedly, a simple alternative would be to simply tell people to simulate half the number of samples for haploid species.*

This issue is linked to the discussions in #1106 #848 and #847

### LINKED ISSUE #1282: should `samples` be diploid?
*Edit:* the question here is whether to make "sample numbers" count numbers of diploids? This would be a substantial change, but arguably less confusing, especially as it's matching msprime and SLiM.

As noted over in #1281, the `samples` argument in msprime now produces by default diploid samples, although we produce individuals with `ploidy=1`, and so sp `write_vcf` now fails if given the `ploidy` argument.

nspope:
- `engine.simulate(samples=...)` now takes a dict of the form `{population : num_individuals}`. `engine.simulate` uses a new function `DemographicModel.get_sample_sets` (that isn't part of the public API) to combine species ploidy with the demographic model.

- The corresponding number of haploid samples (`tree_sequence.num_samples`) is `num_individuals * species.ploidy`

- The CLI for sample specification is now `population_name:num_samples`, like `HomSap -c chr1 -d OutOfAfrica_3G09 YRI:10 CEU:2` (same as for msprime)

- The old way of specifying samples still works (via `DemographicModel.get_samples` in python API, and via positional arguments in the CLI), but throws a `DeprecationWarning`. I opted not to rename the `samples=...` argument to `Engine.simulate` to keep the API less cluttered -- so it now takes either a dict (new behavior) or a list of `msprime.SampleSet` (old behavior).

- Fixes #1282. 

- Fixes #1111 (by adding ploidy as an attribute of species) but doesn't alter the SLiM engine to do anything different for haploids (e.g. it still uses diploid recombination model, gives warning if odd number of samples is requested, etc).

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -73,21 +73,23 @@ different output which shows options for performing the simulation itself and
 the species default parameters. This includes selecting the demographic model,
 chromosome, recombination map, and number of samples.
 
-The most basic simulation we can run is to simulate two (haploid) genomes
-- i.e., two samples -
+The most basic simulation we can run is to simulate a diploid genome
+- i.e., a single individual -
 using the species' defaults as seen in the species help (``stdpopsim HomSap --help``).
-These defaults include constant size population, a uniform recombination map based
+These defaults include constant size population named ``pop_0``, a uniform recombination map based
 on the average recombination rate (either genome-wide or within a chromosome, if
 specified), and the mutation rate shown above.
 To save time we will specify that the simulation use
 chromosome 22, using the ``-c`` option. We also specify that the resulting
 tree-sequence formatted output should be written to the file ``foo.ts`` with the
 ``-o`` option. For more information on how to use tree-sequence files see
-`tskit <https://tskit.dev/tskit/docs/stable/introduction.html>`__.
+`tskit <https://tskit.dev/tskit/docs/stable/introduction.html>`__. Finally, we
+specify that a single (diploid) sample should be simulated for population
+``pop_0`` using the syntax ``<population_name>:<number_of_samples>``.
 
 .. code-block:: console
 
-    $ stdpopsim HomSap -c chr22 -o foo.ts 2
+    $ stdpopsim HomSap -c chr22 -o foo.ts pop_0:1
 
 .. warning:: It's important to remember to either redirect the output of ``stdpopsim``
                 to file or to use the ``-o/--output`` option. If you do not, the
@@ -108,8 +110,8 @@ the two population out-of-Africa :ref:`model <sec_catalog_homsap_models_outofafr
 from `Tennesen et al. (2012) <https://doi.org/10.1126/science.1219240>`_ .
 By looking at the model help we find that the name for this model is
 ``OutOfAfrica_2T12`` and that we can specify it using
-the ``--demographic-model`` or ``-d`` option. We choose to draw two samples from the
-"African American" population and three samples from the "European American" population.
+the ``--demographic-model`` or ``-d`` option. We choose to draw two diploid sample from the
+``AFR`` ("African American") population and three diploids from the ``EUR`` ("European American") population.
 To increase simulation speed we can also choose to simulate a sequence that is
 a fraction of the length of the specified chromosome using the ``-l`` option
 (e.g. 5%). This is just specifying a sequence length, not actually selecting
@@ -118,14 +120,11 @@ other than a uniform recombination map. The command now looks like this:
 
 .. code-block:: console
 
-    $ stdpopsim HomSap -c chr22 -l 0.05 -o foo.ts -d OutOfAfrica_2T12 2 3
+    $ stdpopsim HomSap -c chr22 -l 0.05 -o foo.ts -d OutOfAfrica_2T12 AFR:2 EUR:3
 
-Note that the number of samples from each population are simply specified
-as two numbers at the end of the command. There must be *two* numbers because the
-model has two populations that we can sample from
-The order of those numbers is the same as the order
-specified in the model documentation. In this case, ``2 3`` means
-we are simulating two African American samples and three European American samples.
+Note that the number of samples from each population are simply specified as
+``<population_name>:<number_of_samples>`` at the end of the command. Omitted
+populations will have no samples in the resulting tree sequence.
 
 .. note::
     Many demographic models were inferred or calibrated using a mutation rate that
@@ -150,14 +149,14 @@ the ``-l`` option. (NOTE: this may a minute or so to run).
 
 .. code-block:: console
 
-    $ stdpopsim HomSap -g HapMapII_GRCh37 -c chr22 -o foo.ts -d OutOfAfrica_2T12 2 3
+    $ stdpopsim HomSap -g HapMapII_GRCh37 -c chr22 -o foo.ts -d OutOfAfrica_2T12 AFR:2 EUR:3
 
 For reproducibility we can also choose set the seed for the simulator using the
 ``-s`` flag.
 
 .. code-block:: console
 
-    $ stdpopsim HomSap -s 1046 -g HapMapII_GRCh37 -c chr22 -o foo.ts -d OutOfAfrica_2T12 2 3
+    $ stdpopsim HomSap -s 1046 -g HapMapII_GRCh37 -c chr22 -o foo.ts -d OutOfAfrica_2T12 AFR:2 EUR:3
 
 On running these commands, the CLI also outputs the relevant citations for both
 the simulator used and the resources used for simulation scenario.
@@ -180,15 +179,15 @@ with the ``tskit vcf`` command:
    $ tskit vcf foo.ts > foo.vcf
 
 For this small example (only five samples), the file sizes are similar,
-but the tree sequence is slightly larger
+but the tree sequence is slightly larger than the VCF
 (it does carry a good bit more information about the trees, after all).
-However, if we up the sample sizes to 2000 and 3000
+However, if we up the sample sizes to 1000 and 1500
 (the simulation is still pretty quick)
 the tree sequence is twenty-three times smaller:
 
 .. code-block:: console
 
-   $ stdpopsim HomSap -s 1046 -g HapMapII_GRCh37 -c chr22 -o foo.ts -d OutOfAfrica_2T12 2000 3000
+   $ stdpopsim HomSap -s 1046 -g HapMapII_GRCh37 -c chr22 -o foo.ts -d OutOfAfrica_2T12 AFR:1000 EUR:1500
    $ tskit vcf foo.ts > foo.vcf
    $ ls -lth foo.*
    -rw-r--r-- 1 peter peter 3139M Apr  3 10:40 foo.vcf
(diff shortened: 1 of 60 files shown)


grahamgower:
Shouldn't this be a property of the chromosome rather than the species? I'm thinking of Mt, chrY, and plasmids.

nspope:
Yeah @grahamgower, that makes sense to me. So, store per-chromosome ploidy data as a dict in `catalog/*/species.py`, as is done for per-chromosome mean recombination rate?

nspope:
Hmm, when simulating a generic contig for a species, I think we'd want it to use a default per-species ploidy -- which we wouldn't have if ploidy is strictly a property of chromosome. So, should it be a property of both species and chromosome?

grahamgower:
Ah yes, good point! Well the species property as you have it in this PR looks good. How about having the chromosome property be an optional keyword arg of `stdpopsim.Chromosome()`, which is populated from info in the various genome_data.py:data dicts like the chromosome synonyms? The chromosome ploidy may then be omitted in most cases---maybe the maintenance scripts for generating species can detect "Mt" and "chrY" and set the ploidy=1 on these.

tests/test_dfes.py
```diff
@@ -573,13 +573,14 @@ def test_no_msprime_dfe(self):
...
             mutation_rate=1e-6,
+            ploidy=2,
         )
         contig.add_dfe(
             intervals=np.array([[0, contig.length / 2]], dtype="int"),
             DFE=d,
         )
         model = stdpopsim.PiecewiseConstantSize(1000)
-        samples = model.get_samples(2)
+        samples = {"pop0": 1}
```
grahamgower, line 583: Do we have a mix of conventions for population names: both `pop0` and `pop_0`? I know msprime introduces `pop_j` into population metadata for the j'th population, and SLiM uses `pj` for the j'th. Adding a third convention would be a shame!

nspope: Oh good catch, it should be `pop_0` here. The error that this test checks gets raised before the misnomer gets checked.

nspope: If the generic demographic models API gets reworked at some point (like if we transition to YAML as in #1232), we could make the naming consistent with SLiM.

stdpopsim/catalog/HomSap/species.py
```diff
@@ -76,6 +82,7 @@
             synonyms=data["synonyms"],
             mutation_rate=1.29e-8,
             recombination_rate=_recombination_rate_data[name],
+            ploidy=_ploidy_data[name] if name in _ploidy_data.keys() else None,
```
nspope, line 85: @grahamgower, is this what you mean? If ploidy is not the default (for a species) then set `Chromosome.ploidy` from a dict defined in `catalog/*/species.py` -- and when making a contig use `Species.ploidy` if `Chromosome.ploidy` is `None`.

It'd be nice to have chromosome ploidy automatically generated by the maintenance scripts, but I'm not sure that's available through the REST API.

nspope (commented):
@grahamgower -- makes sense, although I'm not sure about automated retrieval of ploidy per chromosome ... maybe we can punt that down the road for now. Is this mockup for HomSap roughly along the lines of what you're thinking?

stdpopsim/catalog/HomSap/species.py
nspope: ... and per-species QC like,
https://github.com/popsim-consortium/stdpopsim/blob/b63a11a2b6841fd789f8152ea22db8deceaeb536/tests/test_HomSap.py#L41-L48

grahamgower, line 85:
```suggestion
            ploidy=_ploidy_data.get(name),
```

grahamgower: Sorry, I wasn't suggesting to get the ploidy info from ensembl. I was only thinking of doing some dumb name matching magic in the maintenance script---but what you're doing here is just fine.

grahamgower:
Regarding the sampling syntax for the CLI... This PR uses `<name>=<n>`, whereas msprime's `msp ancestry` command uses `<name>:<n>`. Perhaps stdpopsim should follow msprime here? I don't recall the argument for using a colon instead of equals, but maybe @jeromekelleher can enlighten us? 
https://tskit.dev/msprime/docs/stable/cli.html#Positional%20Arguments

jeromekelleher:
This is getting to be a big PR - I wonder if we could decouple the ploidy changes from the sample specification bits to make it easier to review? No worries if they're too intertangled now.

nspope:
Yeah sorry @jeromekelleher -- it's a bit intertwined now especially in the tests, where `ts.num_samples` is frequently checked and depends on both ploidy and number of individuals.

jeromekelleher:
Fair enough - can you ping when you'd like a review please @nspope?

nspope:
Hey @andrewkern, any idea why we're getting period URL timeouts in the mac OS tests? I've been noticing this across different PRs over the past couple days, and often I have to wait a few hours before rerunning to get it through.

nspope:
This is ready for review @jeromekelleher and/or @grahamgower. A couple things,
- I haven't added anything regarding ploidy to the development docs-- will do so in a separate PR
- Sample specification in the CLI with the default constant demography seems kinda unfortunate, e.g. `stdpopsim HomSap pop_0:1` -- I wish we could omit the `pop_0:` but then it's not distinguishable from the positional, haploid sample specification (which is retained with a deprecation warning).

grahamgower:
> Hey @andrewkern, any idea why we're getting periodic URL timeouts in the mac OS tests? I've been noticing this across different PRs over the past couple days, and sometimes it takes a few hours before a rerun will get it through.

The macOS GHA runners just seem to have problems. One can find many reports about slowness (cpu, disk, and/or network). E.g. https://github.com/actions/runner-images/issues/4896

stdpopsim/models.py
```diff
@@ -180,36 +181,92 @@ def register_qc(self, qc_model):
...
                         population=pop_index,
                         time=self.populations[pop_index].default_sampling_time,
-                        ploidy=1,  # Avoid breaking too much at once.
+                        ploidy=1,
                     )
                 )
             elif n > 0:
                 raise ValueError(
-                    "Samples requested from non-sampling population {pop_index}"
+                    "Samples requested from non-sampling population {pop_name}"
```
grahamgower, line 269:
```suggestion
                    f"Samples requested from non-sampling population {pop_name}"
```

stdpopsim/slim_engine.py
```diff
@@ -1363,24 +1363,28 @@ def matrix2str(
...
-            gen = time / demographic_model.generation_time
-            warnings.warn(
-                stdpopsim.SLiMOddSampleWarning(
-                    f"SLiM simulates diploid individuals, so {n_inds} "
-                    f"individuals will be sampled for the {count} haploids "
-                    f"requested from population {pop.name} at time {gen}. "
-                    "See #464."
+        if sample_set.ploidy == 1:
+            # SLiM can only sample individuals, which we assume are diploid.
+            n_inds = (count + 1) // 2
```
grahamgower, line 1371: I guess this results in a tree sequence where pairs of haploid individuals are grouped to look like single individuals? Can this be fixed by post-processing?

grahamgower: Maybe just open an issue for this instead of tackling it here.

docs/tutorial.rst
```diff
@@ -1588,11 +1581,13 @@ and that rates of migration since the split between the populations are both zer
...
     )
 
-We'll simulate 10 chromosomes from each of the populations using the ``msprime`` engine.
+We'll simulate 5 diploids from each of the populations using the
+``msprime`` engine (the populations in this generic model are named
+``pop1`` and ``pop2``).
 
 .. code-block:: python
 
-    samples = model.get_samples(10, 10)
```
jeromekelleher, line 1595: Is this right? Wouldn't the old code be sampling pop_0 and pop_1

stdpopsim/models.py
```diff
@@ -180,36 +181,92 @@ def register_qc(self, qc_model):
...
+        :param dict individuals_per_population: A dict of the form
+            `{population_name:number_of_samples}` giving sample counts per
+            population.
+        :param int ploidy: The ploidy of the samples.
+
+        .. note: This interface is intended for internal use only.
+        """
+
+        sample_sets = []
+        population_indices = {x.name: i for i, x in enumerate(self.populations)}
```
jeromekelleher, line 203: Slightly simpler to do this I think:
```python
pop_map = {pop.name: pop for pop in self.populations}
for name, num_samples in individuals_per_population.items():
      if name not in pop_map: 
                ....
      pop = pop_map[name]
      if pop.allow_samples:
           ...
```
You can use the ``name`` as the population argument for SampleSet.

stdpopsim/utils.py
```diff
@@ -268,3 +268,43 @@ def clip_and_shift_intervals(intervals, left, right):
...
     intervals -= left
     return intervals
+
+
+def parse_population_sample_pairs(pop_sample_pairs):
+    """
+    Parse a list of strings of the form "population_name:integer" into dict
+    {population_name:integer}.
+
+    FIXME: For back-compatibility, return a list of integers if the
```
jeromekelleher, line 278: This isn't really a FIXME, more something that we probably want to do at some point but not quite sure when. 
```suggestion
    For back-compatibility, return a list of integers if the
```

jeromekelleher (approved):
Looks great, this is fantastic work and a real step forward!

Some minor suggestions above.

docs/tutorial.rst
nspope: Nope, looks like the names in the `IsolationWithMigration` model are `pop1`, `pop2`, and `ancestral`.

stdpopsim/models.py
nspope: It turns out that other parts of the code expect integers for the population attribute of SampleSet -- so I'm keeping the use of population indices rather than population names.

stdpopsim/slim_engine.py
nspope: Good point! Mentioned in #1371.

stdpopsim/models.py
```diff
@@ -5,6 +5,11 @@
 import textwrap
 
 import msprime
+import warnings
+
+# DeprecationWarning is filtered by default. The following ensures that these
+# warnings are visible to the user.
+warnings.simplefilter("always", DeprecationWarning)
```
nspope, line 12: Is this a bad idea? Without changing warning filtering settings, the deprecation warning for `model.get_samples` will be filtered by default and not visible to the user. On the other hand, DeprecationWarnings from other packages that are intended for developers will show up. Maybe it would be better to define a custom deprecation warning in `stdpopsim/warning_categories.py`?

Do you have a suggestion for this @jeromekelleher?

jeromekelleher: I'm not sure - @grahamgower has a better handle on how we're using warnings here I think?

grahamgower: Hmm, sorry, I actually don't know how to do sane deprecation warnings in Python. The only thing I do know is that the facilities provided in Python are confusing and that folks complain that warnings are either too noisy and thus ignored, or alternately never reach the right people at the right time.

nspope: ok thanks for the input -- I'll make a `DeprecatedFeatureWarning` that inherits from `UserWarning` into stdpopsim/warning_categories.py ... will be visible to user and won't alter warning behavior from other modules.

PR #1370, opened 2022-09-23
Qc bonobo ghost

kuhlwilm:


kuhlwilm:
@petrelharp 
I think I'm done with the QC - the original and QC model are equal (after some debugging of my QC model and one correction to the original model). One of the checks was not successful, but I'm not sure I can do anything there.
Best,
Martin

petrelharp:
Wow, this is a complex one! Thanks a bunch! And - all of the checks seem to have passed, happily. 🎉 !!

Issue #1349, opened 2022-09-05
deprecate `--length-multiplier` CLI option


grahamgower:
`--length-multiplier` doesn't compose well with other options, and specifying a fixed length is an easier way to get short-running tests.

grahamgower:
Also deprecate/remove the `length_multiplier` options to `Species.get_contig()` and `Contig.species_contig()`. No one should be using these outside our test suite.

nspope:
Length multiplier via `-l` is used pretty heavily in the CLI portion of the tutorial, these examples should use `--left` and `--right` once #1348 is merged.

PR #1391, opened 2022-10-05
Update tutorial.rst for 0.2.0

### LINKED ISSUE #1368: Make examples in tutorial use 'HapMapII_GRCh38' instead of 'HapMapII_GRCh37'
So that these introductory examples don't produce warnings about a mismatch between assembly and genetic map.

### LINKED ISSUE #1386: Provide population names for various demographic models in CLI
We're now specifying population-sample pairs via the CLI syntax `<population_name>:<num_individuals>`. However, population names are not shown via `--help-models`. e.g.,
```
$ stdpopsim HomSap --help-models

All simulation models for Homo sapiens

OutOfAfricaExtendedNeandertalAdmixturePulse_3I21: Three population out-of-Africa with an extended pulse of         Neandertal admixture into Europeans
     Demographic model of an extended admixture pulse from Neandertals
    into Europeans taken from Iasi et al. (2021), specifically the
    simple model of Supplementary Figure 1a with a gamma-shaped pulse.
    This model simulates 3 populations: Africans, Europeans and
    Neandertals with an Out-of-Africa event. The population sizes are
    constant with an unidirectional admixture from Neandertals into
    Europeans after the split between Europeans and Africans. The
    admixture event is modelled as an 800 generation (20 ky) long
    extended admixture pulse.

    Populations:
        0: 1000 Genomes YRI (Yoruba)
        1: 1000 Genomes CEU (Utah Residents (CEPH) with Northern and Western European Ancestry
        2: Neandertals
...
```
instead of integer indices (`0:<whatever>`) we should give the names needed to specify samples.

nspope:
Fixes #1368, partially addresses #1349.

- Modify code blocks in tutorial so that they run/are consistent with text.
- Remove any use of `length_multiplier`, replace with `left`, `right`, or `length`.
- Redo figures to match code blocks.
- Some minor rewording for clarity.
- Fix broken links to tskit.dev subpages


docs/tutorial.rst
```diff
@@ -934,11 +959,11 @@ Next is "recapitation",
 for which the rationale and method is described in detail in the
 `pyslim documentation <https://tskit.dev/pyslim/docs/latest/tutorial.html#recapitation>`__.
 The third (and least crucial) step is to *simplify* the tree sequence.
-If as above we ask for 200 samples from a population whose final size is
+If as above we ask for 200 haploids from a population whose final size is
```
petrelharp, line 962:
```suggestion
If as above we ask for 100 sampled individuals from a population whose final size is
```
The command above has `100`, so saying "200 haploids" is confusing for no reason, I think?

docs/tutorial.rst
```diff
@@ -1092,37 +1126,32 @@ We'll count up the number of neutral and deleterious mutations in the three regi
...
-        )
-        unique_muts = {}
+        region = np.digitize(site.position, gene_interval.flatten())
         for mut in site.mutations:
-            for ds, md in zip(mut.derived_state.split(","), mut.metadata["mutation_list"]):
-                if ds not in unique_muts:
-                    unique_muts[ds] = md["selection_coeff"]
+            for mut_info in mut.metadata["mutation_list"]:
+                if mut_info["selection_coeff"] == 0:
+                    num_neutral[region] += 1
```
petrelharp, line 1135: This is not quite right because a neutral mutation that has another mutation stacked on top of it later will be counted twice (that's the point of the `unique_muts` thing). Instead, my proposal would be to do like
```
mut_s = sum([mut_info["selection_coeff"] for mut_info in mut.metadata["mutation_list"]])
if mut_s == 0:
    num_neutral[region] += 1
```
Oh, wait, this doesn't work either, since a neutral mutation placed on top of a non-neutral one wouldn't be counted. Hm...

nspope: yeah, you're totally right -- the previous version of this code block collapses SLiM mutation ids that occur across multiple `tskit.Mutation`s. So, we'll need to use/explain the `keep_mutation_ids_as_alleles` argument. It'd be great to end up with something a bit less cumbersome than what was in the previous version.

jeromekelleher (approved):
Had a skim through, LGTM

docs/tutorial.rst
nspope: hm, I probably borked a test back in #1356. It's still valid -- just checking for the presence of non-neutral mutations doesn't require that they're correctly collapsed -- but probably not good to have a misleading example floating around in the tests.

docs/tutorial.rst
```diff
@@ -1084,45 +1130,41 @@ are removed from the intervals associated with previous DFEs.
...
-            else:
-                num_del[j] += 1
-
-    for j, (n, d) in enumerate(zip(num_neutral, num_del)):
+            mut_ids = mut.derived_state.split(",")
+            for id, md in zip(mut_ids, mut.metadata["mutation_list"]):
+                selection_coeffs[region][id] = md["selection_coeff"]
+
+    for region, coeffs in enumerate(selection_coeffs):
+        num_neutral = sum([x == 0 for x in coeffs.values()])
```
nspope, line 1152: I think this should now work, @petrelharp

docs/tutorial.rst
petrelharp: How about
```
mut_s = sum([mut_info["selection_coeff"] for mut_info in mut.metadata["mutation_list"]])
if mut.parent != tskit.NULL:
    parent_mut = ts.mutation(mut.parent)
    parent_s = sum([mut_info["selection_coeff"] for mut_info in parent_mut.metadata["mutation_list"]])
    mut_s -= parent_s
if mut_s == 0:
    num_neutral[region] += 1
```
So, if the net selection coefficient of a mutation is the same as it's parent mutation (if there is one), then it's neutral. This will work in all cases except weird ones that couldn't happen in a stdpopsim simulation.

docs/tutorial.rst
```diff
@@ -1092,37 +1126,32 @@ We'll count up the number of neutral and deleterious mutations in the three regi
...
         for mut in site.mutations:
-            for ds, md in zip(mut.derived_state.split(","), mut.metadata["mutation_list"]):
-                if ds not in unique_muts:
-                    unique_muts[ds] = md["selection_coeff"]
+            for mut_info in mut.metadata["mutation_list"]:
+                if mut_info["selection_coeff"] == 0:
+                    num_neutral[region] += 1
                 else:
-                    assert unique_muts[ds] == md["selection_coeff"]
-        for s in unique_muts.values():
```
petrelharp, line 1107: note that previously we were only summing over `unique_muts`, so we weren't double-counting; if we got the same answer previously it means we don't have any mutations happening to stack.

docs/tutorial.rst
nspope: This seems fine -- but I pushed an example using `keep_mutation_ids_as_alleles` a short while ago that gives the right answer when mutations are stacked, and may be a bit clearer. Do you have a preference? (either seem fine to me)

docs/tutorial.rst
nspope: yup, got it!

docs/tutorial.rst
petrelharp: Ah, sorry, saw it now. I vote to not put `keep_mutation_ids_as_alleles` in the docs, since people will do it, and then get bad VCFs. This solution works and is not too bad, I think?

docs/tutorial.rst
```diff
@@ -1393,22 +1431,25 @@ are available:
...
+The next command simulates 2 diploid samples of chromosome 1 from each of the four
 populations, and saves the output to a file called ``afr-america-chr1.trees``.
 For the purposes of this tutorial, we'll also specify a random seed using the
 ``-s`` option.
 To check that we have set up the simulation correctly, we may first wish to perform a
 dry run using the ``-D`` option.
 This will print information about the simulation to the terminal:
 
-.. command-output:: stdpopsim HomSap -c chr1 -o afr-america-chr1.trees -s 13 -g HapMapII_GRCh37 -d AmericanAdmixture_4B11 4 4 4 4 -D
+.. command-output:: stdpopsim HomSap -c chr1 -o afr-america-chr1.trees -s 13 -g HapMapII_GRCh38 -d AmericanAdmixture_4B11 AFR:2 EUR:2 ASIA:2 ADMIX:2 -D
```
petrelharp, line 1443: oh man this sample specification is so much better

petrelharp (approved):
This is *great*. Just that one issue with counting selected mutations yet.

docs/tutorial.rst
nspope: sure, sounds good to me.

nspope: After chatting, we think it'd be a good idea to functionalize this and put it in `ext/selection.py` (i.e. a function that takes a `Mutation` and a `TreeSequence` and returns the selection coefficient).

stdpopsim/ext/selection.py
```diff
@@ -354,3 +355,32 @@ def selective_sweep(
...
     return extended_events
+
+
+def selection_coeff_from_mutation(ts, mutation):
+    """
+    Extract the selection coefficient from a (possibly stacked) mutation.
+
+    :param ts: A ``tskit.TreeSequence`` containing the mutation.
+    :param mutation: A ``tskit.Mutation`` for which to extract the selection coefficient.
+    """
```
nspope, line 366: Do you think we need more explication here @petrelharp?

Looks like the OSX CI environments are down so I'll have to rerun the coverage checks later.

nspope:
Do you see any issues with the new `ext.selection_coeff_from_mutation`, @petrelharp? If not, I'll go ahead and merge.

stdpopsim/ext/selection.py
```diff
@@ -354,3 +355,32 @@ def selective_sweep(
         )
 
     return extended_events
+
+
+def selection_coeff_from_mutation(ts, mutation):
+    """
+    Extract the selection coefficient from a (possibly stacked) mutation.
```
petrelharp, line 362: How's this?
```suggestion
    Returns the selection coefficient of the given mutation.
    
    In nearly all cases, this will just be "the selection coefficient" and users do not need to worry about
    the details. However: mutations produced by the SLiM engine have metadata which includes each mutation's
    selection coefficient. However, mutations may "stack", meaning that one mutation-in-the-tree-sequence
    may in fact represent a superposition of several SLiM-produced-mutations. This method adds up
    all selection coefficients in the metadata of this mutation, and subtracts the sum of the selection
    coefficients of the parent mutation, if there is one. For more information, see the SLiM manual and
    `the pyslim manual <pyslim:sec_extracting-information-about-selected-mutations>`.
```
(I'm sure I"m not getting the intersphinx notation right there in that last line...)

petrelharp (approved):
I wrote some more explanation for that method; if it looks good to you I say we merge this!

PR #1407, opened 2022-10-19
Rescale haploid population size in SLiM engine

### LINKED ISSUE #1406: SLiM and msprime engines handle population sizes differently for haploid species
In #1361, ploidy was added as an attribute of species, but the SLiM engine was left untouched (aside from setting the number of sampled individuals appropriately). However, this means that population size is treated differently between the msprime and SLiM engines for haploid species:
```python
import stdpopsim
import numpy as np

contig = stdpopsim.Contig.basic_contig(length=1, ploidy=1)
model = stdpopsim.PiecewiseConstantSize(1000)
reps = 100 

engine = stdpopsim.get_engine("msprime")
tmrca_msp = []
for _ in range(reps):
    ts_msp = engine.simulate(model, contig, {"pop_0":2})
    tmrca_msp.append(ts_msp.first().tmrca(0,1))
print(f"pair TMRCA (msprime): {np.mean(tmrca_msp)} +/- {np.std(tmrca_msp)/np.sqrt(reps)}")

engine = stdpopsim.get_engine("slim")
tmrca_slim = []
for _ in range(reps):
    ts_slim = engine.simulate(model, contig, {"pop_0":2})
    tmrca_slim.append(ts_slim.first().tmrca(0,1))
print(f"pair TMRCA (slim): {np.mean(tmrca_slim)} +/- {np.std(tmrca_slim)/np.sqrt(reps)}")      

# pair TMRCA (msprime): 1000.116049495669 +/- 93.04547569055264
# pair TMRCA (slim): 2274.7033717826785 +/- 192.0079953821991
```
The solution should be to rescale the demographic model's population size inside `slim_makescript` if the contig is haploid. But, this raises the question, are the population sizes for EscCol / StrAga in the catalog actually # haploids, or are they off by a factor of 2?

nspope:
Fixes #1406.

Should probably add a test for equivalence between msprime and SLiM engines when `contig.ploidy != 2`.

### DIFF
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -1761,6 +1761,9 @@ class _SLiMEngine(stdpopsim.Engine):
             migration_matrix=recap_epoch.migration_matrix,
         )
 
+        # `recap_epoch` contains population sizes from the demographic model,
+        # that are the number of individuals regardless of ploidy. Thus,
+        # ploidy must be set here, as is done in `_MsprimeEngine.simulate`
         ts = msprime.sim_ancestry(
             initial_state=ts,
             demography=demography,
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -1842,26 +1842,7 @@ class _SLiMEngine(stdpopsim.Engine):
 
         # If haploid, rebuild individual table so each sample is an individual
         if contig.ploidy == 1:
-            tables = ts.dump_tables()
-            node_indiv = tables.nodes.individual
-            haploid_indiv = tskit.IndividualTable()
-            haploid_indiv.metadata_schema = tables.individuals.metadata_schema
-            for new_idx, node in enumerate(ts.samples()):
-                old_idx = node_indiv[node]
-                haploid_indiv.add_row(
-                    flags=tables.individuals[old_idx].flags,
-                    location=tables.individuals[old_idx].location,
-                    parents=tables.individuals[old_idx].parents,
-                    metadata=tables.individuals[old_idx].metadata,
-                )
-                assert (
-                    haploid_indiv[new_idx].metadata
-                    == tables.individuals[old_idx].metadata
-                )
-                node_indiv[node] = new_idx
-            tables.nodes.individual = node_indiv
-            tables.individuals.replace_with(haploid_indiv)
-            ts = tables.tree_sequence()
+            ts = stdpopsim.utils.haploidize_individuals(ts)
 
         return ts
 
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -1765,6 +1765,7 @@ class _SLiMEngine(stdpopsim.Engine):
             initial_state=ts,
             demography=demography,
             recombination_rate=contig.recombination_map,
+            ploidy=contig.ploidy,
             random_seed=s0,
         )
         ts = self._simplify_remembered(ts)
--- a/stdpopsim/slim_engine.py
+++ b/stdpopsim/slim_engine.py
@@ -1839,6 +1839,41 @@ class _SLiMEngine(stdpopsim.Engine):
                 )
             )
 
+        # If haploid, rebuild individual table so each sample is an individual
+        if contig.ploidy == 1:
+
+            def to_ragged(x, idx):
+                return [x[idx[i] : idx[i + 1]] for i in range(len(idx) - 1)]
+
+            tables = ts.dump_tables()
+            indiv = tables.individuals.asdict()
+            indiv["flags"] = np.repeat(indiv["flags"], 2)
+            for field in ["location", "parents", "metadata"]:
+                offset = indiv[field + "_offset"]
+                ragged = to_ragged(indiv[field], offset)
+                indiv[field] = np.array([x for x in ragged for _ in range(2)]).flatten()
+                offset_length = np.insert(np.repeat(np.diff(offset), 2), 0, 0)
+                indiv[field + "_offset"] = np.cumsum(offset_length)
+
+            tables.individuals.set_columns(
+                flags=indiv["flags"],
+                location=indiv["location"],
+                location_offset=indiv["location_offset"],
+                parents=indiv["parents"],
+                parents_offset=indiv["parents_offset"],
+                metadata=indiv["metadata"],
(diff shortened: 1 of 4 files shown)


nspope:
Hm -- I think we also need `ploidy=contig.ploidy` here:
https://github.com/popsim-consortium/stdpopsim/blob/51f5a7cf324c6618ee20c192b4b681f0aa41bc63/stdpopsim/slim_engine.py#L1762-L1767

nspope:
I think this is good to go. Looks like the recommended way to model haploids in SLiM with tree sequence recording is via a non-WF model, but that's too much to attempt at the moment.

grahamgower:
This looks good to me in terms of returning the correct number of haploid chromosomes, but I am completely ignorant of how to do non-diploid sims in SLiM. How does the recapitation work here? I.e. isn't there a mismatch with the ploidy of individuals in the tree sequence from SLiM (always diploid I think?) and the ploidy msprime uses? It seems msprime is agnostic to this for recapitation purposes? In any event, should the individual table be rebuilt so that the user gets back a tree sequence with sampled individuals that have the expected ploidy?

stdpopsim/slim_engine.py
```diff
@@ -900,7 +900,9 @@ def fix_time(event):
     growth_rates = np.empty(shape=(dd.num_populations, len(epochs)), dtype=float)
     for j, epoch in enumerate(epochs):
         for i, pop in enumerate(epoch.populations):
-            N[i, j] = int(pop.end_size)
+            # SLiM simulates a diploid population, so rescale population size
+            # depending on contig ploidy.
+            N[i, j] = int(pop.end_size * contig.ploidy / 2)
```
petrelharp, line 905: I agree with this, although we might want to leave a note saying that this should be changed if/when we get SLiM to do haploids correctly (in an appropriate issue).

stdpopsim/slim_engine.py
```diff
@@ -1759,10 +1761,14 @@ def _recap_and_rescale(
             migration_matrix=recap_epoch.migration_matrix,
         )
 
+        # `recap_epoch` contains population sizes from the demographic model,
+        # that are the number of individuals regardless of ploidy. Thus,
+        # ploidy must be set here, as is done in `_MsprimeEngine.simulate`
```
petrelharp, line 1766: I agree

stdpopsim/slim_engine.py
```diff
@@ -1833,6 +1839,41 @@ def _get_msp_rate_map(breaks, is_this_dfe, rate):
                 )
             )
 
+        # If haploid, rebuild individual table so each sample is an individual
+        if contig.ploidy == 1:
+
+            def to_ragged(x, idx):
+                return [x[idx[i] : idx[i + 1]] for i in range(len(idx) - 1)]
```
petrelharp, line 1846: Hm, isn't this related to `tskit.unpack_strings( )` (see [here](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.unpack_bytes))? I guess it's not the same thing, and you probably tried that.

stdpopsim/slim_engine.py
```diff
@@ -1833,6 +1839,41 @@ def _get_msp_rate_map(breaks, is_this_dfe, rate):
...
+                location=indiv["location"],
+                location_offset=indiv["location_offset"],
+                parents=indiv["parents"],
+                parents_offset=indiv["parents_offset"],
+                metadata=indiv["metadata"],
+                metadata_offset=indiv["metadata_offset"],
+                metadata_schema=indiv["metadata_schema"],
+            )
+
+            samples = tables.nodes.flags == tskit.NODE_IS_SAMPLE
```
petrelharp, line 1869: In case there are other flags in the future, it'd be better to do a bitwise or here. (I think this is
```
samples = ( tables.nodes.flags | tskit.NODE_IS_SAMPLE > 0)
```
... ?

stdpopsim/slim_engine.py
```diff
@@ -1833,6 +1839,41 @@ def _get_msp_rate_map(breaks, is_this_dfe, rate):
                 )
             )
 
+        # If haploid, rebuild individual table so each sample is an individual
```
petrelharp, line 1842: This *looks* right, but a less error-prone way to do it would involve lots of calls to `tables.add_row( )`; did you do it this way because it's faster?

petrelharp:
This looks good, but I'm a bit nervous about the rebuilding the individual table thing, especially as we don't have a good way to test that's been done right, I think.

nspope:
I think we do want to rebuild the individual table, though -- otherwise VCF output will look diploid. As you suggest, doing this row-by-row would be safer than tiling those ragged arrays. At the least, I can add an assert checking that the individual metadata matches between diploid and haploid-ized tables.

stdpopsim/slim_engine.py
nspope: agreed, and done

stdpopsim/slim_engine.py
```diff
@@ -1841,37 +1842,25 @@ def _get_msp_rate_map(breaks, is_this_dfe, rate):
...
-
-            samples = tables.nodes.flags == tskit.NODE_IS_SAMPLE
-            tables.nodes.individual = np.where(
-                samples,
-                np.arange(tables.nodes.num_rows, dtype=np.int32),
-                tables.nodes.individual,
-            )
+            node_indiv = tables.nodes.individual
+            haploid_indiv = tskit.IndividualTable()
+            haploid_indiv.metadata_schema = tables.individuals.metadata_schema
```
jeromekelleher, line 1848: Here's a slightly cleaner way:
```python
old_inds = tables.individuals.copy()
tables.individuals.clear() # This keeps the metadata schema intact
# Make sure we don't refer to any individuals that are not samples
node_invid = np.zeros_like(ts.nodes_individual) - 1
for new_idx, node in enumerate(ts.samples()):
     old_idx = ts.nodes_individual[node]
     tables.individuals.append(old_inds[old_idx])
tables.nodes.individual = node_indiv
```

jeromekelleher (approved):
LGTM - some suggested refactoring for individual table rewriting

stdpopsim/slim_engine.py
petrelharp: (a private function)

nspope:
Thanks @jeromekelleher -- shall I go ahead and merge this in?

PR #1416, opened 2022-11-07
Don't use scipy and pandas for OutOfAfricaExtendedNeandertalAdmixture…

grahamgower:
…Pulse_3I21

* Scipy was being used for the Gamma PDF, which has been reimplemented using the math.gamma() function.
* Pandas use here was unnecessary, so has been removed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -12,7 +12,6 @@ channels:
         - bioconda
 dependencies:
         - numpy
-        - pandas
         - msprime
         - scikit-allel
         - matplotlib
--- a/maintenance/annotation_maint.py
+++ b/maintenance/annotation_maint.py
@@ -9,7 +9,7 @@ import tarfile
 import stdpopsim
 
 logger = logging.getLogger(__name__)
-# make root directory for zarr annotations
+# make root directory for annotations
 annot_path = "annotations"
 os.makedirs(annot_path, exist_ok=True)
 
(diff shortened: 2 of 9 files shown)


jeromekelleher:
Thanks @grahamgower - agreed with all the above!

petrelharp:
Me too - seems tricky!

jeromekelleher:
We should totally not be importing the qc models at runtime unless specifically requested - would be happy to help with workarounds to get them out of the ``import stdpopsim`` code flow

grahamgower:
> We should totally not be importing the qc models at runtime unless specifically requested - would be happy to help with workarounds to get them out of the `import stdpopsim` code flow

I think I'd rather have that as a separate PR, and I've opened #1418 to discuss/track.

Maybe for this PR, you'd be happy to look at the CI runs and satisfy yourself that I haven't changed the demographic model (the QC test still passes)? I can then push another commit making the analogous change to remove scipy/pandas from the QC model.

grahamgower:
One additional point: `time python -c 'import stdpopsim'` takes about 1 second; after removing pandas/scipy from the import path, this drops to 500 ms. Loading the QC models (or not) makes negligible difference to import time once pandas/scipy is removed.

grahamgower:
This is good to go now @jeromekelleher.

jeromekelleher (approved):
LGTM

PR #1419, opened 2022-11-09
Only load QC models when running tests

### LINKED ISSUE #1418: load QC models only when the tests are run
Right now we load all QC models from `stdpopsim/__init__.py`, which means `import stdpopsim` loads all the QC models. This is (almost) unnecessary. Only the test suite makes meaningful use of the QC models---to confirm the QC model matches the main model. However, the CLI also looks at the `DemographicModel.qc_model` field to see if a model has been QCed and prints a warning if a user tries to simulate a model that hasn't been QCed. Right now, all our demographic models have been QCed (hooray!), so this warning doesn't apply.

Options:
1. Keep the status quo - loading all QC models at runtime.
2. Remove the warning about non-QCed models from the CLI.
3. Seprate the "this model has been QCed" logic from the QC model itself. This could be done by changing the register_qc_model() method to accept a function that returns the qc model, rather than the model object itself (ie. a very small code change to remove `()` in lots of places). The test suite would call that function. The CLI logic just checks that the qc_model attribute is not None, so this wouldn't need to be changed at all.

Thoughts?

grahamgower:
Closes #1418

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -1284,7 +1284,7 @@ original function by adding this bit of code to ``stdpopsim/qc/<SPECIES_ID>.py``
 
 .. code-block:: python
 
-  _species.get_demographic_model(_MODEL_ID_).register_qc(_your_review_function())
+  _species.get_demographic_model(_MODEL_ID_).register_qc(_your_review_function)
 
 Where ``_MODEL_ID_`` is the string specified by the original contributor as the
 ``id`` of the demographic model, and ``_your_review_function()`` is the function
--- a/stdpopsim/dfe.py
+++ b/stdpopsim/dfe.py
@@ -273,8 +273,8 @@ class DFE:
         """
         Register a QC model implementation for this DFE.
         """
-        if not isinstance(qc_dfe, self.__class__):
-            raise ValueError(f"Cannot register non-DFE '{qc_dfe}' as QC DFE.")
+        if not callable(qc_dfe):
+            raise ValueError(f"'{qc_dfe}' must be a function returning a QC DFE.")
         if self.qc_dfe is not None:
             raise ValueError(f"QC DFE already registered for {self.id}.")
         self.qc_dfe = qc_dfe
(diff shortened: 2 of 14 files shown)


jeromekelleher:
Looks good. Can we also remove the ``from . import qc`` in ``__init__``?

grahamgower:
> Looks good. Can we also remove the `from . import qc` in `__init__`?

That would have been the simpler solution, but the CLI checks if `qc_model is None` to decide whether to print a warning. So we need to flag models as having a qc model available somehow.

https://github.com/popsim-consortium/stdpopsim/blob/b1383cce4eac2a0fea92b660fdca88b8c4ad5928/stdpopsim/cli.py#L729
https://github.com/popsim-consortium/stdpopsim/blob/b1383cce4eac2a0fea92b660fdca88b8c4ad5928/stdpopsim/cli.py#L837-L850

jeromekelleher (commented):
Right, so this prevents the run-time *evaluation* of the QC model but not the importing of the code. So, we couldn't do any extra imports in the qc model (unless we did icky things like put them in the function).

What if we moved the qc package into the ``tests`` directory and imported that at run-time? It could add on the qc_model attribute which we'd use for these tests (or we could  come up with a less icky way of doing it later).

We could even just do something like removing the import from ``stdpopsim/__init__.py`` and explicitly import stdpopsim.qc from ``tests/test_models.py``. I think that would do the trick actually.

Then, we replace the current check to see if there is QC model with a static ``is_qc_checked`` attribute on the class, which we set to True for all existing models and we document the process of setting this to True as part of the QC process?

grahamgower:
Yeah, I agree that the only way to avoid importing the qc module is to change the QC process to manually do something to the model as part of the QC (e.g. setting a model flag as you suggest). Maybe we defer this until after the conversion of models to Demes files though? The model implementation and QC process docs will need modifying then anyhow, and we can probably flag models as having been QCed in other ways then (e.g. put non-QCed model files in "staging" folder or something).

jeromekelleher:
Yeah, sounds good. I just wonder if there's much point in making this change now if it's not really going to affect load time or allow us to include any extra packages for QC.

petrelharp:
I'm going to close this, as it doesn't sound like there's much point.

PR #1454, opened 2023-03-13
QC implemented for MusMus

igronau:
added QC code for MusMus and tweaked the original species.py file. @peterdfields - please check

### DIFF
--- a/stdpopsim/catalog/MusMus/species.py
+++ b/stdpopsim/catalog/MusMus/species.py
@@ -69,8 +69,8 @@ _ploidy = {
 # generation in C57BL/6 laboratory mice." The provided estimate of mtDNA
 # mutation rate is derived from Goios et al. (2007). In this paper, the authors
 # say "we obtained an overall substitution rate for the mouse coding mtDNA
-# of 3.7 × 10−8 substitutions per site per yr." This value is divided by 0.75,
-# an approximate estimate of generations per year, in order to convert the
+# of 3.7 × 10−8 substitutions per site per yr." This value is multiplied by 
+# the average generation time (0.75 ypg) in order to convert the
 # expressed value to an estimate of mutation rate in units of generations.
 
 _overall_rate = 5.4e-09
@@ -96,7 +96,7 @@ _mutation_rate = {
     "19": _overall_rate,
     "X": _overall_rate,
     "Y": _overall_rate,
-    "MT": 4.93e-08,
+    "MT": 3.7e-08*0.75,
 }
 
 _genome = stdpopsim.Genome.from_data(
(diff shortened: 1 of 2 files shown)


stdpopsim/catalog/MusMus/species.py
```diff
@@ -69,8 +69,8 @@
 # generation in C57BL/6 laboratory mice." The provided estimate of mtDNA
 # mutation rate is derived from Goios et al. (2007). In this paper, the authors
 # say "we obtained an overall substitution rate for the mouse coding mtDNA
-# of 3.7 × 10−8 substitutions per site per yr." This value is divided by 0.75,
```
igronau, line 72: In implementing the QC code for this I realized that we should probably be multiplying by the average generation time and not dividing. I changed the text in the documentation and also the mutation rate below.
@peterdfields - please check and confirm or correct

stdpopsim/catalog/MusMus/species.py
```diff
@@ -151,7 +151,8 @@
 # size is generally accepted. Here we assign a value of 500000 for the Ne as described
 # in Booker and Keightley (2018) which is probably most consistent with M. m. castaneus.
 # Users of stdpopsim may wish to modify this value or implement a specific demographic
-# model if they wish to simulate data for a different M. musculus subspecies.
```
igronau, line 154: I added this mention of Fujiwara et al. (2022). Otherwise, we're not using it to set any of the current species parameters. I also re-ordered the citation objects below to reflect their order in the documentation above

igronau:
@nspope Can you see why this PR doesn't pass the tests? I checked this locally and it seems to be okay. What am I missing?

nspope:
It seems like the tests themselves are passing, but coverage report isn't making it through to Codecov (probably a server issue on their or github's end). This has been happening a lot lately. I'll trigger a rerun.

It also seems like the pre-commit hook black is failing (something to do with code formatting). If you try to commit locally with pre-commit hooks installed, these should modify files (or notify you if there's no automatic solution). Then, you'll have to re-commit after black (in this case) automatically fixes the files.

igronau:
My bad. I guess I missed the pre-commit stage. I ran it to fix the code formatting issues and committed a new version.

stdpopsim/catalog/MusMus/species.py
peterdfields: I think my stoichiometry might have lapsed but if we wanted to convert from substitutions per year to substitutions per generation, and if there's 1 year / 0.75 generations,  I think dividing the reported value by 0.75 gives us substitutions per/generation?

stdpopsim/catalog/MusMus/species.py
peterdfields: This sounds fine.

stdpopsim/catalog/MusMus/species.py
igronau: but it's 1 generation per 0.5 years and not 1 year per generation (0.75 is the average generation time, not generations per year).

peterdfields: I think your calculation is correct here, so yes, multiplying by 0.75 seems correct. I think I somehow got mixed up in the very beginning with the numerator and denominator of generation and year and that error carried through. Sorry about that.

nspope:
Thanks @igronau!

Issue #1458, opened 2023-03-24
QC for Oryza sativa


nspope:
**PR for new species:** Oryza sativa #1453 

If you volunteer to QC this species, please use the checklist below.
While this list is intended to be comprehensive, it may not be exhaustive.
Where relevant, the QC reviewer should identify that parameter values match
those given in the linked citation(s).

The QC reviewer should start a pull request that fills out the test stubs
with independently obtained values. The reviewer may look at the python code
for rationale provided in comments, but should ignore the actual code
as much as possible - comments in the code should give enough information
that it's obvious how to get the correct value from the provided references.
(In particular, we shouldn't copy-paste the value from the code into the test!)

- [ ] Recombination rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Average over a recombination map? Redo the calculation.
- [ ] Mutation rate.
  - This might be genome-wide, or per-chromosome. Both are fine.
  - Check there's a comment describing where it came from, and/or how calculated.
  - From a publication? Check the value(s) match the publication.
  - Calculated somehow? Redo the calculation.
- [ ] Recombination map (if present).
  - Does it match the assembly? Liftover is fine, if clearly stated.
  - Is the description/long_description a good summary of how the map was created?
- [ ] Population size.
- [ ] Generation time.

For each citation, check:
- Doi link.
- Is publication a preprint? Is there a peer-reviewed publication instead?
- Is the year correct.
- Is the author correct (spelling, accents/ligatures/etc.)

Citations are required for:
- [ ] Genome reference assembly.
- [ ] Mutation rate.
- [ ] Recombination rate.
- [ ] Recombination map(s) (if relevant).
- [ ] Population size.
- [ ] Generation time.

The final PR should:
- [ ] fill out the test stubs
- [ ] delete the `pytest.mark.skip` lines that make the tests not run
- [ ] make sure they pass, talking to the original author to figure out discrepancies

nspope:
Thanks for volunteering to QC, @minesrebollo!

minesrebollo:
Hi! 

This is my progress so far on checking this:

- [x] Recombination rate
- [x] Genome assembly
- [x] Generation time

[...]

PR #1461, opened 2023-04-17
QC for Oryza Sativa

minesrebollo:
QC of Oryza sativa #1458


minesrebollo:
I can't edit the file with the new Ne, do I need to start a new pull request?

ornobalam:
The tests seem to be failing because of the recombination rate? I think I need to set it per chromosome (and thus, 0 for mitochondria and chloroplast), rather than the current set up with _genome_wide_recombination_rate. Should I make this change and push?

nspope:
@ornobalam That sounds right -- but maybe it's easiest to have @minesrebollo make the change, to avoid having to sync up different PRs.

nspope:
@ornobalam and @minesrebollo it'd also be fine to set an exception for plastid like is done for the Mt genome-- that is, add
```
_recombination_rate_data["Pt"] = 0
```
below [this line](https://github.com/popsim-consortium/stdpopsim/blob/52d75682f47169170684b3b8fc0d5a5653f2c29b/stdpopsim/catalog/OrySat/species.py#L72)

minesrebollo:
> @ornobalam and @minesrebollo it'd also be fine to set an exception for plastid like is done for the Mt genome-- that is, add

Ok, I think I managed to do it!

tests/test_OrySat.py
```diff
@@ -16,65 +16,57 @@ def test_name(self):
...
+        assert self.species.population_size == 46875
 
-    @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 1
 
 
 class TestGenomeData(test_species.GenomeTestBase):
     genome = stdpopsim.get_species("OrySat").genome
```
nspope, line 28: The ploidy test should be an additional function inside this ``TestGenomeData`` class:

```python
 @pytest.mark.parametrize("chrom", [chrom for chrom in genome.chromosomes]) 
 def test_chromosome_ploidy(self, chrom): 
     if chrom.id in ["Mt", "Pt"]: 
         assert chrom.ploidy == 1 
     else: 
         assert chrom.ploidy == 2 
```

nspope (commented):
Awesome, thanks a bunch @minesrebollo!

One thing I noticed is that the automatically-generated test stubs don't include a test for ploidy, which we need (we have this for the other species). I'll fix the test-stub-generator in another PR-- but in the meantime would you mind adding these for OrySat, @minesrebollo? I've indicated what needs to be added, below.

minesrebollo:
Thanks @nspope , I added that!

nspope:
Awesome, thanks @minesrebollo. The only thing that is failing is the pre-commit checks (these look at formatting, etc.). In particular, if you install the [pre-commit hooks](https://popsim-consortium.github.io/stdpopsim-docs/stable/development.html#pre-commit-checks), these'll automatically fix formatting and whatnot every time you commit (or flag issues that can't be automatically corrected).

In a terminal, navigate to your local copy of the stdpopsim repo. Then,
```bash
pre-commit install
git checkout patch-1
pre-commit run --all-files 
```
which should print out something like:
```bash
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted tests/test_OrySat.py

All done! ✨ 🍰 ✨
1 file reformatted, 174 files left unchanged.

flake8...................................................................Passed
blacken-docs.............................................................Passed
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...............................................................Passed
Check for added large files..............................................Passed
Check for merge conflicts................................................Passed
Mixed line ending........................................................Passed
pre-commit hook(s) made changes.
```
This tells you that one of the pre-commit tools (black) has automatically fixed some formatting issues, modifying one of your files in the process. You then need to stage these changes (via `git add`) and commit them:
```bash
git add .
git commit -m 'Fix formatting for precommit'
```
You'll notice that (once installed) the pre-commit checks are automatically run through every time you try to commit. If these fail, then the commit won't go through, and you'll have to `git add` any changes the the tools make (and perhaps manually fix any other issues).

Hope that makes sense, and sorry for the hassle! Let me know if you have questions or run into issues.

minesrebollo:
Hi!

I could install pre-commit but when I try to push I get this error:

> remote: Permission to minesrebollo/stdpopsim.git denied to minesrebollo.
> fatal: unable to access 'https://github.com/minesrebollo/stdpopsim/': The requested URL returned error: 403

I am using a token that is still valid.

Last time I was also getting that same error but pushes were going though anyway. Any clue? Sorry for my GitHub illiteracy.

nspope:
Gosh, not sure -- if you do `git push https://<your_token_here>@github.com/minesrebollo/stdpopsim.git` does it go through? If not, you might try resetting your credentials locally (maybe it's got the wrong token stored, for whatever reason).

minesrebollo:
I got a new token and it is working now :)

.DS_Store
nspope, line 1: Looks great! Could you please remove this newly added temporary file? Then I can merge.

minesrebollo: Done!

PR #1464, opened 2023-04-19
Bump codecov to 2.1.13

nspope:
codecov is [back](https://about.codecov.io/blog/message-regarding-the-pypi-package/) on pypi -- though I'm not clear if it even needs to be in requirements.txt at all.

### DIFF
--- a/requirements/CI/requirements.txt
+++ b/requirements/CI/requirements.txt
@@ -1,4 +1,4 @@
-codecov==2.1.12
+codecov==2.1.13
 coverage==6.3.3
 flake8==5.0.4
 pytest==7.2.1


grahamgower:
> codecov is [back](https://about.codecov.io/blog/message-regarding-the-pypi-package/) on pypi -- though I'm not clear if it even needs to be in requirements.txt at all.

I don't think this is needed, as the codecov upload should only be done through the github actions workflow.

nspope:
makes sense -- i'll remove it in #1465

nspope:
Darn-- looks like there's some setup needed for the coverage reports to upload, when codecov isn't included in the CI requirements. I'll leave it in for now.

grahamgower:
> Darn-- looks like there's some setup needed for the coverage reports to upload, when codecov isn't included in the CI requirements. I'll leave it in for now.

Hmm, strange. None of my other projects use the codecov pypi package - they just use the `codecov/codecov-action` action. E.g. 
https://github.com/RacimoLab/dinf/blob/38da45e51c6e6172e0d391565d533f008457a3d8/.github/workflows/tests.yaml#L77-L84

nspope:
ah -- looks like the upload token is missing from the tests.yml

grahamgower:
The upload token isn't strictly necessary, but without it you'll get regular upload failures. Also, the token can't be put into a github "secret", because forks don't have access to secrets which means PRs can't access the secrets and the token doesn't get used. The only usable option is to just hardcode the token.

nspope:
Good to know-- thanks!

Issue #1477, opened 2023-05-15
Inconsistent factors of 2 in human gamma DFE models?


RyanGutenkunst:
**Short description of the problem:**
For Humans, the Gamma_K17 and Gamma_H17 are inconsistent in their factor of 2 adjustments.

**How to reproduce the problem:**
An obnoxious factor of two to worry about is that SLiM defines 1+s to be the fitness of homozygote, whereas dadi (and some other tools) define 1+2s to the the fitness of the homozygote. So a selection coefficient of 0.01 in dadi is equivalent to a selection coefficient of 0.02 in SLiM.

Thus to convert a DFE from dadi units to SLiM units, the scale/mean of s must be multiplied by 2. This is done in Gamma_K17 https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/catalog/HomSap/dfes.py#L36 but not in Gamma_H17.

The papers do not make it crystal clear (at least to me) whether they define s in the SLiM way or the dadi way, but since they both used dadi for inference, I assume they're using the dadi way.

**Suggested fixes:**
@klohmueller, can you weigh in?
I think Gamma_H17 needs another factor of two in mean s. But currently K17 has a mean of roughly 0.013 versus H17 currently at 0.014, so another factor of two would make them more different.
If I'm wrong, maybe some more comments would help explain the inconsistency here.

klohmueller:
Both Kim 2017 and Huber 2017 are indeed parameterized like Dadi (where 1+2s is the fitness of the homozygote).

These 2 DFEs are in fact different. They were inferred from different populations EA vs. YRI. But, perhaps more importantly, the mutation rates assumed in the inference are different. Kim used 1.5e-8 and the Huber DFE in Gamma_H17 uses 2.5e-8. We also looked at a lower mutation rate of 1.5e-8 in Huber et al. and found E[s] to be 0.00839. This is closer to 0.013/2.

RyanGutenkunst:
Thanks @klohmueller . Should be fixed in https://github.com/popsim-consortium/stdpopsim/pull/1478

petrelharp:
Damn, good catch.

petrelharp:
This was fixed in #1478.

PR #1478, opened 2023-05-15
Fix factor of 2 in Gamma_K17 DFE

RyanGutenkunst:
Confirmed by @klohmueller here: https://github.com/popsim-consortium/stdpopsim/issues/1477


stdpopsim/catalog/HomSap/dfes.py
```diff
@@ -33,6 +33,9 @@ def _KimDFE():
     Na = 12378
     gamma_scale = 875
     gamma_shape = 0.186  # shape
+    # Extra factor of 2 in mean is to account for difference between tools
+    # in how fitness is defined
+    # (1+s for homozygote in SLiM versus 1+2s in dadi)
     gamma_mean = (-gamma_shape * gamma_scale * 2) / (2 * Na)  # expected value
```
petrelharp, line 39: Did you mean to insert another factor of two in this line?

petrelharp:
Could you also fix the QC implementations? (That's what the failing tests are.) You might be thinking a *different* person needs to do this (since like it's QC?), but that would be silly in this case (we've got multiple eyes on it).

petrelharp:
I think the AraTha DFE is off also, but I'll get that fix in.

petrelharp:
This looks good - checks have passed except for OSX, which just didn't run.

PR #1485, opened 2023-05-20
add mus mus domesticus demographic model

peterdfields:
Hi @igronau It's taken me a bit of time to get back to this but you mentioned I should ping you when I submit things for the demographic models and DFE for mouse. Here is a draft demographic model for Mus musculus domesticus from Fujiwara et al. 2022. I have models for the other two subspecies of mouse as well from the same publication, I figured I'd just start with this one to make sure the format was okay. I haven't heard back anything about the genetic map, should I try pinging anyone about that? Anyway, if this looks okay I can add the two other models and then start on the DFE.


stdpopsim/catalog/MusMus/demographic_models.py
```diff
@@ -0,0 +1,344 @@
...
+    populations = [
+        stdpopsim.Population(
+            id="M_musculus_domesticus",
+            description="Mus musculus domesticus German population",
+        )
+    ]
+
+    return stdpopsim.DemographicModel(
+        id="M_musculus_domesticus_Europe",
+        description="M. musculus domesticus piecewise constant size",
```
igronau, line 149: We have certain conventions for the demographic model ids. You can find them described [here](https://popsim-consortium.github.io/stdpopsim-docs/stable/development.html#naming-conventions). According to these conventions, I would suggest something like DomesticusEurope_1F22

stdpopsim/catalog/MusMus/demographic_models.py
```diff
@@ -0,0 +1,344 @@
...
+        long_description="""
+            This model comes from MSMC using four randomly sampled
+            individuals (DEU01,DEU03,DEU04,DEU06) from a German population.
+            The model is estimated with 57 time periods.
+        """,
```
igronau, line 154: I would also explicitly mention Fig 3 of Fujiwara et al 2022 and say that the population sizes and time changes were supplied by the authors (maybe even mention a specific author?).

stdpopsim/catalog/MusMus/demographic_models.py
```diff
@@ -0,0 +1,344 @@
...
+    ]
+
+    return stdpopsim.DemographicModel(
+        id="M_musculus_musculus_East_Asia",
+        description="M. musculus musculus piecewise constant size",
+        long_description="""
+            This model comes from MSMC using four randomly sampled
+            individuals (KOR01,KOR02,KOR03,KOR05) from a Korean population.
+            The model is estimated with 57 time periods.
+        """,
```
igronau, line 323: See comments above about the model id and long description

stdpopsim/catalog/MusMus/demographic_models.py
```diff
@@ -0,0 +1,344 @@
...
+            1915544,
+        ]
+    )
+    sizes = np.array(
+        [
+            2040,
+            2040,
+            3844,
+            90428,
+            145603,
```
igronau, line 78: check that there is no offset here. From examination of fig 3, it looks like between 300-400 generations ago the Ne was ~150K. The way the arrays are set up here, I think that Ne=145,603 is associated with the time range 420-570.

igronau (commented):
Sorry it took me so much time to get to this. I reviewed the Domesticus demographic model and left some comments. Only one has to do with actual content. The other two are more about format and documentation. Let me know if you have any questions.

petrelharp:
Thanks a lot! Also note from the automatic checks that we're missing a table of parameters:

> Skipping model parameters for M_musculus_domesticus_Europe due to missing table

stdpopsim/catalog/MusMus/demographic_models.py
peterdfields: Done. I also added similar names for the other demographic models.

peterdfields:
> Thanks a lot! Also note from the automatic checks that we're missing a table of parameters:
> 
> > Skipping model parameters for M_musculus_domesticus_Europe due to missing table

I have added these parameter tables in and the automatics checks seem to be passing now.

stdpopsim/catalog/MusMus/demographic_models.py
peterdfields: I have added these in.

stdpopsim/catalog/MusMus/demographic_models.py
peterdfields: Added.

stdpopsim/catalog/MusMus/demographic_models.py
peterdfields: When you say offset here do you mean something being added at the stage of plotting?

igronau: There appears to be an inconsistency between the plot in fig 3 of Fujiwara et al 2022 and your implementation. I don't know if the problem is in fig 3 or your implementation. If I understand the implementation correctly, then  each Ne in your table is being associated with the wrong time range (shifted one range back in time). The 5th Ne (145603) is associated in your implementation with the 5th time interval (420 - 570 generations ago). However, in Fig. 3, it appears to be associated  the time range somewhere between 300 to 400 generations ago, which fits your 4th time interval.  If you remove the first element in the ``sizes`` array, this should fix it IMO.

stdpopsim/catalog/MusMus/demographic_models.py
igronau: Looks good.

docs/parameter_tables/MusMus/DomesticusEurope_1F22.csv
```diff
@@ -0,0 +1,116 @@
...
+Time (yrs.),1443,Begining of 48th time interval
+Time (yrs.),1175,Begining of 49th time interval
+Time (yrs.),943,Begining of 50th time interval
+Time (yrs.),743,Begining of 51th time interval
+Time (yrs.),570,Begining of 52th time interval
+Time (yrs.),420,Begining of 53th time interval
+Time (yrs.),291,Begining of 54th time interval
+Time (yrs.),180,Begining of 55th time interval
+Time (yrs.),83,Begining of 56th time interval
+Time (yrs.),0,Begining of 57th time interval
```
igronau, line 114: Parameter table looks okay. I think that the standard is to have one more entry in the list of Nes than in the list of times. This should correspond to the Ne before the first time interval (the ancestral Ne). Other Nes should map to time intervals. This also relates to my comment on the "shift" in the demographic model

igronau (commented):
Looks good. See my comment about the number of Ne entries in the Domesticus model

stdpopsim/catalog/MusMus/demographic_models.py
igronau: Great

igronau (approved):

peterdfields:
> Looks good. See my comment about the number of Ne entries in the Domesticus model

@igronau I removed the first time interval from the demographic history file and tables, please let me know if any additional changes are needed.

igronau:
@peterdfields , Sorry for the delayed response. I just came back from a long vacation. I think that your changes didn't fix the problem. You just removed he last entry in the ``times`` and ``sizes`` arrays, but this shouldn't really change how the more recent time intervals are implemented in the demographic model. What I think needs to be done is removing the **first** entry only in the ``sizes`` array (and keeping the ``times`` array as is). However, seeing as this is a major source of confusion, then maybe we should try to set a time to discuss this (via Zoom or some other interactive medium)? It might be the simplest way to make sure that we get it right. What do you say?

peterdfields:
@igronau No worries. I'm happy to discuss over zoom. What sort of windows of time work best for you?

igronau:
I'm now in UCLA (Pacific time) and quite flexible in my schedule until next Wednesday. What time works best for you?

peterdfields:
I messaged you over slack. Let me know if you would prefer to continue here.

petrelharp:
ping @igronau @peterdfields - what's going on with this?

peterdfields:
@petrelharp @igronau I made some small changes to the configuration and tables files but it's not obvious to me how these changes caused the (runner?) errors. I just saw the update on Slack about getting the demography for mouse figured out and I remembered I needed to implement @igronau suggestions.

petrelharp:
Those errors are not your fault - looks like random CI failure. I'd say (a) implement the changes you need to make (and push); (b) run tests locally to identify errors; then (c) we'll get CI sorted if there's still an issue.

peterdfields:
@petrelharp the updates pushed yesterday seem to be working fine on the functional side. @igronau do the implemented changes fix the problems you identified before?

igronau:
I verified the model for Domesticus and it seems fine. I created a figure below to validate that the pop size between ``times[i]`` and ``times[i+1]`` is indeed ``sizes[i]``. The tables appear to require some tweaking. For example, according to the `DomesticusEurope_1F22.csv`, pop sizes are defined for 55 intervals and not 56, and ``sizes[1]`` (2040) is associated with the 55th interval and not the 56th interval. This can be fixed by adding another 133912 value to the table. There are 3 intervals associated with this Ne in the model. This is in addition to the ancestral Ne (also 133912), which is associated with the time range before year 1915544. 

If my associations are correct,
image

igronau:
I now also validated the demographic model for Mus Mus (red line in fig 3). In particular, I checked the first peak in Ne, which according to the plot occurred around 900-750 generations ago. The first local max in ``sizes`` is ``sizes[11]`` (68111), and  ``times[11]=740`` and ``times[12]=891``, which is consistent with the figure. Also here, the parameter table in `MusculusKorea_1F22.csv` requires tweaking. Again, ``size[1]`` (179912) is associated with the 55th time interval and not the 56th interval. This can be fixed by adding another 152757 value to the table. There are 3 intervals associated with this Ne in the model. This is in addition to the ancestral Ne (also 152757), which is associated with the time range before year 807711.

igronau:
I now also validated the demographic model for Mus Cas (green line in fig 3). I checked the first four intervals and they seem consistent with the figure (although the Ne of 938111 between 1886 and 3011 generations ago is not clear in the figure because it is out of bounds). Also here, the parameter table in CastaneusIndia_1F22.csv requires tweaking. Again, size[1] (64853) is associated with the 55th time interval and not the 56th interval. This can be fixed by adding another 344802 value to the table.

peterdfields:
@igronau I've updated the table files with (I hope correct!) the additional values/intervals you described. Please let me know if I should add any other modifications.

igronau:
Looks good. I think that this PR can finally be merged, and then I can formally QC it

igronau:
@petrelharp - can you merge this?

peterdfields:
> Looks good. I think that this PR can finally be merged, and then I can formally QC it

yay! On to the DFE...I think this one will probably be the most tricky.

petrelharp:
I've run the tests locally, and they've passed. We'll deal with the CI failures elsewhere (e.g., maybe bumping the cache number like Ive just done in #1526 will do it). So, I'll merge this. Can one of you open the QC issue, please?

PR #1492, opened 2023-06-13
add uniform DFE (closes #1481)

### LINKED ISSUE #1481: add `uniform` DFE
In #1470 @klohmueller says

> I would also suggest adding a “discrete DFE” as well. We could use one from Table 2 of Kim et al (2017). I will give the parameters here that we inferred from the EUR 1000 Genomes data with mu=1.5e-8. Note, ranges of |s| correspond to: [0, 10e−5), [10e−5, 10e−4), [10e−4, 10e−3), and [10e−3, 10e−2), respectively. The mass in the [10e−2, 1] range is the complement of the total mass of the four aforementioned categories. We then inferred the proportions of mutations in each bin. We inferred the proportions to be: 0.286, 0.099, 0.222, 0.305.

To do this, we'd just need to add a single "uniform" distribution type to the [list of distribution types](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/dfe.py#L21) (and then the implementation would use four of these). However, @RyanGutenkunst asks

> For this parameterization, 0.088 weight is for the [1e-2, 1] range. Is that uniform on the linear scale or the log scale? In other words, is half the mass assumed to reside between [1e-2, 0.505] or between [1e-2, 1e-1]?

This suggests *also* adding a "log uniform" type? (which would, I think, be a truncated exponential) And I guess also a "negative log uniform"?

petrelharp:
This just adds the Uniform DFE class. In #1481 it was raised whether we also need "log uniform", that we'd do with something like
```
exp(runif(1, logmin, logmax))
```
... but we have no use cases for it, I think, so I'm not adding that.

### DIFF
--- a/stdpopsim/dfe.py
+++ b/stdpopsim/dfe.py
@@ -23,6 +23,7 @@ class MutationType(object):
     - ``g``: gamma, two parameters (mean, shape)
     - ``n``: normal, two parameters (mean, SD)
     - ``w``: Weibull, two parameters (scale, shape)
+    - ``u``: Uniform, two parameters (min, max)
     - ``lp``: positive logNormal, two parameters (mean and sd on log scale; see rlnorm)
     - ``ln``: negative logNormal, two parameters (mean and sd on log scale; see rlnorm)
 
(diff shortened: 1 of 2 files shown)


petrelharp:
Unclear if this is *actually* what we need for the motivation bethind #1481, but I figure we should merge this anyhow?

petrelharp:
@RyanGutenkunst says that people in the literature *do* use piecewise uniform DFEs, so let's merge this.

PR #1498, opened 2023-06-27
h-s relationship (closes #1488)

### LINKED ISSUE #1488: add discretized h-s relationship DFE
We have discussed using the "uniform" DFE (from #1481) to add a DFE with a h-s relationship; Kirk says on Slack that 
> here's the reference for the discritized h-s relationship: https://www.biorxiv.org/content/10.1101/2022.08.12.503792v1.supplementary-material
and that
> The "best available" (also sometimes called "this paper") model is the one that I would recommend using....it's implemented in this SLiM script: https://github.com/ckyriazis/simulations_review/blob/main/simulation_scripts/SFS_scripts/SFS_newDFE_040422.slim

That's for humans; there's also an h-s relationship estimated in arabidopsis, but from the meeting:
> Linh: would this DFE be for human and Arabidopsis?
> Kirk: probably just human? Chris fit h/s distribution from Arabidopsis and tired to fit to human but didn't fit well. Too many weak recessive deleterious mutations present in the data relative to the model. Models that fit better have more additive, weakly del mutations. The model that is posted to slack is fit specifically to humans, checked on 1kg SFS- reasonable fit, also reasonable fit to estimated human inbreed load.

petrelharp:
Here's an implementation of the "discretized h-s relationship" discussed at #1488. There *is* a small backwards incompatibility here; that is that the "slim_mutation_id" entry we record in the top-level metadata is now a list instead of a single value. I think this is okay because we'll be doing a new release (and so can mention this), and our storing of things in metadata is not documented yet anyhow.

Note in particular that with this, we will be decoupling the "mutation type" of SLiM from the "mutation type" of stdpopsim (ie a single stdpopsim mutation type may produce mutations of many SLiM mutation types). This is why it was perhaps a lack of foresight to model the DFE machinery after SLiM's mechanics rather than motivating biology. 

To understand what's going on here, read "10.6 Varying the dominance coefficient among mutations" of the SLiM manual.

### DIFF
--- a/stdpopsim/dfe.py
+++ b/stdpopsim/dfe.py
@@ -37,16 +37,18 @@ class MutationType(object):
     exponential and gamma, a negative mean can be provided, obtaining always
     negative values.
 
-    Instead of a single dominance coefficient, a discretized relationship
-    between dominance and selection coefficient can be implemented:
-    if dominance_coeff_list is provided, then there is a mutations with selection
-    coefficient s with dominance_coeff_breaks[k-1] <= s <= dominance_coeff_breaks[k] will
-    have dominance coefficient dominance_coeff[k]. In other words, the first entry of
-    dominance_coeff_list applies to any mutations with selection coefficient below the
-    first entry of dominance_coeff_breaks; the second entry of dominance_coeff_list
-    applies to mutations with selection coefficient between the first and second entries
-    of dominance_coeff_breaks, and so forth. The list of breaks must therefore be of
-    length one less than the list of dominance coefficients.
+    Instead of a single dominance coefficient (which would be specified with
+    `dominance_coeff`), a discretized relationship between dominance and
+    selection coefficient can be implemented: if dominance_coeff_list is
+    provided, mutations with selection coefficient s for which
+    dominance_coeff_breaks[k-1] <= s <= dominance_coeff_breaks[k] will have
+    dominance coefficient dominance_coeff[k]. In other words, the first entry
+    of dominance_coeff_list applies to any mutations with selection coefficient
+    below the first entry of dominance_coeff_breaks; the second entry of
+    dominance_coeff_list applies to mutations with selection coefficient
+    between the first and second entries of dominance_coeff_breaks, and so
+    forth. The list of breaks must therefore be of length one less than the
+    list of dominance coefficients.
 
     :ivar distribution_type: A one-letter abbreviation for the distribution of
         fitness effects that each new mutation of this type draws from (see below).
--- a/stdpopsim/dfe.py
+++ b/stdpopsim/dfe.py
@@ -8,11 +8,17 @@ import collections.abc
 import numpy as np
 
 
+def _copy_converter(x):
+    if isinstance(x, list):
+        x = x.copy()
+    return x
+
+
 @attr.s(kw_only=True)
 class MutationType(object):
     """
-    Class representing a "type" of mutation.  The design closely (exactly)
-    mirrors SLiM's MutationType class.
+    Class representing a "type" of mutation.  The design closely mirrors SLiM's
+    MutationType class.
 
     The main thing that mutation types carry is a way of drawing a selection
     coefficient for each new mutation. This ``distribution_type`` should be one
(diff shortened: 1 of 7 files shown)


petrelharp:
Perhaps you could have a look at this @ckyriazis?

petrelharp:
Uh oh, we need some CI love. However, we seem to be passing the tests.

ckyriazis (approved):
This looks really great Peter! My python is a bit rusty so I am maybe not the best person for reviewing your code in great detail, but the approach you took here seems like it should work great.

petrelharp:
Closes #1499. (unless someone objects before this goes in)

stdpopsim/dfe.py
```diff
@@ -31,28 +37,97 @@ class MutationType(object):
     exponential and gamma, a negative mean can be provided, obtaining always
     negative values.
 
+    Instead of a single dominance coefficient, a discretized relationship
+    between dominance and selection coefficient can be implemented:
+    if dominance_coeff_list is provided, then there is a mutations with selection
+    coefficient s with dominance_coeff_breaks[k-] <= s <= dominance_coeff_breaks[k] will
```
petrelharp, line 43: missing a `1`

petrelharp:
Note: this will require a change in analysis2 here: https://github.com/popsim-consortium/analysis2/blob/eb16df28c44a1c3a04a1a5801172e806280f5c2f/workflows/ts2fs.py#L54

chriscrsmith:
Hey @petrelharp want me to look at this PR? Or waiting on ADK?

petrelharp:
Sorry, a bit overwhelmed by github notifications - sure, that'd be great!

chriscrsmith:
I think this looks good! Runs good too. (Although I haven't worried about possible interactions with the rest of the code base, since I'm not too familiar with it)

stdpopsim/slim_engine.py
```diff
@@ -503,6 +503,22 @@
         }
     }
 
+    // Setup mutation callbacks.
+    // For each stdpopsim mutation type with an h-s relationship
+    // we have a sequence of assigned SLiM mutation types;
+    // the first is the one that gets produced by mutation,
+    // and the remainder are assigned by a mutation callback.
```
chriscrsmith, line 510:
```suggestion
    // and the remainder are assigned by a mutation callback.
    // This sequence of mutation types is a subset of the combined 
    // mutation types list, e.g., including neutral mutations, etc.
```

stdpopsim/slim_engine.py
```diff
@@ -759,21 +778,36 @@ def _enum_dfe_and_intervals(contig):
...
     `Contig`. This function will return a dictionary with `len(contig.dfe_list)`
     elements, in which the position of the DFE in `contig.dfe_list` is the key.
-    For each DFE, the dictionary holds a list of tuples that contain the an
-    assigned mutation type id (used in  SLiM) and the `MutationType` object.
+    For each DFE, the dictionary holds a list of tuples that contain the
+    assigned mutation type ids (used in SLiM) and the `MutationType` object.
     This is necessary so that we use the same numeric ids to the mutation types
     in the _add_dfes_to_metadata and slim_makescript steps.
+
+    The assigned mutation type ids is a tuple, that is only of length greater
```
chriscrsmith, line 789:
```suggestion
    The assigned mutation type ids are a tuple, that is only of length greater
```

stdpopsim/slim_engine.py
```diff
@@ -759,21 +778,36 @@ def _enum_dfe_and_intervals(contig):
...
+    than one if the mutation type is of a sort that is implemented using more than
+    one SLiM mutation type; currently, these are only mutation types with a
+    discretized h-s relationship, and (a) only the first of these mutation types
+    are actually assigned a positive mutation rate in SLiM, but (b) all such
+    mutations are transformed to the other mutation types in the list, and hence
+    never end up in the final tree sequence. (See Recipe 10.6 in the SLiM manual.)
```
chriscrsmith, line 795:
```suggestion
    discretized h-s relationship, and (a) only the first of these mutation types
    are actually assigned a positive mutation rate in SLiM, but (b) all mutations
    of the first type are transformed to the other mutation types in the list, and 
    hence never end up in the final tree sequence. (See Recipe 10.6 in the SLiM 
    manual.)
```

stdpopsim/slim_engine.py
```diff
@@ -1176,47 +1210,64 @@ def matrix2str(
...
-            # set policy such that there's at most a single mutation per-site
-            # and individual
-            # printsc(f"    m{mid}.mutationStackGroup = 0;")
-            # printsc(f"    m{mid}.mutationStackPolicy = 'l';")
-        mut_types = ", ".join([str(mt) for mt in mut_type_list])
+                distrib_args = "; ".join([f'"{a}"' for a in mt.distribution_args])
+            if mt.dominance_coeff_list is None:
+                h_list = [mt.dominance_coeff]
+            else:
+                h_list = [0.5]  # this value will not matter
```
chriscrsmith, line 1234: is None or -1 or something better?

stdpopsim/dfe.py
```diff
@@ -31,28 +37,97 @@ class MutationType(object):
     exponential and gamma, a negative mean can be provided, obtaining always
     negative values.
 
+    Instead of a single dominance coefficient, a discretized relationship
+    between dominance and selection coefficient can be implemented:
```
chriscrsmith, line 41:
```suggestion
    Instead of a single dominance coefficient (variable 'dominance_coeff'), a 
    discretized relationship between dominance and selection can be 
    implemented:
```

stdpopsim/dfe.py
```diff
@@ -31,28 +37,97 @@ class MutationType(object):
...
+    if dominance_coeff_list is provided, then there is a mutations with selection
```
chriscrsmith, line 42:
```suggestion
    if dominance_coeff_list is provided, then mutations with selection
```

chriscrsmith (commented):
Offered edits on the code-comments that would clarify things for me

stdpopsim/slim_engine.py
petrelharp: I don't actually understand what your clarifying sentence means - which is "this sequence"? What are you trying to clarify?

stdpopsim/slim_engine.py
petrelharp: None will cause an error. And, if it's like `-1` or something weird then I can imagine people looking at that value and thinking "whoa something is wrong" even though it's not?  (Since the value *will* appear in the SLiM script, just for mutations that are never kept around.)

petrelharp: I'll add a more clarifying comment.

petrelharp:
OK - I did most of your suggestions; see my comments for those I didn't. How's it look now, @chriscrsmith ?

petrelharp:
Huh, what happened? The tests *were* passing: https://github.com/popsim-consortium/stdpopsim/actions/runs/5602731956

stdpopsim/slim_engine.py
chriscrsmith: "this sequence [of mutation types]" is "a sequence of assigned SLiM mutation types; the first is the one that gets produced by mutation, and the remainder are assigned by a mutation callback".

The sentence I offered explains that the sequence you wrote about is a subset of a greater sequence, if other types are present.

chriscrsmith:
> OK - I did most of your suggestions; see my comments for those I didn't. How's it look now, @chriscrsmith ?

Looks great!

stdpopsim/slim_engine.py
petrelharp: I tried rewriting this to clarify a few times, but it didn't work - I don't think this is the place to explain that basic concept - it's a thing that we explain elsewhere, and this minor comment in the SLiM code isn't the place to go into a detailed description of which things get slim mutation types.

PR #1535, opened 2023-10-20
Add journal & volume / page to citation

hyanwong:
No idea why this is missing from the BibTeX citation that Elife provides by default


jeromekelleher:
Whoops, thought this was for a different repo! 

Seems uncontroversial though :smile:

hyanwong:
I hit it when copy/pasting the citation to that other repo, so thought it sensible to correct here too.

Issue #1551, opened 2024-02-23
cached files need python version information in filenames


petrelharp:
Over on slack, Alan Izararras pointed out that running
```
species = stdpopsim.get_species("HomSap")
dfe = species.get_dfe("Gamma_K17")
model = species.get_demographic_model("OutOfAfrica_3G09")
contig = species.get_contig("chr22", genetic_map="HapMapII_GRCh37", mutation_rate=model.mutation_rate)
samples = {"YRI":100}
exons = species.get_annotations("ensembl_havana_104_exons")
exon_intervals = exons.get_chromosome_annotations("chr21")
contig.add_dfe(intervals=exon_intervals, DFE=dfe)
```
with version 0.2.0 now errors with
```
ValueError: Expected SHA256=b9df30b3a37cdd26ec625fd80968ff2e412810c045a11dddc458dc606c702c96, but downloaded file has41160d5ad616d25be13e95fb44c523e2d82dffc942b835b979fdc7c4c01f6d8c.
```
I can confirm this.

So, currently doing `pip install stdpopsim` and running that code does not work.

I'm pretty sure that the problem is that we've replaced the files on amazon with new ones that have different hashes; thus breaking previous versions of the software.

I think the solution is to (a) put back the old files, and (b) add a naming convention to the files so we don't clobber previous ones. 

Some suggestions for the naming convention (first suggestion best I think):

0) For each file, just append a file-specific version number, like `filename_v1.tar.gz`. Then we just need to update the version number whenever we update the file, no fragile programming required (like in 1 and 2 below).

1) directory structure on amazon like
```
0.1.0/
0.1.1/
...
0.2.0/
latest/
```
and the files within each of these would be symlinks to the previous one unless they had been updated. This would let us keep the actual file names the same (maybe bad).

2) similar, but have each file present in various versions like `filename_%version%.tar.gz`, again with most of these being symlinks.

petrelharp:
The change in checksums went in  in #1537. @chriscrsmith do you happen to have the old annotations?

petrelharp:
Hm so this means we could recreate the old files by using the pre- #1522 maintenance script.

nspope:
I've got a copy of the old annotation here, I think: 
[cache-stdpopsim-b9df30.tar.gz](https://github.com/popsim-consortium/stdpopsim/files/14382007/cache-stdpopsim-b9df30.tar.gz)

petrelharp:
[...]

PR #1536, opened 2023-10-25
updated to ensembl release 110; did some patching

andrewkern:
okay here is the update to ensembl 110. i had to do a bit of patching along the way, adding or modifying `ensembl_ids` in the `species.py` files of a few species. 

in addition there is a horrible "duct tape" operation i did to patch the `canis_familiaris` situation. i'll point that out below.

Issue #1521, opened 2023-10-10
What to do about shifting ensembl build?


petrelharp:
Over in #1517, @christianhuber has got a PR for Gorilla (yay!). However, it's failing because the autogenerated file [`stdpopsim/catalog/ensembl_info.py`](https://github.com/popsim-consortium/stdpopsim/pull/1517/files#diff-7c0629bfff3b5e2afa2372fecb52a2cb8cfab2eab87dd70b3553d17883ee8bc8L2) has been changed:
```
- release = 103
+ release = 110
```
and in `tests/test_ensembl.py`, we have
```
# Make sure we don't update the release without realising it.
def test_version():
    release = stdpopsim.catalog.ensembl_info.release
    assert release == 103
```

So, looks like ensembl has moved along, and we need to either
1. update all our genomes to build 110
2. make ensembl build a property of Genome/Annotation/Species so that different species are on different ensembl builds
3. decide it doesn't matter, or
4. something else?

I suspect the reason for this is so that we can cite exactly where the info is coming from, but I am not on top of all the ramifications.

jeromekelleher:
It's probably worth trying the build update to 110, like as not nothing will have changed for most species. You'll see with the diff it generates, whether there's anything to worry about. 

The main reason for including the check was to make sure stuff doesn't change without people realising.

petrelharp:
What do you mean "trying the build update"? Do you mean just run `maintenance <species>` for each species, then compare the diffs? (good idea!)

petrelharp:
That won't hit the annotations, though...

reedacartwright:
I don't know if it helps, but Ensembl has archives. 103 is found here: http://feb2021.archive.ensembl.org/index.html Would it be possible to redo the Gorilla PR for 103?

jeromekelleher:
> What do you mean "trying the build update"? Do you mean just run maintenance <species> for each species, then compare the diffs? (good idea!)

Yeah, that's how it was intended. Occasionally update to the new build, and letting git tell you whether anything important has changed. Running ``update-genome-data`` without any arguments should do it for all species in the catalog.

petrelharp:
@andrewkern will give updating to 104 a shot

andrewkern:
ugh running this returns an HTTP error because Ensembl has changed the designation of `canis_familiaris` to `canis_lupus_familiaris` 😵‍💫 

```
[...]

PR #1553, opened 2024-03-11
updates to hash codes and file names

silastittes:
Added `_v1` to annotation files on aws names as discussed in #1551.
Updated hash codes and files names in the respective `annotation.py` files. Tests passed locally — remote tests too, woo!


silastittes:
Note, these are the new annotation files with the new hash codes, not the original ones. I have the originals but wasn't sure if we also wanted to change those file names.

silastittes:
Yes to documentation. Should I write it in a way that applies to all files with a hash? Maybe a "how to update preexisting resources" or something?

petrelharp:
Well, this needs updating at least:
https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#adding-a-genetic-map
Gee, suddenly I'm surprised we haven't encountered this problem with genetic maps?

docs/development.rst
```diff
@@ -1355,6 +1355,12 @@ Once all this is done, submit a PR containing the code changes and wait for dire
 on whom to send the compressed archive of genetic maps to (currently Andrew Kern is the
 primary uploader but please wait to send files to him until directed).
 
+An important note: when an existing resource file (such as a genetic map or annotation)
+is updated and replaces the previous version,
+be sure to give it a new version name (i.e. ``name_{id}_more_name_v1.txt``). 
+This is to ensure that the previous file versions and checksums that are hosted
+remotely remain valid when new releases are made.
+
```
silastittes, line 1363: Is this sufficient and what you had in mind @petrelharp?

petrelharp: Not quite. Can you add, let's see:
1. in what files do you need to put the new name
2. something more prescriptive about how to create the new name (your example is not following how you created the new names for this round - and, they are .tar.gz, not .txt, right?)
3. a note that in the cache the files are renamed to a standard thing
4. something saying why you have to do this?

silastittes: Updates! Better?

silastittes: Went ahead with updating names to match what I described in docs.

petrelharp: Sounds good - that made more sense to me, but I assumed you had a good reason for doing it the other way.

docs/development.rst
```diff
@@ -1355,6 +1355,21 @@ Once all this is done, submit a PR containing the code changes and wait for dire
 on whom to send the compressed archive of genetic maps to (currently Andrew Kern is the
 primary uploader but please wait to send files to him until directed).
 
+**An important note:**
+when an existing resource file (such as a genetic map or annotation)
+is updated and replaces the previous version,
+be sure to update the name of the file pattern with a version number
+(i.e. ``name_{id}_more_name_v1.tar.gz``).
```
petrelharp, line 1362: This seems confused - either `{id}` should not go here (because as above it refers to an individual chromosome, and we *don't* rename the individual chromosome files?) or else this is the wrong thing that needs to be changed (ie we don't change the `file_pattern`)?

petrelharp: Yeah, I think this is wrong. Note [below in your PR](https://github.com/popsim-consortium/stdpopsim/pull/1553/files#diff-7525483765c11f21a5dd836cee64d70bcb66e5f70f2c52246be45d6e649c4989L17); you are not changing `file_pattern`; you're changing like `intervals_url`. Here you should be documenting/explaining the very process you're doing in this PR.

tests/test_ensembl.py
```diff
@@ -4,4 +4,4 @@
 # Make sure we don't update the release without realising it.
 def test_version():
     release = stdpopsim.catalog.ensembl_info.release
-    assert release == 103
+    assert release == 111
```
petrelharp, line 7: Ack, this thing.  See https://github.com/popsim-consortium/stdpopsim/issues/1521

petrelharp: Please don't change this here.

docs/development.rst
```diff
@@ -1355,6 +1355,21 @@ Once all this is done, submit a PR containing the code changes and wait for dire
...
+The file pattern names do not follow a fixed convention throughout the catalog,
+so simply add an underscore and version number to the end of whatever the current
+file pattern is (before the ``.tar.gz`` file extension),
+or increment the version number if the previous file already has one.
+When the file is downloaded locally to the cache, the file is given a standardized name
+that will be the same regardless of which file is pulled from AWS.
+This process is critical to ensure there are no name collisions on AWS server side,
+which would create conflicts between the checksums of the old and new files between
+different version releases of `stdpopsim`.
```
petrelharp, line 1371:
```suggestion
This makes it so that different releases of `stdpopsim` continue to work: since the checksum
is coded into the package, changing the file on AWS would cause an error when the new file's
checksum is compared on download to the checksum in the package.
```

docs/development.rst
silastittes: Yes sorry, thanks for looking carefully!

tests/test_ensembl.py
silastittes: Ok, so best to just let this test fail until the solution is worked out? I tried for follow the threads and it seems like the PR #1536 is still open?

petrelharp: Yes, you should definitely not change this until this is actually fixed. (However, that'd be a great thing to fix next...)

stdpopsim/catalog/ensembl_info.py
```diff
@@ -1,2 +1,2 @@
 # File autogenerated from Ensembl REST API. Do not edit.
-release = 103
+release = 111
```
petrelharp, line 2: please don't update this

petrelharp, line 2:
```suggestion
release = 103
```

petrelharp:
I've edited the docs - please read and see if it makes sense, @silastittes .

And, I verified the only failing tests have to do with the ensembl build.

So, I *think* we can merge this. But: don't we need to do something to the maintenance script? I don't think that the annotation script has to automatically do the version numbers, but it should at least remind people to do this? Or something? I have not digested that side of things - I'm hoping you have the big picture here, @silastittes.

silastittes:
These updates look good to me! My current understanding of the problem did not require any changes to the maintenance script. I used @andrewkern's local `aws_push_annotations.py` to get the old and new versions of the annotation files onto AWS with the correct names. I think as long as developers update the file names with revisions are submitted, that fixes the issue? Automatic updates in the maintenance script or elsewhere seems hard to implement and error prone, but happy to admit I may not have the whole picture either. @andrewkern, thoughts?

petrelharp:
I made some adjustments to the maintenance script. Please see what you think.

maintenance/annotation_maint.py
```diff
@@ -126,9 +128,14 @@ def download_process_annotations():
...
                     np.savetxt(out_file, intervals, fmt="%d")
-                tf = spc_name_path + f"/{an.id}.tar.gz"
+                tf = spc_name_path + f"/{an.id}_vX.tar.gz"
                 make_tarfile(tf, spc_name_path, "")
                 logger.info("made tarball at " + spc_name_path)
+                logger.info(
+                    "ALERT: need to rename the files (replace 'vX' with "
+                    "appropriate version number to not clobber existing files "
+                    "before upload."
+                )
```
silastittes, line 138:
```suggestion
                logger.warning(
                    "ALERT: need to rename the files (replace 'vX' with "
                    "appropriate version number to not clobber existing files "
                    "before upload."
                )
```

silastittes: Ran locally, found it easy to ignore. Worth making a warning or some other more notable flag? Not sure what best practices dictates here.

petrelharp: Well, the filename itself has `_vX` in it, so they should notice that? Hm, maybe instead it should be `_vCHANGEME`?

maintenance/annotation_maint.py
```diff
@@ -126,9 +128,14 @@ def download_process_annotations():
                         spc_name_path, an.file_pattern.format(id=chrom_id)
                     )
                     np.savetxt(out_file, intervals, fmt="%d")
-                tf = spc_name_path + f"/{an.id}.tar.gz"
+                tf = spc_name_path + f"/{an.id}_vX.tar.gz"
```
petrelharp, line 131:
```suggestion
                tf = spc_name_path + f"/{an.id}_vCHANGEME.tar.gz"
```

maintenance/annotation_maint.py
```diff
@@ -126,9 +128,14 @@ def download_process_annotations():
...
                 make_tarfile(tf, spc_name_path, "")
                 logger.info("made tarball at " + spc_name_path)
+                logger.info(
+                    "ALERT: need to rename the files (replace 'vX' with "
+                    "appropriate version number to not clobber existing files "
+                    "before upload."
```
petrelharp, line 137:
```suggestion
                    "ALERT: need to rename the files (replace 'CHANGEME' with "
                    "appropriate version number) to not clobber existing files "
                    "before upload."
```

maintenance/annotation_maint.py
petrelharp: done

maintenance/annotation_maint.py
```diff
@@ -126,9 +128,14 @@ def download_process_annotations():
                         spc_name_path, an.file_pattern.format(id=chrom_id)
                     )
                     np.savetxt(out_file, intervals, fmt="%d")
-                tf = spc_name_path + f"/{an.id}.tar.gz"
+                tf = spc_name_path + f"/{an.id}_vCHANGEME.tar.gz"
                 make_tarfile(tf, spc_name_path, "")
                 logger.info("made tarball at " + spc_name_path)
+                logger.info(
```
petrelharp, line 134:
```suggestion
                logger.warning(
```

petrelharp:
Could you run the maint script again locally to check these changes work?

silastittes:
Seems to be working to me! 
```
ls annotations/*/*vCHANGEME.tar.gz
annotations/AraTha/araport_11_CDS_vCHANGEME.tar.gz
annotations/AraTha/araport_11_exons_vCHANGEME.tar.gz
annotations/DroMel/FlyBase_BDGP6.32.51_CDS_vCHANGEME.tar.gz
annotations/DroMel/FlyBase_BDGP6.32.51_exons_vCHANGEME.tar.gz
annotations/HomSap/ensembl_havana_104_CDS_vCHANGEME.tar.gz
annotations/HomSap/ensembl_havana_104_exons_vCHANGEME.tar.gz
annotations/PhoSin/Phocoena_sinus.mPhoSin1.pri.110_CDS_vCHANGEME.tar.gz
annotations/PhoSin/Phocoena_sinus.mPhoSin1.pri.110_exons_vCHANGEME.tar.gz
```

PR #1556, opened 2024-03-26
Getting started with DFE for MusMus

peterdfields:
@igronau Just getting the pull started for MusMus. Here I've implemented the base DFE for CDS from [here](https://www.biorxiv.org/content/10.1101/2021.06.10.447924v2). Note that there is also a DFE for CNE and UTR type genomic sequence but I wanted to make sure the CDS is parameterized correctly first.

### DIFF
--- a/stdpopsim/catalog/MusMus/__init__.py
+++ b/stdpopsim/catalog/MusMus/__init__.py
@@ -4,3 +4,4 @@ Catalog definitions for MusMus (Ensembl ID='mus_musculus')
 
 from . import species  # noqa: F401
 from . import demographic_models  # noqa: F401
+from . import dfes  # noqa: F401
--- a/stdpopsim/catalog/MusMus/dfes.py
+++ b/stdpopsim/catalog/MusMus/dfes.py
@@ -1,0 +1,51 @@
+import stdpopsim
+
+_species = stdpopsim.get_species("MusMus")
+
+###########################################################
+#
+# DFEs
+#
+###########################################################
+
+
+def _BookerDFE():
+    id = "Gamma_B21"
+    description = "Deleterious Gamma DFE CDS"
+    long_description = """
+    Return negative MutationType()s representing a Mus
+    musculus subsp. castaneous DFE for protein coded exons or CDS. Booker et al. (2021),
+    https://doi.org/10.1101/2021.06.10.447924
+    DFE parameters are based on an analysis of the unfolded site frequency spectrum
+    (uSFS) using polyDFE v2 (Tataru and Bataillon 2019) as presented in Booker
+    et al. (2021).
+    """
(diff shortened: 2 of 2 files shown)


petrelharp:
I've rebased this; we'll see what needs to be done?

petrelharp:
Okay - looking at the paper, I've got a few questions: should we be (also?) implementing a DFE with positive selection? And, should we be worried this paper from 2021 is still a preprint?

peterdfields:
@petrelharp I think we could implement a couple of different scenarios, I just wanted to make sure this simplest scenario, or at least a scenario that is present for a few other species in the catalog, passed spec before trying others. An alternative citation that could be included as well because it includes some of the same/similar simulations is [here](https://academic.oup.com/g3journal/article/10/7/2317/6026211).

petrelharp:
Sounds good. If the intention is to implement more than one from the same paper, perhaps the name needs additional information? Like I don't know maybe `GammaDeleteriousCDS_B21`? Hm, but actually this is the "just gamma" version, so maybe *this* one doesn't need anything extra.

I checked with Tom Booker, btw, who says we're good to use this - there's  no issue with this part of the analysis.

petrelharp:
Okay, I think this looks good! Mind opening a QC issue, @peterdfields?

peterdfields:
@petrelharp Sure, I can open the QC issue. 

Regarding the other scenarios, I should open another branch for these?

I definitely figured we were on strong footing for the parameter values/details given Tom had worked them up :)

petrelharp:
Great, thanks! And yes - if you'd like to add more, open a new PR. (Should be easy now?)

petrelharp:
I started this in #1649, so it wouldn't be forgotten - could you provide more info there, please?

clararehmann:
Hey @peterdfields, I'm working on the QC for this model and our neutral:negative proportions don't agree - where in the paper did you get yours from?

I'm using the transition:transversion ratio at zero-fold sites for the Mmc model from Supplementary Table 1 to estimate the neutral:negative proportion:

```
tstv = 3.3833
prop_nonsynonymous = 1 / (1 + tstv)
```

however my code gives neutral:negative proportions of `[0.772,  0.228)]` while yours is `[0.334, 0.666]`.  Could you clarify where these parameters are coming from?

petrelharp:
Ah - I should have noticed this earlier - but I don't think transition:transversion is the right number to be using, since that's biochemical, not having to do with synonymous-nonsynonymous?

peterdfields:
@petrelharp The numbers in the model are also included in Tom's repo [here](https://github.com/TBooker/MuridRodentProject/blob/master/bin/SlimFunctions.py) but maybe we need to message Tom to clarify the details a bit more?

PR #1560, opened 2024-05-06
added QC for vaquita DFE

ckyriazis:


petrelharp:
This is perfect, thanks, @ckyriazis! All the DFEs get automatically tested [here](https://github.com/popsim-consortium/stdpopsim/blob/main/tests/test_dfes.py#L828), so there's no need to write specific tests for them.

igronau:
Yes, @ckyriazis is right. I guess I picked up the wrong value from the next line in the text. The correct shape should be -0.0257. Thanks for finding this and correcting it. I now wonder why the shape parameter is slightly different for each mutation rate (0.1314, 0.1315, and 0.1316). In the DFE model I used 0.131 as specified in the results of the inference on the top of page 11.

Issue #1548, opened 2024-02-20
QC for DFE model Gamma_R22 for Vaquita (PhoSin)


igronau:
**PR for new model:** [#1547](https://github.com/popsim-consortium/stdpopsim/pull/1547)
**Original paper:** [Robinson et al. (2022)](https://www.science.org/doi/10.1126/science.abm1742)
**Parameter values:** taken from pages 9-13 of [supplement of Robinson et al. (2022)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9881057/bin/NIHMS1864134-supplement-Supplementary_material.pdf)
**Population indices:** {description of population indices}. See comments below

**Potential issues that might lead to differences between the model implementations:**
The gamma shape was taken directly from the reported estimate.
The gamma mean was taken from the value reported in simulations. This was derived from the inferred shape and scale + the estimated Na, but since the estimated Na is not reported, I couldn't confirm the reported mean.
The h-s relationship is based on what was used in simulations (see top of page 13 in supp).

**QC'er requests:** @ckyriazis / Kirk Lohmueller / @petrelharp

ckyriazis:
Hi Ilan - sorry for being slow to get to this. This looks good to me with one correction - the gamma mean should be -0.0257 rather than -0.00971. As noted near the bottom of page 12 in the supplement, -0.0257 is the value when assuming the estimated mutation rate of 5.8-9 (the default rate used throughout the paper) whereas -0.00971 is used under sensitivity analysis when assuming an alternative mutation rate of 2.2e-9.

Everything else looks correct, including the h-s relationship parameters and the ratio of synonymous to nonsynonymous mutations. Let me know if there are any other questions!

petrelharp:
Thanks, Chris. Hm, so this was a code review more than a QC. (Recall that a
QC reimplements the models blind.) Any chance you could get the QC code in,
@ckyriazis?

ckyriazis:
Hi Peter - sorry for the misunderstanding...I looked at the docs for info on how to [QC a DFE model](https://popsim-consortium.github.io/stdpopsim-docs/stable/development.html#reviewing-a-dfe-model) and it seemed to suggest that the current process is a simple code review. I might have time next week to put the QC code in, but given that I've already looked at Ilan's code, it wouldnt really be a blind implementation. What do u suggest?

petrelharp:
[...]

PR #1561, opened 2024-05-07
move PhoSin QC where it should go

petrelharp:
There was a whoopsie in #1560 I didn't notice - the code was in `qc/PhoSin.py` not `stdpopsim/qc/PhoSin.py`. Fixing this.


petrelharp:
Looks like the parameters for the gamma disagree:
```
>           assert np.allclose(mt1[i].distribution_args, mt2[i].distribution_args)
E           AssertionError: assert False
E            +  where False = <function allclose at 0x7fb9680175e0>([-0.00971, 0.131], [-0.0257, 0.131])
E            +    where <function allclose at 0x7fb9680175e0> = np.allclose
E            +    and   [-0.00971, 0.131] = MutationType(dominance_coeff=None, distribution_type='g', distribution_args=[-0.00971, 0.131], convert_to_substitution=True, dominance_coeff_list=[0.0, 0.01, 0.1, 0.4], dominance_coeff_breaks=[-0.1, -0.01, -0.001]).distribution_args
E            +    and   [-0.0257, 0.131] = MutationType(dominance_coeff=None, distribution_type='g', distribution_args=[-0.0257, 0.131], convert_to_substitution=True, dominance_coeff_list=[0.0, 0.01, 0.1, 0.4], dominance_coeff_breaks=[-0.1, -0.01, -0.001]).distribution_args
```
This issue was flagged by @igronau in #1548:
```
The gamma shape was taken directly from the reported estimate.
The gamma mean was taken from the value reported in simulations. This was derived from the inferred shape and scale + the estimated Na, but since the estimated Na is not reported, I couldn't confirm the reported mean.
```
@ckyriazis, can you confirm/clarify?

ckyriazis:
ah oops sorry about the directory issue! yes, this is what I was expecting - the gamma mean in the original PR was incorrect, -0.0257 is the correct value.

petrelharp:
Hm, I do see that in [the paper](file:///tmp/mozilla/NIHMS1864134-supplement-Supplementary_material.pdf):
```
Selection coefficients (s) for deleterious mutations were drawn from a gamma
distribution estimated using the genomic data (see Inference of the distribution of selection
coefficients) with mean = -0.0257 and shape parameter = 0.1314. 
```

Thanks! I will fix this and merge; if @igronau disagrees, we can discuss and fix it up separately.

petrelharp:
Ah, it looks like it was a case of just grabbing the wrong number - the number previously in the code was from the lower mutation rate:
```
For sensitivity analyses
varying the mutation rate, we used the parameters inferred under the respective mutation rate
(mean = -0.00971 and shape = 0.1316 for μ = 2.2x10-9 and mean = -0.0476 and shape = 0.1315
for μ = 1.08x10-8; fig. S13).
```

stdpopsim/catalog/PhoSin/dfes.py
```diff
@@ -48,7 +48,7 @@ def _RobinsonDFE():
     #   is defined in fitDadi and in SLiM  ]
     # Since the inferred value of Na is not reported in the supplement, we use
     # here the value of the mean as reported in the supp info
-    gamma_mean = -0.00971  # mean of gamma          (see bottom of page 12 in supp)
+    gamma_mean = -0.0257  # mean of gamma          (see bottom of page 12 in supp)
```
igronau, line 51: Yes, this is correct. Thanks, @petrelharp for fixing this.

igronau (approved):
Looks good.

PR #1567, opened 2024-07-05
Fix SLiM time units warning, bump minimum version to 4.0

nspope:


nspope:
I'm going ahead and merging this -- includes fix for the macos runner which was preventing the CI from completing

PR #1569, opened 2024-07-05
Move sweep API and extended events from `stdpopsim.ext` to `stdpopsim`

nspope:


nspope:
Does the added doc blurb look OK here @petrelharp ? I added in the API for extended events, but maybe we don't actually want these as they're not mentioned anywhere else in the docs (e.g. we could just include `selective_sweep` in `api.rst`)

petrelharp:
I'm going to merge, and will consider the docs later.

petrelharp (commented):
Thanks!

PR #1570, opened 2024-07-05
Remove shifting of coordinates

### LINKED ISSUE #1404: When using `left`/`right`, output coordinate system is relative to start of contig
The options `left`, `right` in `Species.get_contig` allow a portion of a chromosome to be simulated (in contrast to simulating an entire chromosome and "masking" everything outside of `[left, right]`). When adding DFEs/inclusion masks/sweeps/etc to the contig, the coordinate system used is that of the original chromosome.

However, the coordinate system of the output tree sequence will be relative to the start of the contig, not of the original chromosome. So for example, if I simulate chr22 from 30 Mb to 50 Mb then dump a VCF, the coordinates of variants won't reflect their positions on the original chromosome.

Instead, should the output tree sequence span the entire chromosome and have missing data outside of `[left, right]`?

petrelharp:
As discussed in #1404.

Closes #1404.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,7 +2,27 @@
 [0.2.1] - 2023-XX-XX
 --------------------
 
+**Bug fixes**:
+
+- Updates to SLiM support: updated the `active` flag in the SLiM code to be integer.`
+
 **Breaking changes**:
+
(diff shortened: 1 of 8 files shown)


nspope (commented):
LGTM!

PR #1574, opened 2024-07-06
more windows CI

petrelharp:
DO NOT MERGE


petrelharp:
The logfile issue looks to be a SLiM bug: https://github.com/MesserLab/SLiM/issues/452 maybe

PR #1587, opened 2024-10-24
Illustrate use of both dtwf and hudson msprime models

hyanwong:
I always forget what the syntax is here! I think others might find it useful.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -684,17 +684,15 @@ defined above. The simulation results are recorded in a tree sequence object
 .. code-block:: python
 
     ts = engine.simulate(
-        model, contig, samples,
-        msprime_model="dtwf", msprime_change_model=[(20, "hudson")]  # optional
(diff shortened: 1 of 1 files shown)


petrelharp:
I like the idea, but sticking this in the "quickstart" doesn't seem like the right place for it? Perhaps under a new subheader "Advanced: switching demographic models"?

hyanwong:
This addition isn't in the first introduction to stdpopsim example code (which is under the heading "Running a published model"). Rather it is in the (more advanced) "Running a generic model" section which follows that. So I reckon it's not really treated as a "Quickstart" example IMO, and is probably fine where it is?

petrelharp:
It's still fairly "quickstart"-ish; what about in the new section ("tips and tricks and gotchas") I'm proposing adding in #1601?

petrelharp:
I changed my mind here - this is good practice, and as you say this isn't the very first demonstration. But, I've rewritten it a bit and added a link to the DTWF paper - what do you think?

hyanwong:
Yes, this LGTM. Merge away!

PR #1599, opened 2024-12-12
Update CI versions

benjeffery:


petrelharp:
Ping me on slack if I should jump in here, @benjeffery - I may or may not notice what happens here. Godspeed!

nspope:
The issue is that `tests/test_dfes.py::TestRegisterQCDFE::test_register_qc` assigns an object to be an attribute of itself, which causes the recursion issue. I'm completely puzzled as to why this hasn't surfaced before, but is pretty easy to fix by changing the test to:
```python
    def test_register_qc(self):
        dfe = self.make_dfe("test")
        dfe_two = self.make_dfe("test")
        dfe.register_qc(dfe_two)   # was previously dfe.register_qc(dfe)                                                                        
        assert dfe.qc_dfe == dfe_two
```

nspope:
If you rebase on main hopefully that'll fix the error here, @benjeffery

petrelharp:
Rebased.

petrelharp:
Now biopython is failing to compile?!?!

nspope:
It's a Biopython version thing-- the installed version is using CPython's deprecated unicode API, which was removed in python 3.12

benjeffery:
Ben Out

Thanks for the assist @nspope!

nspope (approved):
Beautiful, thanks @benjeffery!!

petrelharp:
Yes! Thanks, both!!!

PR #1600, opened 2024-12-12
Fix self-reference causing recursion depth error

nspope:
Fixes error popping up in #1599


nspope:
I'm merging this as the tests are passing on ubuntu and windows, so that #1599 can go in

PR #1601, opened 2024-12-13
docs on missing data and coordinate systems, closes #1584

### LINKED ISSUE #1584: beware of recapitation and missing data
A thing that caught me by surprise: now, if you simulate a chunk of a contig, you'll get 'missing data' outside that contig. This is represented as, basically, "no trees there". However, this will have `ts.first().num_roots == ts.num_samples` (and this is the reason for the `root_threshold` argument).  This is fine and expected, if confusing - so maybe needs some documentation.

A particular gotcha is that if someone notices the ts doesn't look recapitated (although it is!), and recapitates again with `pyslim.recapitate(ts, ...)`, then this will wrongly simulate coalescent history on the bits of chromosome that should be missing. I'm not sure if there's anything we can do to prevent this besides some documentation.

petrelharp:
In perusing #1584 I realized we hadn't written down this thing about the coordinate system being fixed even when we simulate a chunk. Now it's written down - hopefully without assuming the reader understands tskit too much?

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1439,6 +1439,78 @@ To make the example quick, we've only simulated the first 100Kb;
 a more realistic example would apply it to the exons, available
 as a :ref:`annotation <sec_catalog_phosin_annotations>`.
 
+.. _sec_tute_recapitation:
+
+Tips, tricks, and gotchas
+=========================
+
+Here are a few things about the whole process that it might be useful to know.
+Maybe this will save you some time,
+or let you do new things!
+
+.. _sec_tute_missing_data:
+
+Missing data and coordinates
+----------------------------
+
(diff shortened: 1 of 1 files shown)


petrelharp:
Here's the result - maybe too verbose?
Screenshot from 2024-12-12 21-57-44
Screenshot from 2024-12-12 21-57-38

petrelharp:
Thanks! Well, I think these docs are aiming to get the reader to - say - VCFs as quickly and painlessly as possible. Hopefully people doing that will realize they don't need to worry about that. *Hm:* maybe I'll just add an explicit note saying so.

petrelharp:
Added 
Screenshot from 2024-12-12 22-27-57

PR #1604, opened 2024-12-15
Update tutorial.rst - fix for #1592

### LINKED ISSUE #1592: Typo in wiki catalog.
Hi! I found a typo in the wiki. Browning et al. 2011 should be Briowning et al. 2018. 

image

gregorgorjanc:
Closes #1592

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1555,7 +1555,7 @@ models available to us:
 This prints detailed information about all of the available models to
 the terminal.
 In this tutorial, we will use the model of African-American admixture from
-`2011 Browning et al <http://dx.doi.org/10.1371/journal.pgen.1007385>`_.
+`Browning et al. (2018) <http://dx.doi.org/10.1371/journal.pgen.1007385>`_.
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -512,8 +512,9 @@ This provides both a ``size_history`` plot and a ``tubes`` visualization. E.g.
 .. code-block:: python
 
     import demesdraw
+
     graph = model.model.to_demes()
-    demesdraw.tubes(model.model.to_demes());
+    demesdraw.tubes(model.model.to_demes())
```
petrelharp, line 517: do you actually want that change? in jupyter notebooks anyhow, the `;` makes it not print out the returned like <matplotlib.pyplot> object? (I'm not checking the compiled docs...)

petrelharp:
Awesome; thanks! Just checking, though - did you compile the docs and check that it looks good? (see my comment?)

docs/tutorial.rst
gregorgorjanc: I changed that because `black` was otherwise complaining and suggested this change.

nspope: These code blocks aren't actually being run during doc compilation though? So it shouldn't matter

PR #1616, opened 2024-12-16
Fix to tutorial.rst for precommit black

nspope:


nspope:
This was fixed by #1604, whoops

PR #1626, opened 2024-12-22
Squashing unintended test warnings due to model recombination rate

gregorgorjanc:
As triggered by https://github.com/popsim-consortium/stdpopsim/issues/1598#issuecomment-2551983948 I looked into the test warnings due to model specific mutation and recombination rate.

One test warning was not intended, so I fixed that.

I am a bit confused what to do with these ones:

```
> python3 -m pytest tests/test_models.py
========================================================================= test session starts =========================================================================
platform darwin -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0
rootdir: /Users/ggorjanc/Storages/GitBox/popsim/stdpopsim/stdpopsim_fork
configfile: setup.cfg
plugins: xdist-3.6.1
4 workers [275 items]   
............................................................................................................................................................... [ 57%]
....................................................................................................................                                            [100%]
========================================================================== warnings summary ===========================================================================
tests/test_models.py::TestBosTauHolsteinFriesian_1M13::test_simulation_runs
  /Users/ggorjanc/Storages/GitBox/popsim/stdpopsim/stdpopsim_fork/stdpopsim/engines.py:125: UserWarning: The demographic model has recombination rate 1e-08, but this simulation used the contig's recombination rate 0.0. Patterns of linkage may be different than expected for this species. For details see documentation at https://popsim-consortium.github.io/stdpopsim-docs/stable/tutorial.html
    warnings.warn(

tests/test_models.py::TestBosTauAngus_1B16::test_simulation_runs
  /Users/ggorjanc/Storages/GitBox/popsim/stdpopsim/stdpopsim_fork/stdpopsim/engines.py:125: UserWarning: The demographic model has recombination rate 5e-09, but this simulation used the contig's recombination rate 0.0. Patterns of linkage may be different than expected for this species. For details see documentation at https://popsim-consortium.github.io/stdpopsim-docs/stable/tutorial.html
    warnings.warn(

tests/test_models.py::TestBosTauHolsteinFriesian_1B16::test_simulation_runs
  /Users/ggorjanc/Storages/GitBox/popsim/stdpopsim/stdpopsim_fork/stdpopsim/engines.py:125: UserWarning: The demographic model has recombination rate 3.66e-09, but this simulation used the contig's recombination rate 0.0. Patterns of linkage may be different than expected for this species. For details see documentation at https://popsim-consortium.github.io/stdpopsim-docs/stable/tutorial.html
    warnings.warn(

[...]

PR #1630, opened 2024-12-23
Squash unintended recombination rate test warnings

gregorgorjanc:
Following upon https://github.com/popsim-consortium/stdpopsim/pull/1626 I now figured out that it was not mitochondrion that was causing the errors, but the fact that model tests in `DemographicModelTestMixin:test_simulation_runs` are ran without recombination, which rightly issued the unintended warning when a model has it's specific rec rate.

We test these warnings in `TestRecombinationRates:test_recombination_rate_warning` and `TestRecombinationRates:test_recombination_rate_match` so I provide this fix to get rid of the unintended recombination rate test warnings.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -34,10 +34,19 @@ class DemographicModelTestMixin:
 
     @pytest.mark.filterwarnings("ignore:.*Zigzag_1S14.*:UserWarning")
     def test_simulation_runs(self):
-        # With a recombination_map of None, we simulate a coalescent without
-        # recombination in msprime, with mutation rate equal to rate from model.
+        # When model has no specific recombination rate,
+        #   we simulate coalescent without recombination.
+        # Otherwise, we use the rate from the model.
+        # We use mutation rate equal to rate from model
+        #  (or species when the model does not have one.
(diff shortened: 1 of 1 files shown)


tests/test_models.py
```diff
@@ -34,10 +34,19 @@ def test_debug_runs(self):
 
     @pytest.mark.filterwarnings("ignore:.*Zigzag_1S14.*:UserWarning")
     def test_simulation_runs(self):
-        # With a recombination_map of None, we simulate a coalescent without
-        # recombination in msprime, with mutation rate equal to rate from model.
+        # When model has no specific recombination rate,
+        #   we simulate coalescent without recombination.
+        # Otherwise, we use the rate from the model.
+        # We use mutation rate equal to rate from model
+        #  (or species when the model does not have one.
```
petrelharp, line 41:
```suggestion
        # For the purposes of testing the model, we use the
        # model's mutation and recombination rate if it has them;
        # otherwise, we use zero recombination and the species
        # mutation rate.
```

petrelharp:
Gotcha - looks good. I just tried to make the motation even more clear.

PR #1638, opened 2024-12-24
more messing around with CI, take 4

petrelharp:


petrelharp:
note these tests (finally!!) passed; just tmate failed

PR #1479, opened 2023-05-16
AraTha DFE QC

petrelharp:
This is a work in progress; see #1344 for discussion.

PR #1647, opened 2025-01-08
QC for AraTha DFE

### LINKED ISSUE #1344: QC for AraThal DFE
**PR for new model:** 
#1324 

**Original paper:**
[Huber et al., 2018](https://doi.org/10.1038/s41467-018-05281-7)

**Parameter values:** 
I took the gamma parameters from Supplementary Table 4, the genome-wide, additive-only model for A. LYRATA- due to challenges with simulating with selfing for A. thaliana. Not certain this is what the group had in mind, also there might be updates on simulating selfing.

**Potential issues:**
- I used the estimates from A. Lyrata.
- I'm only 90% certain there were zero neutral mutations estimated as part of this DFE.

**QC'er requests:** 
Kirk Lohmueller  :)

clararehmann:
Closes #1344.
 
I used the gamma distribution parameters from the additive *A. lyrata* model from Table S4 of [Huber et al., 2018](https://doi.org/10.1038/s41467-018-05281-7) to QC the model.

The scale parameter of the distribution is misidentified as `gamma_mean` in the catalog, so I modified the Gamma_H18 DFE in the AraTha catalog to match so that `gamma_mean` is now the product of the shape and scale parameters.

### DIFF
--- a/stdpopsim/catalog/AraTha/dfes.py
+++ b/stdpopsim/catalog/AraTha/dfes.py
@@ -10,19 +10,15 @@ _species = stdpopsim.get_species("AraTha")
 
 
 def _HuberDFE():
-    id = "Gamma_H18"
-    description = "Deleterious Gamma DFE"
+    id = "GammaAdditive_H18"
+    description = "Deleterious additive Gamma DFE"
     long_description = """
-    Return negative (no neutral, in this case) MutationType()s
-    representing an Arabidopsis DFE. From Huber et al. (2018).
-    Gamma parameters are based on Supplementary Table 4,
-    the genome-wide, additive-only model for A. LYRATA- due to
-    challenges with simulating with selfing for A. thaliana.
-    The Supplementary Table 4 DFEs are not noted to contain
-    any neutral mutations (in contrast, Supplementary Table 3
-    notes neutral proportions for two other DFEs), and in the
-    main text including neutral mutations seems like a
-    supplementary analysis rather than the main strategy.
+    Additive, deleterious DFE from Huber et al. (2018) for Arabidopsis,
+    estimated from the SFS of A. lyrata as a Gamma distribution
+    of deleterious effects. Parameters are from Supplementary Table 4,
+    the "genome-wide, additive-only model for A. lyrata". The DFE for
+    A. lyrata (rather than A. thaliana) is provided due to challenges
+    with simulating selfing.
     """
     citations = [
         stdpopsim.Citation(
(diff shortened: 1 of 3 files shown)


petrelharp:
I've discussed this with @clararehmann; here's the summary.

As discussed in #1344, this DFE is arguably *not* the "best" one to use from this paper. However, it is a totally reasonable one, and who says we need just one DFE. So, the plan is to (1) add this additive DFE (and, this PR finalizes that); and (2) also add the "genome-wide both species h-s relationship" DFE (I will open another issue for that).

With that in mind: @clararehmann - how about let's change the name of this DFE, maybe to `GammaAdditive_H18`, as we discussed? (Or, make another proposal?)

petrelharp:
Ah, uh-oh, @clararehmann - maybe there is a [factor of 2 missing?](https://github.com/popsim-consortium/stdpopsim/pull/1479/files#diff-e6c3e602a1101170bc823c22d6220983cfa1b9752ceff0ac0a41dcb5da0ded8eR222-R224) as well?

stdpopsim/catalog/AraTha/dfes.py
```diff
@@ -10,8 +10,8 @@
 
 
 def _HuberDFE():
-    id = "Gamma_H18"
-    description = "Deleterious Gamma DFE"
+    id = "GammaAdditive_H18"
+    description = "Deleterious additive Gamma DFE"
     long_description = """
```
petrelharp, line 15: Ah - could you update the long description, also? How about
```
    Additive, deleterious DFE from Huber et al. (2018) for Arabidopsis, 
    estimated using <software> from A. lyrata as a Gamma distribution
    of deleterious effects. Parameters are from Supplementary Table 4,
    the "genome-wide, additive-only model for A. lyrata". The DFE for
    lyrata (rather than thaliana) is provided due to challenges with
    simulating selfing.
```

stdpopsim/qc/AraTha.py
```diff
@@ -208,3 +208,29 @@ def HuberThreeEpoch():
 
 
 _species.get_demographic_model("African3Epoch_1H18").register_qc(HuberThreeEpoch())
+
+
+def HuberDFE():
+    id = "QC-Gamma_H18"
```
petrelharp, line 214:
```suggestion
    id = "QC-GammaAdditive_H18"
```

petrelharp:
Looks good! I've moved the improved description to the catalog, rather than the QC implementation (which no one will ever see); and squashed the commits. When the tests finish we can merge (assuming I didn't muck something up.)

stdpopsim/catalog/AraTha/dfes.py
petrelharp: ah, I see `<software>` wasn't some named software package, they just did it using Methods from the paper. I rephrased this to "using the SFS"

Issue #1252, opened 2022-05-16
Add assembly name to standard naming for annotation


petrelharp:
We have the assembly name at the end of every genetic map; this should probably be the case for annotations also, so it's obvious if they don't match.

nspope:
@andrewkern suggests adding a property to the class that carries the annotation name

silastittes:
Happy to work on this. Which class specifically should carry the annotation name?

petrelharp:
I would think [`Annotation`](https://popsim-consortium.github.io/stdpopsim-docs/latest/api.html#stdpopsim.Annotation); are there any other candidates, @silastittes?

silastittes:
`Annotation` has an `id` field already, and the full file name is in the url field. What info would go in the annotation name?

nspope:
That's my bad -- "annotation name" ==> "assembly name"

silastittes:
Should an assembly name be added to genetic maps too then?

petrelharp:
I think figuring that out is part of this issue.

silastittes:
It does seem like it would be nice to be able to do things like 
```
assert genetic_map.assembly == annotation.assembly
```

petrelharp:
Ping @silastittes - might you be able to figure this out (in the next week)?

silastittes:
Yeah happy to work on this.

nspope:
This should hopefully fix #719

petrelharp:
hey @silastittes - any progress on this?

silastittes:
No! Sorry, other things including edits on the manuscript took priority. After reviewing conversation, I'm a bit confused. Is this about adding new attributes to classes or just having conventions for the file names? Both?

silastittes:
Ok, re-re-reading ... We don't currently have a genome version attribute in Annotation, or GeneticMap, but we do in [Genome](https://github.com/popsim-consortium/stdpopsim/blob/d07f5e59cae2b02a2628111776f0797bd180cfb0/stdpopsim/genomes.py#L44). So a potential solution could be to just carry over the same formatting to the two other classes to facilitate checks that they match, and enforce naming conventions that include the genome assembly in map and annotation files. Does that cover the concerns?

petrelharp:
Ah, okay - first off, many apologies; my issue was totally misguided. Here's what's going on, I think (please check?):

I was talking about the *files*, but neglected to say so. For these:

[...]

PR #1650, opened 2025-01-09
adding assembly_name attribute

silastittes:
Here's a draft solution for #1252

@grahamgower made a patch described here dealing with genetic map intervals that don't match that of the assembly, which will happen. All described further in #719

I've started the process of adding the assembly names to the Annotation and GeneticMap classes, which would facilitate checking that the resources all belong on the same assembly. 

It's hard to write explicit tests for this using the interval information, especially for annotations, because the tests likely won't fail if the annotation intervals are still within the bounds of the (incorrect) assembly. I think this largely comes down to humans being careful to check that the annotation/genetic maps match the assembly. Adding the attribute really just serves as a helpful reminder during the QC process to verify this, and do a liftover if it's wrong.      

If this is the direction we want to go a few more things need to be done.

- Add `assembly_name` to stubs for generating new species in the catalogue
- Finish population the genetic maps and annotations (I think all annotations are done, but should verify) 
- Flesh out documentation about making sure the assemblies match between resources (I belive there's some stuff about this already concerning the liftover process).
- A test or two checking that assembly name matches for Annotation, GeneticMap, and Genome.
OR 
- Add some warnings when the assembly_name doesn't match between the components.

### DIFF
--- a/stdpopsim/annotations.py
+++ b/stdpopsim/annotations.py
@@ -27,6 +27,9 @@ class Annotation:
     :vartype citations: list of :class:`.Citation`
     :ivar file_pattern: The pattern used to map individual chromosome id strings
         to files
+        :ivar assembly_name: The name of the genome assembly.
+    :vartype assembly_name: str    
+    :ivar str assembly_name: The name of the assembly the annotation is based on
     """
 
     id = attr.ib()
@@ -40,6 +43,7 @@ class Annotation:
     file_pattern = attr.ib()
     annotation_source = attr.ib()
     annotation_type = attr.ib()
+    assembly_name = attr.ib()
 
     def __attrs_post_init__(self):
         self._cache = stdpopsim.CachedData(
--- a/stdpopsim/catalog/AraTha/annotations.py
+++ b/stdpopsim/catalog/AraTha/annotations.py
@@ -27,6 +27,7 @@ _an = stdpopsim.Annotation(
     file_pattern="araport_exons_{id}.txt",
     annotation_source="araport11",
     annotation_type="exon",
+    assembly_name = "TAIR10"
 )
 _species.add_annotations(_an)
 
@@ -56,5 +57,7 @@ _an2 = stdpopsim.Annotation(
     file_pattern="araport_CDS_{id}.txt",
     annotation_source="araport11",
     annotation_type="CDS",
+    assembly_name = "TAIR10"
+
 )
 _species.add_annotations(_an2)
--- a/stdpopsim/catalog/DroMel/annotations.py
+++ b/stdpopsim/catalog/DroMel/annotations.py
@@ -28,6 +28,7 @@ _an = stdpopsim.Annotation(
     file_pattern="flybase_exons_{id}.txt",
     annotation_source="FlyBase",
     annotation_type="exon",
+    assembly_name="BDGP6.32.51",
 )
 _species.add_annotations(_an)
 
@@ -57,5 +58,6 @@ _an2 = stdpopsim.Annotation(
     file_pattern="flybase_CDS_{id}.txt",
     annotation_source="FlyBase",
     annotation_type="CDS",
+    assembly_name="BDGP6.32.51",
 )
 _species.add_annotations(_an2)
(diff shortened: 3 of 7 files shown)


petrelharp:
Hm, ineteresting. We didn't make `assembly` a property of `GeneticMap`: see [genetic_map.py](https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/genetic_maps.py); should we have? What's your rationale for adding this new thing?

petrelharp:
Okay - we need @andrewkern's input here, I think. Observations:

1. An annotation is associated with a given assembly; we ought to be able to know which one.
2. Same for a genetic map.
3. Similarly, we ought to be able to know that the genetic map and annotation being used together are with reference to the same assembly!
4. So, maybe `assembly` should be a property of `Genome` (is it already?)?
5. Right now @andrewkern is putting in a property of `Genome` that gives (for instance) ensembl build number; this is similar to but not the same as assembly. It seems to me that `assembly` is what we *actually* want, not ensembl build?

So: should we have `assembly` as a property of `Genome`? Then check that annotations and genetic maps match that assembly? To check, we don't need to have `assembly` as a property of `Annotation` and `GeneticMap`, but we could. Should we?

I feel like I'm over-complicating things, maybe?

andrewkern:
so in looking this over, do we need the assembly in the 

> Okay - we need @andrewkern's input here, I think. Observations:
> 
> 1. An annotation is associated with a given assembly; we ought to be able to know which one.
> 2. Same for a genetic map.
> 3. Similarly, we ought to be able to know that the genetic map and annotation being used together are with reference to the same assembly!
> 4. So, maybe `assembly` should be a property of `Genome` (is it already?)?
> 5. Right now @andrewkern is putting in a property of `Genome` that gives (for instance) ensembl build number; this is similar to but not the same as assembly. It seems to me that `assembly` is what we _actually_ want, not ensembl build?
> 
> So: should we have `assembly` as a property of `Genome`? Then check that annotations and genetic maps match that assembly? To check, we don't need to have `assembly` as a property of `Annotation` and `GeneticMap`, but we could. Should we?
> 
> I feel like I'm over-complicating things, maybe?

so the `assembly_name` is already a property of `Genome` like [here](https://github.com/popsim-consortium/stdpopsim/blob/a344665be1f442fe150b32ccb9c3cbc341c19b4a/stdpopsim/catalog/HomSap/genome_data.py#L4). 

the `GeneticMap` files are named based on their assemblies, as are the `Annotations`, so maybe this is redundant? 

Having said that, I do like the idea of a test that could yield a warning if the user chose to say use a genetic map from an assembly that was mismatched with the genome, but maybe we should tag this as a TODO and put it off until after the release we are working on?

petrelharp:
I think I'm in the same place as @andrewkern here - this would be a nice thing to do, but let's put this off and not do it right now. There's a bunch of other things to think through (and implement) here, and we'd like to do that properly, but it's not urgent. I think the only things that we *do* want to do now are:

1. make sure the assembly is mentioned in the description of the PhoSin annotation (as with the other annotations)
2. make the assembly property of `Genome` visible in the catalog

I'll open issues for those.

petrelharp:
So - if you agree, let's close this, @silastittes? And, I'll close #1252 as well, but if you'd like to write up a summary in another issue for the future (which can refer to this PR), that'd be helpful?

silastittes:
That sounds good to me! 

Regarding your two points. 

1. make sure the assembly is mentioned in the description of the PhoSin annotation (as with the other annotations)
The current assembly_name is [`mPhoSin1.pri`](https://github.com/popsim-consortium/stdpopsim/blob/3879666523fddd3cbb2cc3621ffaf8459b911e82/stdpopsim/catalog/PhoSin/genome_data.py#L4), 

The [annotation files](https://github.com/popsim-consortium/stdpopsim/blob/3879666523fddd3cbb2cc3621ffaf8459b911e82/stdpopsim/catalog/PhoSin/annotations.py#L11) for PhoSin include the assembly nam (`Phocoena_sinus.mPhoSin1.pri.110_CDS_v1.tar.gz`) and otherwise follow the conventions.

2. the following returns the assembly `species.get_contig("chr20", genetic_map = "HapMapII_GRCh38").genetic_map.assembly_name`.  Is that what you mean by visible in the catalog @petrelharp?

Seems like we're good to close?

PR #1654, opened 2025-01-17
lizard recomb rate estimate

### LINKED ISSUE #1365: Recombination rate placeholder used for AnoCar
https://github.com/popsim-consortium/stdpopsim/blob/eccf60ce12c8cde4d3550ea48a75948ff57be42d/stdpopsim/catalog/AnoCar/species.py#L23-L26

Is there a better placeholder that could be used here -- e.g. from another Anolis or reptile?

andrewkern:
recombination rate estimate for AnoCar from a linkage map for the common lizard, Zootoca vivipara, from this paper: https://doi.org/10.1093/gbe/evaa161

closes #1365

### DIFF
--- a/stdpopsim/catalog/AnoCar/species.py
+++ b/stdpopsim/catalog/AnoCar/species.py
@@ -20,10 +20,21 @@ _BourgeoisEtAl = stdpopsim.Citation(
     },
 )
 
-# No recombination rate yet for this species.
-# Author of BourgeoisEtAl is sending the recombination map
-# Placeholder rate of 1cM/Mb used
-_recombo_rate = 1e-8
+_YurchenkoEtAl = stdpopsim.Citation(
+    doi="https://doi.org/10.1093/gbe/evaa161",
+    year=2020,
+    author="Yurchenko et al.",
+    reasons={stdpopsim.CiteReason.REC_RATE},
+)
+
+# Recombination rate from Yurchenko et al.
+# which was calculated from a linkage map
+# from a different lizard species, Zootoca vivipara
+# the "common lizard"
+# they estimated male and female recombination rates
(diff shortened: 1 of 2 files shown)


petrelharp:
Hm - what do you think about this estimate, actually from Anolis carolinensis? [This paper](https://onlinelibrary.wiley.com/doi/full/10.1111/1755-0998.13780) says it's 0.52e-8 based on
Screenshot from 2025-01-18 07-10-46
which comes from [this paper](https://compcytogen.pensoft.net/articles.php?id=10916); however, the latter paper points out it's just in a male, and females often have higher recombination rates?

andrewkern:
I prefer the linkage map data I found above for a few reasons, I think.

Linkage map data is still the gold standard for estimating genome-wide rates for our uses I reckon, as they incorporate recombination in a number of individuals over generations. The immuno-histological / microscopy approach of the original paper here is good and I know people have shown they are correlated with linkage maps, but I get nervous about a sample of N=2 males (for instance we know the female map will be longer). 

OTOH the paper I found was a different species, but their linkage map was really dense and sampling seems excellent - 205 half-sib progeny collected from 20 wild caught dams.

petrelharp:
Hm - I still would guess the rougher estimate from the taxon in question is probably closer to the truth than this estimate (which differs threefold) in distant species? True that we expect females to be higher, but judging from the 'common lizard', not a ton higher (nothing approaching threefold).

andrewkern:
that's fine by me- we can use this other estimate if you prefer

dschride:
Given that the data seem to be higher quality for the common lizard, the rate is more in line with what we see in other vertebrates, and the genome sizes do not differ that much between Anolis and the common lizard, my thinking is that common lizard estimate is probably closer to the truth. Just my guess though.

PR #1655, opened 2025-01-17
exposed assembly name in catalog docs

### LINKED ISSUE #1651: add assembly name to the catalog
Right now each `Genome` object has an `assembly` property, but it's not visible in the catalog. It should be!

Also: PhoSin's annotation does not mention which assembly it is on in the description, unlike other annotations. (It's in the name, though.)

andrewkern:
this PR adds the `assembly_name` as a field of the `Genome` in the catalog docs. when rendered it looks like this:

image

this PR will close #1651 when merged


petrelharp (approved):
lgtm!

PR #1657, opened 2025-01-19
add pos & neg DFE from Rodrigues et al (PosNeg_R24 in HomSap)

petrelharp:
Motivation: we'd like to have a DFE containing positive mutations that aren't all at one fixed value of `s`.

[Rodrigues, Kern & Ralph](https://academic.oup.com/genetics/article/226/4/iyae006/7577595) doesn't exactly infer such a DFE, in the sense of using a validated inference procedure, but does explore parameter space and find a best-fitting DFE - so, I think it counts, and since it is simulation-based, we know it produces roughly reasonable simulations. The targets of inference are only the proportions of positive and negative selection; the negative part of the DFE is from [Castellano et al](https://academic.oup.com/genetics/article/226/4/iyae006/7577595), which I also cite here; the positive part is based on "expert knowledge" (i.e., something that seems reasonable).

So, in the popsim spirit of "best guess based on current state of knowledge", I think this is a good one to put in. Other opinions welcome (esp if they propose a better alternative such DFE?).

Note that the method uses all the great apes, and so could reasonably be applied to any of them.

### DIFF
--- a/stdpopsim/catalog/HomSap/dfes.py
+++ b/stdpopsim/catalog/HomSap/dfes.py
@@ -207,7 +207,6 @@ def _RodriguesDFE():
     description = "Deleterious Gamma and Beneficial Exponential DFE"
     long_description = """
     The best-fitting DFE simulated by Rodrigues et al (2024),
-    https://doi.org/10.1093/genetics/iyae006,
     from among 57 simulated scenarios with varying amounts of positive
     and negative selection. Fit was based on similarity of diversity
     and divergence across the great ape clade in simulations; the shape
--- a/stdpopsim/catalog/HomSap/dfes.py
+++ b/stdpopsim/catalog/HomSap/dfes.py
@@ -200,3 +200,66 @@ def _KyriazisDFE():
 
 
 _species.add_dfe(_KyriazisDFE())
+
+
+def _RodriguesDFE():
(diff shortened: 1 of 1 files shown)


petrelharp:
(Actually I was hoping @mufernando would QC this - I'll get a code review from someone else?)

petrelharp:
I've removed you from the assignee list, @mufernando, since I'm hoping you'll QC this? Say if you are actually going to review this PR instead.

mufernando:
I will QC this instead

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

stdpopsim/catalog/HomSap/dfes.py
```diff
@@ -200,3 +200,66 @@ def _KyriazisDFE():
...
 
 _species.add_dfe(_KyriazisDFE())
+
+
+def _RodriguesDFE():
+    id = "PosNeg_R24"
+    description = "Deleterious Gamma and Beneficial Exponential DFE"
+    long_description = """
+    The best-fitting DFE simulated by Rodrigues et al (2024),
+    https://doi.org/10.1093/genetics/iyae006,
```
petrelharp, line 210:
```suggestion
```

PR #1658, opened 2025-01-19
include long description in DFE catalog entries

petrelharp:
In doing this, I've also edited these descriptions to be readable (they were not). Here's what it looks like.
Screenshot from 2025-01-19 10-45-58


mufernando (approved):
LGTM. It's nice to be able to see the long descriptions in the docs!

PR #1663, opened 2025-01-20
clarify add_dfe docs (closes #1580)

### LINKED ISSUE #1580: Document what sort of intervals `contig.add_dfe` expects
In the docs https://popsim-consortium.github.io/stdpopsim-docs/stable/api.html#stdpopsim.Contig.add_dfe I think we should make it extra clear what sort of interval is expected; e.g. 0-based coordinate system and inclusive endpoint (I think that's the case based on the current text but am not entirely sure?)

petrelharp:
Also fixed the code display:
Screenshot from 2025-01-20 07-59-32

tldr; is that stdpopsim, like msprime/tskit, uses [a,b); while SLiM uses [a,b].

### DIFF
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -710,14 +710,17 @@ class Contig:
         object is added more than once.
 
         For instance, if we do
-        ```
-        a1 = np.array([[0, 100]])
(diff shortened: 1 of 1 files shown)


mufernando (approved):
Important clarification! LGTM.

Issue #1100, opened 2021-11-10
QC for Gamma_H17 (HomSap)


petrelharp:
**PR for new model:** https://github.com/popsim-consortium/stdpopsim/pull/1099
**Original paper:** https://doi.org/10.1073/pnas.1619508114

**Potential issues that might lead to differences between the model implementations:**
@chriscrsmith couldn't find the `proportions` that @xin-huang used in the paper.

**QC'er requests:** @xin-huang

xin-huang:
Should I QC this DFE? The proportions is in the supplementary https://www.pnas.org/doi/full/10.1073/pnas.1619508114#supplementary-materials

xin-huang:
Hi @petrelharp 

how to qc a DFE model? Should I follow [the review process for demographic models](https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#demographic-model-review-process) and put the codes in `./stdpopsim/qc/HomSap.py` ?

grahamgower:
This is a dupe of #1054.

xin-huang:
ok, but what are the codes [here](https://github.com/popsim-consortium/analysis2/tree/main/QC) for ?

petrelharp:
> ok, but what are the codes [here](https://github.com/popsim-consortium/analysis2/tree/main/QC) for ?

That's doing a comparison of the output of a simulation with a DFE to what's expected in some other way. Instead, we're writing code so that someone can re-implement the DFEs independently and then the code will automatically compare the original implementation to the QC implementation, and error on discrepancies. How about I'll get this started?

xin-huang:
> > ok, but what are the codes [here](https://github.com/popsim-consortium/analysis2/tree/main/QC) for ?
> 
> That's doing a comparison of the output of a simulation with a DFE to what's expected in some other way. Instead, we're writing code so that someone can re-implement the DFEs independently and then the code will automatically compare the original implementation to the QC implementation, and error on discrepancies. How about I'll get this started?

sounds good

petrelharp:
Well, that was annoying, but I finally figured out where the proportions [0.3, 0.7] come from. It's here:
![Image](https://github.com/user-attachments/assets/534b23e5-f40f-41c9-8d0a-5e8720440941)
confirmation:
```
> 1/(1 + 2.31)
[1] 0.3021148
```

The code should be changed to reflect this; i.e., instead of
```
        proportions=[
            0.3,
            0.7,
[...]

Issue #1054, opened 2021-10-22
QC for Gamma_H17 DFE in DroMel


petrelharp:
**PR for new DFE:** #1046
**Original paper:** [Huber et al 2017](https://www.pnas.org/content/114/17/4465)

petrelharp:
Note that the proportions of neutral & gamma ought to come from this snippet (see [this explanation](https://github.com/popsim-consortium/stdpopsim/issues/1100#issuecomment-2599989906) in #1100).
![Image](https://github.com/user-attachments/assets/5162ba6c-59d4-4dfa-83a3-73a67a9981cf)

(I have not checked.)

andrewkern:
hey @petrelharp @clararehmann -- is this issue now complete with the merge of #1668?

petrelharp:
Yes! Thanks.

PR #1668, opened 2025-01-21
Dromel qc

clararehmann:
Addresses #1054

I used the gamma distribution parameters from the full model (no singletons, recent mutation rate estimates) in table S2 and calculated the proportions of neutral and gamma from Peter's comment:

> Note that the proportions of neutral & gamma ought to come from this snippet (see [this explanation](https://github.com/popsim-consortium/stdpopsim/issues/1100#issuecomment-2599989906) in #1100).
> Image
> 
> (I have not checked.) 

 _Originally posted by @petrelharp in [#1054](https://github.com/popsim-consortium/stdpopsim/issues/1054#issuecomment-2599994485)_


petrelharp:
Looking at the "details" on the failing tests (above), there's one failing test:
```
=========================== short test summary info ============================
FAILED tests/test_dfes.py::TestDroMelGamma_H17::test_mutation_types_match - AssertionError: assert False
 +  where False = <function allclose at 0x7ff425f22c30>([-0.000396, 0.33], [-0.000792, 0.33])
 +    where <function allclose at 0x7ff425f22c30> = np.allclose
 +    and   [-0.000396, 0.33] = MutationType(dominance_coeff=0.5, distribution_type='g', distribution_args=[-0.000396, 0.33], convert_to_substitution=True, dominance_coeff_list=None, dominance_coeff_breaks=None).distribution_args
 +    and   [-0.000792, 0.33] = MutationType(dominance_coeff=0.5, distribution_type='g', distribution_args=[-0.000792, 0.33], convert_to_substitution=True, dominance_coeff_list=None, dominance_coeff_breaks=None).distribution_args
===== 1 failed, 2599 passed, 53 skipped, 20 warnings in 349.21s (0:05:49) ======
```
You can run (only) this test locally by doing
```
python -m pytest tests/test_dfes.py::TestDroMelGamma_H17::test_mutation_types_match
```
- it should give the same error on your computer. And, the discrepancy looks like a factor of two! Could you double-check, and if you still agree with your version, briefly explain?

clararehmann:
I get an error locally, I think it is a factor of two thing:

In my code, I use a -2 multiplier for the `gamma_mean` parameter to adjust it for `SLiM`, since per my reading it seems like the authors are usign `DaDi` for their simulations (Text S2):

```
gamma_shape = 0.33
gamma_scale = 1.2e-3
gamma_mean = gamma_shape * gamma_scale
h = 0.5  # dominance coefficient
negative = stdpopsim.MutationType(
        dominance_coeff=h,
        distribution_type="g",  # gamma distribution
        # (1+s for homozygote in SLiM versus 1+2s in dadi)
        distribution_args=[-2 * gamma_mean, gamma_shape],
    ...
    )
```
And in the original code for the DFE the mean is unscaled:

```
gamma_shape = 0.33  # shape
gamma_mean = -3.96e-04  # expected value
h = 0.5  # dominance coefficient
negative = stdpopsim.MutationType(
        dominance_coeff=h,
        distribution_type="g",  # gamma distribution
        distribution_args=[gamma_mean, gamma_shape],
    )
```
I'm not sure which is more correct based on the paper!

petrelharp:
> I'm not sure which is more correct based on the paper!

Hm, why aren't you sure? If they're using dadi, there should be a factor of 2 (as you say); is it not clear that the parameters are from dadi? (I'm not looking at Text S2 right now; you could paste in a screenshot of the paragraph in question?)

clararehmann:
This is the main text (referring to Text S2) where they say they use DaDi to infer _s_:

Screenshot 2025-01-22 at 12 21 45 PM

Based on this, it seems to me that there should be a factor of two in our code to convert to SLiM units?

petrelharp:
Ah, I see - they used dadi to fit the *demographic* model, but then they "infer the DFE from the nonsynonymous SFS". Looks like we need to have a look in that Appendix (and, hopefully not the code).

petrelharp:
Ah-ha: but *in* Text S2, they say
Screenshot from 2025-01-22 14-02-57
... in other words, they used dadi to do the calculations having to do with selection also. So, I agree with you about the factor of two.

petrelharp:
So - could you fix up the original implementation with this factor of two as well? (And, the way you've coded it is better: writing down explicitly the shape/scale values that are in the paper and multiplying them to get the mean.) Then this should pass!

clararehmann:
lol - I just took the same screenshot! I'll fix the implementation in the catalog as well :)

stdpopsim/qc/DroMel.py
```diff
@@ -115,3 +115,39 @@ def SheehanSongThreeEpic():
...
+def Gamma_H17():
+    id = "Gamma_H17"
+    description = "Deleterious Gamma DFE"
+    long_description = """
+    Return neutral and negative MutationType()s representing a drosophila DFE.
+    Huber et al. (2017), https://doi.org/10.1073/pnas.1619508114.
+    DFE parameters are based on the Full model described in Table S2, in which
+    singletons are excluded and a recent mutation rate estimate is used
+    (mu=3x10e-9, Keightley 2014).
+    """
```
petrelharp, line 129:
```suggestion
    long_description = "QC version"
```

petrelharp:
hah I was faster =)

stdpopsim/catalog/DroMel/dfes.py
```diff
@@ -29,20 +29,24 @@ def _HuberDFE():
...
     negative = stdpopsim.MutationType(
         dominance_coeff=h,
         distribution_type="g",  # gamma distribution
-        distribution_args=[gamma_mean, gamma_shape],
+        # (1+s for homozygote in SLiM versus 1+2s in dadi)
+        distribution_args=[-2 * gamma_mean, gamma_shape],
     )
-
+    # LNS = 2.85 * LS
+    # prop_synonymous = 1/(1+2.85) = 0.26
```
petrelharp, line 42: a bit more detail:
```suggestion
    # p. 2 in supplement says that the total sequence length of synonymous sites LS
    # related to the total sequence length of nonsynonymous sites LNS by LNS = 2.85 * LS;
    # so this is 1 / (1 + 2.85) = 0.2597402597402597
```

petrelharp:
Yay! Have a look at my suggestion; looks like we can merge once you accept that (or not?).

Issue #1649, opened 2025-01-08
QC for Gamma_B21 (MusMus)


petrelharp:
**PR for new model:** #1556 
**Original paper:** [Booker et al](https://www.biorxiv.org/content/10.1101/2021.06.10.447924v2)

@peterdfields - could you provide additional context for where to find parameter values in the paper, and additional issues QC'ers might encounter?

PR #1669, opened 2025-01-23
gamma b21 QC

clararehmann:
Addresses #1649

I've replicated the Gamma DFE from [Booker et al. (2021)](https://doi.org/10.1101/2021.06.10.447924), using the shape and mean parameters from Table S2 and basing the neutral:negative proportion off of the transition:transversion ratio in Table S1. 

The authors use PolyDFE to estimate the scaled selection coefficient:

> Using polyDFE, the DFE is estimated in terms of the scaled selection coefficient for deleterious mutations, 2Nesd, where sd is the reduction in fitness experienced by an individual homozygous for the mutation (which is assumed to be semi-dominant).

I divided the scaled mean of the gamma distribution by the effective population size quoted in the paper (420,000) to obtain 2sd, then divided by two to get the _s_ coefficient used for homozygotes in SLiM.


clararehmann:
It seems like the tests are failing bc I included more sig figs - is there a standard for how many to use? I just copy and pasted values from the provided supplementary tables

petrelharp:
Ah, I see: the second one is yours.
```
FAILED tests/test_dfes.py::TestMusMusGamma_B21::test_mutation_types_match - AssertionError: assert False
 +  where False = <function allclose at 0x7fc47dd367b0>([-0.0596, 0.186], [-0.05957688452380952, 0.18617976])
```

We don't have a standard. In general I'd think "as many digits as they report"? But, the third digit or so here is not going to make a difference, so what they had isn't *wrong*. I'd say change the main implementation to agree with yours, after checking they agree up to rounding? The other consideration is if the catalog looks ugly - but I think the values printed there are rounded? If not, maybe we shoudl round?

petrelharp:
That's not the isuse now:
```
=========================== short test summary info ============================
FAILED tests/test_dfes.py::TestMusMusGamma_B21::test_proporitions_match - assert False
 +  where False = <function allclose at 0x104c3b1b0>([0.334, 0.666], [np.float64(0.772), np.float64(0.228)])
 +    where <function allclose at 0x104c3b1b0> = np.allclose
===== 1 failed, 2601 passed, 53 skipped, 20 warnings in 278.46s (0:04:38) ======
```

petrelharp:
If you look at your parameters again and still don't agree with what the original implementer put in, then ping them here (and maybe elsewhere too) to consult?

clararehmann:
Okay I figured out where the 0.33:0.66 ratio for neutral and negative mutations is coming from, it's in the original way they parse DFE info, line 74 in https://github.com/TBooker/MuridRodentProject/blob/master/bin/SlimFunctions.py:

```
## Now, we read in a file that contains the DFE information for each of the genomic elements featured in your BED file
## This parser assumes that there are only deleterious mutations and that all the DFE are gamma distributions
## The parse also assumes that there are 1/3 neutral sites in the CDS class. However, this is a variable and you can change it if you wish
```

So I'll fix that up in my QC and things should be good!

stdpopsim/qc/MusMus.py
```diff
@@ -286,3 +286,40 @@ def QC_CastaneusIndia():
...
+        dominance_coeff=h,
+        distribution_type="g",
+        # PolyDFE gives two times the selection coefficient on a homozygote,
+        # so divide mean by two
+        distribution_args=[
+            np.around(gamma_mean / 2, decimals=4),
+            np.around(gamma_shape, decimals=3),
+        ],
+    )
+    prop_neutral = 0.334
```
petrelharp, line 315: Could you stick in a comment with some explanation of where `0.334` comes from?

petrelharp:
Great, nice work! I think we just need an explanatory comment (since as you've found out it's not obvious...).

petrelharp:
Looks good, thanks! I will squash & merge this.

PR #1674, opened 2025-01-29
get release / version info from setuptools_scm

### LINKED ISSUE #680: docs/conf.py has a hardcoded (and wrong) version string
This should be read from the version in `stdpopsim/_version.py`. E.g.
https://github.com/tskit-dev/tskit/blob/main/docs/conf.py#L52-L53

andrewkern:
this changes the `doc/conf.py` to pull the version and release info automatically from the `setuptools_scm` interface as requested in #680 

will close #680

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,8 +12,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
+import datetime
 import os
 import sys
+from setuptools_scm import get_version
 
 sys.path.insert(0, os.path.abspath(".."))
 # Add the path for local extensions
@@ -22,13 +24,16 @@ sys.path.append(os.path.abspath("./_ext"))
 # -- Project information -----------------------------------------------------
 
 project = "stdpopsim"
-copyright = "2019-2022, PopSim Consortium"
+copyright = f"2019-{datetime.datetime.now().year}, PopSim Consortium"
 author = "PopSim Consortium"
 
-# The short X.Y version
-version = ""
-# The full version, including alpha/beta/rc tags
-release = "0.2.0"
+# get the release version from the setuptools_scm
+try:
+    release = get_version(root="..", relative_to=__file__)
+    version = ".".join(release.split(".")[:2])
+except Exception:
+    release = "unknown"
+    version = "unknown"
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -93,7 +98,10 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 #
-# html_theme_options = {}
+
+# Show the version number in the navigation bar
```
nspope, line 102: Version doesn't seem to be printed anywhere in local doc build. Seems like how versioning is handled may have changed: https://stackoverflow.com/a/79192065

and in particular, that it may now be only via a version selector that is only available if docs are hosted by RTD. Guess we could merge and see if this is the case (?)

I think we want the version information printed somewhere. One option would be to manually print the version somewhere, by putting `:Version: |version|` into the rst

docs/conf.py
```diff
@@ -22,13 +24,16 @@
...
 author = "PopSim Consortium"
 
-# The short X.Y version
-version = ""
-# The full version, including alpha/beta/rc tags
-release = "0.2.0"
+# get the release version from the setuptools_scm
+try:
+    release = get_version(root="..", relative_to=__file__)
+    version = ".".join(release.split(".")[:2])
```
nspope, line 33: What's the `[:2]` intended to chop out here? (Note that I can't get the version number to display anywhere in a local docs build). But, if I run this via the REPL, I get that `release == '0.2.1a2.dev65+g2d5ba65'` and `version == 0.2`. 

I don't think we want to chop off this much, if anything? That is, stable builds should have a nice looking version number (so no need to chop anything off). Unstable builds will have the commit hash, which looks ugly but is providing necessary information IMO (so also don't want to chop that off).

nspope (commented):
Looks good but a couple points to clarify on the actual display

docs/conf.py
andrewkern: yeah that's what i was intending-- that version would be `version == 0.2` and release would be that blob. I've pushed a commit that manually adds a footer with all this info displayed. also I *think* once we merge the version selector business will appear on RTD

andrewkern:
okay i've added a custom footer now which adds this to the bottom of the RTD pages:

image

@nspope note that the version / release is now correct. when I showed you this earlier my feature branch did not have the tags properly fetched on to it, which is what `setuptools_scm` looks at. 

I'm guessing that when we merge this the version selector option of the sphinx_rtd_theme will also just work, so I think this is ready to merge

Issue #320, opened 2019-12-10
Reconsider "sampling population" terminology


jeromekelleher:
We're using ``sampling_time=None`` currently to denote populations that are never sampled from. In reality we should be able to sample from any population, it's mainly just a case of what time thoses samples should be at. This gets complicated though, since someone may want to sample at multiple times from the same population. Then, we can just use the standard msprime API for creating samples.

So, perhaps the current approach is a good compromise, covering the majority of cases people will want?

ndukler:
I agree. We can even put together a tutorial on how to customize sampling.

jeromekelleher:
@grahamgower, this has come up as the main sticking point in #764. Is there a good reason for never allowing samples from a population, or is it just because we don't have the right times?

grahamgower:
I think we should be able to sample from any population at any time over it's period of existence. But there's a problem if we change to this now, because we already have a notion of populations that can't be sampled. The real kicker is that this has implications for the number and order of CLI parameters used for sampling.

jeromekelleher:
OK, thanks. I guess we'll need to think about the models in question so that we can be sure of keeping compatibility with the CLI.

petrelharp:
Some notes, since I ran across this tracking something else down.
Take [the human `AncientEurope_4A21` model](https://github.com/popsim-consortium/stdpopsim/blob/a011bf190a1f386ac1a17c35959b05a24da81283/stdpopsim/catalog/HomSap/demographic_models.py#L1793).
This has ancestral populations, for instance:
```
eur = stdpopsim.get_species("HomSap").get_demographic_model("AncientEurope_4A21")
print(eur.model.populations[0])
# Population(initial_size=5000, growth_rate=0, name='OOA', description='Basal/OOA', extra_metadata={}, default_sampling_time=1500, initially_active=False, id=0)
```
Note it has `initially_active=False`. However, we can sample from it; in fact, adding a `print(samples)` to `test_simulation_runs` gets
```
# from tests/test_models.py::TestHomSapAncientEurope_4A21::test_simulation_runs
{'OOA': 2, 'NE': 2, 'WA': 2, 'CHG': 2, 'ANA': 2, 'WHG': 2, 'EHG': 2, 'YAM': 2, 'NEO': 2, 'Bronze': 2}
```
[...]

PR #1676, opened 2025-02-03
adds easier access to sampling populations

kevinkorfmann:
Adds function to get list valid populations to sample from. Previously, we had `model.populations,` now have `model.sampling_populations` as well.

### DIFF
--- a/stdpopsim/models.py
+++ b/stdpopsim/models.py
@@ -171,6 +171,10 @@ class DemographicModel:
         return self.model.populations
 
     @property
+    def sampling_populations(self):
+        return [pop for pop in self.populations if pop.allow_samples]
+
+    @property
     def num_populations(self):
         return self.model.num_populations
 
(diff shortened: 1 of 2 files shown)


petrelharp:
Good idea; what's your thoughts on how this interacts with #320?  (If this new method will soon become deprecated, I'm not very enthusiastic about adding it...)

kevinkorfmann:
the idea was to add a tiny section in the docs under the tutorial section next, for sampling, e.g. 1 MB segments for all demographic models randomly distributed throughout the genome (under the wider context of creating diverse datasets quickly), using this random_sample_counts method - if it will be possible to sample from all time points in the future, then of course we would need to change model.sampling_populations to model.populations, or adapt the method otherwise.

petrelharp:
> Which lead to very large memory footprints, during attempting to simulate a large dataset with slim engine, so feel free to close it.

That's odd - I don't see why archaic populations specifically would lead to a large memory footprint? I'll close this (you should be able to), but feel free to re-open.

Issue #1056, opened 2021-10-22
QC for Gamma_K17 DFE in HomSap


petrelharp:
The original paper is [Kim et al. (2017)](https://doi.org/10.1534/genetics.116.197145), and this is the "deleterious gamma" model.  This issue obviously depends on #1055.

petrelharp:
Closed in #1687

PR #1687, opened 2025-02-22
Kim 2017 DFE

clararehmann:
QC for the gamma distribution DFE from Kim et al. 2017 (https://doi.org/10.1534/genetics.116.197145). I got the distribution parameters from the gamma model in table S4 for the 1kG EUR dataset and scaled them according to the ancestral population size for the same dataset in table S2 (where I also got the neutral:negative proportions from).


clararehmann:
Oh and this addresses #1056 :)

petrelharp:
Nice! Looks good!

Issue #1690, opened 2025-02-24
QC for GammaPos_H17 (DroMel)


petrelharp:
**PR for new model:** #1656 
**Original paper:** https://genome.cshlp.org/content/31/1/110.abstract
**Parameter values:** see discussion in #1469 

**Potential issues that might lead to differences between the model implementations:**
none; note that this has essentially been QC'ed as I independently got the parameter estimates

**QC'er requests:**

PR #1695, opened 2025-02-25
DroMel positive DFE

clararehmann:
Addresses #1690 

I used the Huber et al. 2017 gamma-distributed DFE with a proportion of nonsynonymous mutations being positively selected. Parameters for the positively-selected mutations were from Table S6 in the [Zhen et al. paper.](https://genome.cshlp.org/content/31/1/110.abstract)

### DIFF
--- a/stdpopsim/catalog/DroMel/dfes.py
+++ b/stdpopsim/catalog/DroMel/dfes.py
@@ -77,12 +77,13 @@ def _ZhenDFE():
     ]
     neutral = stdpopsim.MutationType()
     gamma_shape = 0.33  # shape
-    gamma_mean = -3.96e-04  # expected value
+    gamma_scale = 6.01e-4
+    gamma_mean = gamma_shape * gamma_scale
     h = 0.5  # dominance coefficient
     negative = stdpopsim.MutationType(
         dominance_coeff=h,
         distribution_type="g",  # gamma distribution
-        distribution_args=[gamma_mean, gamma_shape],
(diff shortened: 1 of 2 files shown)


clararehmann:
@petrelharp - I noticed that in this implementation you used the Huber 2017 model w/o singletons, but in the implementation of the HomSap DFE with positive mutations (#1694) you used the full model. Is this on purpose?

petrelharp:
Gee - I don't remember this at all, and haven't been able to reconstruct what this means - could you bump my memory? (What  "singletons"?)

petrelharp:
The models seem very parallel to me?
Screenshot from 2025-02-24 17-13-06

petrelharp:
Oh wait I understand. Hm let's see.

petrelharp:
For reference, table S2 from H17: 
Screenshot from 2025-02-24 17-17-37

petrelharp:
And, here's the table from Zhan et al showing which parameters they actually used:
Screenshot from 2025-02-24 17-27-52

As you note, Zhen et al (according to this table) are using the "all data" estimates for both species, but for some reason I've taken the "no singletons" numbers for DroMel. Good catch! 

I think the parameters for DroMel should be
```
    gamma_shape = 0.35  # shape
    gamma_mean = -1.33e-04  # expected value
```

I'm getting the "mean" values from the H17 table, rather than taking beta from the Z24 table and multiplying. Please check! (I also got worried I mixed up 'mean' and 'beta' for HomSap, but it looks okay!)  If you agree, could you swap those in, for both the implementation and QC?

Issue #1507, opened 2023-08-14
New DFE: "generic" human DFE


chriscrsmith:
**PR for new model:** 
#1505 

**Original paper:** 
preprint: [Kyriazis et al. 2022](https://www.biorxiv.org/content/10.1101/2022.08.12.503792v1)
ahead of print: [Kyriazis et al. 2023](https://doi.org/10.1086/726736)

**Parameter values:**
Using values from "Description of simulation models" in supp material section 4:
- Gamma mean and shape: 0.0131 and 0.186
- Discretized h-s relationship: coefficients [0.45,  0.2,  0.05, 0] and breaks [0.001, 0.01, 0.1]
- proportion_deleterious = 0.63 / (3e7 * mu * 2)
- proportion 0.003 lethal mutations

**Potential issues:**
- Called the DFE "Mixed_K23", because it's not simply gamma, but it's gamma+other.
- The proposed model calls for a deleterious mutation rate of 0.63. My interpretation is, the 0.63 deleterious rate is agnostic to the neutral mutation rate and the proportion of deleterious mutations, and it's up to the practitioner to decide which parameter to change. I figured the mutation rate is a more confident estimate than the proportion deleterious, so leave the mutation rate as 1.29e-8 (from the human catalog), and instead adjust the proportion of deleterious mutations to get 0.63. And this value was calculated on 30Mb of diploid sequence.
- The proportion 0.003 lethal is maybe up to interpretation: proportion of deleterious mutations, versus proportion of all mutations?

PR #1696, opened 2025-02-25
generic DFE from kyriazis

clararehmann:
Addresses #1507 -- I used the parameters (with discrete dominance coefficients) from supplemental section S6 in [Kyriazis et al. 2023](https://doi.org/10.1086/726736), with a 0.3% of *nonsynonymous* mutations being lethal. 
I used the 2.31:1 nonsynonymous:synonymous ratio described in Huber et al. as per the discussion in #1507.

### DIFF
--- a/stdpopsim/qc/HomSap.py
+++ b/stdpopsim/qc/HomSap.py
@@ -1784,3 +1784,45 @@ def ZhenPos():
 
 
 _species.get_dfe("GammaPos_Z21").register_qc(ZhenPos())
+
+def genericDFE():
+    """
+    gamma distribution based on Kyriazis et al. 2022
+    w/ discrete dominance coefficients
+    and lethal mutations
+    """
+    id = "generic_gamma_dfe"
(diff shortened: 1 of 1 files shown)


stdpopsim/qc/HomSap.py
```diff
@@ -1747,3 +1747,46 @@ def Kim2017():
...
+        distribution_type="f",
+        distribution_args=[-1],
+        dominance_coeff=0,
+    )
+
+    ns_proportion = 2.31 / (2.31 + 1.0)
+    lethal_proportion = 0.003 * ns_proportion
+    ns_proportion = ns_proportion - lethal_proportion
+    return stdpopsim.DFE(
+        id=id,
```
petrelharp, line 1780: I think the error is because you haven't defined `id`, and
```
$ python3
Python 3.13.1 (main, Jan  3 2025, 10:26:34) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> id
<built-in function id>
```

Issue #1484, opened 2023-05-18
QC for LogNormalDFE (HomSap)


RyanGutenkunst:
**PR for new model:** [PR](https://github.com/popsim-consortium/stdpopsim/pull/1480)
**Original paper:** [paper](https://doi.org/10.1073/pnas.1619508114)
**Parameter values:** DFE parameters are based off the Full model in Table S3, using recent mutation rate estimates
**Population indices:** N/A

**Potential issues that might lead to differences between the model implementations:**
SLiM and dadi use different conventions for selection coefficients. In SLiM, the homozygote has fitness 1+s; in dadi it is 1+2s. Since this model was inferred using dadi, to map it to SLiM we need to double the selection coefficient, by shifting the center of the lognormal distribution.

**QC'er requests:** @petrelharp

petrelharp:
Closed in #1698

PR #1698, opened 2025-02-25
LogNormal DFE for humans

clararehmann:
addresses #1484

I used table S3 from [Huber et al. 2017](https://doi.org/10.1073/pnas.1619508114) for the parameters and refactored the mu value of the LogNormal distribution to account for DaDi vs. SLiM scaling. I also modified the DFE tests so that they can accept LogNormal distribution arguments (which are strings, not floats).


petrelharp:
Excellent!

Issue #1697, opened 2025-02-25
QC for LognormalPlusPositive_R16 (DroMel)


petrelharp:
**PR for new model:** #1178 
**Original paper:** https://doi.org/10.1534/genetics.115.184812
**Parameter values:** Table S1

**Potential issues that might lead to differences between the model implementations:**

The re-scaling of the log-normal scale parameter should be double-checked, since the mean of the log-normal is given as 2Ne*s. Pretty sure I just need to divide the scale by 2Ne, but this is based on the scipy definition of the log-normal (which should be standard), and I'm not 100% sure if there are differences in the implementation here.

**QC'er requests:** @clararehmann

petrelharp:
Also note the discussion around scaling by Ne in #1178.

PR #1699, opened 2025-02-26
dromel lognormal DFE

clararehmann:
Addresses #1697

I used the parameters from tables S1 (LogNormal distribution parameters, positive selection coefficient) and S2 (proportion of NS mutations that are positive). Because the parameters in S1 are scaled by 2Ne, I divided them by 2Ne using the effective population size from Huber et al. 

Since the paper used DaDi to infer their DFEs, I rescaled the mu term so that the expected mean (X = exp(mu + (sigma^2)/2)) of the LogNormal distribution was two times the inferred mean (after scaling by 2Ne).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@
   in growing/declining populations not matching (by a small amount) what is listed
   in the model; improved discretization scheme (:user:`petrelharp`, :pr:`1622`)
 
+- Incorrect scaling of the DroMel/LognormalPlusPositive_R16 DFE led to negative selection
+  coefficients that were too large (on the order of 1e-3); scaling was fixed for the
+  log scale (:user:`clararehmann`, :pr:`1699`)
+
 **Breaking changes**:
 
 - The 2018 Browning et al HomSap demographic model previously named
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,7 @@
   in the model; improved discretization scheme (:user:`petrelharp`, :pr:`1622`)
 
 - Incorrect scaling of the DroMel/LognormalPlusPositive_R16 DFE led to negative selection
-  coefficients that were too large (on the order of 1e-3); scaling was fixed for the
+  coefficients that were too large (many on the order of -1e3); scaling was fixed for the
   log scale (:user:`clararehmann`, :pr:`1699`)
 
 **Breaking changes**:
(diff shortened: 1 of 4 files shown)


clararehmann:
@petrelharp I'm not sure this is the correct way to rescale this sort of distribution?

petrelharp:
Looks like you should not be doing exp( ), since [the parameters are on the log scale](https://popsim-consortium.github.io/stdpopsim-docs/latest/api.html#stdpopsim.MutationType) - it is parameterized in the same way as SLiM.

(However, now I'm confused why your other lognormal one passed - got to go look at that.)

petrelharp:
Oh, I see! Sorry, I'm moving too quickly here. And, very nice catch! You're right on, there's just a slightly simpler way to do it.

Here you've got
```
    muval = -5.42 / Ne / 2
    sigmaval = 3.36 / Ne / 2
    expectedmean = math.exp(muval + (sigmaval**2 / 2))
    targetmean = expectedmean * 2
    muval = math.log(targetmean) - (sigmaval**2 / 2)
```
So: note that if
$$x = \exp(\mu + \sigma^2/2)$$
then
$$2x = \exp(\log(2) + \mu + \sigma^2/2)$$,
so a better way to do this is just to add `math.log(2)` to `muval`.

And, this is in fact what we do in the original code - for instance, here's the QC code for the HomSap lognormal DFE compared to the library code:
Screenshot from 2025-02-26 11-53-38

So, actually - I think you should just leave it the way you have it! It's a better test, because we see we're getting the same answer doing it a different way.

petrelharp:
Okay - the difference is coming because you're dividing sigma by 2Ne, but the original code does not. Huh - now I need to go back and lookat Ragsdale to figure out what the right thing is.

petrelharp:
Putting this here for reference:
Screenshot from 2025-02-26 12-07-55

petrelharp:
Great! This one needs a rebase and an entry in the `CHANGELOG.rst` for that bugfix.

CHANGELOG.rst
```diff
@@ -16,6 +16,10 @@
   in growing/declining populations not matching (by a small amount) what is listed
   in the model; improved discretization scheme (:user:`petrelharp`, :pr:`1622`)
 
+- Incorrect scaling of the DroMel/LognormalPlusPositive_R16 DFE led to negative selection
+  coefficients that were too large (on the order of 1e-3); scaling was fixed for the
+  log scale (:user:`clararehmann`, :pr:`1699`)
```
petrelharp, line 21:
```suggestion
- Incorrect scaling of the DroMel/LognormalPlusPositive_R16 DFE led to negative selection
  coefficients that were too large (many on the order of -1e3); scaling was fixed for the
  log scale (:user:`clararehmann`, :pr:`1699`)
```

petrelharp:
Small edit to the CHANGELOG; then it's ready to go!

petrelharp:
Whoops - looks like it needs a rebase.

PR #1700, opened 2025-03-01
Rattus norvegicus

kevinkorfmann:
This contains the basic information for the brown rat.


andrewkern:
i see the recombination rates vary per chromosome here @kevinkorfmann -- is there a finer scale map that we could include?

kevinkorfmann:
There is a genetic map from 2018 consisting of 95k markers, where I got the average recombination rates per chromosome above, which is from the assembly Rnor_6.0. 

Also, I found a newer paper published last year reporting a new assembly (mRatBN7.2) and genetic map. However, this new genome, despite containing improvements, is not a telomere to telomere assembly, and they have still some disassembled regions in their data which cause some issues for their genetic map. 

The first genetic map, is already corrected and can be use directly (and also lifted over to mRatBN7.2), while the newer version would require some correction or filtering to make it useable. **So, I would suggest sticking to the correct 2018 version for now.**

2018 map: https://pmc.ncbi.nlm.nih.gov/articles/PMC6027877/
2024 map: https://www.sciencedirect.com/science/article/pii/S2666979X24000697

kevinkorfmann:
I've uploaded the genetic map as well as the liftover command to: https://github.com/kevinkorfmann/genetic_map_rn6

andrewkern:
okay let's hold off on getting the genetic map in until the species gets added to the catalog. 

One thing I see missing is a citation for the generation time. Can you please add one and then we can incorporate this species and trigger the QC process?

Also a note-- you've filled in values in the QC tests. Please remove those for this initial PR so that they can be filled in by the QC reviewer blind to what you've come up with. The idea behind the QC is that the reviewer follows the citations for each of the quantities and independently fills in the quantities. This turns out to be a good error check in practice.

stdpopsim/catalog/RatNor/species.py
```diff
@@ -0,0 +1,144 @@
...
+)
+
+_species = stdpopsim.Species(
+    id="RatNor",
+    ensembl_id="rattus_norvegicus",
+    name="Rattus norvegicus",
+    common_name="Rat",
+    genome=_genome,
+    ploidy=_species_ploidy,
+    generation_time=0.5,
```
andrewkern, line 132: need a citation here

kevinkorfmann: Thanks -- added two citations for the generation time and resetted the test_file now

stdpopsim/catalog/RatNor/species.py
```diff
@@ -0,0 +1,156 @@
...
+            author="Littrell et al.",
+            year=2018,
+            doi="https://doi.org/10.1534/g3.118.200187",
+            reasons={stdpopsim.CiteReason.REC_RATE},
+        ),
+        stdpopsim.Citation(
+            author="Deinum et al.",
+            year=2015,
+            doi="https://doi.org/10.1093/molbev/msv126",
+            reasons={stdpopsim.CiteReason.MUT_RATE},
```
andrewkern, line 120: sorry one more, nitpick here-- can you also add the citation to the mtDNA rate as a separate citation and list it's reason as MUT_RATE? It's great that you've added mito mutation rate separately but we want to be able to track that citation in the same way. It's fine btw to have two citations listed for the same 'reason'

andrewkern:
added one more comment @kevinkorfmann -- once you make that small addition I think we can get this in and start QC

stdpopsim/catalog/RatNor/species.py
```diff
@@ -0,0 +1,156 @@
...
+        stdpopsim.Citation(
+            author="Deinum et al.",
+            year=2015,
+            doi="https://doi.org/10.1093/molbev/msv126",
+            reasons={stdpopsim.CiteReason.POP_SIZE},
+        ),
+        stdpopsim.Citation(
+            author="Deinum et al.",
+            year=2015,
+            doi="https://doi.org/10.1093/molbev/msv126",
```
petrelharp, line 144: Is this the right citation? They just cite [Ness](https://academic.oup.com/g3journal/article/2/12/1661/6028550) and [Halligan](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1003995) for the generation time of 2/year.

petrelharp: and hm, those two papers just assert it:
Screenshot From 2025-03-02 17-33-30
Screenshot From 2025-03-02 17-33-04

stdpopsim/catalog/RatNor/species.py
```diff
@@ -0,0 +1,156 @@
...
+            reasons={stdpopsim.CiteReason.GEN_TIME},
+        ),
+        stdpopsim.Citation(
+            author="Jing et al.",
+            year=2022,
+            doi="https://doi.org/10.1002/ece3.9409",
```
petrelharp, line 150: And this one just cites Deinum (the previous citation), so no need for it at al, AFAICT?
Screenshot From 2025-03-02 17-34-14

stdpopsim/catalog/RatNor/species.py
petrelharp: well this makes 2/year plausible
Screenshot From 2025-03-02 17-37-28

petrelharp: this also makes that seem plausible
Screenshot From 2025-03-02 17-40-47

petrelharp: For sure an estimate of actual generation time could be obtained from [this gold mine of information](https://www.journals.uchicago.edu/doi/10.1086/399860) - for instance, I see that about 29.1% of female rats are pregnant at any given time; that sort of number combined with numbers about age of first litter and average lifespan would I think do it? This may be beyond the scope, but it'd be worth seeing if this is easily obtainable from that paper.

andrewkern: wow that's an awesome resource -- will be great to have rat in the catalog

stdpopsim/catalog/RatNor/species.py
kevinkorfmann: Thanks for mentioning, Peter. Removed this citation now.

stdpopsim/catalog/RatNor/species.py
kevinkorfmann: no problem, thanks for the advice -- I added the citation as well

stdpopsim/catalog/RatNor/species.py
andrewkern: everyone seems to be using 2 generations per year back to the Halligan et al. 2010 paper. I think that's the citation we should use. In looking at the Davis (1953) paper there isn't much concrete other than getting some demographic parameters which will give us a number that doesn't quite agree with what people in popgen have been doing.

andrewkern: more research on this.... this paper nails down two spikes in reproduction per year in a wild rat population: https://doi.org/10.1674/0003-0031(2006)155[0221:GEOARP]2.0.CO;2

these two papers seem to again assert 2gens/yr: [DOI: 10.1126/science.290.5490.331](https://doi.org/10.1126/science.290.5490.331), [10.1073/pnas.2233252100](https://doi.org/10.1073/pnas.2233252100), but the first one cites a couple of mouse ecological studies. e.g., https://doi.org/10.2307/1379439

going in to older literature it seems that Wen-Hsiung Li and colleagues were using ~2gen per year to calibrate generation time effects in the 1990s e.g., doi.org/10.1006/mpev.1996.0012

my take home-- we have a choice here to either cite the older popgen literature (perhaps Keightley and Eyre-Walker) or we could cite that ecological paper above. Both justify this parameter choice

kevinkorfmann: appreciate you digging into it. I really like the ecological justification after having had a look at the figures in Fig 2 and Fig 3 of McGuire et al. (2006)

PR #1517, opened 2023-09-18
Adding GorGor species and demography

ChristianHuber:

PR #1701, opened 2025-03-02
QC for GorGor

andrewkern:
**QCing the species addition for GorGor #1517** 

this pertains to issue #1659 as well

I've gone through and everything looks good except for the mutation rates initially drafted by @ChristianHuber 

The mutation rate estimate in the catalog currently is 1.235e-08. Looking at the [citation](https://www.nature.com/articles/s41559-018-0778-x) I see that they sequenced two gorilla trios, one of which had a rate of 1.03e-08 and the other which was at 1.22e-08. Averaging these I came up with a rate of 1.125e-08. 

@ChristianHuber -- do you see different numbers reported?

### DIFF
--- a/tests/test_GorGor.py
+++ b/tests/test_GorGor.py
@@ -18,51 +18,46 @@ class TestSpeciesData(test_species.SpeciesTestBase):
         assert self.species.common_name == "Gorilla"
 
     # QC Tests. These tests are performed by another contributor
-    # independently referring to the citations provided in the
-    # species definition, filling in the appropriate values
-    # and deleting the pytest "skip" annotations.
-    @pytest.mark.skip("Population size QC not done yet")
     def test_qc_population_size(self):
-        assert self.species.population_size == -1
+        assert self.species.population_size == 25200
 
-    @pytest.mark.skip("Generation time QC not done yet")
     def test_qc_generation_time(self):
-        assert self.species.generation_time == -1
+        assert self.species.generation_time == 19.0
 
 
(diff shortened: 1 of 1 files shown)


andrewkern:
Also for recombination rates, I see that chromosome X was given the genome average rate (which we agree on) but no data in the citation was given about X-linked recombination rates. In my QC PR I've set it to zero, so this will fail tests. I'm not sure as to the best way to proceed here. Perhaps @petrelharp has an opinion of what to do in this case?

petrelharp:
Lacking other information, we set the rate on the X to the autosomal rate, I'd say - pretty sure this is what we do elsewhere, and this will be closer to correct than 0.

petrelharp:
I know what it is! Those two reported mutation rates are for particular gorillas with particular parental ages; they've then used assumptions about how the mutation rate varies with age to extrapolate to an average mutation rate across time. They report 0.65e-9 mutations/year and a generation time of 19 years (as in [the catalog](https://popsim-consortium.github.io/stdpopsim-docs/latest/catalog.html#sec_catalog_GorGor)); and 1.235e-8 = 0.65e-9 * 19.

Screenshot From 2025-03-02 17-26-53

This seems pretty reasonable to me, FWIW. (Clearly, though, the calculation should be documented.)

petrelharp (commented):
LGTM except for issues discussed

andrewkern:
Nice! So I'll make these changes on the QC and we should be good to go here

PR #1646, opened 2025-01-07
changes to maintenance script / ensembl push

andrewkern:
This PR addresses (closes?) #1521.

think i've finally cleaned this up a bit. this PR overhauls the `update-genome-data` portion of the maintenance script such that:

- species whose genomes were manually created are skipped
- species where the ensembl Rest API returns mismatching chromosome names with the current release are skipped
- detailed logging warnings of these issues are created, along with a final report of which species were skipped in the update and why. 

the final report looks like this:

```bash
=== Species Update Summary ===
2025-01-07 14:34:46,744 [91493] WARNING  maint: The following species were not updated:
2025-01-07 14:34:46,744 [91493] WARNING  maint:   - AnoCar (Ensembl ID: anolis_carolinensis):
2025-01-07 14:34:46,744 [91493] WARNING  maint:     Chromosome names mismatch.
2025-01-07 14:34:46,744 [91493] WARNING  maint:     Existing chromosomes: ['1', '2', '3', '4', '5', '6', 'LGa', 'LGb', 'LGc', 'LGd', 'LGf', 'LGg', 'LGh', 'MT']
2025-01-07 14:34:46,745 [91493] WARNING  maint:     New chromosomes: ['1', '2', '3', '4', '5', '6', 'a', 'b', 'c', 'd', 'f', 'g', 'h']
2025-01-07 14:34:46,745 [91493] WARNING  maint:   - CanFam (Ensembl ID: canis_lupus_familiaris):
2025-01-07 14:34:46,745 [91493] WARNING  maint:     Chromosome names mismatch.
2025-01-07 14:34:46,745 [91493] WARNING  maint:     Existing chromosomes: ['1', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '2', '20', '21', '22', '23', '24', '25', '26', '27', '28', '29', '3', '30', '31', '32', '33', '34', '35', '36', '37', '38', '4', '5', '6', '7', '8', '9', 'MT', 'X']
2025-01-07 14:34:46,745 [91493] WARNING  maint:     New chromosomes: ['1', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '2', '20', '21', '22', '23', '24', '25', '26', '27', '28', '29', '3', '30', '31', '32', '33', '34', '35', '36', '37', '38', '4', '5', '6', '7', '8', '9', 'X', 'Y']
[...]

PR #1709, opened 2025-03-11
changelog update

petrelharp:
Here's the updated changelog for the next release, which we were planning would be 0.2.1.

However, @andrewkern is proposing it's 0.3, which seems like a good idea?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,6 @@
----------------------
+--------------------
 [0.3.0] - 2025-03-18
----------------------
+--------------------
 
 Major feature release adding documented, stable support for simulating selection (via SLiM),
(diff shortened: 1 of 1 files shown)


andrewkern:
a couple of things:
- yes to version 0.3
- i'm not sure about what all gets included in the change log but do we want to add in changes to docs like #1655? or changes to maintenance stuff like #1646

petrelharp:
No need for those minor docs/maintenance stuff, I think.

petrelharp:
Also FWIW I don't see what you mean about selection stuff not being visible in the docs, @andrewkern - here I'm seeing selection in `stable` which points to 0.2.0?
Screenshot From 2025-03-18 19-12-03

Issue #1713, opened 2025-03-18
Add selection paper citation to the docs!


mufernando:
We can't forget to add the <s>new preprint</s> selection paper citation to the docs once it's in.

*Edit:* the preprint is in; leaving this open for the publisehd version.

gregorgorjanc:
Done in https://github.com/popsim-consortium/stdpopsim/pull/1719.

Best to leave this open to update the pre-print to published paper - once that is achieved.

gregorgorjanc:
Done in #1760 and #1769 so this issue can be closed;)

PR #1719, opened 2025-03-31
Added selection pre-print to the docs #1713

gregorgorjanc:
Added selection pre-print to the docs #1713


petrelharp:
Love it, thanks!

Issue #1740, opened 2025-06-11
add to docs if a model hasn't been QC'ed


petrelharp:
This would help us catch things that haven't been, for one thing.

Issue #1738, opened 2025-06-08
Citation should be shown for species mutation/recombination rates


bhaller:
While working with a HomSap model, I noticed that its Genome section (https://popsim-consortium.github.io/stdpopsim-docs/main/catalog.html#sec_catalog_HomSap_genome) where it shows a table of the chromosomes and their lengths, recombination rates, and mutation rates, does not provide a citation for those pieces of information.  I asked, on the stdpopsim newbies channel, where the mutation rate of 1.29e-08 shown there came from, and Ilan gave me the cite: https://doi.org/10.1016/j.ajhg.2019.09.012.

That cite also doesn't appear in the notes emitted by stdpopsim, at least for the way I'm using it right now (I'm specifying a demographic model, which I guess overrides the mutation rate with a different rate given by that demographic model, but I'm not actually _using_ the demographic model, I just put it there because the tutorial had it... hard to explain :->).  Can one omit the demographic model -- could I leave that out?  Hmm, maybe so, and then maybe stdpopsim would use the rate of 1.29e-08, and so then maybe the citation for that rate would get emitted.  Well, I just tried that with:

`stdpopsim -e slim --slim-script HomSap -g HapMapII_GRCh38 -c chr2 --dfe PosNeg_R24 --dfe-annotation ensembl_havana_104_CDS -o foo.ts`

That gave me the error:

`stdpopsim HomSap: error: the following arguments are required: samples`

Anyway, point being, maybe I could have dug up this cite somehow, with a different usage pattern, but it seems like it ought to be shown in the doc, as cites for other things already are.  :->  I mentioned this on Slack and Ilan said

> If it's a bug, then it applies to all other species. I browsed a few and saw that mutation and recombination rates are typically not accompanied with their citation in the docs. You can open an issue, but this will need to be fixed in the script that generates the docs from the python code.

It does seem like a bug, so I opened this issue.  :->

PR #1741, opened 2025-06-11
two updates to the docs

petrelharp:
This:

1. adds warning to models and DFEs in the docs without QC, closing #1740
2. adds the list of citations to the genome section of the docs, closing #1738

For instance:
Screenshot From 2025-06-11 06-30-49
and
Screenshot From 2025-06-11 06-31-15

Styling/wording suggestions welcome.


gregorgorjanc:
Can you please add citations also to each demographic model (since model values&citations can differ from species values&citations)? Apologies if this is already done (can’t see well as I am travelling).

Issue #1312, opened 2022-07-29
Add pig species and published demographies (SusScr)


gregorgorjanc:
- **Name:** Sus Scrofa
- **Common name:** pig

Just starting this - will continue in time

https://en.wikipedia.org/wiki/Sus_(genus)

gregorgorjanc:
Adding some pig demography papers

2012: ***Analyses of pig genomes provide insight into porcine demography and evolution***
https://www.nature.com/articles/nature11622

For 10,000 years pigs and humans have shared a close and complex relationship. From domestication to modern breeding practices, humans have shaped the genomes of domestic pigs. Here we present the assembly and analysis of the genome sequence of a female domestic Duroc pig (Sus scrofa) and a comparison with the genomes of wild and domestic pigs from Europe and Asia. Wild pigs emerged in South East Asia and subsequently spread across Eurasia. Our results reveal a deep phylogenetic split between European and Asian wild boars ∼1 million years ago, and a selective sweep analysis indicates selection on genes involved in RNA processing and regulation. Genes associated with immune response and olfaction exhibit fast evolution. Pigs have the largest repertoire of functional olfactory receptor genes, reflecting the importance of smell in this scavenging animal. The pig genome sequence provides an important resource for further improvements of this important livestock species, and our identification of many putative disease-causing variants extends the potential of the pig as a biomedical model.

2015: ***Evidence of long-term gene flow and selection during domestication from analyses of Eurasian wild and domestic pig genomes***
https://www.nature.com/articles/ng.3394

Traditionally, the process of domestication is assumed to be initiated by humans, involve few individuals and rely on reproductive isolation between wild and domestic forms. We analyzed pig domestication using over 100 genome sequences and tested whether pig domestication followed a traditional linear model or a more complex, reticulate model. We found that the assumptions of traditional models, such as reproductive isolation and strong domestication bottlenecks, are incompatible with the genetic data. In addition, our results show that, despite gene flow, the genomes of domestic pigs have strong signatures of selection at loci that affect behavior and morphology. We argue that recurrent selection for domestic traits likely counteracted the homogenizing effect of gene flow from wild boars and created 'islands of domestication' in the genome. Our results have major ramifications for the understanding of animal domestication and suggest that future studies should employ models that do not assume reproductive isolation.

2022: *** Revisiting the evolutionary history of pigs via De Novo mutation rate estimation in a three-generation pedigree***
https://doi.org/10.1016/j.gpb.2022.02.001

[...]

PR #1752, opened 2025-09-24
Sus scrofa demography from Piecewise size model for wild boar (Wang et al. 2025)

AprilYUZhang:


stdpopsim/catalog/SusScr/demographic_models.py
```diff
@@ -0,0 +1,3462 @@
...
+_species = stdpopsim.get_species("SusScr")
+
+
+def _WildBoar_4Z22():
+    id = "WildBoar_4Z22"
+    description = "Piecewise size model for wild boar (Zhang et al. 2022)"
+    long_description = """
+    This demographic model is a piecewise size model for wild boar
+    from Zhang et al. (2022). Effective population sizes were estimated by
+    PSMC with parameters Tmax = 20, n = 64 (4 + 50*1 + 4 + 6)
```
gregorgorjanc, line 14: @AprilYUZhang change 50*1 to just 50?

gregorgorjanc, line 14: @AprilYUZhang PSMC is run on just one genome (as far as I know), but you state n=64, so I am confused. Will read the paper, but you might get back to this comment faster.

gregorgorjanc:
I am happy to work with @AprilYUZhang on revising this PR and the follow-up QC

gregorgorjanc:
Tagging the issue that this PR addresses https://github.com/popsim-consortium/stdpopsim/issues/1312

stdpopsim/catalog/SusScr/demographic_models.py
```diff
@@ -0,0 +1,3462 @@
...
+    return stdpopsim.DemographicModel(
+        id=id,
+        description=description,
+        long_description=long_description,
+        populations=populations,
+        citations=citations,
+        generation_time=3,
+        mutation_rate=3.6e-9,
+        population_configurations=[
+            stdpopsim.PopulationConfiguration(
```
gregorgorjanc, line 32: @AprilYUZhang there is a lot of code here due to many timesteps. I suggest you follow the style from https://github.com/popsim-consortium/stdpopsim/blob/main/stdpopsim/qc/BosTau.py

stdpopsim/catalog/SusScr/demographic_models.py
AprilYUZhang: About n, this is a time pattern syntax of PSMC and MSMC
I quoted PSMC's explanation here.
> In particular, the `-p'
> option specifies that there are 64 atomic time intervals and 28 (=1+25+1+1)
> free interval parameters. The first parameter spans the first 4 atomic time
> intervals, each of the next 25 parameters spans 2 intervals, the 27th spans 4
> intervals and the last parameter spans the last 6 time intervals. The `-p' and
> `-t' options are manually chosen such that after 20 rounds of iterations, at
> least ~10 recombinations are inferred to occur in the intervals each parameter
> spans. Impropriate settings may lead to overfitting. The command line in the
> example above has been shown to be suitable for modern humans.
> 
I quoted the original text of Zhang's paper

> Then we used the tool ‘fq2psmcfa’ from the PSMC package to create the input file. We used Tmax = 20, n = 64 (4 + 50*1 + 4 + 6) following the study by Groenen and colleagues [2].

stdpopsim/catalog/SusScr/demographic_models.py
```diff
@@ -0,0 +1,3462 @@
...
+                time=957493.3833541665, initial_size=177796.99883124998, population_id=3
+            ),
+        ],
+    )
+
+
+_species.add_demographic_model(_WildBoar_4Z22())
+
+_WangEtAl = stdpopsim.Citation(
+    doi="https://doi.org/10.1101/2025.02.05.636574",
```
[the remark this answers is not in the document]
AprilYUZhang: Agree, have updated now

Issue #1753, opened 2025-09-30
add additional properties to species and chromosomes


petrelharp:
Species:
- sexual or hermaphroditic? (#1146)
- recombination in which sexes? (#1743)

Chromosome:
- chromosome type

Other things?

Also note #383, where there's discussion on how to handle sex chromosomes with msprime.

Issue #1743, opened 2025-07-23
should Drosophila recomb rates be divided by two?


petrelharp:
In Drosophila, males don't do recombination; and AFAICT the rates [we have](https://popsim-consortium.github.io/stdpopsim-docs/stable/catalog.html#sec_catalog_DroMel) are untransformed from [Comeron et al](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1002905), which are reporting crossovers per *female* meiosis.

However, our simulations have recombination happening in all individuals.

A better alternative would be to implement "females-only" recombination, which would not be that hard (it would require an additional species-level property, though). But either way the first question is: is this right, that we're getting twice as much recombination as we should?

andrewkern:
i _think_ what we are doing is correct because Comeron estimated the number of recombinations in a cross. If males were also to recombine Comeron's estimates would be ~2x higher.

dschride:
As far as I can tell from the phrasing in the Comeron et al. paper, they are reporting the rates per female meiosis.

petrelharp:
I think we're all in agreement - reported numbers are number of recombinations in a cross (which = number of recombs per female meiosis). So, are we in agreement that this means wer'e simulating twice as much recombination as we should?

RyanGutenkunst:
My group made the same mistake years ago using the Comeron map (which a reviewer caught!). The do report rates per female meiosis, so if the map is applied to all individuals, it is twice as much recombination as there should be.

andrewkern:
okay! this is a great catch @petrelharp

petrelharp:
Okay - here's my proposal for how to fix this:
1. divide the rates by two and add a prominent message explaining this to the docs for them
2. put on the TODO list to do stuff with sexes appropriately

Alternative route:
1. leave rates as-is (with prominent message)
2. hack something in temporary that checks if it's Drosophila and if so, divides by two
3. longer-term, add a property that says "sexes recombination occurs in" and uses that property appropriately

Advantages to dividing by two:
- rates are correct for msprime and non-sexual SLiM models
- definition of rates agree with other species (they are sex-averaged)

Advantages for leaving as-is:
- rates are correct for the arguably-more-correct to-be-implemented-recomb-only-in-females SLiM model

I think it's pretty clear then that we just want to divide all the rates by two.

PR #1766, opened 2025-10-27
include separate sexes property

petrelharp:
This is an attempt at adding the "separate sexes" property, referred to in #1753 and needed by (the correct fix to) #1743.
This needs some testing on the SLiM side before merging.

This sounds minor but actually hits a lot of things:
0. All species now have a `separate_sexes` property, which is boolean. Even honeybees.
1. This will change the simulation results of any sexual species when using the SLiM engine. Not by much, but it is a change.
2. Since `simulate` only gets `demographic_model` and `contig` as arguments (not `species`), to get the species-specific information into `simulate` to allow  the recipe to depend on species attributes like this, I have added the property `species` to the `Contig` class.
3. We can't make `species` a *required* argument to `Contig`, since that would be backwards-incompatible: people make genetic contigs. So, my compromise is that if a Contig is created with `species=None` then it gets an ad-hoc created genetic Species that will have any required parameters at default values (eg `separate_sexes=False`).

Other notes:
- thanks to the new chromosome support in SLiM we don't need to specify sex determination system
- msprime does nothing with this information.

### DIFF
--- a/docs/_ext/speciescatalog.py
+++ b/docs/_ext/speciescatalog.py
@@ -102,6 +102,7 @@ class SpeciesCatalogDirective(SphinxDirective):
                 ],
             ),
             ("Ploidy", species.ploidy, None),
+            ("Separate sexes", species.separate_sexes, None),
             (
                 "Population size",
                 species.population_size,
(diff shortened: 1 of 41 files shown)


gregorgorjanc:
@petrelharp @andrewkern does this mean sex-specific genetic/recombination maps could/should be added too?

petrelharp:
Ah, I see what you mean. You're right, then - it seems less wrong to leave bees as `False` (and that would perhaps get changed when implementing haplodiploidy).

PR #1768, opened 2025-10-28
Fix infinite hang in test_incorrect_url

### LINKED ISSUE #1767: `test_incorrect_url` hangs indefinitely
Running
```
python -m pytest tests/test_annotations.py::TestAnnotationDownload::test_incorrect_url
```
hangs indefinitely. The place it hangs is when doing `urllib.request.urlopen` [here](https://github.com/popsim-consortium/stdpopsim/blob/be24f27f33d801dd9bdd6a39782fb5733f4014b5/stdpopsim/utils.py#L85). I thought at first the problem was exception handling, but changing that function to just do the urlopen call produces the same problem. I haven't turned up any helpful results yet.

This is happening to me locally, but also causing our CI to hang indefinitely, eg over in #1766. (There was another test or two as well that I'm dealing with over there.)

andrewkern:
The test was attempting to connect to a fake domain, causing DNS
resolution to hang indefinitely. Changed to use localhost:0 which
fails fast without DNS lookup.

Also add URLError handling to utils.download() to properly catch
DNS and connection failures.

Fixes #1767

### DIFF
--- a/stdpopsim/utils.py
+++ b/stdpopsim/utils.py
@@ -89,6 +89,10 @@ def download(url, filename):
         # Amend the error message to include the url.
         e.msg += f": {url}"
         raise e
+    except urllib.error.URLError as e:
+        # URLError can occur for DNS failures, connection errors, etc.
+        # Wrap it with additional context about the URL.
+        raise OSError(f"Failed to download {url}: {e.reason}") from e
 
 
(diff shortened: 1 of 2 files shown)


petrelharp (approved):
LGTM!

PR #1782, opened 2025-12-10
Enhance tutorial with SLiM ticks and output details

hyanwong:
Added explanation for checking simulated ticks in SLiM and how to output SLiM script information.


andrewkern:
this looks good to me, but I see there was a pre-commit linting error. Any chance you could fix that @hyanwong ?

hyanwong:
Hmm, OK. Is there any other way to figure out how long the SLiM simulation was, given a tree sequence from stdpopsim?

I guess this could be a discussion point on GitHub instead, anyway.

PR #1792, opened 2026-01-21
Traits

petrelharp:
This started with the code from @jeffspence's `master` branch that he and @roshnipatel got together.

Proposed rough low-level API:
```
# set up traits
traits = stdpopsim.TraitsModel(
    [(name, link_function), (name, link_function), ... ],
    default_environmental_sigma = 0,  # if no environments are added, they are all Gaussian with this sigma
    default_fitness_sigma = Inf, # similarly, default Gaussian
)
traits.add_environment( ... ) # optionally
traits.add_fitness_function(...) # optionally
# maybe add more of each

# set up "distribution of mutational effects" which is like DFE:
esd = stdpopsim.traits.DistributionOfMutationEffects(  # maybe? they will be (mostly?) generic
   mutation_types = [ some mutation types ],
   proportions = [ should add to 1 ],
   ]
)
contig.add_dme(esd, ...) # like add_dfe

# use them
engine.simulate(contig, traits)
```
*Note:* we could alternatively instantiate `Environment` and `FitnessFunction`s to pass into the `Trait` constructor,
but these have to refer to "which traits do I work with", so it's cleaner to have the traits set up
before we make them.

*Notes:*

- terminology: "trait" is the concept; "phenotype" is the actual measured value (or maybe "trait value"?)

# `TraitsModel`:

Encompasses all the stuff: how they are constructed, how they change with time. (Thus, "model".)

Has
- a list of `Trait`s, each of which has a link function
- a list of `Environment`s (each of which can affect a collection of the traits) 
- a list of `FitnessFunction`s (each of which operate on a collection of the traits) 

# `FitnessFunction`

Has:
- an id (important for debugging)
- trait type: quantitative, binary, fitness, or neutral
  * this is just basically "which fitness function"
- when and where this applies

# `Environment`

This affects how the genetic value maps to trait phenotypic values,
including the part that changes with time/space.
Environments can affect more than one trait (so you have correlated noise).
Possibly not public?

Adds "noise" to the genetic value. Has:
- and id (important for debugging)
- distribution and parameters
- when and where this applies

# Trait

This describes how to produce the *observed* biological trait value (ie, the phenotype):
for instance, describes how disease liability translates to disease incidence,
so it should not change with time or location.
Link functions map only a single trait, independently of others.

`GeneticValueTransform`? `PhenotypeTransform`? Well, this applies to "genetic value plus environment" so it's not a "genetic value transform"? Something else?

This transforms the underlying (genetic value plus environemtn) to the "observed trait". Has:
- an id
- whether it's multiplicative or additive
- a function
- maybe some randomness (but NOT what could be put in with environment)

Examples:
- identity
- threshold
- liability: trait equals 1 with probability logit(value)

# `DistributionOfMutationEffects`

Maps mutations to traits, so has
- a list of `MutationType`s
- and corresponding "what proportion of mutations"

# `MutationType`

A (biologically motivated, hopefully) class of mutations. As [before](https://popsim-consortium.github.io/stdpopsim-docs/stable/api.html#stdpopsim.MutationType), but now there's more than one trait, so also needs to know:

- which traits this affects (defaults to `"fitness"`)

# List of standard use cases:

See [below](https://github.com/popsim-consortium/stdpopsim/pull/1792#issuecomment-3780886616)

We should make it very easy to produce one of each of these,
something like `stdpopsim.traits.StabilizingSelectionOnTrait(n, sigma)`

Miscellaneous TODOs that may require getting into some weeds, and should turn into issues when we merge this:
- [ ] a special trait is named "fitness" currently; deal with this fact or change it
- [ ] figure out about `convert_to_substitution` *edit: I don't think we need an issue for this*
- [ ] switch dominance over to being "truly additive" and explain #1808 
- [ ] refactor the "check that the arguments and stuff related to this distribution or functional form are right" code (eg, determine if code should be shared between the FitnessFunction and the MutationType paths) (`_check_distribution` and related code)
- [ ] do scaling (with `Q`) only for fitness-affecting mutation effects #1818
- [ ] move code that changes lognormal and uniform to Eidos to `slim_engine.py` #1819 
- [ ] ideally we'd not need to define a new `mvn` distribution type since we already have Gaussian; but the existing `distribution_type="n"` uses a (mean, sd) parameterization, while the `mvn` type uses (mean, (co)variance); maybe this is just how things has to be but this needs to be super clear if so
- [ ] <s>do we need to keep the `stdpopsim.dfe` module?</s> *edit: probably not but let's leave it for backwards compatibility*
- [ ] revisit the `contig.is_neutral()` method; this previously just looked at `MutationType`s but now needs to include the `TraitsModel`. #1820 
- [ ] make other distributions multivariate (by scale transformation of normal?) #1821


petrelharp:
Outline: 
board1
board2

petrelharp:
I'm thinking that the plan here should be:

1. sketch out the higher-level API (like "give me two traits under stabilizing selection");
2. make sure the conceptual framework here works good for those;
3.  get the tests to pass; 
4.  merge this and open a bunch of issues for future work.

Here's brainstorming of those issues, besides the checklist in the original post here:

- [x] up code coverage on this code
- [ ] write method that applies a DME to a contig + annotation #1822
- [ ] write SLiM code for constructing traits  #1823 
- [ ] write SLiM code for distribution of mutation effects (DMEs) #1823
- [ ] figure out what fitness functions we want #1824
- [ ] write SLiM code for fitness function #1824 
- [ ] move the code for lognormal and uniform to slim_engine #1819
- [ ] write the spec for restricting environments and fitness functions to a particular time and/or location(s)
- [ ] write the method to check wether the spacetime specs are consistent with a particular demography
- [ ] write `TraitsModel.print(demographic_model)` that visualizes how the environments and fitness functions apply across a mutation model, to check for correctness with

So, this might be close, but we need (a) some eyes on it, and (b) to sketch out that higher-level API, to make sure?

gregorgorjanc:
> * `DirectionalSelection(num_traits, ...)` - @gregorgorjanc maybe you can advise here?

@roshnipatel: after talking with @petrelharp, I think we should separate two mechanisms that are often conflated in breeding contexts:

* Truncation selection: pick the top (or bottom) proportion of individuals, then generate offspring from that selected set (with a specified mating scheme).

* Directional selection: assign fitness directly as a monotonic function of trait value (no explicit truncation step), so higher-trait (or lower-trait) individuals contribute more offspring.

We could parameterise directional selection like `w(z) = a + b z` or `w(z) = exp(a + b z)` (Malthusian-style).

So maybe API-wise:

`DirectionalSelection(num_traits, a, b, fitness_model="linear"|"exponential", direction="top"|"bottom")`

@petrelharp if you recall any canonical references for this formulation, that would help us pick default forms, language etc.

Thoughts?

roshnipatel (commented):
@jeffspence and I did a code review together and had some comments:

- There's some ambiguity around whether multivariate fixed effects are allowed in `MutationType` -- given lines 740-744, it seems they ought to be, and we should provide that functionality/make it explicit in the docstring
- `_check_args_list` is pretty inflexible and we might eventually want to provide distribution params that do not match the total number of dimensions (e.g. weights for a mixture of Gaussians) -- I think we ought to remove this function and outsource args-checking to the relevant functions (i.e. `_check_distribution` and the `FitnessFunction` initialization)
  - Also, the `dim` argument is not being used at the moment
- A longer conversation for later: `MutationType` allows sampling from a distribution that could give you negative fitness (which might have been less of an issue when DFEs were largely catalog-provided, but feels like something to address in our new era of user-generated distributions)

Minor notes:
- We ought to check for nonempty trait/environment/fitness IDs
- Re: line 469 I think I agree this should happen in `slim_engine.py`
- We might want to expand the list of neutral DFEs in `MutationType.is_neutral( )`
- Are we requiring that a multivariate normal distribution has 2+ dimensions? (We don't seem to explicitly check this anywhere...)
- Line 351 has an incomplete docstring (but I think we actually have decided what it's supposed to be?)
- `test_traits.py` requires function arguments are provided as lists, but that seems rather strict -- maybe we should just check that things can be cast to a numpy array?

petrelharp:
I'm going to merge this so it doesn't get in the way of others, but still TODO for me is to make issues for the things above.

Issue #1801, opened 2026-02-19
Question about migration rate units in EarlyWolfAdmixture_6F14 (Freedman et al. 2014)


David-Peede:
**Short description of the problem:**
Hi, I was going over the demographic parameters for the `EarlyWolfAdmixture_6F14` model from [Freedman et al. (2014)](https://doi.org/10.1371/journal.pgen.1004016), and after spending a couple of hours wrapping my head around what `G-PhoCS` is actually reporting, I believe the per-generation continuous migration rate parameters in the current implementation may be incorrect, or my understanding of `G-PhoCS` is completely wrong.

**How to reproduce the problem:**
After reading [Gronau et al. (2011)](https://www.nature.com/articles/ng.937), it is my understanding that in `G-PhoCS`, each migration band from source population $S$ to target population $T$ has a scaled migration rate $m_{ST} = M_{ST} \div \mu$, where $M_{ST}$ represents the per-generation instantaneous migration rate, and backward-in-time migration along the band $\tau_{ST}$ (i.e., the time interval where both populations $S$ and $T$ exists) is modeled as a constant-rate Poisson process with rate $m_{ST}$ (see “The migration model” in section S4.2). The total (cumulative) migration rate corresponds to the product of the estimated scaled migration rate and the duration of the band $m_{ST}^{tot} =  m_{ST} \times \tau_{ST}$, which can be used as an approximation for the probability a lineage in the target population $T$ will migrate (backward-in-time) into the source population $S$ and I believe this can be computed as $1 - e^{-m_{ST}^{tot}}$ (see the second paragraph in section S5.2 and section S6.4). Note that I got the migration probability equation for the approximation not directly from the original `G-PhoCS` paper, but from the methods section of [vonHoldt et al (2016)](https://www.science.org/doi/10.1126/sciadv.1501714).

Moving to [Freedman et al. (2014)](https://doi.org/10.1371/journal.pgen.1004016), I will use the gene flow event from the Israeli wolf population (ISW) to Basenji population (BSJ), i.e., ISW $\rightarrow$ BSJ as a motivating example, but I believe the same logic would apply to all of the current migration rates in `EarlyWolfAdmixture_6F14`. From the Figure 5 caption, we are told:

> “Migration bands are shown in green with associated values indicating estimates of total migration rates, which equal the probability that a lineage will migrate through the band during the time period when the two populations co-occur.”

This is further elaborated in the first paragraph from supplemental section S9.2.3:

[...]

PR #1802, opened 2026-02-23
Update to EarlyWolfAdmixture_6F14 (CanFam)

David-Peede:
Following the discussion in #1801, I fixed the divergence times, per generation migration rates, and sanity checked the directionality of gene flow.


docs/parameter_tables/CanFam/EarlyWolfAdmixture_6F14.csv
```diff
@@ -10,18 +10,18 @@ Population size,"1,393","Ancestral (Israeli, Croatian) wolf pop. size"
...
-Time (gen.),"14,874","Ancestral (dog, wolf) split time"
-Time (gen.),"398,262","Ancestral ((dog, wolf), golden jackal) split time"
+Migration rate,0.18,"Israeli wolf -> Basenji total migration rate"
+Migration rate,0.07,"Basenji -> Israeli wolf total migration rate"
+Migration rate,0.03,"Chinese wolf -> Dingo total migration rate"
+Migration rate,0.04,"Dingo -> Chinese wolf total migration rate"
+Migration rate,0.00,"Golden jackal -> Israeli wolf total migration rate"
+Migration rate,0.05,"Israeli wolf -> Golden jackal total migration rate"
+Migration rate,0.02,"Golden jackal -> Ancestral (dog, wolf) total migration rate"
+Migration rate,0.99,"Ancestral (dog, wolf) -> Golden jackal total migration rate"
```
petrelharp: Looking from the outside, I'm not sure what "total migration rate" means. Is this something like "probability of migration in total across this branch"? Seems like we should either say that more precisely, or instead insert here the per-gen migration rate; either way we should put the units on "Migration rate".

petrelharp (commented):
Looks good, see suggestions. I tried to think of a more accurate thing to say than "total migration rate" - e.g., "total migration probability", but it's not a probability - and didn't come up with anything.

David-Peede:
Just dropping in to see if there is anything else I can do to help out with this :)

petrelharp:
Well, yes, there is: could you remind us what needs to be done and who said they'd do it? (guessing, was it "@igronau update the QC model"? I'm not sure if that's right or if there's anything else?)

David-Peede:
Of course! I believe @igronau originally agreed to QC the model, but then you also asked @gregorgorjanc afterwards, since @igronau was traveling at the time (I think, this is why you asked Gregor). So perhaps there was some confusion on who would QC the model? In any case, if either Ilan or Gregor (or someone else) is able to QC the model, the instructions were to read the discussion in  #1801 so the QC'er knows how things are parameterized, while trying to their best not to remember the exact numbers, and then update the QC code. 

P.S. I know people are busy, and this is not a rush for me personally, but I wanted to follow-up so this PR doesn't go stale.

petrelharp:
@igronau says he'll QC this. Thanks!

petrelharp:
ping @igronau

igronau:
I didn't forget, but it'll take me 1/2 day to clean my head and do this properly, and this semester I just didn't get a chance to have a clean 1/2 day. It's high up on my priority list.

PR #1810, opened 2026-03-03
AWS maintenance

andrewkern:
this is a workflow that aims to securely deal with uploading new artifacts (genetic maps / annotations) to the AWS site. 
 - Adds a local helper script (`maintenance/upload_to_s3.sh`) that validates uploads against the
  catalog before triggering the workflow
  - AWS credentials are now stored as GitHub repo secrets, accessible only through a protected
  environment requiring admin approval
  
   A maintainer runs the local script pointing at a tarball and the catalog entry it corresponds to:

  `./maintenance/upload_to_s3.sh data/HapMapII_GRCh38.tar.gz HomSap genetic_map HapMapII_GRCh38`

  The script:
  1. Looks up the species, genetic map/annotation in the stdpopsim catalog
  2. Verifies the local file's SHA256 matches the catalog entry
  3. Creates a temporary draft GitHub release to transport the file
  4. Triggers the GitHub Actions workflow

  The workflow then:
  1. Checks out the repo and re-validates against the catalog 
  2. Downloads the asset and verifies SHA256 again
  3. Checks that the S3 destination doesn't already exist (no overwrites)
  4. Uploads to S3
  5. Cleans up the draft release

This is in a draft state currently and will need some review

### DIFF
--- a/.github/workflows/upload-to-s3.yml
+++ b/.github/workflows/upload-to-s3.yml
@@ -27,6 +27,11 @@ on:
         description: "Resource ID as defined in the catalog (e.g. HapMapII_GRCh38)"
         required: true
         type: string
+      dry_run:
+        description: "If true, run all validation but skip the actual S3 upload"
+        required: false
+        type: boolean
+        default: false
 
 concurrency:
   group: s3-upload
@@ -187,6 +192,7 @@ jobs:
           echo "SHA256 verified."
 
       - name: Check if S3 object already exists
+        if: ${{ inputs.dry_run != true }}
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
@@ -201,6 +207,7 @@ jobs:
           echo "Confirmed: s3://stdpopsim/$DEST does not exist yet."
 
       - name: Upload to S3
+        if: ${{ inputs.dry_run != true }}
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
@@ -213,6 +220,20 @@ jobs:
           echo "Upload complete."
           echo "S3 URL: https://stdpopsim.s3-us-west-2.amazonaws.com/$DEST"
           echo "SHA256: ${{ inputs.expected_sha256 }}"
+
+      - name: Dry run summary
+        if: ${{ inputs.dry_run == true }}
+        run: |
+          echo "=== DRY RUN COMPLETE ==="
+          echo "All validation passed. Skipped S3 upload."
+          echo ""
+          echo "  Species:  ${{ inputs.species_id }}"
+          echo "  Type:     ${{ inputs.resource_type }}"
+          echo "  Resource: ${{ inputs.resource_id }}"
+          echo "  S3 dest:  s3://stdpopsim/${{ inputs.s3_destination }}"
+          echo "  SHA256:   ${{ inputs.expected_sha256 }}"
+          echo ""
+          echo "To perform the actual upload, re-run without --dry-run."
 
       - name: Clean up draft release
         if: always()
--- a/maintenance/upload_to_s3.sh
+++ b/maintenance/upload_to_s3.sh
@@ -3,15 +3,20 @@
 # Upload a tarball to the stdpopsim S3 bucket via GitHub Actions.
 #
 # Usage:
-#   ./maintenance/upload_to_s3.sh <local-file> <species-id> <genetic_map|annotation> <resource-id>
+#   ./maintenance/upload_to_s3.sh [--dry-run] <local-file> <species-id> <genetic_map|annotation> <resource-id>
 #
 # Examples:
 #   ./maintenance/upload_to_s3.sh data/HapMapII_GRCh38.tar.gz HomSap genetic_map HapMapII_GRCh38
 #   ./maintenance/upload_to_s3.sh data/ensembl_havana_104_exons_v1.tar.gz HomSap annotation ensembl_havana_104_exons
+#   ./maintenance/upload_to_s3.sh --dry-run data/test.tar.gz HomSap genetic_map HapMapII_GRCh38
 #
 # The script looks up the resource in the stdpopsim catalog to find the
 # expected S3 URL and SHA256, verifies the local file matches, then uploads
 # via a GitHub Actions workflow.
+#
+# Options:
+#   --dry-run  Run all validation (local + remote) but skip the actual S3 upload.
+#              The workflow will still run to verify everything end-to-end.
 #
 # Prerequisites:
 #   - GitHub CLI (gh) installed and authenticated
@@ -22,16 +27,24 @@ set -euo pipefail
 
 REPO="popsim-consortium/stdpopsim"
 WORKFLOW="upload-to-s3.yml"
+DRY_RUN=false
 
 die() { echo "ERROR: $*" >&2; exit 1; }
 
+# --- Parse --dry-run flag ---
+if [ "${1:-}" = "--dry-run" ]; then
+    DRY_RUN=true
+    shift
+fi
+
 # --- Validate arguments ---
 if [ $# -ne 4 ]; then
-    echo "Usage: $0 <local-file> <species-id> <genetic_map|annotation> <resource-id>" >&2
+    echo "Usage: $0 [--dry-run] <local-file> <species-id> <genetic_map|annotation> <resource-id>" >&2
     echo "" >&2
     echo "Examples:" >&2
     echo "  $0 data/HapMapII_GRCh38.tar.gz HomSap genetic_map HapMapII_GRCh38" >&2
     echo "  $0 data/exons_v1.tar.gz HomSap annotation ensembl_havana_104_exons" >&2
+    echo "  $0 --dry-run data/test.tar.gz HomSap genetic_map HapMapII_GRCh38" >&2
     exit 1
 fi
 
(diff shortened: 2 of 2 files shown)


andrewkern:
@petrelharp -- heads up, it looks like the pyslim docs site is returning a 404

maintenance/upload_to_s3.sh
```diff
@@ -0,0 +1,211 @@
...
+    SHA256=$(sha256sum "$LOCAL_FILE" | awk '{print $1}')
+elif command -v shasum >/dev/null 2>&1; then
+    SHA256=$(shasum -a 256 "$LOCAL_FILE" | awk '{print $1}')
+else
+    die "Neither sha256sum nor shasum found"
+fi
+
+# --- Validate against stdpopsim catalog ---
+echo "Validating against stdpopsim catalog..."
+VALIDATION=$(python3 << PYEOF
```
nspope, line 89: i'm probably being dense here, but why do we need to recreate this validation here, if the same code is being run as an action? Shouldn't it be one or the other?

maintenance/upload_to_s3.sh
```diff
@@ -0,0 +1,211 @@
...
+# Usage:
+#   ./maintenance/upload_to_s3.sh [--dry-run] <local-file> <species-id> <genetic_map|annotation> <resource-id>
+#
+# Examples:
+#   ./maintenance/upload_to_s3.sh data/HapMapII_GRCh38.tar.gz HomSap genetic_map HapMapII_GRCh38
+#   ./maintenance/upload_to_s3.sh data/ensembl_havana_104_exons_v1.tar.gz HomSap annotation ensembl_havana_104_exons
+#   ./maintenance/upload_to_s3.sh --dry-run data/test.tar.gz HomSap genetic_map HapMapII_GRCh38
+#
+# The script looks up the resource in the stdpopsim catalog to find the
+# expected S3 URL and SHA256, verifies the local file matches, then uploads
```
nspope, line 14: "verifies the local file matches" makes it seem like the file has already been uploaded; but that can't be the case right?

I think it'd help clarify if you spelled out the workflow on a bit more, something like
1. Create a catalogue entry for the new resource, in a feature branch
2. Use this script and point it towards the local path to the resource, it will validate it against the feature branch and upload it to AWS
3. Merge the feature branch

Maybe this should go in the developer docs as well?

maintenance/upload_to_s3.sh
```diff
@@ -0,0 +1,211 @@
...
+    echo "Workflow triggered successfully."
+fi
+echo ""
+echo "Monitor the upload with:"
+echo "  gh run list --repo $REPO --workflow $WORKFLOW --limit 3"
+echo ""
+echo "Or watch the latest run:"
+echo "  gh run watch --repo $REPO \$(gh run list --repo $REPO --workflow $WORKFLOW --limit 1 --json databaseId --jq '.[0].databaseId')"
+echo ""
+echo "Final S3 URL: $S3_URL"
```
nspope, line 211: Do you think it's worth uploading some metadata or otherwise trying to version what's on AWS? So that, if someone uses this script, we'll have a record of who/when/what? Or, I guess this would be tracked by the feature branch that adds the new resource?

nspope (commented):
I get the gist here (and it's a clever way to automate uploading without sharing credentials), but have a couple questions about the big picture.

maintenance/upload_to_s3.sh
andrewkern: No, this is a  fair point. My thinking was the local validation gives fast feedback before creating a draft release and triggering the workflow. The workflow validation is the security gate (since someone could bypass the script). But it is duplicated code. This is the "belt and suspenders" defensive approach I was talking to you about. 

We could remove the local validation and let the workflow be the single source of truth, but this would be a slower feedback loop

maintenance/upload_to_s3.sh
andrewkern: yeah this comment is ambiguous.... agreed this should be changed and added to the dev docs

maintenance/upload_to_s3.sh
andrewkern: good idea! I looked in to this quickly and see that we can include metadata with a call like:

```bash
  aws s3 cp file.tar.gz s3://bucket/path.tar.gz \
      --metadata '{"uploaded-by":"andrewkern","pr":"1810","timestamp":"2026-03-05"}'
 ```

File: maintenance/upload_to_s3.sh (hunk 1 above)
Comment (petrelharp) [reply maintenance/upload_to_s3.sh:89]: I vote for checking this twice!

Comment (petrelharp) [reply maintenance/upload_to_s3.sh:89]: For instance, it's relatively easy for the workflow to be accidentally turned off; doing it locally is I think great.

File: maintenance/upload_to_s3.sh (hunk 3 above)
Comment (petrelharp) [reply maintenance/upload_to_s3.sh:211]: Where does that metadata go? I was imagining @nspope was suggesting also uploading a JSON file.

File: maintenance/upload_to_s3.sh [hunk 4]
```diff
@@ -0,0 +1,211 @@
...
+        avail = list(resources.keys())
+        print(json.dumps({"error": f"Annotation '{resource_id}' not found for "
+                           f"{species_id}. Available: {avail}"}))
+        sys.exit(0)
+    resource = resources[resource_id]
+    s3_url = resource.intervals_url
+    expected_sha256 = resource.intervals_sha256
+
+# Extract S3 path from URL (handles both s3-us-west-2 and s3.us-west-2 styles)
+m = re.match(r"https://stdpopsim\.s3[.-]us-west-2\.amazonaws\.com/(.*)", s3_url)
```
petrelharp, line 130: I'm unclear why this is necessary?

maintenance/upload_to_s3.sh
```diff
@@ -0,0 +1,211 @@
...
+    --field "s3_destination=$S3_DEST" \
+    --field "expected_sha256=$SHA256" \
+    --field "species_id=$SPECIES_ID" \
+    --field "resource_type=$RESOURCE_TYPE" \
+    --field "resource_id=$RESOURCE_ID" \
+    --field "dry_run=$DRY_RUN"
+
+echo ""
+if [ "$DRY_RUN" = true ]; then
+    echo "Dry-run workflow triggered. It will validate everything but skip the S3 upload."
```
petrelharp, line 200: should the dry-run also skip the release?

maintenance/upload_to_s3.sh
```diff
@@ -0,0 +1,211 @@
...
+# --- Create draft release as file transport ---
+TAG="s3-upload-$(date +%Y%m%d%H%M%S)-$RANDOM"
+FILENAME=$(basename "$LOCAL_FILE")
+
+echo "Creating draft release ($TAG) to transport the file..."
+gh release create "$TAG" \
+    --repo "$REPO" \
+    --draft \
+    --title "S3 upload: $SPECIES_ID/$RESOURCE_TYPE/$RESOURCE_ID" \
+    --notes "Temporary draft release for S3 upload. Will be cleaned up automatically." \
```
petrelharp, line 180: Where does the advertised cleaning up happen?

.github/workflows/upload-to-s3.yml
```diff
@@ -0,0 +1,252 @@
...
+            exit 1
+          fi
+
+          # S3 destination must start with genetic_maps/ or annotations/
+          if [[ ! "$DEST" =~ ^(genetic_maps|annotations)/ ]]; then
+            echo "ERROR: S3 destination must start with 'genetic_maps/' or 'annotations/'"
+            exit 1
+          fi
+
+          # S3 destination must end with .tar.gz or .tgz
```
petrelharp, line 64: How about we also validate it's of the form `_vN.tar.gz` (as [required](https://popsim-consortium.github.io/stdpopsim-docs/stable/development.html#adding-a-recombination-genetic-map-or-annotation))?

.github/workflows/upload-to-s3.yml
```diff
@@ -0,0 +1,252 @@
...
+      - name: Check if S3 object already exists
+        if: ${{ inputs.dry_run != true }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-west-2
+        run: |
+          DEST="${{ inputs.s3_destination }}"
+          if aws s3api head-object --bucket stdpopsim --key "$DEST" 2>/dev/null; then
+            echo "ERROR: s3://stdpopsim/$DEST already exists. Refusing to overwrite."
```
petrelharp, line 203:
```suggestion
            echo "ERROR: s3://stdpopsim/$DEST already exists. Refusing to overwrite."
            echo "Do not replace or remove previously-released files!"
```

maintenance/upload_to_s3.sh
petrelharp: ah I see, [here](https://github.com/popsim-consortium/stdpopsim/pull/1810/changes#diff-841f4193c6752405ddb52bd2e332ad95ddbf9a449b1eea5215dcaa46282fb5d3R238). But it won't get cleaned up if someone doesn't run the aws script. I'm a little unclear about what the "release" actually is, though.

.github/workflows/upload-to-s3.yml
```diff
@@ -0,0 +1,252 @@
...
+          echo ""
+          echo "To perform the actual upload, re-run without --dry-run."
+
+      - name: Clean up draft release
+        if: always()
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAG="${{ inputs.release_tag }}"
+          echo "Cleaning up draft release: $TAG"
```
petrelharp, line 244: we should at least validate the form of `$TAG` here to make sure this isn't deleting something it shouldn't?

.github/workflows/upload-to-s3.yml
```diff
@@ -0,0 +1,252 @@
...
+          DEST="${{ inputs.s3_destination }}"
+          echo "Uploading to s3://stdpopsim/$DEST"
+          aws s3 cp "$DOWNLOADED_FILE" "s3://stdpopsim/$DEST"
+          echo ""
+          echo "Upload complete."
+          echo "S3 URL: https://stdpopsim.s3-us-west-2.amazonaws.com/$DEST"
+          echo "SHA256: ${{ inputs.expected_sha256 }}"
+
+      - name: Dry run summary
+        if: ${{ inputs.dry_run == true }}
```
petrelharp, line 225: If someone runs `--dry-run` then won't the draft release will get deleted, so they won't be able to continue on to do the wet run?

petrelharp:
Rebasing should make the pyslim error go away.

Issue #1822, opened 2026-03-18
Mechanism for adding DMEs to a contig


petrelharp:
In the API sketched out in #1792, one step was going to be 
```
contig.add_dme(dme, ...)
```
where `dme` is a `DistributionOfMutationEffects` object. Besides actually doing this, here we need to:

- add a `dme_list` to `Contig` (probably?)
-  figure out how much of the `contig.add_dfe()` API to use
- if we need additional arguments

Ideally, we can just move most of the `add_dfe` code over into `add_dme` and then have `add_dfe` call `add_dme`.

Currently, [Contigs have a `dfe_list` property](https://popsim-consortium.github.io/stdpopsim-docs/stable/api.html#stdpopsim.Contig). I don't think we want to keep both a `dfe_list` and (presumably) a `dme_list`, so -without thinking about it- I'm hoping that we can just make `dfe_list` a `@property` that just returns `self.dme_list` and maybe a deprecation warning.

jeffspence:
Should be closed by https://github.com/popsim-consortium/stdpopsim/pull/1827

PR #1827, opened 2026-04-10
Adding DMEs to a contig

roshnipatel:
Proposed fix for #1822 - swaps DFEs for DMEs, and preserves backwards compatibility in most places.

Two minor notes to add per @jeffspence:
- in `__str__` and `mutation_types` we break backwards compatibility by replacing "dfe_id" and "dfe_list" (appearing in the string representation) by "dme_id" and "dme_list".  Should we worry about this?
- Do we want to update `add_single_site` to allow a trait effect?

### DIFF
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -772,7 +772,9 @@ class Contig:
         self.dme_list = value
 
     def add_dfe(self, intervals, DFE):
-        """Alias for :meth:`add_dme`, see there for more details."""
+        """
+        Alias for :meth:`add_dme`, see there for more details.
+        """
         self.add_dme(intervals, DFE)
 
     def add_dme(self, intervals, DME):
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -785,7 +785,7 @@ class Contig:
                 "add_dfe is deprecated. Please use add_dme."
             )
         )
-        self.add_dme(self, intervals, DFE)
+        self.add_dme(intervals, DFE)
 
     def add_dme(self, intervals, DME):
         """
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -253,6 +253,13 @@ class Contig:
     mutational effects (DMEs), gene conversion rates and recombination rates
     that are needed to simulate this region.
 
+    Note that distributions of fitness effects (DFEs) are a special case
+    of DMEs, and so DFEs are supported and can be can be used anywhere that
+    calls for a DME. Additionally, all functions and attributes based on DMEs
+    have an aliased version with `dfe` instead of `dme`. For example,
+    ``dfe_list`` is an alias of ``dme_list``, :meth:`.add_dfe` is an alias of
+    :meth:`.add_dme`, and so on.
+
     Information about variants that affect traits
     or fitness are contained in the ``dme_list``
     and ``interval_list``. These must be of the same length, and the k-th DME
@@ -307,6 +314,9 @@ class Contig:
     :ivar dme_list: A list of :class:`.DME` objects.
         By default, the only DME is completely neutral and affects no traits.
     :vartype dme_list: list
+    :ivar dfe_list: Alias of the `dme_list` keyword argument. Note that only
+        one of `dme_list` or `dfe_list` may be specified.
+    :vartype dfe_list: list
     :ivar coordinates: The location of the contig on a named chromosome,
         as a tuple of the form `(chromosome, left, right)`. If `None`, the contig
         is assumed to be generic (i.e. it does not inherit a coordinate system
@@ -342,16 +352,19 @@ class Contig:
     inclusion_mask = attr.ib(default=None)
     exclusion_mask = attr.ib(default=None)
     dme_list = attr.ib(factory=list)
-    _dfe_list = attr.ib(factory=list, alias="dfe_list")
     interval_list = attr.ib(factory=list)
     coordinates = attr.ib(default=None, type=tuple)
     species = attr.ib(default=None, kw_only=True)
+    # _dfe_list only gets used in the initialization and subsequently
+    # gets deleted. It should not be used anywhere else.
+    _dfe_list = attr.ib(factory=list, alias="dfe_list")
 
     def __attrs_post_init__(self):
-        if len(self._dfe_list) > 0 and len(self.dme_list) > 0:
-            raise ValueError("Cannot specify both dme_list and dfe_list.")
         if len(self._dfe_list) > 0:
+            if len(self.dme_list) > 0:
+                raise ValueError("Cannot specify both dme_list and dfe_list.")
             self.dme_list = self._dfe_list
+        del self._dfe_list
 
         if self.coordinates is None:
             self.coordinates = (None, 0, int(self.length))
@@ -689,6 +702,7 @@ class Contig:
             return f"{chromosome}:{left}-{right}"
 
     def dfe_breakpoints(self, *, relative_coordinates=None):
+        """Alias for :meth:`dme_breakpoints`, see there for more details."""
         return self.dme_breakpoints(relative_coordinates=relative_coordinates)
 
     def dme_breakpoints(self, *, relative_coordinates=None):
@@ -734,6 +748,7 @@ class Contig:
         return breaks, dme_labels
 
     def clear_dfes(self):
+        """Alias for :meth:`clear_dmes`, see there for more details."""
         self.clear_dmes()
 
     def clear_dmes(self):
@@ -753,6 +768,7 @@ class Contig:
         self.dme_list = value
 
     def add_dfe(self, intervals, DFE):
+        """Alias for :meth:`add_dme`, see there for more details."""
         self.add_dme(intervals, DFE)
 
     def add_dme(self, intervals, DME):
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -748,7 +748,9 @@ class Contig:
         return breaks, dme_labels
 
     def clear_dfes(self):
-        """Alias for :meth:`clear_dmes`, see there for more details."""
+        """
+        Alias for :meth:`clear_dmes`, see there for more details.
+        """
         self.clear_dmes()
 
     def clear_dmes(self):
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -314,7 +314,7 @@ class Contig:
     :ivar dme_list: A list of :class:`.DME` objects.
         By default, the only DME is completely neutral and affects no traits.
     :vartype dme_list: list
-    :ivar dfe_list: Alias of the `dme_list` keyword argument. Note that only
+    :ivar dfe_list: Alias of the `dme_list` keyword argument. Only
         one of `dme_list` or `dfe_list` may be specified.
     :vartype dfe_list: list
     :ivar coordinates: The location of the contig on a named chromosome,
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -348,6 +348,18 @@ class Contig:
     species = attr.ib(default=None, kw_only=True)
 
     def __attrs_post_init__(self):
+        if self._dfe_list and self.dme_list:
+            raise ValueError(
+                "Cannot specify both dme_list and dfe_list."
+            )
+        if len(self._dfe_list) > 0:
+            warnings.warn(
+                stdpopsim.DeprecatedFeatureWarning(
+                    "Use of dfe_list is deprecated. Please use dme_list."
+                )
+            )
+            self.dme_list = self._dfe_list
+
         if self.coordinates is None:
             self.coordinates = (None, 0, int(self.length))
         _, left, right = self.coordinates
@@ -364,17 +376,6 @@ class Contig:
                 genome=None,
                 ensembl_id="none",
             )
-        if self._dfe_list and self.dme_list:
-            raise ValueError(
-                "Cannot specify both dme_list and dfe_list."
-            )
-        if len(self._dfe_list) > 0:
-            warnings.warn(
-                stdpopsim.DeprecatedFeatureWarning(
-                    "Use of dfe_list is deprecated. Please use dme_list."
-                )
-            )
-            self.dme_list = self._dfe_list
 
     @staticmethod
     def basic_contig(
--- a/stdpopsim/genomes.py
+++ b/stdpopsim/genomes.py
@@ -250,12 +250,13 @@ class Contig:
     """
     Class representing a contiguous region of genome that is to be simulated.
     This contains the information about mutation rates, distributions of
-    fitness effects (DFEs), gene conversion rates and recombination rates
+    mutational effects (DMEs), gene conversion rates and recombination rates
     that are needed to simulate this region.
 
-    Information about targets of selection are contained in the ``dfe_list``
-    and ``interval_list``. These must be of the same length, and the k-th DFE
-    applies to the k-th interval; see :meth:`.add_dfe` for more information.
+    Information about variants that affect traits
+    or fitness are contained in the ``dme_list``
+    and ``interval_list``. These must be of the same length, and the k-th DME
+    applies to the k-th interval; see :meth:`.add_dme` for more information.
 
     One attribute of the Contig is ``species``, which is necessary to pass
     on aspects of life history important for simulation. However, generic
@@ -299,12 +300,13 @@ class Contig:
     :ivar exclude: If True, ``mask_intervals`` specify regions to exclude. If False,
         ``mask_intervals`` specify regions in keep.
     :vartype exclude: bool
-    :ivar dfe_list: A list of :class:`.DFE` objects.
-        By default, the only DFE is completely neutral.
-    :vartype dfe_list: list
     :ivar interval_list: A list of :class:`np.array` objects containing integers.
-        By default, the inital interval list spans the whole chromosome with the
-        neutral DFE.
+        These specify where each DME provided in dme_list is applied.
+        By default, the inital interval list spans the whole chromosome with
+        DME where variants have no effects on traits or fitness.
+    :ivar dme_list: A list of :class:`.DME` objects.
+        By default, the only DME is completely neutral and affects no traits.
+    :vartype dme_list: list
     :ivar coordinates: The location of the contig on a named chromosome,
         as a tuple of the form `(chromosome, left, right)`. If `None`, the contig
         is assumed to be generic (i.e. it does not inherit a coordinate system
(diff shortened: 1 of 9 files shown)


roshnipatel:
@petrelharp -- @jeffspence wrote pretty much all of this, and I reviewed it before opening the PR, but maybe you or someone else with more institutional knowledge would like to review as well?

jeffspence:
To give a little bit of context, the main changes are:

In `genomes.py`:
* References to distributions of fitness effects (DFEs) in docstrings have been changed to distributions of mutational effects (DMEs)
* `Contig` now has a `dme_list` attribute, which should be used in place of `dfe_list`
* To maintain back-compatibility, the constructor in `Contig` still allows `dfe_list` as a kwarg but internally it stores this in `dme_list` (and raise a `stdpopsim.DeprecatedFeatureWarning`).  `Contig.dfe_list` can still be accessed (it is now a `@property`) and set (these actually access / set `Contig.dme_list` and again raise warnings)
* Several functions have had their names changed, with the original versions raising a `stdpopsim.DeprecatedFeatureWarning` and then calling the new version:
    * `Contig.dfe_breakpoints` --> `Contig.dme_breakpoints`
    * `Contig.clear_dfes` --> `Contig.clear_dmes`
    * `Contig.add_dfe` --> `Coting.add_dme`
* As @roshnipatel mentioned, `Contig.mutation_types` previously returned a list of dictionaries where one of the keys was `"dfe_id"` -- now it's `"dme_id"`.  This will technically break back-compatibility.  If we're concerned about this, we could have the dictionary have both keys.
* The other place where back-compatibility is broken is in `Contig.__str__` where the string `"dfe_list"` appeared in the string representation of the Contig, which again is now changed to `"dme_list"`.

Beyond these main changes, the rest are essentially tests of these in `test_genomes.py` (generally just backing sure the "dfe" methods raise deprecation warnings, and that the dfe versions and dme versions return the same things), along with moving over from the now deprecated functions to their dme versions in: `cli.py`, `slim_engine.py`, `traits.py`, `test_dfes.py`, `test_engines.py`, and `test_slim_engine.py`.

petrelharp:
Yes, I'll have a look. Thanks!!

petrelharp:
> in `__str__` and `mutation_types` we break backwards compatibility by replacing "dfe_id" and "dfe_list" (appearing in the string representation) by "dme_id" and "dme_list". Should we worry about this?

I think in `__str__` we don't have to worry about it. I don't have the context for `mutation_types` though - where's that?

stdpopsim/genomes.py
```diff
@@ -299,12 +300,13 @@ class Contig:
...
     :vartype exclude: bool
-    :ivar dfe_list: A list of :class:`.DFE` objects.
-        By default, the only DFE is completely neutral.
-    :vartype dfe_list: list
     :ivar interval_list: A list of :class:`np.array` objects containing integers.
-        By default, the inital interval list spans the whole chromosome with the
-        neutral DFE.
+        These specify where each DME provided in dme_list is applied.
+        By default, the inital interval list spans the whole chromosome with
+        DME where variants have no effects on traits or fitness.
```
petrelharp, line 306:
```suggestion
        By default, the initial interval list spans the whole chromosome with
        a DME whose variants have no effects on traits or fitness.
```

stdpopsim/genomes.py
```diff
@@ -339,18 +341,29 @@ class Contig:
     genetic_map = attr.ib(default=None)
     inclusion_mask = attr.ib(default=None)
     exclusion_mask = attr.ib(default=None)
-    dfe_list = attr.ib(factory=list)
+    dme_list = attr.ib(factory=list)
+    _dfe_list = attr.ib(factory=list, alias="dfe_list")
```
petrelharp, line 345: What's the thinking here? How will this work? (I don't know quite what `alias` does.)

stdpopsim/genomes.py
```diff
@@ -339,18 +341,29 @@ class Contig:
...
     interval_list = attr.ib(factory=list)
     coordinates = attr.ib(default=None, type=tuple)
     species = attr.ib(default=None, kw_only=True)
 
     def __attrs_post_init__(self):
+        if self._dfe_list and self.dme_list:
```
petrelharp, line 351: We like to make things explicit rather than relying on truthiness of things (eg is an empty list True or False? I'd have to check), so how about
```suggestion
        if self._dfe_list is not None and self.dme_list is not None:
```

petrelharp:
Could you explain the main changes and additions here? Like, it looks like this is this deprecating `Contig.dfe_list`, is that the intention? An outline of the intention will make it easier to review, eg separating out the "just changed dfe to dme" from actual functionality changes.

*Edit:* Okay, by my reading this does:

1. observes that `Contig.add_dfe` mostly just adds the DFE to `dfe_list`.
2. so, this changes over `Contig`s to have a `dme_list` instead of a `dfe_list`, since DMEs are generalizations of DFEs anyhow.
3. And, this puts a deprecation warning in all the `Contig.xxx_dfe_yyy` things.

An alternative to the deprecation warnings would be to basically just have all the `dfe_list` stuff be an alias for `dme_list` and not bother with the deprecation warnings. It would be cleaner to deprecate and eventually remove it, but we'll be throwing warnings in  lots of people's code that works just fine, and DFE is a term they are familiar with (unlike DME). So I'm tempted to *not* throw those warnings? Thoughts?

stdpopsim/traits.py
```diff
@@ -603,7 +603,7 @@ class DFE(DistributionOfMutationEffects):
     Class representing a "Distribution of Fitness Effects", i.e., a DFE.
     The class records the different *mutation types*, and the *proportions*
     with which they occur. The overall rate of mutations will be determined
-    by the Contig to which the DFE is applied (see :meth:`.Contig.add_dfe`).
+    by the Contig to which the DME is applied (see :meth:`.Contig.add_dme`).
```
petrelharp, line 606: I don't think this change is correct (at least the first one)?

stdpopsim/genomes.py
jeffspence: `alias="dfe_list"` causes `attr` to populate `_dfe_list` with the keyword arg `dfe_list` instead of a `_dfe_list` keyword arg.  This keeps back compatibility by allowing users to instantiate a `Contig` with a `dfe_list` keyword arg without needing to have a `dfe_list` attribute.  If we had both a `dme_list` and a `dfe_list` attribute, we would need some way of enforcing that they remain consistent.

stdpopsim/genomes.py
jeffspence: Thanks for this -- they both get instantiated by default to empty lists (by `attr.ib(factory=list)`), so I think testing if they're `None` is the wrong behavior, but I will explicitly test that they're greater than length 0 instead.

jeffspence:
Thanks for taking a careful look @petrelharp! I really appreciate it.  I had outlined the major changes [here](https://github.com/popsim-consortium/stdpopsim/pull/1827#issuecomment-4230857019) but it seems like you figured out the gist.

Regarding your comments: I like the idea of _not_ throwing deprecation warnings.  I had added them because I was worried about having to maintain `dme` and `dfe` versions of each function in `Contig`, but I agree that the overall maintenance burden should be low and we probably don't want users getting deprecation warnings for using the `dfe` versions.  If maintenance becomes a pain in the future, we can deprecate them then.  Otherwise, I fixed the typos/error in the docstrings (thanks!!!), and changed `if self._dfe_list and self.dme_list:` to `if len(self._dfe_list) > 0 and len(self.dme_list) > 0:`.

I think it's ready for you take another look, and I'm happy to make further changes.

stdpopsim/genomes.py
```diff
@@ -339,18 +341,24 @@ class Contig:
...
+    _dfe_list = attr.ib(factory=list, alias="dfe_list")
     interval_list = attr.ib(factory=list)
     coordinates = attr.ib(default=None, type=tuple)
     species = attr.ib(default=None, kw_only=True)
 
     def __attrs_post_init__(self):
+        if len(self._dfe_list) > 0 and len(self.dme_list) > 0:
+            raise ValueError("Cannot specify both dme_list and dfe_list.")
+        if len(self._dfe_list) > 0:
+            self.dme_list = self._dfe_list
```
petrelharp:
```suggestion
        if len(self._dfe_list) > 0:
            if len(self.dme_list) > 0:
                raise ValueError("Cannot specify both dme_list and dfe_list.")
            self.dme_list = self._dfe_list
```

stdpopsim/genomes.py
petrelharp: Let's add a comment explaining that `_dfe_list` is only used at initialization, and it should never be used after that?

And hm, having these both default to empty lists seems like it might lead to odd things down the road. What if the default for `_dfe_list` was `None`?

tests/test_genomes.py
```diff
@@ -18,14 +18,70 @@ class TestContig(object):
...
     )
 
-    def verify_dfe_breakpoints(self, contig):
-        breaks, dfe_labels = contig.dfe_breakpoints()
+    def test_dfe_and_dme_initalization(self):
+        chr_id = "chr2"
+        species = stdpopsim.get_species("HomSap")
+        genmap = "HapMapII_GRCh38"
+        chrom = species.get_contig(chr_id, genetic_map=genmap)
+        with pytest.raises(ValueError):
```
petrelharp:
```suggestion
        with pytest.raises(ValueError, match="Cannot specify both"):
```
to make sure we're getting the *right* ValueError

tests/test_genomes.py
```diff
@@ -18,14 +18,70 @@ class TestContig(object):
...
+    def test_dfe_list_aliasing(self):
+        chr_id = "chr2"
+        species = stdpopsim.get_species("HomSap")
+        genmap = "HapMapII_GRCh38"
+        chrom = species.get_contig(chr_id, genetic_map=genmap)
+        contig = stdpopsim.Contig(
+            recombination_map=chrom.recombination_map,
+            mutation_rate=1e-8,
+            dfe_list=[self.example_dfe],
+        )
```
petrelharp, line 43: Can we also `assert contig.dfe_list == contig.dme_list`? (I'm not sure, I'd have to check.)

petrelharp: We should also be testing this with a few more options, like for instance an empty list (in which case we definitely want `contig.dfe_list == contig.dme_list`) and a list of more than one element.

tests/test_genomes.py
```diff
@@ -18,14 +18,70 @@ class TestContig(object):
...
+        dfe_id = contig.dfe_list[0].id
+        assert dfe_id == "abc"
+        contig.dfe_list = ["a"]
+        assert len(contig.dme_list) == 1 and contig.dme_list[0] == "a"
+
+    def test_dfe_breakpoints(self):
+        contig = stdpopsim.Contig.basic_contig(length=100)
+        dfe_res1, dfe_res2 = contig.dfe_breakpoints()
+        dme_res1, dme_res2 = contig.dme_breakpoints()
+        assert np.allclose(dfe_res1, dme_res1)
```
petrelharp: these should be just plain `==`, right? byte-for-byte, no need for floating point caveats?

petrelharp:
> I had outlined the major changes https://github.com/popsim-consortium/stdpopsim/pull/1827#issuecomment-4230857019 but it seems like you figured out the gist.

Gee, so you did. I probably forgot, sorry! (Although github has been temporarily losing comments sometimes lately.)

We discussed this at the meeting briefly, and we think that the docstring probably needs some updating to make it clear that `dfe` stuff is still fine and still works good. The concern here is that people are familiar with DFEs and if they only see this DME thing then they'll be confused. I'm not sure the best way to do this, but brainstorming, maybe:
- put back `dfe_list` as a documented argument to `Contig` and say "alias to `dme_list`" or something
- put a short paragraph in the docstring saying that DFEs are supported but are a specialization of DMEs and so that's why it's like this
- put back a short docstring for things like `add_dfe`, like maybe "Adds a DFE to the contig. This is an alias for :meth:`add_dme`, see there for more details."

What do you think?

jeffspence:
Thanks, @petrelharp.  I think I addressed all of these suggestions, so feel free to take another look when you've got a chance.  Happy to change the wording in any docstrings if you want me to (or anything else).  The macos tests failed, but I don't think that's my fault (looks like errors while compiling scipy?)

I also got an email with a comment from you about initializing `Contig._dfe_list` to `None` instead of an empty list (but I'm not seeing it here).  I'm not sure I see what the issue is, but in either case, I update the initialization so that `_dfe_list` gets deleted at the end of the initialization, and the code where it shows up is commented to that effect. I think that should be enough to prevent anyone from accidentally using it.

gregorgorjanc:
> Although github has been temporarily losing comments sometimes lately.

I am having same issues too in several projects - I find that if click on the > character next to file name (to fold the file changes) and then click it again (to unfold the file changes) then I see the comments.

petrelharp:
> I also got an email with a comment from you about initializing Contig._dfe_list to None instead of an empty list (but I'm not seeing it here).

Scrolling up, I see that comment. But thinking more about this, I think it's weird to have the default in one place being None and in the other, `[]`. So never mind.

tests/test_genomes.py
petrelharp: What about this? I've tested the following, and it would be a stronger (and more concise) test of what we want?
```
assert contig1.dme_list == contig1.dfe_list
```

petrelharp: So then I think we can just do like
```
for dl in [[], [self.example_dfe], [example_dfe2, self.example_dfe]]:
    # make the contig
    assert contig.dfe_list == contig.dme_list
```
without most of the rest of this function? We could test for further correctness (eg `assert len(contig.dme_list) == len(dl) + 1`) but that's the job for other tests, this one is supposed to be testing the "aliasing".

tests/test_genomes.py
```diff
@@ -18,14 +18,104 @@ class TestContig(object):
...
+            assert contig.dfe_list == contig.dme_list
+            with pytest.raises(AttributeError):
+                contig._dfe_list
+
+    def test_dfe_breakpoints(self):
+        contig = stdpopsim.Contig.basic_contig(length=100)
+        dfe_res1, dfe_res2 = contig.dfe_breakpoints()
+        dme_res1, dme_res2 = contig.dme_breakpoints()
+        assert all(dfe_res1 == dme_res1)
+        assert all(dfe_res2 == dme_res2)
```
petrelharp:
```suggestion
        np.testing.assert_equal(dfe_res1, dme_res1, strict=True)
        np.testing.assert_equal(dfe_res2, dme_res2, strict=True)
```
This is because, for instance,
```
np.array([1,1]) == 1
array([ True,  True])
```
(and the `strict=True` is necessary to avoid the same problem).

stdpopsim/genomes.py
```diff
@@ -250,12 +250,20 @@ class Contig:
...
 
-    Information about targets of selection are contained in the ``dfe_list``
-    and ``interval_list``. These must be of the same length, and the k-th DFE
-    applies to the k-th interval; see :meth:`.add_dfe` for more information.
+    Note that distributions of fitness effects (DFEs) are a special case
+    of DMEs, and so DFEs are supported and can be can be used anywhere that
+    calls for a DME. Additionally, all functions and attributes based on DMEs
+    have an aliased version with `dfe` instead of `dme`. For example,
+    ``dfe_list`` is an alias of ``dme_list``, :meth:`.add_dfe` is an alias of
+    :meth:`.add_dme`, and so on.
```
petrelharp, line 261:
```suggestion
    :meth:`.add_dme`, and so on.
    We encourage new code to use `dme` rather than `dfe`, however.
```

stdpopsim/genomes.py
```diff
@@ -299,12 +307,16 @@ class Contig:
...
     :ivar interval_list: A list of :class:`np.array` objects containing integers.
-        By default, the inital interval list spans the whole chromosome with the
-        neutral DFE.
+        These specify where each DME provided in dme_list is applied.
+        By default, the initial interval list spans the whole chromosome with
+        a DME whose variants have no effects on traits or fitness.
+    :ivar dme_list: A list of :class:`.DME` objects.
+        By default, the only DME is completely neutral and affects no traits.
+    :vartype dme_list: list
+    :ivar dfe_list: Alias of the `dme_list` keyword argument. Note that only
```
petrelharp:
```suggestion
    :ivar dfe_list: Alias of the `dme_list` keyword argument. Only
```

stdpopsim/genomes.py
```diff
@@ -717,27 +742,43 @@ def dfe_breakpoints(self, *, relative_coordinates=None):
...
+    @property
+    def dfe_list(self):
+        return self.dme_list
+
+    @dfe_list.setter
+    def dfe_list(self, value):
+        self.dme_list = value
+
     def add_dfe(self, intervals, DFE):
+        """Alias for :meth:`add_dme`, see there for more details."""
```
petrelharp: The rest of this code has this (IMO) tidier and more readable format.
```suggestion
        """
        Alias for :meth:`add_dme`, see there for more details.
        """
```

stdpopsim/genomes.py
```diff
@@ -681,29 +702,33 @@ def origin(self):
             return f"{chromosome}:{left}-{right}"
 
     def dfe_breakpoints(self, *, relative_coordinates=None):
+        """Alias for :meth:`dme_breakpoints`, see there for more details."""
```
petrelharp:
```suggestion
        """
        Alias for :meth:`dme_breakpoints`, see there for more details.
        """
```

stdpopsim/genomes.py
```diff
@@ -717,27 +742,43 @@ def dfe_breakpoints(self, *, relative_coordinates=None):
...
-        dfe_labels = np.full(len(breaks) - 1, -1, dtype="int")
+        dme_labels = np.full(len(breaks) - 1, -1, dtype="int")
         for j, intervals in enumerate(self.interval_list):
-            dfe_labels[np.isin(breaks[:-1], intervals[:, 0], assume_unique=True)] = j
-        return breaks, dfe_labels
+            dme_labels[np.isin(breaks[:-1], intervals[:, 0], assume_unique=True)] = j
+        return breaks, dme_labels
 
     def clear_dfes(self):
+        """Alias for :meth:`clear_dmes`, see there for more details."""
```
petrelharp:
```suggestion
        """
        Alias for :meth:`clear_dmes`, see there for more details.
        """
```

jeffspence:
Thanks, @petrelharp.  I'm confused about the one line docstrings... [This PEP](https://peps.python.org/pep-0257/#one-line-docstrings) suggests that one line docstrings should literally go on one line, but happy to defer to you on that one.

Otherwise, the rest of the changes looked good to me and I committed them.  I simplified the alias testing.  It used to also test accessing, setting, and that `_dfe_list` gets deleted after initialization.  I think those are still important to test, but following your suggestion, I've moved them to separate tests.

petrelharp:
> [This PEP](https://peps.python.org/pep-0257/#one-line-docstrings) suggests that one line docstrings should literally go on one line, but happy to defer to you on that one.

Ah, well, that was just my gut (and looking elsewhere in the code I didn't find any)?

>  It used to also test accessing, setting, and that _dfe_list gets deleted after initialization. I think those are still important to test, but following your suggestion, I've moved them to separate tests.

Oh! I see - sorry, I didn't realize that was the intention. (Please do tell me when I'm missing silly things?) But the way it is now is also good.

And okay, the tests are failing on macos because scipy needs a fortran complier to install, and I guess there is no longer one available.

petrelharp (approved):
LGTM, thanks for all the back-and-forth.

petrelharp:
Ah, and the codecov error is due to the tests not running on macos. So I think this is good (and we can fix the macos thing elsewhere).

Issue #1807, opened 2026-03-03
allow/demosntrate smc(k) model


petrelharp:
The new SMC(k) model in msprime should allow doing neutral models a lot faster!

PR #1829, opened 2026-04-30
Enable smc(k) model

currocam:
Related to #1807

Hi! I’m interested in using the new SMC(k) model in msprime

### (1) Passing strings

The current code relies in many places on being able to specify models with strings (e.g. the command line), whereas the new "msprime.SMCK" requires a `k=X` argument. To get the (significant) speedup of the new SMCK, one could simply update

https://github.com/popsim-consortium/stdpopsim/blob/c96504f30f9191445e5a3adabc5bfcc8186453c5/stdpopsim/engines.py#L149-L154

into:

    model_class_map = {
        "hudson": msprime.StandardCoalescent,
        "dtwf": msprime.DiscreteTimeWrightFisher,
        "smc": msprime.SMCK(k=0),
        "smc_prime": msprime.SMCK(k=1),
    }

but, of course, doesn't follow the current `msprime` convention. Do you know if there are plans to "deprecate" legacy msprime.SmcPrimeApproxCoalescent and msprime.SmcApproxCoalescent in favour of `msprime.SMCK(k=0)` and `msprime.SMCK(k=1)`? 

If not, another option would be to come up with string aliases for the models (smc_k0 vs. smc_k1?) but it doesn't feel right to come up with aliases outside of the `msprime` (although it would be neat for the CLI). 

### (2) Passing `msprime.AncestryModel` (this PR)

The last alternative (I can think of) would be to allow passing msprime.AncestryModel or [msprime.AncestryModel, ...] directly. This fits nicely, but doesn't play nicely with the command-line nor the `msprime_change_model` argument. I've outlined this option in the PR. It's not complete, but perhaps a good start to start thinking about how to enable this feature?

## Quick test (Vaquita2Epoch_1R22, chr1, n=4 lineages)

I made a quick test and it looks like it’s working. 

The smc(k=1) is very fast for this long chromosome (1.86M)! Almost 1000x speedup compared to the naive rejection SMC' and a 10X speedup compare to Hudson. 

msprime_model_benchmark

msprime_model_num_trees

### DIFF
--- a/stdpopsim/engines.py
+++ b/stdpopsim/engines.py
@@ -1,11 +1,13 @@
+import copy
 import logging
+import math
 import warnings
 
 import attr
 import msprime
-import stdpopsim
 import numpy as np
-import math
+
+import stdpopsim
 
 logger = logging.getLogger(__name__)
 
@@ -145,7 +147,7 @@ class _MsprimeEngine(Engine):
         )
     ]
 
-    # We default to the first model in the list.
+    # We default to ``hudson`` if no model is specified.
     model_class_map = {
         "hudson": msprime.StandardCoalescent,
         "dtwf": msprime.DiscreteTimeWrightFisher,
@@ -166,40 +168,73 @@ class _MsprimeEngine(Engine):
 
     @property
     def supported_models(self):
+        # NOTE: this list does not include models without a string alias.
+        # Only msprime models with a string alias are supported in the CLI
+        # (and this method is only referenced in the CLI logic).
         return list(self.model_class_map.keys())
 
-    def _convert_model_spec(self, model_str, model_changes):
+    def _convert_model_spec(self, model, model_changes):
         """
         Convert the specified model specification into a form suitable
-        for sim_ancestry. The model param is a string or None. The
-        model_changes is either None or list of (time, model_str) tuples.
+        for sim_ancestry. The model param is a string, msprime.AncestryModel
+        instance, or None. The model_changes is either None or list of
+        (time, string | msprime.AncestryModel) tuples.
         Also return the appropriate extra citations.
         """
-        citations = []
-        if model_str is None:
-            model_str = "hudson"
-        else:
-            if model_str not in self.model_class_map:
-                raise ValueError(f"Unrecognised model '{model_str}'")
-            if model_str in self.model_citations:
-                citations.extend(self.model_citations[model_str])
 
-        if model_changes is None:
-            model = model_str
-        else:
+        def _from_model_spec(model):
+            # We default to the Hudson model if no model is specified
+            if model is None:
+                return _from_model_spec("hudson")
+            elif isinstance(model, str):
+                if model not in self.model_class_map:
+                    raise ValueError(f"Unrecognised model '{model}'")
+                return self.model_class_map[model](duration=None)
+            elif isinstance(model, msprime.AncestryModel):
+                model = copy.deepcopy(model)
+                model.duration = None
+                return model
+            raise TypeError("`model` must be a string or msprime.AncestryModel")
+
+        citations = []
+        model = _from_model_spec(model)
+        if model.name in self.model_citations:
+            citations.extend(self.model_citations[model.name])
+
+        if model_changes is not None:
             model_list = []
             last_t = 0
-            last_model = model_str
-            for t, model in model_changes:
-                if model not in self.supported_models:
-                    raise ValueError(f"Unrecognised model '{model}'")
-                if model in self.model_citations:
-                    citations.extend(self.model_citations[model])
+            last_model = model
+
+            def _from_tuple(change):
+                if len(change) != 2:
+                    raise ValueError(
+                        "`model_changes` must be a list of (time, model) tuples"
+                    )
+                t, next_model = change
+                try:
+                    t = float(t)
+                    assert t >= 0 and np.isfinite(t)
+                except (ValueError, AssertionError, TypeError):
+                    raise ValueError(
+                        f"Specified time {t} is not a valid non-negative number"
+                    )
+                if next_model is None:
+                    raise ValueError("Unrecognised model: `None`")
+                return t, _from_model_spec(next_model)
+
+            for change in model_changes:
+                t, next_model = _from_tuple(change)
                 duration = t - last_t
-                model_list.append(self.model_class_map[last_model](duration=duration))
-                last_model = model
+                if duration < 0:
+                    raise ValueError("Tuples in `model_changes` must be sorted by time")
+                last_model.duration = duration
+                model_list.append(last_model)
+                last_model = next_model
+                if last_model.name in self.model_citations:
+                    citations.extend(self.model_citations[last_model.name])
                 last_t = t
-            model_list.append(self.model_class_map[last_model](duration=None))
+            model_list.append(last_model)
             model = model_list
 
         return model, citations
@@ -222,12 +257,15 @@ class _MsprimeEngine(Engine):
         for all engines.
 
         :param msprime_model: The msprime simulation model to be used.
-            One of ``hudson``, ``dtwf``, ``smc``, or ``smc_prime``.
+            One of ``hudson``, ``dtwf``, ``smc``, or ``smc_prime``, or a
+            ``msprime.AncestryModel`` instance.
             See msprime API documentation for details.
-        :type msprime_model: str
+        :type msprime_model: str or msprime.AncestryModel
         :param msprime_change_model: A list of (time, model) tuples, which
-            changes the simulation model to the new model at the time specified.
-        :type msprime_change_model: list of (float, str) tuples
+            changes the simulation model to the new model at the time
+            specified. Each model may be a string as in ``msprime_model``
+            or a ``msprime.AncestryModel`` instance.
+        :type msprime_change_model: list of (float, str or msprime.AncestryModel) tuples
         :param dry_run: If True, ``end_time=0`` is passed to :meth:`msprime.simulate()`
             to initialise the simulation and then immediately return.
         :type dry_run: bool
--- a/tests/test_engines.py
+++ b/tests/test_engines.py
@@ -173,8 +173,8 @@ class TestBehaviour:
 
     def check_execution_order(self, expected, msprime_model, msprime_change_model=None):
         """
-        Run a short simulation and verify that the model execution order
-        recorded in the tree sequence provenance matches `expected`.
+        Small helper function to verify that the model execution order
+        recorded in the simulated tree sequence provenance matches `expected`.
         Each entry in `expected` is a (ClassName, duration) tuple.
         """
         species = stdpopsim.get_species("AraTha")
@@ -184,9 +184,8 @@ class TestBehaviour:
             contig=species.get_contig(length=100),
             samples={"pop_0": 5},
             msprime_model=msprime_model,
+            msprime_change_model=msprime_change_model,
         )
-        if msprime_change_model is not None:
-            kwargs["msprime_change_model"] = msprime_change_model
         ts = engine.simulate(**kwargs)
         prov = ts.provenance(0)
         record = json.loads(prov.record)
@@ -201,66 +200,34 @@ class TestBehaviour:
             actual.append((name, duration))
         assert expected == actual
 
-    def test_msprime_ancestry_model(self):
-        engine = stdpopsim.get_engine("msprime")
-        species = stdpopsim.get_species("HomSap")
-        kwargs = dict(
-            demographic_model=species.get_demographic_model("AshkSub_7G19"),
-            contig=species.get_contig("chr1"),
-            samples={"YRI": 5, "CHB": 5, "CEU": 5},
-            dry_run=True,
-        )
-        # Valid single None, string or `msprime.AncestryModel`
+    def test_check_execution_order(self):
+        # If no `msprime_change_model` is provided, we expect a single
+        # msprime ancestry model with duration=None
+        # msprime_model=None defaults to StandardCoalescent
         self.check_execution_order([("StandardCoalescent", None)], None)
+        # string alias
         self.check_execution_order([("StandardCoalescent", None)], "hudson")
+        # `msprime.AncestryModel`
         self.check_execution_order(
             [("StandardCoalescent", None)], msprime.StandardCoalescent()
         )
         self.check_execution_order([("SMCK", None)], msprime.SMCK(k=1))
 
-        # Fail if provided an invalid `msprime.AncestryModel` subclass
-        class MyClass:
-            pass
-
-        with pytest.raises(TypeError):
-            engine.simulate(**kwargs, msprime_model=MyClass())
-
-        # Fail if msprime_model is a list — passing a list is not supported;
-        # use msprime_change_model instead.
-        with pytest.raises(TypeError):
-            engine.simulate(
-                **kwargs,
-                # Note: this is valid input for `msprime`
-                msprime_model=[
-                    msprime.DiscreteTimeWrightFisher(duration=10),
-                    msprime.SMCK(k=1),
-                ],
-            )
-
-        # Test that we do not mutate the input model
-        model = msprime.SMCK(k=1, duration=10)
-        engine.simulate(
-            **kwargs,
-            msprime_model=model,
-            msprime_change_model=[(12, model), (99, model)],
-        )
-        assert model.duration == 10
-        model = msprime.SMCK(k=1, duration=10)
-        engine.simulate(**kwargs, msprime_model=model)
-        assert model.duration == 10, "model should not be mutated"
-
-    def test_msprime_change_model(self):
+        # Test that msprime_change_model results in the correct sequence of
+        # msprime ancestry models with the corresponding durations
         expected = [
             ("DiscreteTimeWrightFisher", 20),
             ("StandardCoalescent", None),
         ]
+        # Input as string aliases
         self.check_execution_order(expected, "dtwf", [(20, "hudson")])
+        # Input as `msprime.AncestryModel`
         self.check_execution_order(
             expected,
             msprime.DiscreteTimeWrightFisher(),
             [(20, msprime.StandardCoalescent())],
         )
-
+        # Or a combination of both
         model_changes = [
             (20, msprime.SMCK(k=1)),
             (21, msprime.SMCK(k=0)),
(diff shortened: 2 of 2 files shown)


gregorgorjanc:
FYI we have done some testing and for large number of samples SMC is in fact slower than Hudson, but the opposite is true with small number of samples, as you see here.

currocam:
Thanks @gregorgorjanc. Yes, I have also observed that. 

I have a very limited understanding of Hudson and the generalised SMC algorithm, but, tbh, I thought it was some implementation issue (e.g. the msprime Hudson being far more optimised than SMCK).  Also, for a fixed number of lineages, say, n=100, I often see that if I increase Ne enough, eventually smc(k=1) gets faster than Hudson, although I haven't tested this idea rigorously  at all. Do you know if this is something expected / intrinsic to complexity the algorithm?

P.S sorry for the digression.

gregorgorjanc:
I have only tested one specific model, so can't comment more. Konrad L tells me that there is something O(n^2) in the SMC(k) which makes it slow in large regime, but you should talk to him, or Jerome.

petrelharp:
> FYI we have done some testing and for large number of samples SMC is in fact slower than Hudson, but the opposite is true with small number of samples, as you see here.

What's a "large number"? (Sounds like it depends on the ratio of sample size squared and Ne, but, roughly?)

I think changing this to allow passing an msprime.AncestryModel is a good idea. We could also parse a string of the form `"smc(4)"`, say, but maybe we don't need to worry about making it available on the CLI for now because of these questions about efficiency and sample size? (Is it clear that people would def want to be using it for some common range of parameter values?)

stdpopsim/engines.py
```diff
@@ -168,28 +169,47 @@ class _MsprimeEngine(Engine):
...
+                        citations.extend(self.model_citations[str(m.name)])
+            else:
+                raise ValueError(
+                    "`model` must be a string, msprime.AncestryModel instance,"
+                    " or a list of msprime.AncestryModel instances"
+                )
 
         if model_changes is None:
-            model = model_str
+            model = model
```
petrelharp: um

petrelharp:
About the PR - I don't think you've got the API for model changes right yet? I think (based on a quick read) that what happens there is that `model` applies from time 0 back to the first model change time (if any); and then at each `time` the model changes to the corresponding `model`, for the `(time, model)` tuples in `model_changes`. So, I think that `model` should not be a list; and you should extend the `model` in those tuples to be AncestryModels as well as strings. (But I'm sure you'd have got that after a quick read.)

tests/test_engines.py
```diff
@@ -167,3 +168,35 @@ def test_msprime_bad_samples(self):
...
+            demographic_model=model,
+            contig=species.get_contig("chr1"),
+            samples={"YRI": 5, "CHB": 5, "CEU": 5},
+            dry_run=True,
+        )
+        import msprime
+
+        # Valid `msprime.AncestryModel`
+        engine.simulate(**good_kwargs, msprime_model=msprime.StandardCoalescent())
+        # TODO: here we should test we are adding the citation
```
petrelharp: I suspect that elsewhere we already test adding citations; here we just need to test the ability to specify models as AncestryModels.

tests/test_engines.py
```diff
@@ -167,3 +168,35 @@ def test_msprime_bad_samples(self):
...
+        engine = stdpopsim.get_engine("msprime")
+        species = stdpopsim.get_species("HomSap")
+        model = species.get_demographic_model("AshkSub_7G19")
+        good_kwargs = dict(
+            demographic_model=model,
+            contig=species.get_contig("chr1"),
+            samples={"YRI": 5, "CHB": 5, "CEU": 5},
+            dry_run=True,
+        )
+        import msprime
```
petrelharp: ???

petrelharp:
I'm happy to review this when you've got a reasonable PR? LMK.

currocam:
> About the PR - I don't think you've got the API for model changes right yet? I think (based on a quick read) that what happens there is that `model` applies from time 0 back to the first model change time (if any); and then at each `time` the model changes to the corresponding `model`, for the `(time, model)` tuples in `model_changes`. So, I think that `model` should not be a list; and you should extend the `model` in those tuples to be AncestryModels as well as strings. (But I'm sure you'd have got that after a quick read.)

Yeah, I tried to avoid dealing with it. Conceptually, my problem is `msprime.AncestryModel` takes a `duration` argument,   which might disagree with the `stdpopsim` specific tuples `msprime_change_model`. 

```
# current API (I think)
engine.simulate(msprime_model="dtwf", msprime_change_model = [(100, "hudson")],**kwargs)
# This could also work now
engine.simulate(msprime_model="dtwf", msprime_change_model = [(100, msprime.SMCK(k=1))],**kwargs) 
# But what if the user also uses duration? User might specify an invalid state
engine.simulate(msprime_model=msprime.SMCK(k=1, duration=50), msprime_change_model = [(100, msprime.SMCK(k=0))],**kwargs) 
```
IMO, copying the MSPrime API would be cleanest (having a list of MSPrime.AncestryModel with durations), but that breaks the existing signature. But I guess it's OK to assert duration is set to None here and force the user to use the tuple signature.


> I'm happy to review this when you've got a reasonable PR? LMK.

Yes, let me get back to you :=)

jeromekelleher:
Regarding the complexity question, the SMC(k) implementation is best used for small n (sample size). This is an inherent tradeoff in tracking pairs of lineages that can potentially coalesce, given the overlap constraints. If there's a lot of such lineages, it slows down. So it's great for small sample sizes and large chromosomes, but it's not a silver bullet, and good ole Hudson still wins generally.

currocam:
@petrelharp I think I now got `model_changes` right. Ready for review (I think?)

stdpopsim/engines.py
```diff
@@ -1,11 +1,13 @@
 import logging
+import math
 import warnings
+from copy import deepcopy
```
petrelharp: It's generally code style in this repo to just use the full name (`copy.deepcopy`), let's stick to that?

stdpopsim/engines.py
```diff
@@ -166,40 +168,53 @@ class _MsprimeEngine(Engine):
...
+        def _from_model_spec(model):
+            # We default to the Hudson model if no model is specified
+            if model is None:
+                return _from_model_spec("hudson")
+            elif isinstance(model, str):
+                if model not in self.model_class_map:
+                    raise ValueError(f"Unrecognised model '{model}'")
+                return self.model_class_map[model](duration=None)
+            elif isinstance(model, msprime.AncestryModel):
+                model = deepcopy(model)
```
petrelharp: reminder to make sure we're testing this by passing in a model and then changing it or something

stdpopsim/engines.py
```diff
@@ -166,40 +168,53 @@ class _MsprimeEngine(Engine):
...
-                citations.extend(self.model_citations[model_str])
+        model = _from_model_spec(model)
+        if model.name in self.model_citations:
+            citations.extend(self.model_citations[model.name])
 
-        if model_changes is None:
-            model = model_str
-        else:
+        if model_changes is not None:
+            # Note: one could validate that times in model_changes are
```
petrelharp: are what?

stdpopsim/engines.py
```diff
@@ -222,12 +237,15 @@ def simulate(
...
+            ``msprime.AncestryModel`` instance.
             See msprime API documentation for details.
-        :type msprime_model: str
+        :type msprime_model: str or msprime.AncestryModel
         :param msprime_change_model: A list of (time, model) tuples, which
-            changes the simulation model to the new model at the time specified.
-        :type msprime_change_model: list of (float, str) tuples
+            changes the simulation model to the new model at the time
+            specified. Each model may be a string or a
+            ``msprime.AncestryModel`` instance.
```
petrelharp:
```suggestion
            specified. Each model may be a string as in ``msprime_model``
            or a ``msprime.AncestryModel`` instance.
```

tests/test_engines.py
```diff
@@ -167,3 +168,79 @@ def test_msprime_bad_samples(self):
...
+        engine.simulate(
+            **kwargs,
+            msprime_model=msprime.SMCK(k=1),
+            msprime_change_model=[(1000, msprime.SMCK(k=0)), (2000, "hudson")],
+        )
+
+        # Verify model_changes produces the correct model order
+        def _check_model_order(initial, changes, expected_names):
+            engine_obj = stdpopsim.get_engine("msprime")
+            model_list, _ = engine_obj._convert_model_spec(initial, changes)
```
petrelharp: Wait, we should not be testing this private function. Is there not some other way to check that the correct sequence of models has been run? I haven't yet looked through the previous tests to see how we deal with this?

petrelharp (commented):
The code looks right to me, but we need a better way of testing we're doing the right models at the right times. (Also a test of whatever we're needing to `deepcopy` for.) Could you see whether this (model switching in msprime) a thing we were testing robustly before?

petrelharp:
Also, I should  have said - that code looks a tricky; nice job sorting that out.

stdpopsim/engines.py
currocam: *are valid (i.e. monotonically increasing) and provide a better error message. 

Right now, running something like 

```
engine.simulate(
    demographic_model=species.get_demographic_model("AshkSub_7G19"),
    contig=species.get_contig("chr1"),
    samples={"YRI": 5, "CHB": 5, "CEU": 5},
    dry_run=True,
    msprime_model="dtwf",
    msprime_change_model=[(2, "smc_prime"), (1, "hudson")],
)
```

does not fail with `dry_run=True`. When false, it fails with `ValueError: Model durations must be >= 0`. 

We can validate it during initialisation and provide a better error message. Happy to add here or in a separate PR.

currocam:
I took a closer look and I don’t think `msprime_model_changes` was being tested directly. I’ve added a couple of tests that directly test the execution order is correct (parsing the tree-sequence provenance, which is a bit hacky…) and added some better (IMP) error handling.

tests/test_engines.py
currocam: Fixed in a later commit.

petrelharp:
Thanks! I'll have a look.

> parsing the tree-sequence provenance, which is a bit hacky

Yeah, but I'm not sure where else we'd get it? I guess there are things we can check for most models, like if times are node integer for DTWF.  This seems fine, anyhow.

stdpopsim/engines.py
```diff
@@ -202,15 +202,35 @@ def _from_model_spec(model):
...
+                        f"Specified time {t} is not a valid non-negative number"
+                    )
+                return t, _from_model_spec(next_model)
+
+            for change in model_changes:
+                t, next_model = _from_tuple(change)
                 duration = t - last_t
+                if duration < 0:
+                    raise ValueError(
+                        "Tuples in `model_changes` must be sorted time `time`"
```
petrelharp: "must be sorted by time"?

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+        )
+        with pytest.raises(
+            ValueError,
+            match=r"`model_changes` must be a list of \(time, model\) tuples",
+        ):
+            engine.simulate(
+                **kwargs,
+                msprime_model="dtwf",
+                msprime_change_model=[
+                    # (10, "hudson")
```
petrelharp:
```suggestion
```

stdpopsim/engines.py
```diff
@@ -166,40 +168,73 @@ class _MsprimeEngine(Engine):
...
+                return t, _from_model_spec(next_model)
+
+            for change in model_changes:
+                t, next_model = _from_tuple(change)
                 duration = t - last_t
-                model_list.append(self.model_class_map[last_model](duration=duration))
-                last_model = model
+                if duration < 0:
+                    raise ValueError(
+                        "Tuples in `model_changes` must be sorted time `time`"
```
petrelharp:
```suggestion
                        "Tuples in `model_changes` must be sorted by time"
```

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+            engine.simulate(
+                **kwargs,
+                msprime_model="dtwf",
+                msprime_change_model=[
+                    (-1, "hudson"),
+                ],
+            )
+        with pytest.raises(
+            ValueError,
+            match=r"Tuples in `model_changes` must be sorted time `time`",
```
petrelharp:
```suggestion
            match=r"Tuples in `model_changes` must be sorted",
```

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+            )
+        with pytest.raises(ValueError, match=r"Unrecognised model"):
+            engine.simulate(
+                **kwargs,
+                msprime_model="dtwf",
+                msprime_change_model=[
+                    (10, "notvalid"),
+                ],
+            )
+        with pytest.raises(
```
petrelharp: let's get some more invalid times in here, like `None`, `"hudson"`, `np.inf`

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+        ):
+            engine.simulate(
+                **kwargs,
+                msprime_model="dtwf",
+                msprime_change_model=[
+                    # (10, "hudson")
+                    ("hudson"),
+                ],
+            )
+        with pytest.raises(ValueError, match=r"Unrecognised model"):
```
petrelharp: and let's get some more invalid models, like `None` and `100`

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+            engine.simulate(
+                **kwargs,
+                msprime_model="dtwf",
+                msprime_change_model=[
+                    (2, "hudson"),
+                    (1, "dtwf"),
+                ],
+            )
+
+    def test_msprime_ancestry_model(self):
```
petrelharp: How about factor out the "check execution order" code above into a method (like maybe `self.check_execution_order(ts, expected)`) and apply it to these, so we're checking more than just "did this run"?

tests/test_engines.py
petrelharp: This is still here? Did you miss a commit?

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+                **kwargs,
+                # Note: this is valid input for `msprime`
+                msprime_model=[
+                    msprime.DiscreteTimeWrightFisher(duration=10),
+                    msprime.SMCK(k=1),
+                ],
+            )
+
+        # Test that we do not mutate the input model
+        # Internally, we update the duration, but we should not mutate the input model
```
petrelharp: what about something like this:
```suggestion
        model = msprime.SMCK(k=1, duration=10)
        engine.simulate(**kwargs, msprime_model=model, msprime_change_model=[(12, model), (99, model)])
        assert model.duration == 10
```

petrelharp (commented):
I like the testing; but there's still the non-public function being used in the testing code.

tests/test_engines.py
```diff
@@ -167,3 +170,184 @@ def test_msprime_bad_samples(self):
...
+        assert expected == _get_execution_order("dtwf", model_changes)
+
+    def test_msprime_change_model_errors(self):
+        engine = stdpopsim.get_engine("msprime")
+        species = stdpopsim.get_species("HomSap")
+        kwargs = dict(
+            demographic_model=species.get_demographic_model("AshkSub_7G19"),
+            contig=species.get_contig("chr1"),
+            samples={"YRI": 5, "CHB": 5, "CEU": 5},
+            dry_run=True,
```
petrelharp: I know we don't need to have a real simulation to trigger this, but we want to make sure that it works in the usual code path.
```suggestion
```

tests/test_engines.py
```diff
@@ -167,3 +170,163 @@ def test_msprime_bad_samples(self):
...
+        species = stdpopsim.get_species("AraTha")
+        engine = stdpopsim.get_engine("msprime")
+        kwargs = dict(
+            demographic_model=stdpopsim.PiecewiseConstantSize(species.population_size),
+            contig=species.get_contig(length=100),
+            samples={"pop_0": 5},
+            msprime_model=msprime_model,
+        )
+        if msprime_change_model is not None:
+            kwargs["msprime_change_model"] = msprime_change_model
```
petrelharp: `None` is the default for this parameter, so this should work?
```suggestion
            msprime_model = msprime_change_model,
        )
```

tests/test_engines.py
```diff
@@ -167,3 +170,163 @@ def test_msprime_bad_samples(self):
...
+    def test_msprime_ancestry_model(self):
+        engine = stdpopsim.get_engine("msprime")
+        species = stdpopsim.get_species("HomSap")
+        kwargs = dict(
+            demographic_model=species.get_demographic_model("AshkSub_7G19"),
+            contig=species.get_contig("chr1"),
+            samples={"YRI": 5, "CHB": 5, "CEU": 5},
+            dry_run=True,
+        )
+        # Valid single None, string or `msprime.AncestryModel`
```
petrelharp: This test needs some refactoring so it's clear what's going on: currently it first sets up the kwargs for a HomSap simulation, then it calls `check_execution_order` a few times, which doesn't use those `kwargs`, then it goes on to use the kwargs for something else.

It looks like it should be three tests: checking execution order, testing errors, and testing we don't mutate the model?

tests/test_engines.py
```diff
@@ -167,3 +170,163 @@ def test_msprime_bad_samples(self):
...
+
+        # Test that we do not mutate the input model
+        model = msprime.SMCK(k=1, duration=10)
+        engine.simulate(
+            **kwargs,
+            msprime_model=model,
+            msprime_change_model=[(12, model), (99, model)],
+        )
+        assert model.duration == 10
+        model = msprime.SMCK(k=1, duration=10)
```
petrelharp:
```suggestion
```

tests/test_engines.py
petrelharp: I'm also not sure what's the distinction between the `check_execution_order` tests here and in the next one?

petrelharp:
Almost there - the tests need a bit of organization. I've gone ahead and rebased and squashed this also, so the tests would run; I hope that doesn't mess you up!

currocam:
No worries 🙂 Thanks for keep the PR going :)

petrelharp:
lmk when I should take a look again!

currocam:
I think the PR is ready :)

andrewkern:
@petrelharp -- looks like you were doing the review on this. want to take another look?

petrelharp:
Lol I missed that message, sorry!

petrelharp (approved):
LGTM! Thanks a bunch, @currocam!

andrewkern:
@mergifyio queue

PR #1834, opened 2026-05-27
fix sign error in AraTha DFE (closes #1833)

### LINKED ISSUE #1833: Sign error in DFE for Arabidopsis thaliana
Hi all! 

I was working on some DFE inference validation in polyploids and was using the stdpopsim catalog for _Arabidopsis thaliana_ to output a SLiM script as a starting point for simulations. It seems that the script simulates mutations drawn from the correct gamma distributed DFE, but for beneficial instead of deleterious mutations (which is an easy fix). (See code snippet below). 

Python code to output _Arabidopsis_ SLiM script with DFE:
```
import stdpopsim
import numpy as np
from contextlib import redirect_stdout

species = stdpopsim.get_species("AraTha")
model = species.get_demographic_model("African3Epoch_1H18")
contig = species.get_contig("chr1")
samples = {'SouthMiddleAtlas' : 10}
dfe = species.get_dfe('GammaAdditive_H18')
contig.add_dfe(intervals=np.array([[0, int(contig.length)]]), DFE=dfe)
engine = stdpopsim.get_engine("slim")

with open("arabidopsis.slim", "w") as f:
    with redirect_stdout(f):
        ts = engine.simulate(
            model,
            contig,
            samples,
            slim_script=True,
            verbosity=2,
            slim_scaling_factor=10,
        )
```

SLiM output bug (line 41 of arabidopsis.slim):
```
initializeMutationType(2, 0.5, "g", Q * 0.00021600000000000002, 0.27);
```

petrelharp:
See #1833 for context. I don't think this needs a separate QC (but I'll want someone else to take a look at this to make sure I'm not being silly).

### DIFF
--- a/stdpopsim/catalog/AraTha/dfes.py
+++ b/stdpopsim/catalog/AraTha/dfes.py
@@ -38,7 +38,7 @@ def _HuberDFE():
         distribution_type="g",  # gamma distribution
         # extra factor of 2 is to convert dadi to SLiM
         # (1+s for homozygote in SLiM versus 1+2s in dadi)
-        distribution_args=[-2 * gamma_mean, gamma_shape],
+        distribution_args=[2 * gamma_mean, gamma_shape],
     )
 
     return stdpopsim.DFE(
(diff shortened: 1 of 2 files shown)


mufernando (approved):
This looks good to me. It's weird that two different people parametrizing the DFE in different ways made the same error!

PR #1835, opened 2026-05-27
update github actions

petrelharp:
Since they're all failing, I'm copying some code from the tskit setup, which was examined more recently.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,20 +35,18 @@ jobs:
         uses: actions/cache@v4.2.0
         env:
           # Increase this to reset the cache if the key hasn't changed.
-          CACHE_NUM: 0
+          CACHE_NUM: 1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -30,36 +30,49 @@ jobs:
...
+
       - name: Install conda deps
         if: steps.cache.outputs.cache-hit != 'true'
         shell: bash -l {0} #We need a login shell to get conda
         run: |
-          conda install --yes --file=requirements/CI/conda.txt
+          micromamba install --yes --file=requirements/CI/conda.txt
 
       - name: Install pip deps
         if: steps.cache.outputs.cache-hit != 'true'
```
petrelharp, line 63: okay so this is not cacheing pip now: https://github.com/mamba-org/setup-micromamba/issues/195
maybe I should go back to the cache

PR #1839, opened 2026-05-29
bump msprime and slim versions for CI

petrelharp:
*Edit:* This is also bumping the max python version tested.


petrelharp:
Should we bump the minimum supported python version to 3.11, as I've done here? I think so, since msprime did so (and 3.10 is EOL in October).

PR #1840, opened 2026-05-29
build slim from source

### LINKED ISSUE #1809: Switch to development SLiM 5.2
We need this for a lot of things - traits! - so first step would be to switch to using it (and that currently means the development version).

petrelharp:
Closes #1809. Note that we'll want to switch back to the SLiM conda release after the next SLiM release and before we release stdpopsim. (And I've opened #1841 to remind us.)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,10 +65,26 @@ jobs:
         run: |
             pip install -r requirements/CI/requirements.txt
 
+      - name: Install development SLiM
+      # This should be COMMENTED OUT for release versions,
+      # since this builds SLiM from github head.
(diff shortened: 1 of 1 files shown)


petrelharp:
It worked! I'm getting
```
SLiM version 5.2, built May 29 2026 21:15:10
Git commit SHA-1: b7435e243ede9ce519b872e603a63ad1931f0313
```
which is indeed [the latest commit](https://github.com/MesserLab/SLiM/commit/b7435e243ede9ce519b872e603a63ad1931f0313)

We'll probably need to bump the cache number in [.github/workflows/tests.yml](https://github.com/popsim-consortium/stdpopsim/blob/5793804fd1e1d9e4a8ff693c8c21d1d51b0e3795/.github/workflows/tests.yml#L38) if we want to re-build SLiM from a more recent commit.

PR #1843, opened 2026-05-30
drop Python 3.10 support

andrewkern:
Drops Python 3.10 support so the dependabot bumps can be merged. A lot of our dependencies now require >= 3.11 so I guess it's time


petrelharp (approved):
LGTM!

Issue #1854, opened 2026-06-17
Bug in Genome-Wide Recombination Rate for A. thaliana


srgibbon17:
I came across what I am (almost) certain is a bug in the genome-wide recombination rate for _A. thaliana_. For context, I was working on validating a DFE inference method and was getting very strong signals of BGS from my SLiM simulations based on demographic models and other parameters from stdpopsim. The led to me doing some investigating, and I think the primary culprit is the recombination rate for _A. thaliana_ which is listed as ~8e-10 (or .08 cM/Mb) and I think should be ~3.6e-8.

First, from the recombination rate map provided in stdpopsim for _A. thaliana_, the first few recombination rates for chromosome 1 are much higher than the genome wide estimate: e.g., 5.0161340000000004e-08, 4.579948999999998e-08, 4.021042000000001e-08 ... (these can be gathered from the SLiM script output by the code snippet below) and, taking a weighted average across all the chromosomes in the genome gives an average rate of ~3.67e-8.

Second, I took a look at the [Salome et al. 2012](https://doi.org/10.1038/hdy.2011.95) reference for the rate map and they estimate r=3.6e-8 or 3.6 cM/Mb (see the Mean - Whole Genome entry from Table 1). 

The issue in the [stdpopsim code](https://github.com/popsim-consortium/stdpopsim/blob/05f33ad9cad479d42a8fe3132f0d6f0fa20f08c3/stdpopsim/catalog/AraTha/species.py#L5-L9) seems to stem from a typo in the [Huber et al. paper (MBE 2014)](http://doi.org/10.1093/molbev/msu247) used to estimate r... there, they choose 2.5 cM/Mb for their simulations which, with Ne=124,000 should give a population scaled recombination rate rho = 2rNe = 2(2.5/100)(124,000) = 6200 (not 200, as written in the paper!) or r = 2.5e-8 (which is much closer to the 3.6e-8 estimate from Salome et al.).

Using the updated recombination rate (or the recombination rate map, which seems to be good to go) produced much more reasonable/expected B-values for the strength of BGS, so I just wanted to make sure no one else gets bitten by this in simulations!

Code snippet to output recombination rate map for chromosome 1
```
import stdpopsim
import numpy as np
from contextlib import redirect_stdout

species = stdpopsim.get_species("AraTha")
model = species.get_demographic_model("African3Epoch_1H18")
contig_rmap = species.get_contig("chr1", genetic_map="SalomeAveraged_TAIR10")
[...]

Issue #857, opened 2021-04-08
add selfing rate


petrelharp:
Many species are partial selfers; for instance *C. elegans* (see #834) has a selfing rate of about 99.9%. SLiM has a "selfing rate" option that we can use directly; for msprime we'd need to rescale things (see for instance [Nordborg & Donnelly](https://academic.oup.com/genetics/article-abstract/146/3/1185/6053972\)). So, maybe we need to 

- [ ] add `selfing_rate` as a new attribute to `Species` objects
- [ ] modify engines to use this information

grahamgower:
FYI, [`demes`](https://github.com/popsim-consortium/demes-python) includes a `selfing_rate` (and a `cloning_rate`).

grahamgower:
This is a duplicate of #715. @petrelharp, I'm not sure which issue you'd prefer to keep open?

petrelharp:
I'l close the other, but quote @pblischak from it here:
> I think the main issue for a coalescent simulator and selfing is that it can't capture the increase in homozygosity within individuals if chromosomes, rather than individuals, are randomly sampled from the population. Both theta and rho can be scaled by 1+F like Ryan mentioned but the non-random sampling is what may cause the bigger issue. People have typically dealt with this by only sampling one allele per SNP (or some version of this), effectively "haploidizing" their data

petrelharp:
Discussed today - implementing selfing means that we *also* need to figure out what the demographic models inferred for selfing species actually mean. The current proposal is (very roughly) - if we turned on selfing we should also multiply all the population sizes by 2, for demographic inferences that don't explicitly include selfing?

petrelharp:
See discussion in #1858 also!

PR #1855, opened 2026-06-29
AraTha recombination rate bug fix + selfing

srgibbon17:
This is in response to #1854 (see the discussion there for more context). 

Essentially there was a discrepancy between the rate map and genome-wide rates, which seems to stem from whether or not selfing was accounted for in the recombination rate.

Below, I start from the recombination rates in the map from [Salome et al. 2012](https://doi.org/10.1038/hdy.2011.95) which are calculated from F2 populations (so no selfing). Taking those rates, I then adjust for selfing following [Nordborg 2000](https://doi.org/10.1093/genetics/154.2.923) (which is also done [here for Oryza sativa](https://github.com/popsim-consortium/stdpopsim/blob/05f33ad9cad479d42a8fe3132f0d6f0fa20f08c3/stdpopsim/catalog/OrySat/species.py#L59-L66)). 

I'll open a separate QC issue and tag it here.


srgibbon17:
QC request is in #1856

silastittes:
Happy to help! Typically we would separate the issue creation and PR between two people to quality assurance, but this isn't a new addition to a catalogue, but a patch. Do @andrewkern and @petrelharp or others have thoughts on the process here? Are there additional tests, or changes to code we need for this patch to be consistent with other parts of the cartalogue?

petrelharp:
Awesome, thanks @srgibbon17! I'll have a look.

> Typically we would separate the issue creation and PR between two people to quality assurance

I'm not sure what you mean here, @silastittes - the only policy I'm aware of here is separating implementation from QC? Opening a QC issue is different than actually doing QC? In https://github.com/popsim-consortium/stdpopsim/issues/1854#issuecomment-4734587013 I was proposing that someone (maybe you? =) do the QC after this PR?

silastittes:
Ah ok! I think maybe I'm confused about the process. Sure, happy to do QC after this PR.

petrelharp:
Let's see: Some comments: @srgibbon17 - could you perhaps

- put some more words into this PR to explain exactly what's being done; I think this should include some fairly extensive comments explaining things, and a short summary in the `"long_description"`? This should be pretty self-contained. For instance, citing an equation number in Nordberg.

- put in how the mean recombination rates were calculated from the Salome2012 map?

Your solution to adding the selfing correction to the maps is clever. I was trying to think of a more elegant solution- like maybe including selfing rate or something as an argument- but actually doing selfing right is a can of worms: see #857. So I guess it's okay?

petrelharp:
Oh and: in the QC issue, please put the text describing any choices you made, with the rationale, so the QC person knows what to do.

srgibbon17:
Thanks for the suggestions and review, @petrelharp! I agree a selfing argument would be preferable, but is, perhaps, outside the scope of this issue. 

I added some comments and a brief addendum to the `"long_description"`. Unfortunately, Nordborg has no equation numbers, so I referenced the section and page number instead (but happy to do something else if there is a different standard/preference). 

I also gave a (somewhat lengthy) description of my rationale in the QC issue and a few ideas for what needs to be checked.

stdpopsim/catalog/AraTha/genetic_maps.py
```diff
@@ -9,7 +11,28 @@
     reasons={stdpopsim.CiteReason.GEN_MAP},
 )
 
-_gm = stdpopsim.GeneticMap(
+
+# define a new subclass to handle the selfing rate correction
```
petrelharp: We don't need this explanation of what a subclass is - we need the scientific explanation, which you've got part of  (referring to Nordborg) but maybe it should also say that Salome estimated recombination rates under outcrossing?

Also we can delete the per-line comments that just rephrase the operation done in that line of code?

stdpopsim/catalog/AraTha/species.py
```diff
@@ -1,15 +1,27 @@
 import stdpopsim
 from . import genome_data
 
+# correction to the recombination rate to account for selfing
+# sigma = selfing rate, which here is 0.97 (Platt et al. 2010 PLOS Genetics)
```
petrelharp, line 5: Mind adding the citation to Platt et al in the appropriate places?

stdpopsim/catalog/AraTha/species.py
```diff
@@ -1,15 +1,27 @@
...
-_rho = 200 / 1e6  # 200/Mb
-_Ne = 124000
-_mean_recombination_rate = _rho / (2 * _Ne)
-_recombination_rate = {str(j): _mean_recombination_rate for j in range(1, 6)}
-_recombination_rate["Mt"] = 0
-_recombination_rate["Pt"] = 0  # JK Is this correct??
+# Mean recombination rates were calculated from the Salome2012 map for F2
+# populations by taking a weighted average over non-missing windows.
+# (See: `msprime.RateMap.mean_rate` for details).
+# Then, the selfing correction is applied to each non-zero recombination rate.
```
petrelharp:
```suggestion
# populations by taking a weighted average over non-missing windows,
# using `msprime.RateMap.mean_rate` .
```

petrelharp (commented):
A few more comments/suggestions!

stdpopsim/catalog/AraTha/genetic_maps.py
srgibbon17: Thanks for clarifying! I trimmed this down to focus just on the relevant scientific explanation and added a cite here to Platt et al.

stdpopsim/catalog/AraTha/species.py
srgibbon17: I added the Platt et al. reference anywhere the selfing rate was mentioned and swapped the Salomé 2012 paper in for the Huber paper for the recombination rate citation. 

I don't know what the standard practice is, but I have not added a `stdpopsim.Citation` for either Nordborg 2000 or Platt et al. 2010 (in part because it is not clear what the correct `CiteReason` would be).

petrelharp: Good question; I think we're using those to compute the recombination rate so let's cite Platt for RecombinationRate. (We aren't citing general theory, so no need for Nordborg I think.)

petrelharp:
Whoops - could you remove the `.DS_Store` file here?

stdpopsim/catalog/AraTha/genetic_maps.py
```diff
@@ -9,7 +11,24 @@
...
+class _AraThaGeneticMap(stdpopsim.GeneticMap):
+    def get_chromosome_map(self, id):
+        rate_map = super().get_chromosome_map(id)
+        return msprime.RateMap(
+            position=rate_map.position,
+            rate=rate_map.rate * _selfing_correction,
+        )
+
+
+# Here, we use `_AraThaGeneticMap` in place of `stdpopsim.GeneticMap`
```
petrelharp:
```suggestion
```

stdpopsim/catalog/AraTha/species.py
```diff
@@ -1,15 +1,26 @@
...
-# associated with all recombining chromosomes
-_rho = 200 / 1e6  # 200/Mb
-_Ne = 124000
-_mean_recombination_rate = _rho / (2 * _Ne)
-_recombination_rate = {str(j): _mean_recombination_rate for j in range(1, 6)}
-_recombination_rate["Mt"] = 0
-_recombination_rate["Pt"] = 0  # JK Is this correct??
+# Mean recombination rates were calculated from the Salome2012 map for F2
+# populations by taking a weighted average over non-missing windows,
+# using `msprime.RateMap.mean_rate` .
```
petrelharp:
```suggestion
# Mean recombination rates were calculated from the Salome2012 map for F2
# populations by taking a weighted average over non-missing windows,
# using `msprime.RateMap.mean_rate` . This was done using the uncorrected
# map returned by stdpopsim through version 0.3.0.
```

petrelharp:
Ah and *one* last thing: could you add somethign to the `CHANGELOG`?

stdpopsim/catalog/AraTha/species.py
srgibbon17: I added a citation to Platt, so this should be good to go now!

CHANGELOG.rst
```diff
@@ -11,6 +11,11 @@ Upcoming
 - The 'positive' and 'negative' proportions were swapped in the PosNeg_R24 DFE.
     (:user:`bhaller`, :user:`petrelharp`, :pr:`1736`)
 
+- The recombination rates and rate map for AraTha have been updated to
+    include an adjustment for selfing.
```
petrelharp:
```suggestion
- The recombination rates and rate map for AraTha have been updated to
    include an adjustment for selfing, making them about 5.8% of their previous
    values. The recombination map used was estimated from F2s and so would
    overestimate long-term recombination in a selfing population.
```

petrelharp:
One last suggestion - the changelog for an important thing like this should say concretely what happened.

silastittes:
Sorry, deleting my concern here to keep the convo relevant on this PR.

petrelharp:
This was rolled into #1858.

Issue #1830, opened 2026-04-30
Should it be DME or DES?


petrelharp:
@bhaller points out that he's used DES in SLiM, and put a lot of thought into that.

I feel like we considered that and didn't, for some reason?

gregorgorjanc:
DME came from January discussions in Eugene. We were discussing in passing how to turn DFE (Distribution of Fitness Effects, so the Dist of X Effects, where X is fitness) into "a trait thing" and during this we had to coin a name, so Distribution of Mutation Effects (DME) was proposed to cover both fitness and traits and we ran with it without giving it much thought. I am saying the later, because if we would have followed the logic in DFE, we would have called it DTE (for Distribution of Trait Effects), but I think we wanted something more generic, which is where (I think) DME came in. 

Later @petrelharp pointed out that DES (Distribution of Effect Sizes) is used in SLiM. We actually both knew this, but completely forgot about it at the time - sorry @bhaller!

As a non-native speaker I am quite term-agnostic about which term we should use (I always keep at least two terms for the same concept in my head anyway!). Pro for DME is that is says it is distribution of mutation effects and not any other effects, say environmental effects. Con for DME is that SLiM already uses DES. Opposite pro and con then apply to DES.

bhaller:
To me, a big pro for DES is that it emphasizes "effect _size_".  As I wrote in the Slack thread, in SLiM effect _size_ is without dominance or other considerations; it is _s_ and _a_ (for multiplicative and additive traits, respectively).  A mutation _effect_ is the final effect of the mutation with dominance and other considerations; it is 1+_s_ or 1+_hs_, 2 _a_ or 2 _ha_.  Those things do need to be differentiated terminologically.  DME is on the wrong side of that line, at least as far as the terminology I chose for SLiM is concerned; it suggests that it is the distribution of _effects_ – 1+_s_ and 1+_hs_ values, or 2 _a_ and 2 _ha_ values – but it is not that at all, it is the distribution of effect _sizes_.  Now, maybe the terminological choices I made for SLiM are imperfect in some way; but they seem good to me...?  (And there was a very long period for discussion and feedback on those choices.  :->)  And there is value is consistency across our software ecosystem, as has been noted; I'm just pointing out here that it extends also to this terminology of "effects" versus "effect sizes".  :->

petrelharp:
I think my only complaint about "effect size" is that "size" usually means absolute value. But that's minor. And are we meaning the same thing in both places?

[...]

PR #1860, opened 2026-08-11
Getting traits into slim_engine.py

jeffspence:
## Work in progress (not ready for primetime yet)

This is where @roshnipatel and I have been working on getting traits implemented in slim_engine.

### This is what we've done:

- in `slim_engine.py`:
    - we simulate effect sizes for mutations from the DME using mutation callbacks
    - we simulate environments in late blocks
    - add environment to phenotype and then potentially transform
    - apply fitness functions in fitness callbacks
    - Checks for consistency between `contig` and `TraitsModel`
    - Checks for consistency between `TraitsModel` and demographic model
- Updated `Environment` and `FitnessFunction` in `traits.py` to specify the times and populations to which they apply (which is then used in `slim_engine.py`)
- We've implemented tests for the changes in `traits.py`, and have some hacky, non-functional tests for a couple of `slim_engine.py` things.

### Here are our TODOs:
(edit: crossing these out as we get them done)
- ~We currently implement everything in `late()` blocks or `fitnessEffect()` code blocks.  This is inconsistent with the ethos of the rest of `slim_engine.py` which uses `registerLateEvent()` to programmatically register demographic events based on tables of values defined at the beginning of the SLiM script.~
    - ~This also causes some issues for us because registered events always happen _after_ `late()` code blocks.  This is a problem when new populations are created by population splits, because we want to access traits in populations that don't exist yet in their first generation.  For now we've switch population splits to `registerEarlyEvent()`s but this is probably _bad_.~
- ~Tests for things in slim_engine.~
- ~We implemented a way to check that `Environment`s and `FitnessFunction`s are consistent with the demography, but it would be cool to be able to print out a "decorated" demography somewhere, where it prints out simultaneously the demography and fitness functions and environments. @andrewkern -- would you be interested in tackling this?~
- We have a bunch of TODOs sprinkled throughout, for things like implementing more fitness functions (we only implemented gaussian) or environmental distributions (we only implemented MVN)
- ~Testing and writing metadata.~
- ~Update environments to use SLiM's individual offsets instead of doing it all ourselves.~

### Other things:
- ~For testing purposes I added `generate_slim.py` and `generate_slim.sh` which make `test_script.slim` (all in the `stdpopsim/` directory.  These are just temporary and I'm using them to look at the SLiM code generated by `slim_engine.py`.  Once the PR is ready, they'll get removed.~


roshnipatel:
Also, we'd love input on certain decisions we've made about how generation times get interpreted since we newly allow users to specify generation times for both environments and fitness functions.

In general, the decisions can be summarized as:
- (Positive) infinity is interpreted as the start of burn-in
- Burn-in is scheduled to end at the earliest (finite) time specified in the demographic model, the environments, and the fitness functions
- If no population is specified, the condition (environment or fitness function) applies to all active populations; if no time is specified, the condition applies to all times.
- Finite intervals are required to be fully contained within the intervals on which a population exists; infinite intervals are interpreted more generously.

For some concrete examples, suppose we have a demographic model that begins 5,000 generations ago.
- The interval `[0, inf)` describes an event that occurs throughout burn-in and continues to the present.
- The interval `[0, 5000)` describes an event that begins at the start of the demographic model (but NOT during burn-in) and continues to the present.
- The interval `[0, 6000)` describes an event that begins 1,000 generations before the demographic model and continues to the present; burn-in is then scheduled to end 6,000 generations ago (rather than 5,000 generations ago, as would be implied by the demographic model). 

Also suppose that in this demographic model, pop B splits 3,000 generations ago, i.e. only exists on the interval `[0, 3000)`.
- With `population_list = [popB]`, specifying `time_intervals` as `[0, inf)` and `[0, 3000)` has the same result: the condition will apply to population B in its entirety.
- With `population_list = [popB]`, specifying `time_intervals` as `[0, 4000)` will error -- this is a finite interval, so it is required to fully overlap the time(s) that pop B exists.

stdpopsim/slim_engine.py
```diff
@@ -1521,9 +1780,275 @@ def matrix2str(
...
+            # will be properly indented.
+            env_code_spacing = "        "
+
+        if env.population_list is None:
+            printsc(env_code_spacing + 'affected_inds = inds;')
+        else:
+            pop_ind_str = []
+            for pop_id in env.population_list:
+                pop_ind_str.append(f'sim.subpopulations[{pop_id}].individuals')
+            pop_ind_str = ", ".join(pop_ind_str)
```
andrewkern: i'm a bit worried about this implementation choice. here subpopulations are referred to by index, so e.g., SLiM  will go and pull subpop '2'. But if the number of active populations is changing over time, this could be an issue.

elsewhere I see that pop_ids (strings) are used, for example in the fitness function generator. I think that's the thing to do here too?

Like line 1831 could use `f'p{pop_id}.individuals'`?

stdpopsim/slim_engine.py
```diff
@@ -910,12 +942,146 @@ def msprime_rm_to_slim_rm(recombination_map):
...
+        )
+
+
+def _collect_valid_population_intervals(demographic_model):
+    # First, figure out when populations are around
+    # TODO: this might be wrong if there's a population that's not active at
+    # the present.
+    accessible_demes = demographic_model.model.debug().possible_lineage_locations(
+        [
+            msprime.SampleSet(1, population=p.id, time=0)
```
andrewkern: the TODO note is right here-- I think this will fail if the population isn't active, like with models with ancestral populations not present at sampling time. I _think_ the fix is to set `time=None` and then msprime will sample each population at its default sampling time. I _think_....

andrewkern: actually i went and tested this and I'm wrong about `time=None` -- that doesn't help.

instead i think we need to use `demography.debug().epochs` to walk through each to see which population is around

stdpopsim/traits.py
```diff
@@ -15,6 +15,39 @@ def _copy_converter(x):
...
+            )
+        if not isinstance(interval[0], (int, float)):
+            raise ValueError("Intervals must be numeric.")
+        if not isinstance(interval[1], (int, float)):
+            raise ValueError("Intervals must be numeric.")
+        if interval[0] < 0:
+            raise ValueError(
+                "Intervals must start at the present or some more ancient time."
+            )
+        if interval[0] > interval[1]:
```
andrewkern, line 39: i think we want to check that interval[0] is finite -- this would be a user error that hopefully we wouldn't see, but stuff would blow up if the user handed [inf, inf)

also we should check for zero length intervals like [4, 4) -- those would slip by, but probably not affect things?

andrewkern:
generally this looks really good! I like the decisions you've outlined above @roshnipatel. I've placed some comments on some nitpicks I saw along the way, but haven't done a complete review of the code. Happy to do that whenever

andrewkern:
and yeah -- i could take a swing at the decorated demography thing @jeffspence. are you thinking a function that produces a table _a la_ msprime's demography debugger?

jeffspence:
@andrewkern yeah! something like that would be great.  Just something showing the populations and times of events (both demographic and traits related).

bhaller:
Hi folks!  A few superficial comments, not having looked at the underlying code:

- A while back, we discussed the terminology of "DME" versus "DES", and my (perhaps erroneous) recollection is that you guys decided to follow SLiM in using "DES", for consistency; I think that'd be a good idea.  See section 1.7 of the draft SLiM manual for terminological discussion; it'd probably be good for stdpopsim to match SLiM's terminology on this stuff, to try to minimize confusion.

- You write "add environment to phenotype" as something you've done, and you have "Update environments to use SLiM's individual offsets instead of doing it all ourselves" but it's crossed out.  What's the situation there?  If the way SLiM does individual offsets is not fit for its purpose, and so you have to roll your own equivalent mechanism for some reason, I'd like to hear about that.

- You write "we simulate effect sizes for mutations from the DME using mutation callbacks"; is this because you are supporting distributions that SLiM doesn't support built-in?  Or because you want to have correlations between effects of different traits?  Again, if SLiM's built-in facilities are not sufficient I'd like to know why.  If there's a distribution you want added, just ask.  :->  (On the other hand, if it is for reasons of correlated effect sizes, then yeah, SLiM doesn't support that built-in.)

- You write "apply fitness functions in fitness callbacks"; doing this in a vectorized fashion and putting the results into `fitnessScaling` would probably be more efficient (as in the examples in chapter 13 of the draft manual, I think), but perhaps this is not a huge deal since the callback(s) would only be called once per individual.  If you've got a clear reason for your design, then don't worry about it, I doubt the performance hit will be large for most models.

- You write "This also causes some issues for us because registered events always happen _after_ late() code blocks".  Yes, hmm.  SLiM's scheduling internals are not very exposed right now.  I could conceivably add a parameter to the various `registerX()` methods that would let you specify whether to add the new X to the start or to the end of the queue, if that would be useful...?  Adding more flexibility than a simple start/end choice would be more complex, and probably would probably be overkill but could be discussed.

That's all I see for right now.  Great to see the progress!  :->  You guys are kind of the guinea pigs for SLiM 6, since I haven't managed to get the SLiM 6 beta out yet; basically nobody is using it yet except you.  So feedback on anything that seems non-optimal would be super useful.  Thanks!

jeffspence:
@bhaller thanks for all of this!!  We are very happy to be guinea pigs, and everything has been going pretty smoothly so far.

> A while back, we discussed the terminology of "DME" versus "DES", and my (perhaps erroneous) recollection is that you guys decided to follow SLiM in using "DES", for consistency; I think that'd be a good idea. See section 1.7 of the draft SLiM manual for terminological discussion; it'd probably be good for stdpopsim to match SLiM's terminology on this stuff, to try to minimize confusion.

I believe that we're discussing that in this issue: https://github.com/popsim-consortium/stdpopsim/issues/1830 .  I'm inclined to stick with DME for now here, but am open to changing it in another PR.

> You write "add environment to phenotype" as something you've done, and you have "Update environments to use SLiM's individual offsets instead of doing it all ourselves" but it's crossed out. What's the situation there? If the way SLiM does individual offsets is not fit for its purpose, and so you have to roll your own equivalent mechanism for some reason, I'd like to hear about that.

Crossed out because I committed some changes to switch everything over to using the individual offsets.  We rolled our own because I didn't read the SLiM documentation carefully enough, and then we switched it over when @petrelharp pointed us to the offsets.  I think that the existing offset mechanism in SLiM works perfectly for us.

> You write "we simulate effect sizes for mutations from the DME using mutation callbacks"; is this because you are supporting distributions that SLiM doesn't support built-in? Or because you want to have correlations between effects of different traits? Again, if SLiM's built-in facilities are not sufficient I'd like to know why. If there's a distribution you want added, just ask. :-> (On the other hand, if it is for reasons of correlated effect sizes, then yeah, SLiM doesn't support that built-in.)

Yes, sorry -- I was a bit sloppy writing this up.  We use the SLiM built-ins for mutations that only affect one trait and mutation callbacks for mutations that affect multiple traits.

> You write "apply fitness functions in fitness callbacks"; doing this in a vectorized fashion and putting the results into `fitnessScaling` would probably be more efficient (as in the examples in chapter 13 of the draft manual, I think), but perhaps this is not a huge deal since the callback(s) would only be called once per individual. If you've got a clear reason for your design, then don't worry about it, I doubt the performance hit will be large for most models.

Thanks for pointing this out.  I'll take a look and see if we can switch to using `fitnessScaling`

> You write "This also causes some issues for us because registered events always happen after late() code blocks". Yes, hmm. SLiM's scheduling internals are not very exposed right now. I could conceivably add a parameter to the various `registerX()` methods that would let you specify whether to add the new X to the start or to the end of the queue, if that would be useful...? Adding more flexibility than a simple start/end choice would be more complex, and probably would probably be overkill but could be discussed.

I think we settled on switching over all of the traits stuff to also use various `registerX()` methods to be consistent with the way we are handling demography.  So at least for now, I think that this will be a non-issue once we've switched over to doing that, but we will let you know if we end up needing this kind of functionality.

Thanks again!!

stdpopsim/slim_engine.py
```diff
@@ -1326,12 +1579,71 @@ def matrix2str(
...
                 f"    initializeGenomicElement({j}, {element_starts}, {element_ends});"
             )
+
+    # Now we deal with the multivariate mutation types. Here we print out the
+    # information that we will need later to generate mutation callbacks.
+    # TODO: this set up is (probably) fine for now, but assumes that everything
+    # is multivariate normal. I think this should probably mimic what we do
+    # with transformations below, where we store the transformation type in a
+    # SLiM list and then store the relevant parameters in a SLiM Dictionary()
+    # That would "futureproof" us to implementing new mutation types
```
petrelharp, line 1589: I think no need to over-complicate to prepare for things we aren't doing now (it's hard to predict what exactly we'd need in the future). However, if that change would make the code more obvious, then sure.

stdpopsim/traits.py
petrelharp: but we catch that here:
```
>>> np.inf > np.inf
False
```
?

stdpopsim/traits.py
```diff
@@ -28,26 +56,60 @@ def _check_trait_ids(trait_ids):
...
+        If multiple fitness functions are added that overlap in time or space,
+        then their effects will combine multiplicatively.
+
         :ivar traits: List of :class:`Trait` objects, with unique IDs.
+            or ``None``.
         :vartype traits: list
         """
-        pids = [p.id for p in traits]
+        # we'll put fitness first, BUT DO NOT RELY ON THIS
+        self.traits = [Trait(id="fitness", type="multiplicative")]
```
petrelharp, line 107: Do we want to revisit this choice (or at least the name of this trait)? What are we using this for (if anything)?

tests/test_dfes.py
```diff
@@ -790,17 +93,26 @@ def test_simulation_runs(self):
...
         nonneutral = np.repeat(False, ts.num_mutations)
+
+        mut_metadata = pyslim.mutation_metadata(ts)
+        assert ts.metadata["SLiM"]["traits"][0]["name"] == "fitnessT"
         for mut in ts.mutations():
-            for j, md in zip(
-                mut.derived_state.split(","), mut.metadata["mutation_list"]
-            ):
+            for j in mut.derived_state.split(","):
+                md = mut_metadata[int(j)]
```
petrelharp: Now this is: (and we can drop the `keep_mutation_ids_as_alleles` above)
```suggestion
            for j in mut.metadata["derived_states"]:
                md = mut_metadata[j]
```

petrelharp: this will still work, TBC

tests/test_slim_engine.py
```diff
@@ -379,7 +380,7 @@ def test_stacked_mutations(self):
                 slim_burn_in=0,
                 keep_mutation_ids_as_alleles=False,
             )
-            is_stacked = [len(m.metadata["mutation_list"]) > 1 for m in ts.mutations()]
+            is_stacked = [m.parent > -1 for m in ts.mutations()]
```
petrelharp: I think probably having a parent is the same as being stacked for us, but in principle that's not the same thing, so:
```suggestion
            is_stacked = [len(m.metadata["derived_states"]) for m in ts.mutations()]
```

tests/test_slim_engine.py
```diff
@@ -1332,16 +1334,20 @@ def test_slim_simulates_dfes(self):
                 )
             )[0]
             assert len(sites) > 0
+            mut_metadata = pyslim.mutation_metadata(ts)
+            assert ts.metadata["SLiM"]["traits"][0]["name"] == "fitnessT"
             for k in sites:
                 s = ts.site(k)
                 for mut in s.mutations:
-                    for md in mut.metadata["mutation_list"]:
+                    for slim_idx in map(int, mut.derived_state.split(",")):
```
petrelharp:
```suggestion
                    for slim_idx in mut.metadata["derived_states"]:
```

tests/test_slim_engine.py
```diff
@@ -1874,6 +1890,7 @@ def test_dominance_coeff_list(self):
             samples=self.samples,
             verbosity=3,  # to get metadata output
             seed=260,
+            keep_mutation_ids_as_alleles=True,
```
petrelharp: again let's drop this and use metadata as above

tests/test_slim_engine.py
```diff
@@ -2772,7 +2792,7 @@ def test_sweep_meets_min_freq_at_start(self, tmp_path):
                 slim_burn_in=1,
                 logfile=logfile,
                 logfile_interval=1,
-                seed=654,
+                seed=654 + idx,  # otherwise we just repeat the same simulation?
```
petrelharp, line 2797: huh this is puzzling, good call

tests/test_slim_engine.py
```diff
@@ -3386,3 +3412,1116 @@ def test_separate_sexes_default(self):
...
+            tol2 = 5*np.sqrt(variances[k][1]/num_inds)
+
+            assert obs1 >= v[0] - tol1 and obs1 <= v[0] + tol1
+            assert obs2 >= v[1] - tol2 and obs2 <= v[1] + tol2
+
+        for k, v in variances.items():
+            obs1 = np.var(list(zip(*observations[k]))[0])
+            obs2 = np.var(list(zip(*observations[k]))[1])
+            num_inds = len(observations[k])
+            # these are off by a factor of (n-1)/n or something, but
```
petrelharp, line 3969: if so add `ddof=1` to the call to `np.var( )`?

petrelharp: (generally I agree with this comment tho)

tests/test_slim_engine.py
```diff
@@ -3386,3 +3412,1116 @@ def test_separate_sexes_default(self):
...
+            # being within 5 standard deviations should be fine regardless of
+            # these details. Same with using a biased estimator.
+            tol1 = 5*np.sqrt(2*variances[k][0]**2/num_inds)
+            tol2 = 5*np.sqrt(2*variances[k][1]**2/num_inds)
```
petrelharp:
```suggestion
            tol2 = 5*np.sqrt(2*variances[k][1]**2/num_inds)
            assert v[0] - tol1 > 0
            assert v[1] - tol2 > 0
```

tests/test_slim_engine.py
petrelharp: um wait but you've not got standard deviations here? you've got `5 * sqrt(2/n) * variances[k]`? I'm not following where `tol1` and `tol2` are coming from?

tests/test_traits.py
```diff
@@ -522,3 +672,724 @@ def test_gaussian_errors(self):
...
             stdpopsim.FitnessFunction(**bad_args)
+
+
+class TestCreateMutationType:
+    """
+    Tests for creating a MutationType instance.
+
+    TODO: move this to `tests/test_traits.py`; I'm leaving it here
+    for now so that the diff is cleaner.
+    """
```
petrelharp, line 699:
```suggestion
```

petrelharp:
Notes on how this works from me and @roshnipatel going over things. Looking at the SLiM script, we:

1. initialize traits, including a fitnessT trait that wasn't part of the traits model, and is the only one with a diirect fitness effect
2. sets up mutation times with no effect (?)
3. and then sets up effect size distributions for traits (?) including fitness
4. sets up dictionaries of things: `multivar_mut_*` for later use?
5. defines the "oldest time" and the eopch boundaries `_T`
6. defines some lists and dictionaries for environment and fitness and transforms `env_`, `trait_transform`, `fit_func`, `fitness_callbacks`
7. pop size change events are registered in late
8. migration rate change events registered in late, then admixture pulses then fitness callbacks (for what??)
9. set up trait mutation callbacks for correlated mutational effects using parameters stored in `multivar_mut_*`
10. set up environment script blocks, which are late events that draws `env_effects` and uses these to *add* to the individual's offsets for the traits (*note* this relies on being a WF simulation!)
11. trait transformation script blocks, in late, which use things in `trat_transforms_*` and do `inds.setPhenotypeForTrait`
12. sets up trait finess effect callbacks (which are all dnorm-based) 
13. mutation callbacks for the h-s relationship

Notes:
- Each script block first does something like `check_if_in_interval` to see if they're in the right interval and if they're in the right subpopulation; so they're active in all ticks but don't do anything except when they're supposed to.
- In setting up trait mutation callbacks we're looking things up in eg `multivar_mut_means` but could instead just stick in the actual number, avoiding the lookup for every mutation? But then we couldn't change these things with time, do we need to do that?
- If a fitness function applies to every population, this ends up being one fitness function for each population.

petrelharp:
TODO:

- [ ] get pyslim in CI using development head
- [ ] currently environments have to be MVN (open issue for subsequent work?)
- [ ] fitness functions have to be Gaussian (open issue for subsequent work?)
- [ ] what do we want to get out and how do we give it to the users? a utility to get things into plink? some way to get a nice file out of mutation effects? or in the `INFO` column about effect sizes? (TODO: make issues in pyslim)
- [ ] scaling? (we have an open issue about this) but it's in here and there's no testing for it! (open issue!)
- [ ] maybe don't create the `fitnessT` trait if there's not a DFE?
- [ ] make sure the phenotype transforms are in effect when fitness effect callbacks happen, and for tree seq output (we think it's okay; double-check)

Current testing strategy:
- sets up situations where the effect is OBVIOUS and in limited times/places and checks that it's happening
- and some weak statistical tests

bhaller:
I've got a couple comments from the peanut gallery here.

Re: (1), it feels a bit weird to always have this `fitnessT` trait.  The new SLiM 6 design allows you to have zero, one, or many traits with direct effects on fitness.  Why not define the actual traits that you need to model the particular scenario at hand, rather than always having a catch-all "fitness" trait?  That kind of feels like trying to wedge the SLiM 5 paradigm into SLiM 6.  Won't you sometimes have more than one phenotypic trait with a direct effect on fitness (two different Mendelian loci, associated with different phenotypic traits, that are both undergoing a selective sweep, say), and sometimes none (just a quantitative trait subject to a fitness function, with no direct fitness effect from anything)?  Defining traits unnecessarily will have a large performance impact, since SLiM then keeps mutation effect sizes and effects on that (unused, unnecessary) trait for every mutation in the model.

Re: (11), you can use `setPhenotypeForTrait()`, but be very aware of the caveats on doing that.  In the end, SLiM owns the trait values.  It will invalidate them when it feels that that is appropriate, and then recalculate them from its own data the next time that demand is expressed for them, and so if you put values into them those values might get overwritten with different values at unexpected times unless you _really_ understand the whole model of when and why SLiM invalidates trait values (and there's no guarantee that those rules won't change as SLiM evolves).  If you want to start with trait values as calculated by SLiM and do further calculations based upon them to figure out derived stuff (transformed values, derived/composite traits, ...), you might consider using setValue() to store that kind of additional information instead, and leaving the trait values themselves alone.

Re: (13), ok, we know that an h-s relationship exists empirically.  But that relationship might be entirely an emergent consequence of the effect of selection on a distribution that is initially unbiased.  I'm not sure it's a good idea to build some empirical h-s relationship into a DES in a SLiM model, because then selection will act upon the mutations generated from _that_, and you'll get some _other_ distribution of the realized h-s relationship that is different from the empirical one you plugged in.  I imagine you're already aware of all this, but I thought maybe it'd be helpful to point it out.  You might want to use a different (perhaps unbiased) h-s relationship and allow the realized/empirical h-s relationship to emerge as a consequence of selection.

> Each script block first does something like check_if_in_interval to see if they're in the right interval and if they're in the right subpopulation; so they're active in all ticks but don't do anything except when they're supposed to.

Is there a strong reason not to generate them with the tick range they apply to?

> In setting up trait mutation callbacks we're looking things up...

In my experience new mutations are a rare enough event that a little extra overhead in a `mutation()` callback isn't a big deal.  Extra overhead in a `mutationEffect()` callback, on the other hand, kills you.  (Using `mutationEffect()` callbacks at all kills you!)

There's my two cents.  Disregard if it's not useful.  :->

bhaller:
OK, and re: the checklist you just posted:

> what do we want to get out and how do we give it to the users?

If adding something like plink output to SLiM makes sense, feel free to open an issue.  And note that SLiM's built-in output format, as well as VCF, will provide a way to get multi-trait output in future; that just hasn't been implemented yet.  (I've been regarding that as a post-beta polish thing, but I could raise its priority.)

More broadly, this is just to say: when there are things that stdpopsim needs that would probably be of broader utility to other users as well, let's talk about whether it makes sense to add them to SLiM directly.  You guys are on the bleeding edge, which is great; the rough edges that you experience might well be rough edges that other people are going to experience down the road.  :-O

petrelharp:
(editing the TODO above to reflect some of Ben's comments)

petrelharp:
> Is there a strong reason not to generate them with the tick range they apply to?

Well, kinda: what with rescaling, and times not being integers, this turns out to be a nicer and cleaner way to do things that to figure out how to calculate the ranges.

bhaller:
> > Is there a strong reason not to generate them with the tick range they apply to?
> 
> Well, kinda: what with rescaling, and times not being integers, this turns out to be a nicer and cleaner way to do things that to figure out how to calculate the ranges.

OK.  But keep in mind that tick ranges on script blocks can now involve user-defined functions.  You could have:

```
SPS_SCHEDULE(x, y, z) late() {
   ...
}
```

where `SPS_SCHEDULE()` is a user-defined function.  The virtue of this is that it just does the scheduling once and then SLiM tracks things for you after that, with no scripting overhead involved just to answer the question "am I active?" in every tick.  And the x, y, z in the call there could involve defined constants like the rescaling factor, etc.  It _should_ all work the way that you would want it to.  But I can accept that you folks might have reasons for want to roll your own scheduling, given how complex the stdpopsim internals are.  Just pointing out the new-ish SLiM functionality that you might not be aware of.  :->

petrelharp:
Good idea, that might help with some things.

bhaller:
> Good idea, that might help with some things.

Let me know if you hit any roadblocks with it.  :->

petrelharp:
Well, this is now using development pyslim!

jeffspence:
@petrelharp -- do you know what's happening with the new assertion errors in `pyslim` that are getting triggered?  I could get all of the tests to pass before on my local machine using the version of `pyslim` I initially submitted in [this PR](https://github.com/tskit-dev/pyslim/pull/434).

petrelharp:
Well, that code isn't very transparent since it has to be vectorized to be sufficiently efficient on large tree sequences. The error means that something went wrong with it? Happily, Ben's just committed a change that can let me make that code a lot simpler. I'll look into it.

petrelharp:
Just to check, do you get these errors locally?

jeffspence:
Yeah -- I just updated to the latest dev `pyslim` commit and get the same errors.  I also updated to the latest commit on the multitrait branch of SLiM and still get them.

petrelharp:
I get them locally, that's good news.

petrelharp:
The problem was the code was still using `SLiMMutationModel`. I've made a note that pyslim should notice this and raise a more informative error: https://github.com/tskit-dev/pyslim/issues/450. (Too bad that the metadata parses!)

petrelharp:
I've also got the windows CI using development SLiM (it wasn't because of a hold-over) and inserted a linebreak to make the docs build happy (to check this locally just run `make` inside of `docs/`). *edit:* whoops, no I didn't

bhaller:
> The problem was the code was still using `SLiMMutationModel`. I've made a note that pyslim should notice this and raise a more informative error: [tskit-dev/pyslim#450](https://github.com/tskit-dev/pyslim/issues/450). (Too bad that the metadata parses!)

Yeah, this is the kind of thing that using SLiMv5MutatiomModel instead would deal with.  Sometimes it really is better to break people's code than to silently run with something that is likely to be wrong.

petrelharp:
> Yeah, this is the kind of thing that using SLiMv5MutatiomModel instead would deal with. Sometimes it really is better to break people's code than to silently run with something that is likely to be wrong.

See discussion at https://github.com/tskit-dev/pyslim/issues/450 for what we can actually do about this.

petrelharp:
The tests are passing 🎉 ! Let's not worry about the Windows tests in this PR: we can copy things over from the SLiM CI, but maybe a better way to go is to migrate us from conda to uv (see the pyslim CI machinery, mostly at https://github.com/tskit-dev/.github).

But let's do get test coverage on those last few lines!

jeffspence:
I can take a crack at writing some of these tests today, unless @petrelharp or @roshnipatel are planning to do so.

petrelharp:
Do it; next on my list is the code review!

stdpopsim/traits.py
```diff
@@ -15,6 +16,33 @@ def _copy_converter(x):
...
+        if not isinstance(interval[0], (int, float)):
+            raise ValueError("Intervals must be numeric.")
+        if not isinstance(interval[1], (int, float)):
+            raise ValueError("Intervals must be numeric.")
+        if interval[0] < 0:
+            raise ValueError(
+                "Intervals must start at the present or some more ancient time."
+            )
+        if interval[0] > interval[1]:
+            raise ValueError("Intervals must be specified as (lower, upper).")
```
petrelharp:
```suggestion
            raise ValueError("Lower end of interval must be strictly less than upper end.")
```

stdpopsim/traits.py
```diff
@@ -28,26 +56,60 @@ def _check_trait_ids(trait_ids):
...
+    oldest_event = max(e.start_time for e in demography_debugger.epochs)
+    oldest_time = oldest_event
+    oldest_id = ""
+    conditions = traits_model.environments + traits_model.fitness_functions
+    for condition in conditions:
+        for interval in condition.time_intervals:
+            for t in interval:
+                if np.isfinite(t) and t > oldest_time:
+                    oldest_time = t
+                    oldest_id = condition.id
```
petrelharp, line 77: In the case where one condition is [a,b) and another one is [b, inf) then which ID do we want? I'm assuming the latter? This code will do that if the conditions are sorted, which they aren't guaranteed to be since we just concatenate environments and fitness functions. Hm, and what should happen if there's more than one condition with the same (maximal) endpoint?

petrelharp: Ah, I see what this is for, okay; I'm going to rearrange some here.

jeffspence:
Okay -- if I understand correctly, the last bit of uncovered code in codecov will get covered when we get the windows tests work.  The new uncovered lines are all related to this in `cli.py`:

```
# resource is from the standard library, but it's not available on
# Windows. We break from the usual import grouping conventions here
# to avoid lots of pep8 complaints about mixing imports and code.
_resource_module_available = False
try:
    import resource

    _resource_module_available = True
except ImportError:
    pass
```

PR #1864, opened 2026-08-26
Add check on upper limit for supported SLiM version, update lower bound to 5.0

### LINKED ISSUE #1863: broken sweep example in tutorial
I ran the tutorial as specified in https://popsim-consortium.github.io/stdpopsim-docs/stable/tutorial.html#selective-sweeps.

It failed with the following error: 
```
File "/Users/zchen13/miniforge3/envs/stdpopsim/lib/python3.14/site-packages/stdpopsim/slim_engine.py", line 1709, in simulate
self._run_slim(
~~~~~~~~~~~~~~^
script_filename,
^^^^^^^^^^^^^^^^
...<3 lines>...
verbosity=verbosity,
^^^^^^^^^^^^^^^^^^^^
)
^
File "/Users/zchen13/miniforge3/envs/stdpopsim/lib/python3.14/site-packages/stdpopsim/slim_engine.py", line 1808, in _run_slim
raise SLiMException(
f"{slim_path} exited with code {proc.returncode}.\n{errs}"
)
stdpopsim.slim_engine.SLiMException: slim exited with code 1.
ERROR (EidosValue_Object::GetPropertyOfElements): property genomes is not defined for object element type Subpopulation.

Error on script line 194, character 24:

targets = sample(pop.genomes, 1);
```
any ideas for how to fix it? It seems to be a version incompatibility issue, that the stdpopsim is generating a slim script with pop.genomes, which is a syntax no longer supported by the current slim (conda installed with stdpopsim).

nspope:
Closes #1863


nspope:
Hm, the conda env is building with development slim (now erroring out given the version bounds), which makes sense---that's needed for the trait work. Presumably that means the incompatibilities from #1863 have been already fixed in development ~~tskit~~ stdpopsim, and the bounding in this PR isn't needed (or can be relaxed). Either that, or the incompatibility isn't being hit by a test.

petrelharp:
The problem with #1863 was fixed in stdpopsim in  #1751, I believe. (This has nothing to do with tskit, it's about SLiM.)

petrelharp:
Hm, and that was a while ago but also more recent than our most recent release. We probably should have done a 0.3.1 release to sync with SLiMv5.

nspope:
Slip of the keyboard --- I meant development stdpopsim

stdpopsim/slim_engine.py
```diff
@@ -63,14 +63,18 @@
 logger = logging.getLogger(__name__)
 
 
+# clamp SLiM version to [SLIM_MIN_VERSION, SLIM_MAX_VERSION)
+SLIM_MIN_VERSION = "4.0"
```
petrelharp: I don't think we can do v4, because of the genome->haplosome change at least?

stdpopsim/slim_engine.py
```diff
@@ -63,14 +63,18 @@
 logger = logging.getLogger(__name__)
 
 
+# clamp SLiM version to [SLIM_MIN_VERSION, SLIM_MAX_VERSION)
```
petrelharp, line 66:
```suggestion
# clamp SLiM version to [SLIM_MIN_VERSION, SLIM_MAX_VERSION)
# (note that the MAX is assumed to be *not* compatible)
```

environment.yml
```diff
@@ -16,6 +16,7 @@ dependencies:
         - scikit-allel
         - matplotlib
         - demesdraw
+        #- slim >=4.0,<5.0
```
petrelharp: is this a thing we can put in here?

petrelharp (commented):
LGTM, mod suggestions. thanks!

environment.yml
nspope: yes, we can pin min/max versions here, but no, this comment won't survive (it's me debugging stuff)

nspope:
I'm going to go ahead and merge this---the test failure is a sporadic "fetch bibtex" curl failure on a single runner

Issue #1865, opened 2026-09-04
update CI


petrelharp:
We need to get the SLiM development build working on Windows: see #1860 test failures, for instance. This could be done by following what's done for SLiM. (And some simple attempts at this failed.) But: would it be better to migrate everything (CI and maybe dev docs to) over to `uv`? This would probably involve moving from `requirements/` to putting everything in `pyproject.toml` and following what [tskit](https://github.com/tskit-dev/.github) does. But I don't know the pros and cons here, I'm just following the lead of more skillful folks who set up tskit. (And observing how fast uv is!)

andrewkern:
okay i'm going to work on this. migrating to `uv` would indeed be a good idea and consolidating into a model pyproject.toml.

PR #1868, opened 2026-09-15
bump deps for CI

andrewkern:
trying to update the dependency chain here before moving over to a `uv` workflow as suggested in #1865


nspope (approved):
LGTM, lights are all green

petrelharp (approved):
LGTM!

PR #1872, opened 2026-09-15
moving ci to uv + docs

### LINKED ISSUE #1865: update CI

andrewkern:
closes #1865

this PR aims to move the CI infra over to `uv` following the example of `tskit`:

-  `pyproject.toml` now holds all metadata
-   The `requirements/` directory is gone. Dev tools live in dependency groups: test, docs, lint, maintenance, and dev (all of them). Plain uv sync installs dev.
-  uv.lock pins every version
- ` docs/development.rst` now describes the uv workflow

Note:  I haven't tried the windows SLiM build or the macOS runner. This PR should test them.


andrewkern:
tests are failing... working on it

.github/workflows/docs.yml
```diff
@@ -21,20 +21,19 @@ jobs:
...
+          enable-cache: true
 
-      - name: Install deps
-        run: |
-          python -m pip install wheel
-          python -m pip install -r requirements/CI/requirements.txt
-          # Install the local package so that stdpopsim is in the path,
-          # which is needed for generating docs `command-output`.
-          python -m pip install .
+      - name: Install docs deps
```
petrelharp, line 30: my understanding of how uv works would be that this isn't necessary? but that's a poor understanding, so maybe it is?

.github/workflows/pre-commit.yml
```diff
@@ -11,7 +11,21 @@ jobs:
...
         with:
-          python-version: 3.12
-      - uses: pre-commit/action@v3.0.1
+          python-version: "3.12"
+          enable-cache: true
+
+      - name: Install lint deps
+        run: uv sync --locked --only-group lint
+
+      - name: Cache pre-commit hook environments
```
petrelharp, line 24: but uv is also cached?

.github/workflows/tests.yml
```diff
@@ -30,62 +31,72 @@ jobs:
...
-        run: |
-          micromamba install --yes --file=requirements/CI/conda.txt
-
-      - name: Install pip deps
-        if: steps.cache.outputs.cache-hit != 'true'
-        shell: bash -l {0}
-        run: |
-            pip install -r requirements/CI/requirements.txt
+      # SLiM is not a Python package, so it is built from source below.
+      # Only the binary is cached, keyed on the runner and SLIM_CACHE_NUM.
```
petrelharp: It *is* a python package, on conda; but we need to build it from github head for now. (So we plan to comment this out at the next SLiM release.)

I don't see how only the binary is cached? but I don't know at all how the cacheing works.

.github/workflows/docs.yml
petrelharp: oh I see: is this letting us see failures in installation separate from running things?

docs/development.rst
```diff
@@ -1825,10 +1817,10 @@ Here is a list of things to do when making a new release:
 1. Update the changelog and commit
 2. Create a release using the GitHub UI
 3. `git fetch upstream` on your local branch.
-    Then check out `upstream/main` and create a release tarball
-    (with `python setup.py sdist`).
-    Setuptools_scm will detect the version appopriately.
-4. Upload to PyPI: `twine upload dist/{version just tagged}.tar.gz`
+    Then check out `upstream/main` and build the release sdist and wheel
+    with `uv build`.
```
petrelharp: We should add the instructions for testing things out on TestPyPI here - see the tskit dev docs on this.

.github/workflows/docs.yml
andrewkern: Right, uv run would do it on demand. The separate step keeps install failures out of the docs build log and checks the  lockfile. Added a comment.

.github/workflows/pre-commit.yml
andrewkern: yep but different caches. setup-uv caches Python packages. pre-commit builds its own environment per hook (blech), and that is what this caches. Added a comment.

.github/workflows/tests.yml
andrewkern: Good catch. I meant that SLiM is not on PyPI, so it cannot come from uv.lock.

for the cache: the path for the cache step is SLiM/Release/slim*, so only the built binary is saved and restored,  not the whole build tree.

Rather than commenting the build step out at the next SLiM release, this is built as a switch. There is a SLIM_VERSION variable at the top of the workflow. Empty means build from github. Set it to a released version, e.g. "5.2", and the workflow installs that release from conda-forge with micromamba instead of building. I tested both paths across the the different OSs and it passes with either.

docs/development.rst
andrewkern: Added. I followed the tskit-dev release language/process. take a look

tests/test_version.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+import importlib
+import sys
+
+import stdpopsim
+
+
+class TestVersion:
+    def test_version_is_string(self):
```
petrelharp: perhaps instead follow [this check](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_version.py)?

tests/test_version.py
```diff
@@ -0,0 +1,36 @@
...
+        assert isinstance(stdpopsim.__version__, str)
+        assert len(stdpopsim.__version__) > 0
+
+    def test_fallback_without_version_file(self):
```
petrelharp: Took me a minute to get my head around why we wanted this test. It seems a bit pedantic: isn't this just testing whether the version will be "undefined" if loaded from the source repo? Which is a basic aspect of how python works (apparently)?

If you think it's good though, then fine. However, shouldn't this use `mock.patch` instead of actually modifying the package? (We use this in a bunch of other test files.)

.readthedocs.yml
```diff
@@ -5,16 +5,25 @@
 # Required
 version: 2
 
+build:
```
petrelharp: do we actually need this file any more? we're not using readthedocs.com, we're building our own (but using the RTD theme)

pyproject.toml
```diff
@@ -1,3 +1,109 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Bio-Informatics",
+]
+keywords = ["simulations", "selection", "recombination map", "models"]
+requires-python = ">=3.11"
+dependencies = [
+    "msprime>=1.0.4",
```
petrelharp:
```suggestion
    "msprime>=1.4.4",
```
We need 1.4.3 to get SLiMv6MutationModel, and 1.4.4 was a bugfix.

pyproject.toml
```diff
@@ -1,3 +1,109 @@
...
+    "tskit",
+    "pyslim>=1.0.4",
```
petrelharp:
```suggestion
    "pyslim>=1.1.1",
```
This is confusing: we don't require SLiM, and we don't need pyslim if we don't use SLiM? No harm in requiring it, though. But the version of pyslim depends on the version of SLiM: 1.1.0 was the pyslim update for SLiMv5, and it'll be 1.2.0 for SLiMv6 (which we're using currently on HEAD).  Maybe we just require pyslim, and then wherever we test for "is slim installed" we check that pyslim has an appropriate version?

petrelharp (commented):
LGTM, some minor suggestions/questions

.github/workflows/tests.yml
```diff
@@ -30,62 +39,85 @@ jobs:
...
-      # since this builds SLiM from github head.
-        if: (matrix.os == 'macos-latest' || matrix.os == 'ubuntu-24.04') && steps.cache.outputs.cache-hit != 'true'
-        # If we want to re-build slim from a new commit to the slim repo
-        # we may need to bump the cache key above.
-        shell: bash -l {0}
+      - name: Build development SLiM (Linux, macOS)
+        if: env.SLIM_VERSION == '' && runner.os != 'Windows' && steps.cache-slim.outputs.cache-hit != 'true'
         run: |
-          git clone https://github.com/messerlab/SLiM.git
+          git clone --depth 1 https://github.com/messerlab/SLiM.git
```
petrelharp: so currently over in #1860 we also need this:
```suggestion
          git clone --depth 1 https://github.com/messerlab/SLiM.git
          git checkout multitrait
```
so we *could* add this as a separate option? (also see [tskit](https://github.com/tskit-dev/.github/blob/main/.github/actions/build-docs/action.yml#L63)) But it seems maybe just as clean to put in this single line into this file when we need it, so: never mind?

petrelharp:
Ah, one more thing: we'll need to be building SLiM from source for the development docs here shortly; maybe it makes sense to get that in here now?

pyproject.toml
petrelharp: (maybe the current stdpopsim HEAD only needs 1.0.4, though? so maybe leave it?)

docs/development.rst
petrelharp: Hold on, these docs don't give instructions for how to do the test release? And I'm confused now - I think that over in  [`.github/workflows/wheels.yml`](https://github.com/popsim-consortium/stdpopsim/pull/1872/changes#diff-75627befeafda60a526c116b32f6fdef9bf57bcf456f840d7592c1d6f13facb4R57) it's set up so that pushing to the `test-publish` branch uploads to TestPyPI? But that's not talked about here?

andrewkern: i think this is covered?

pyproject.toml
andrewkern: `stdpopsim/slim_engine.py` imports pyslim so we want the requirement.

andrewkern:
> Ah, one more thing: we'll need to be building SLiM from source for the development docs here shortly; maybe it makes sense to get that in here now?

Done. The docs workflow now uses the same shared install-slim action as the tests

PR #1874, opened 2026-09-21
Handle single-row BED masks

### LINKED ISSUE #1174: read_bed does not work with one-line bed files
To reproduce:
```
import stdpopsim
lines = [
       "\t".join(["chr22", "1000", "1450"]),
]
bedfile = open("ex.bed", "w")
for lin in lines:
            bedfile.write(lin + "\n")

bedfile.close()

stdpopsim.utils.read_bed("ex.bed", "chr22")
```
gets
```
  File "/home/peter/projects/popsim-consortium/stdpopsim/stdpopsim/utils.py", line 147, in read_bed
    lines = lines.compress(in_chrom)
ValueError: condition must be a 1-d array
```

A one-line bed file might be used with the `--inclusion-mask` command-line argument.

VishnuR23:
A BED file containing one interval is loaded as a scalar structured array, causing chromosome filtering to raise `ValueError: condition must be a 1-d array`. Request at least one dimension from NumPy so single-row files follow the same filtering path as multi-row files.

Regression tests cover matching and absent chromosomes, BED3 and extra columns, and single-interval inclusion/exclusion masks through `Species.get_contig()`. All six cases reproduce the failure before the fix.

Validation: masking unit tests and utility tests pass (59 passed, 1 skipped); pre-commit checks pass. The broader masking run also attempted the SLiM integration test, which could not complete because the SLiM executable is not installed locally.

Fixes #1174.


petrelharp:
Well, this seems useful. How did you find this, @VishnuR23?

petrelharp:
Oh right, sorry, via the issue #1174, I suppose.

